PR #1142, opened 2014-11-06
WIP: (PDB-950) Catch and express postgresql regexp failures as better errors

DemonShi:
This commit adds a validation for regexps in queries as was proposed in PDB-950

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ env:
 script: ./ext/travisci/test.sh
 notifications:
   email: false
+addons:
+  postgresql: "9.3"
 services: postgresql
 before_install:
   - rvm use 1.9.3
--- a/src/puppetlabs/puppetdb/query_eng.clj
+++ b/src/puppetlabs/puppetdb/query_eng.clj
@@ -16,7 +16,16 @@
             [puppetlabs.puppetdb.query :as query]
             [net.cgrand.moustache :refer [app]]
             [puppetlabs.puppetdb.jdbc :as jdbc]
-            [puppetlabs.puppetdb.http :as http]))
+            [puppetlabs.puppetdb.http :as http]
+            [clojure.tools.logging :as log]))
+
+(defmacro process-psql-invalid-regexp-or-rethrow
+  [e & body]
+  `(if (= (.getSQLState ~e) "2201B")
+    (do
+      (log/debug ~e "Caught PSQL processing exception")
+      ~@body)
+    (throw ~e)))
 
 (defn ignore-engine-params
   "Query engine munge functions should take two arguments, a version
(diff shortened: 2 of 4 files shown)


kbarber:
@DemonShi I've been absent and missed your last question: I have no specific deadline on this one. In regards to performance testing, we have nothing specific for queries today that we use, its usually bespoke - one could use an external testing tool, or do the test in clojure without the networking traffic so you can iterate in a tight loop. I'm not sure if this change needs something heavier than this today, its easy enough to observe a potential change and conjecture as well :-).

DemonShi:
@kbarber Oh, I've been pretty busy and also had pretty long holidays here too.

According to the PR: I developed a solution that tries to take a first item from result-set lazy sequence under try/catch and resolves a promise with nil. If there is an exception, I resolve that promise with exception from DB.
Then from the "ring" thread we can wait until promise is ready (basically, that is the moment right after we took a first item from result-set or nothing - when result-set is empty) and say whether query was successful or not. If query failed - we re-throw the exception and follow normal error processing (the same way as if there was a count-query). If query is okay - continue execution.

Here is code snippet just to get your opinion of the approach:
```
      (let [{[sql & params] :results-query
             count-query :count-query
             projections :projections} (query->sql version query
                                                   paging-options)
             resp-error (promise)
             resp (output-fn
                   (fn [f]
                     (try
                         (query/streamed-query-result version sql params
                           (comp
                             f
                             #(do
                                (first %)
                                (deliver resp-error nil)
                                %)
                             (munge-fn version projections))))
                       (catch Exception e
                         (deliver resp-error e)
                         (throw e)))))]
        (when @resp-error
          (println (str "SQL error occured: " (.getMessage @resp-error)))
          (throw @resp-error))
        (if count-query
          (http/add-headers resp {:count (jdbc/get-result-count count-query)})
          resp)))))
```

I still need to refactor it a bit, check corner cases and cover it with tests, but the idea of the approach should be clear.

What do you think?

DemonShi:
@puppetcla, I signed CLA

kbarber:
@DemonShi I think this sounds interesting, it looks to me like this should be okay in theory - and it looks like this is only done on the first item in the response only which is great. As long as it doesn't try to realize the lazy-seq at any point (ie. it keeps it as a lazy-seq and just does this on the first item for example) then it should be fine.

DemonShi:
@pljenkinsro test this please

DemonShi:
Couple of questions:
1. Should I write http tests for each endpoint that supports regexps?
2. Should I combine my main commit with travis-related commit?

kbarber:

kbarber:
@DemonShi to answer your questions:

1. no, this is enough
2. no, this is fine for now
3. yes, well - you should be able to? No idea, I can look into github permissions.

kbarber:
thats a PASS, only a rubygems transient is being thrown ^^.

kbarber:
Merged manually, with some small cleanups and the extra tests.

PR #1164, opened 2014-11-26
WIP: (PDB-950) Added upfront validation for regexps in queries in API v4 for HSQLDB

DemonShi:
This comes from [PR-1142](https://github.com/puppetlabs/puppetdb/pull/1142) to simplify merge/discussion.

This PR adds a simple validation for regexps in queries in all APIs v4.
It validates regexps using Java rules, so there might be some constructions that pass this validation but won't be valid in Postgres.
Here is an example of such construction:
curl -v 'http://localhost:8080/v4/fact-contents?query=\["~","name","\\p\{Alnum\}"\]'

Those regexp differences are discussed in original [PR-1142](https://github.com/puppetlabs/puppetdb/pull/1142).

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -3,7 +3,7 @@
             [puppetlabs.puppetdb.zip :as zip]
             [puppetlabs.puppetdb.scf.storage-utils :as su]
             [clojure.string :as str]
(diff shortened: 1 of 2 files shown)


kbarber:
@DemonShi this patch needs rebasing, and there are a few open comments on it.

DemonShi:
I will do this pretty soon. Sorry, I was busy the last days.

kbarber:
@DemonShi re-open this one when you are ready.

PR #1175, opened 2014-12-10
NEEDS INFO: Allow tilde in resource name, like in puppet

saimonn:
tilde character is allowed in the resource and tag names in puppet, but not in puppetdb.
This PR fixes this.


kbarber:
@saimonn any luck on reproducing this again?

PR #1179, opened 2014-12-11
(PDB-1034) Ezbake source based testing

kbarber:
methodology can be retired.

### DIFF
--- a/ext/jenkins/beaker-tests-source.sh
+++ b/ext/jenkins/beaker-tests-source.sh
@@ -32,6 +32,16 @@ mkdir vendor
 
 bundle install --path vendor/bundle --without test
 
+# Install a copy of leiningen
+if [ -d "leiningen" ];
+then
+  rm -rf leiningen
+fi
+mkdir leiningen
+
+wget https://raw.githubusercontent.com/technomancy/leiningen/stable/bin/lein -O leiningen/lein
+chmod +x leiningen/lein
+
 # Now run our tests
-PUPPETDB_REPO_PUPPETDB="${REPO_URL}#${sha1}" \
+PATH=$PATH:$(pwd)/leiningen \
 bundle exec rake test:beaker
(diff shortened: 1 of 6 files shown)


kbarber:

kbarber:

kbarber:

kbarber:

ext/jenkins/beaker-tests-source.sh
```diff
@@ -32,6 +32,16 @@ mkdir vendor
...
 bundle install --path vendor/bundle --without test
 
+# Install a copy of leiningen
+if [ -d "leiningen" ];
+then
+  rm -rf leiningen
+fi
+mkdir leiningen
+
+wget https://raw.githubusercontent.com/technomancy/leiningen/stable/bin/lein -O leiningen/lein
```
waynr: It is definitely best to pin to a specific version of leiningen; with this line leiningen stable can change at any time regardless of whether it's been tested against puppetdb and therefore may introduce bugs that would break puppetdb's CI pipeline asynchronous to any changes in this repo or our CI system. QE has actually configured jenkins slaves to install a specific version of leiningen to address this kind of problem. I believe the current version of leiningen availabe on jenkins slaves is 2.4.3.

In my opinion, this kind of script should not be in project-specific repos exactly because of snippets like this--as a QE automation engineer I can't reasonably be expected to audit all projects for quirks like this which may blow up at any time. I realize that addressing this issue is well outside the scope of the current PR or any currently ticketed work. Just want to begin socializing some of what I perceive to be the negative impact of storing CI scripts outside of CI.

kbarber: Fair points, the reason why we had this script in the first place was the differences between stable & master and leiningen requirements now - I guess I overlooked the fact the slaves already have a working leiningen today. But I can adjust this, shouldn't be a problem.

kbarber:
Closing this PR for now, as I plan on making changes and don't want to create testing noise here. I'll re-open it later when its ready.

PR #1192, opened 2014-12-26
Firewall doc

rnelson0:
(PDB-137) Document firewall and selinux requirements for PuppetDB.

### DIFF
--- a/documentation/connect_puppet_master.markdown
+++ b/documentation/connect_puppet_master.markdown
@@ -118,7 +118,16 @@ Create it if necessary, and add the following:
         terminus: puppetdb
         cache: yaml
 
-## Step 3: Restart Puppet Master
+## Step 3: Set Security Policy
+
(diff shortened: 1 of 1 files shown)


documentation/connect_puppet_master.markdown
```diff
@@ -118,7 +118,17 @@ Create it if necessary, and add the following:
         terminus: puppetdb
         cache: yaml
 
-## Step 3: Restart Puppet Master
+## Step 3: Set Security Policy
+
+PuppetDB listens on tcp ports 8080 (http) and 8081 (https). Ensure these ports
```
kbarber: We don't support talking to HTTP from the master, so this isn't entirely true or necessary to open up 8080. The only port that requires public access is generally 8081, unless someone is doing something special.

rnelson0:
Removed port 8080 reference.

PR #1193, opened 2014-12-26
(PDB-1072) Add endpoint to mirror report command structure.

wkalt:
This commit adds an endpoint called event-reports, the return value of which
can be easily munged for resubmission as a report. In an HA context, this will
allow us to transfer reports from one database to another over HTTP without
needing a second query for the events associated with a report.


src/puppetlabs/puppetdb/http/event_reports.clj
```diff
@@ -0,0 +1,37 @@
...
+  (:require [puppetlabs.puppetdb.http.query :as http-q]
+            [puppetlabs.puppetdb.query.paging :as paging]
+            [puppetlabs.puppetdb.query.factsets :as fs]
+            [puppetlabs.puppetdb.query-eng :refer [produce-streaming-body]]
+            [puppetlabs.puppetdb.query :as query]
+            [net.cgrand.moustache :refer [app]]
+            [puppetlabs.puppetdb.middleware :refer [verify-accepts-json validate-query-params
+                                                    wrap-with-paging-options]]
+            [puppetlabs.puppetdb.jdbc :as jdbc]
+            [puppetlabs.puppetdb.http :as http]))
```
ajroetker: Are all these dependencies necessary?

kbarber:
So here is the thing, this new endpoint looks almost exactly like /reports ... but without the downstream information like events ... I'm wondering if we should just call this /reports instead, and let people use the extract field selector capability if they want the old report format? That would save us from having to create a brand new end-point.

wkalt:
that's fine with me in concept, but we'll have to extend extract to cover non-queryable fields, since the event-reports field of this endpoint is not queryable. Not sure what that will take; I'll give it some thought.

wkalt:
actually, that's not going to be an issue in the slightest. Happy to take this route.

wkalt:

kbarber:

wkalt:

wkalt:

wkalt:

wkalt:

test/puppetlabs/puppetdb/http/reports_test.clj
```diff
@@ -45,20 +46,18 @@
...
 
 (defn reports-response
   [version reports]
-  (set (map (case version
-              :v3 (comp #(remove-status % :v3) #(remove-environment % :v3) report-response)
-              report-response) reports)))
+  (set (map report-response reports)))
 
-(defn remove-receive-times
+(defn treat-reports-response
```
senior: Can you come up with a more descriptive function name here? Maybe munge-for-compare, or prep-for-compare or something like that?

test/puppetlabs/puppetdb/query/reports_test.clj
```diff
@@ -2,13 +2,22 @@
...
             [puppetlabs.puppetdb.testutils.reports :refer :all]
             [puppetlabs.puppetdb.time :refer [to-secs]]
+            [clj-time.coerce :refer  [to-string]]
             [clj-time.core :refer [now ago days]]))
 
 (use-fixtures :each with-test-db)
 
+(defn reports-equal
+  [r1 r2]
+  (let [norm (fn [x] (sort-by :hash (map #(update-in % [:resource-events] munge-resource-events) x)))]
```
senior: Have you seen this fail? I think it will be opaque in that it will just say reports-equal is false. To diagnose the failure you'll have to go in and change the test to give more information (i.e. println here or something). I think if you take your anon-function above, give it a name and then use it directly below like:

```
(is (= (order-events expected) (order-events actual)))
```

You'll see what failed

test/puppetlabs/puppetdb/query/reports_test.clj
```diff
@@ -142,4 +151,85 @@
                   actual   (reports-query-result :v4
                                                  ["=" "certname" "foo.local"]
                                                  {:order-by [[:report-format order]] :offset offset})]
-              (is (= actual expected)))))))))
+              (is (reports-equal actual expected)))))))))
+
+
+
+(def data-seq (-> (slurp "./test-resources/puppetlabs/puppetdb/cli/export/reports-query-rows.json")
+                  (json/parse-string)
```
senior: no need for parens here and below

wkalt:

test/puppetlabs/puppetdb/query/reports_test.clj
wkalt: @senior fixed your comments -- not sure why this one isn't updating

senior:
+1 from me

PR #1194, opened 2014-12-27
(maint) Allow factset ordering on hash

wkalt:
This corrects an oversight in adding the hash field to factsets that left
ordering results on hash impossible. I'm not a fan of the solution here, but it
follows our approach for environments and I can't think of a better way ATM.
The constraints are:

a) The factsets query must be "doubled up" in jdbc.clj so that paging options are applied
on an object basis instead of a by-row basis (a factset comprises multiple rows.)

b) Both environment and hash are nullable fields in factsets, so a COALESCE is
necessary to prevent a "WHERE null IN null" scenario.

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -139,7 +139,8 @@
   (case entity
     :factsets
     (-> clause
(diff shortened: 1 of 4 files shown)


senior:

PR #1196, opened 2015-01-03
(maint) use latest lein

wkalt:
This will hopefully fix the acceptance test failures that cropped up today in relation to OPS-5175


wkalt:

PR #1197, opened 2015-01-03
Apply the TK hook config to the test instances too

rbrw:
Previously the jetty testutils service instances' configs were only
adjusted by process-config!, but the actual server's config was also
adjusted by hook-tk-parse-config-data.  Change it so that
adjust-tk-config is applied in both cases.

Doing so exposes a bug -- a [global] url-prefix setting with a relative
path like "foo" will prevent the server from responding on /foo.  That's
because add-web-routing-config (via the tk hook) wasn't normalizing the
url-prefix before stashing it in in the web-router-service config.  To
fix that, normalize the url-prefix in add-web-routing-config.

### DIFF
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -356,23 +356,26 @@
 
 (defn add-web-routing-config
   [config-data]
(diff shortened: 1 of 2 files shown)


ajroetker:

PR #1198, opened 2015-01-05
(PDB-1070) Deprecate url-prefix in global

ajroetker:
This commit deprecates the use of the `url-prefix` config setting under
the `[global]` header, favoring the use of the webrouting service to
configure the url-prefix. This commit uses the WebroutingService
get-route function to retrieve the url-prefix from the config data,
using the hook in config.clj to add a default or copy the config from
global url-prefix to the web-routing config.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -180,6 +180,8 @@ When this is set to true, debugging information will be written to `<vardir>/deb
 
 ### `url-prefix`
 
+> **Deprecated:** This setting will be removed in the future and url-prefixes should be configured using the `web-router-service`.
+
 This optional setting may be used to mount the PuppetDB web application at a URL other than "/".  This should not be necessary
 unless you intend to run additional web applications in the same server with your PuppetDB instance.  **NOTE:** if you change
 this setting, you must also set the corresponding setting in your Puppet Master's [puppetdb.conf][puppetdb.conf] file.
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -180,7 +180,7 @@ When this is set to true, debugging information will be written to `<vardir>/deb
 
 ### `url-prefix`
 
-> **Deprecated:** This setting will be removed in the future and url-prefixes should be configured using the `web-router-service`.
+> **Deprecated:** This setting will be removed in the future and url-prefixes should be configured using the [`web-router-service`](#:web-router-service).
 
 This optional setting may be used to mount the PuppetDB web application at a URL other than "/".  This should not be necessary
 unless you intend to run additional web applications in the same server with your PuppetDB instance.  **NOTE:** if you change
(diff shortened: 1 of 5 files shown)


test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -17,11 +17,12 @@
   "Creates a default config, populated with a temporary vardir and
    a fresh hypersql instance"
   []
-  {:repl {}
+  {:nrepl {}
```
ajroetker: I changed this because with the `adjust-tk-config` now being used in the tests we see a the deprecation warning in our tests.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -356,19 +361,22 @@
...
   [config-data]
-  (let [prefix (normalize-url-prefix (get-in config-data [:global :url-prefix]
-                                             ""))]
-    (assoc-in config-data [:web-router-service
-                           :puppetlabs.puppetdb.cli.services/puppetdb-service] prefix)))
+  (let [url-prefix-path [:web-router-service :puppetlabs.puppetdb.cli.services/puppetdb-service]
+        prefix (or (get-in config-data url-prefix-path)
+                   ;; Setting url-prefix in [global] is deprecated
+                   (normalize-url-prefix (get-in config-data [:global :url-prefix]))
+                   "")]
```
ajroetker: @rbrw I know we discussed the fact that the two config options needed different defaults but I think the the one we we're using for the web-routing config `"/"` was just wrong after reading the tk docs. `"/"` has the desired behavior (mounting the app with no prefix) for trapperkeeper but not for `compojure/context` which gives the app the prefix `/`. If you're mounting the app without a prefix you probably won't be using the option at all, while if you are mounting with a prefix you'll probably be using something more than "/" such as "/pdb".

documentation/configure.markdown
```diff
@@ -180,6 +180,8 @@ When this is set to true, debugging information will be written to `<vardir>/deb
 
 ### `url-prefix`
 
+> **Deprecated:** This setting will be removed in the future and url-prefixes should be configured using the `web-router-service`.
```
senior: I think we need to document the web-router-service, I don't think we do that currently

ajroetker: Should I link to the docs in https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/webrouting-service.md or should we have an example in our docs? Also, should we have that example be in hocon or ini?

senior: In other cases we duplicate the docs from there. Unfortunately TK docs aren't currently being published to the docs site (is that still the case @cprice404 ?). I think web routing stuff is painful to write in ini, is that right? The rest of the docs are in ini, probably makes sense to keep it that way unless we switch all of it over.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -338,6 +336,13 @@
         (println "`ssl-protocols` contains SSLv3, a protocol with known vulnerabilities and should be removed from the `ssl-protocols` list"))))
   config-data)
 
+(defn warn-url-prefix-deprecation
```
senior: I think you can use [this](https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/utils.clj#L31) instead

test/puppetlabs/puppetdb/testutils/jetty.clj
senior: Good catch

senior:
Looks good to me, just minor feedback on the docs and println-err

documentation/configure.markdown
cprice404: still the case; will chat with @nickf about whether we have a plan for that.

nfagerlund: Not to be confused with! (Although I got the note in hipchat.)

cprice404: oops, sorry about that.

ajroetker: Anything but the simplest configuration of the web-router-service is hard to do painful because it's nested https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/webrouting-config.md. I doubt anyone will do anything more complicated than the simplest configuration atm since there isn't really a need, but it seems like our plans for the web-router-service, mounting metrics next to pdb next to "foo", would require some of that fine tuning. All this to say, I think it should probably be done in hocon if we think that's ok.

Is the work to split out metrics etc. to their own services planned for 3.0.0?

rbrw:
Is it OK that http/server.clj still refers to [:globals :url-prefix]?

ajroetker:
@rbrw thats a config map that gets put together in `start-puppetdb`

ajroetker:
@senior Added the docs for a web-router-config section!

PR #1199, opened 2015-01-05
(PDB-1052) fix puppet 4 directory environment failures

wkalt:
In Puppet 4, directory environments are enabled by default and any environment
specified in a config must have a corresponding directory in the environmentpath.

This patch brings us into accordance so spec tests can pass on Puppet master.

### DIFF
--- a/puppet/lib/puppet/util/puppetdb.rb
+++ b/puppet/lib/puppet/util/puppetdb.rb
@@ -32,6 +32,15 @@ module Puppet::Util::Puppetdb
     defined?(Puppet::Parser::AST::HashOrArrayAccess)
   end
 
+  def self.create_environmentdir(environment)
+    if not puppet3compat?
+      envdir = File.join(Puppet[:environmentpath], environment)
+      if not Dir.exists?(envdir)
+        Dir.mkdir(envdir)
+      end
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


wkalt:

puppet/lib/puppet/util/puppetdb.rb
```diff
@@ -32,6 +32,15 @@ def self.puppet3compat?
     defined?(Puppet::Parser::AST::HashOrArrayAccess)
   end
 
+  def self.create_environmentdir(environment)
```
kbarber: Since this is test only, it shouldn't live in real code.

puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
```diff
@@ -11,6 +11,7 @@
   before :each do
     Puppet::Util::Puppetdb.stubs(:server).returns 'localhost'
     Puppet::Util::Puppetdb.stubs(:port).returns 0
+    Puppet::Util::Puppetdb.create_environmentdir("my_environment")
```
kbarber: I'm curious, in the real world with Puppet 4, if someone doesn't have a directory ... lets say they haven't created it yet. How does Puppet and our code behave? Is this scenario possible, like if I just use environment=foo in my puppet.conf.

PR #1200, opened 2015-01-06
(PDB-1052) fix spec tests on puppet master

wkalt:
In Puppet 4 directory environments are enabled by default, which means that
environments referenced in a config file must have a corresponding directory
in the environment path or an error will be thrown. This patch creates a
my_environment directory for facts and catalogs to bring us into accordance.

### DIFF
--- a/puppet/spec/spec_helper.rb
+++ b/puppet/spec/spec_helper.rb
@@ -10,9 +10,17 @@ require 'puppetlabs_spec_helper/puppet_spec_helper'
 require 'tmpdir'
 require 'fileutils'
 require 'puppet'
+require 'puppet/util/puppetdb'
 require 'puppet/util/log'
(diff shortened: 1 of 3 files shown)


puppet/spec/spec_helper.rb
```diff
@@ -10,9 +10,17 @@
...
 require 'fileutils'
 require 'puppet'
+require 'puppet/util/puppetdb'
 require 'puppet/util/log'
-require 'puppet/util/puppetdb/command'
 
+def create_environmentdir(environment)
+  if not Puppet::Util::Puppetdb.puppet3compat?
+    envdir = File.join(Puppet[:environmentpath], environment)
+    if not Dir.exists?(envdir)
```
senior: Should we be using exist? instead of exists? here? It looks like exists? is deprecated

PR #1203, opened 2015-01-07
(EZ-15) Update project.clj with lein-ezbake snippets.

waynr:
@kbarber 
Hey thanks for setting up those jobs. I've modified the uberbuild job to use `lein-ezbake` and kicked off a run of that pipeline using my fork of puppetdb. I'll probably go to sleep before the acceptance matrix finishes but here it is: 

https://jenkins.puppetlabs.com/view/PuppetDB/view/lein-ezbake/job/platform_puppetdb_intn-sys_lein-ezbake/2/

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -2,6 +2,8 @@
 (use '[clojure.java.shell :only (sh)]
      '[clojure.java.io :only (file)])
 
+(def pdb-version "3.0.0-SNAPSHOT")
+
 (defn deploy-info
   "Generate deployment information from the URL supplied and the username and
    password for Nexus supplied as environment variables."
@@ -15,7 +17,7 @@
 (def tk-jetty9-version "0.9.0")
 (def ks-version "0.7.2")
 
-(defproject puppetlabs/puppetdb "3.0.0-SNAPSHOT"
+(defproject puppetlabs/puppetdb pdb-version
   :description "Puppet-integrated catalog and fact storage"
 
   :license {:name "Apache License, Version 2.0"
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -96,6 +98,15 @@
   :lein-release {:scm        :git
                  :deploy-via :lein-deploy}
 
+  :uberjar-name "puppetdb-release.jar"
+  :lein-ezbake {:vars {:user "puppetdb"
+                       :group "puppetdb"
+                       :build-type "foss"
```
waynr: I noticed that the only difference between `puppetdb` and `pe-puppetdb` is that `:build-type` is `pe` and the user/group names have the `pe-` prefix. I'm thinking about ways to accomodate this that aren't totally janky:

* We could take arbitrary key-value pairs on at the end of the `lein ezbake <action>` command line that would shadow the values set in this group of vars. 
* We could use `pe` and `foss` leiningen profiles that contain different `:lein-ezbake` values. Then the command to build pe-puppetdb would be `lein with-profile ezbake,pe ezbake build` as opposed to `lein with-profile ezbake ezbake build`.

kbarber: I think I like approach #2 ... keeps it altogether I guess.

So to let you know, pe-puppetdb doesn't 'work' today, we have another ticket tracking the effort to get that working: https://tickets.puppetlabs.com/browse/PDB-1032 so we could punt this until later, as long as we think we have a way that won't dramatically change the lein-ezbake plugin API too much, and its in our vision this is probably okay for now.

PR #1206, opened 2015-01-08
(PDB-1068) Recognize prefix urls in import export

rbrw:
Note that there are 4 commits, which might or might not be easier to review individually, and I wondered about defaulting the base-url :protocol to "http".  Right now it has to be specified every time.

(This series is based on the PDB-1070 PR and so may need to be rebased once that goes in.)

### DIFF
--- a/src/puppetlabs/puppetdb/cli/benchmark.clj
+++ b/src/puppetlabs/puppetdb/cli/benchmark.clj
@@ -199,7 +199,8 @@
   * Submit the resulting catalog"
   [{:keys [host lastrun catalog report factset puppetdb-host puppetdb-port run-interval rand-percentage] :as state} clock]
   (if (> (- clock lastrun) run-interval)
-    (let [catalog (if catalog (maybe-tweak-catalog rand-percentage catalog))
+    (let [base-url {:protocol "http" :host puppetdb-host :port puppetdb-port}
+          catalog (when catalog (maybe-tweak-catalog rand-percentage catalog))
           report (and report (update-report-run-fields report))
           factset (and factset (update-factset rand-percentage factset))]
       ;; Submit the catalog and reports in separate threads, so as to not
@@ -207,21 +208,21 @@
       (when catalog
         (future
           (try
-            (client/submit-catalog puppetdb-host puppetdb-port 5 (json/generate-string catalog))
+            (client/submit-catalog base-url 5 (json/generate-string catalog))
             (log/info (format "[%s] submitted catalog" host))
             (catch Exception e
               (log/error (format "[%s] failed to submit catalog: %s" host e))))))
       (when report
         (future
           (try
-            (client/submit-report puppetdb-host puppetdb-port 3 (json/generate-string report))
+            (client/submit-report base-url 3 (json/generate-string report))
             (log/info (format "[%s] submitted report" host))
             (catch Exception e
               (log/error (format "[%s] failed to submit report: %s" host e))))))
       (when factset
         (future
           (try
-            (client/submit-facts puppetdb-host puppetdb-port 3 (json/generate-string factset))
+            (client/submit-facts base-url 3 (json/generate-string factset))
             (log/info (format "[%s] submitted factset" host))
             (catch Exception e
               (log/error (format "[%s] failed to submit factset: %s" host e))))))
@@ -236,15 +237,16 @@
    similar to timed-update-host, but always sends the update (doesn't run/skip
    based on the clock)"
   [{:keys [host lastrun catalog report factset puppetdb-host puppetdb-port run-interval rand-percentage] :as state}]
-  (let [catalog (and catalog (maybe-tweak-catalog rand-percentage catalog))
+  (let [base-url {:protocol "http" :host puppetdb-host :port puppetdb-port}
+        catalog (and catalog (maybe-tweak-catalog rand-percentage catalog))
         report (and report (update-report-run-fields report))
         factset (and factset (update-factset rand-percentage factset))]
     (when catalog
-      (client/submit-catalog puppetdb-host puppetdb-port 5 (json/generate-string catalog)))
+      (client/submit-catalog base-url 5 (json/generate-string catalog)))
     (when report
-      (client/submit-report puppetdb-host puppetdb-port 3 (json/generate-string report)))
+      (client/submit-report base-url 3 (json/generate-string report)))
     (when factset
-      (client/submit-facts puppetdb-host puppetdb-port 3 (json/generate-string factset)))
+      (client/submit-facts base-url 3 (json/generate-string factset)))
     (assoc state :catalog catalog)))
 
 (defn submit-n-messages
--- a/src/puppetlabs/puppetdb/cli/export.clj
+++ b/src/puppetlabs/puppetdb/cli/export.clj
@@ -13,13 +13,15 @@
             [clojure.java.io :as io]
             [clj-http.client :as client]
             [puppetlabs.puppetdb.archive :as archive]
-            [puppetlabs.puppetdb.schema :as pls]
+            [puppetlabs.puppetdb.schema :refer [defn-validated]]
             [schema.core :as s]
             [clojure.string :as str]
             [puppetlabs.puppetdb.utils :as utils]
-            [slingshot.slingshot :refer [try+]]
+            [slingshot.slingshot :refer [throw+ try+]]
             [clj-time.core :refer [now]]
             [clj-http.util :refer [url-encode]]))
+
+(def ^:private api-version :v4)
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Internal Schemas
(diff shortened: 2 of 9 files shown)


senior:
@rbrw Now that @ajroetker PR has been merged, this needs a rebase

wkalt:

rbrw:
Rebased to master.

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -229,15 +210,25 @@
...
 
-(defn -main
+(defn- main
   [& args]
-  (let [[{:keys [outfile host port]} _] (validate-cli! args)
-        nodes (get-nodes host port)]
+  (let [[{:keys [outfile host port url-prefix] :as opts} _] (validate-cli! args)
+        src {:protocol "http" :host host :port port :prefix url-prefix}
+        _ (if (utils/report-bad-base-url src "Invalid source")
+            (throw+ {:type :exit-process :status 1}))
```
senior: I think exit-process should be something different from :type. We usually use type to indicate the type of error that occurred, rather than what do do about the error. So in this case maybe invalid-source is the error type? I also prefer the namespace keywords in these potentially global maps, so we know where it came from, i.e. ::invalid-source.

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -229,15 +210,25 @@
...
          :puppetlabs.kitchensink.core/cli-help  (System/exit 0))))))
 
-(defn -main
+(defn- main
   [& args]
-  (let [[{:keys [outfile host port]} _] (validate-cli! args)
-        nodes (get-nodes host port)]
+  (let [[{:keys [outfile host port url-prefix] :as opts} _] (validate-cli! args)
+        src {:protocol "http" :host host :port port :prefix url-prefix}
+        _ (if (utils/report-bad-base-url src "Invalid source")
```
senior: Convention here is when

src/puppetlabs/puppetdb/client.clj
```diff
@@ -6,78 +6,80 @@
...
+(defn-validated submit-command-via-http!
   "Submits `payload` as a valid command of type `command` and
   `version` to the PuppetDB instance specified by `host` and
   `port`. The `payload` will be converted to JSON before
   submission. Alternately accepts a command-map object (such as those
   returned by `parse-command`). Returns the server response."
-  ([host port command version payload]
-     {:pre [(string? command)
-            (integer? version)]}
+  ([base-url
```
senior: Might be useful to distinguish between "version" below, which is command version and version in base-url. Maybe just rename version to command-version?

src/puppetlabs/puppetdb/client.clj
```diff
@@ -6,78 +6,80 @@
...
      (->> payload
           (command/assemble-command command version)
-          (submit-command-via-http! host port)))
-  ([host port command-map]
-     {:pre [(string? host)
-            (integer? port)
-            (map? command-map)]}
+          (submit-command-via-http! base-url)))
+  ([base-url :- (s/both utils/base-url-schema
+                        (s/pred #(if-let [v (:version %)]
```
senior: Seems like if you moved this pred over to utils as a def, you could name it the same as the description you have below, or something similar and it becomes more intention revealing and more concise in the function definition

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -134,3 +134,36 @@
...
+   (s/optional-key :version) (s/both
+                              s/Keyword
+                              (s/pred #(and (keyword? %)
+                                            (re-matches #"v\d+" (name %)))
+                                      'valid-version?))})
+
+(pls/defn-validated base-url->str :- s/Str
+  "Converts the `base-url' map to an ASCII URL.  May throw
+   MalformedURLException or URISyntaxException."
+  [base-url :- base-url-schema]
```
senior: If you destructure base-url here, you'll save yourself some typing below

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -134,3 +134,36 @@
...
+  (-> (URL. (:protocol base-url) (:host base-url) (:port base-url)
+            (str (when-let [p (:prefix base-url)]
+                   (str "/" p))
+                 "/" (name (or (:version base-url) :v4))))
+    .toURI .toASCIIString))
```
senior: Why create a URL object here?

test/puppetlabs/puppetdb/cli/import_export_roundtrip_test.clj
```diff
@@ -153,7 +179,32 @@
...
+       (is (thrown-with-msg?
+            java.util.concurrent.ExecutionException #"Results not found"
+            @(block-until-results 5
+                                  (json/parse-string
+                                   (export/catalog-for-node *base-url*
+                                                            "foo.local")))))))))
+
+(defn- check-invalid-url-handling [cmd expected-msg-re]
+  (is (re-find expected-msg-re
+               (with-out-str
```
senior: I think [this](https://github.com/puppetlabs/puppetdb/blob/master/test/puppetlabs/puppetdb/testutils.clj#L291) would simplify the code here.

test/puppetlabs/puppetdb/cli/import_export_roundtrip_test.clj
```diff
@@ -153,7 +179,32 @@
...
+                 (binding [*err* *out*]
+                   (is (try+
```
senior: I think you may want to use (is (thrown-with-msg+?...)) here as it will assert your specific slingshot error is thrown.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -199,7 +199,8 @@
   * Submit the resulting catalog"
   [{:keys [host lastrun catalog report factset puppetdb-host puppetdb-port run-interval rand-percentage] :as state} clock]
   (if (> (- clock lastrun) run-interval)
-    (let [catalog (if catalog (maybe-tweak-catalog rand-percentage catalog))
+    (let [base-url {:protocol "http" :host puppetdb-host :port puppetdb-port}
+          catalog (if catalog (maybe-tweak-catalog rand-percentage catalog))
```
senior: when

src/puppetlabs/puppetdb/cli/import.clj
```diff
@@ -44,15 +49,15 @@
...
+(defn-validated process-tar-entry
   "Determine the type of an entry from the exported archive, and process it
   accordingly."
-  [^TarGzReader tar-reader ^TarArchiveEntry tar-entry host port metadata]
-  {:pre  [(instance? TarGzReader tar-reader)
-          (instance? TarArchiveEntry tar-entry)
-          (string? host)
-          (integer? port)
-          (map? metadata)]}
+  [^TarGzReader tar-reader :- (s/pred #(instance? TarGzReader %)
```
senior: Not sure if you knew about this schema feature, but you can give it a class here and it will "do the right thing". 
```
tar-reader :- TarGzReader
```
It also gives a slightly better error message since tar-gz-reader? isn't an actual function.

src/puppetlabs/puppetdb/utils.clj
rbrw: It's just a hack for the error checking side-effect -- I may have overlooked it, but I didn't see a way to get URL/URI to do that otherwise.

src/puppetlabs/puppetdb/cli/export.clj
rbrw: Right -- just an oversight.

senior:
This is a good PR, I really like the throw/exit pattern by the way, I think it's better than the thunk passing that I did in the services ns. At first I wasn't sure about the base-url map stuff, but after reading through more of the change, I think it's an improvement over just a bare URL string. We have some utility functions built up around it and it allows us to give better, more specific error messages.

src/puppetlabs/puppetdb/client.clj
rbrw: Sounds reasonable, though I'm wondering how to do it in a composable way that's still compact enough (i.e. probably want to keep using s/both so that if we ever need to check the version and the protocol, we can compose without duplicating the base-url-schema checks...).  In this case, I was just trying to preserve the original semantics where the version had to be v4 if it was specified.

rbrw: As we discussed, I'll change this so that the function just relies on the base-url version.

rbrw:
OK, thanks for all the suggestions.  Hopefully the push above address the issues.  It also adds some minor additional tests, and a hopefully better wrapper for import/export main.

I've rebased the previous push as ticket/master/pdb-1068-recognize-prefix-urls-in-import-export-prev so that you should be able to get a pretty good sense of the changes via

    git diff rbrw/ticket/master/pdb-1068-recognize-prefix-urls-in-import-export-prev..rbrw/ticket/master/pdb-1068-recognize-prefix-urls-in-import-export

if you fetch both branches.  The previous push itself is also still available as

    rbrw/ticket/master/pdb-1068-recognize-prefix-urls-in-import-export-v1

in case the "prev" rebase did something unexpected.

Thanks

senior:
Error messages are much better with this patch, before:
```
$ lein run export -o the-file.tar.gz --host foo:bar
java.lang.NumberFormatException: For input string: "bar:8080"
```
after:
```
lein run export -o the-file.tar.gz --host foo:bar
clojure.lang.ExceptionInfo: Invalid source (Malformed IPv6 address at index 8: http://[foo:bar]:8080//v4)
```
I'm looking into it further, not sure why there's a double slash there if a url-prefix isn't specified.

rbrw:
The double slash was caused by not checking for empty? in base-url->str, and I've fixed the prn.

This series should fix both issues, and the series should be available as https://github.com/rbrw/puppetdb/tree/ticket/master/pdb-1068-recognize-prefix-urls-in-import-export-v2

rbrw:

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -219,7 +198,9 @@
   [args]
   (let [specs    [["-o" "--outfile OUTFILE" "Path to backup file (required)"]
                   ["-H" "--host HOST" "Hostname of PuppetDB server" :default "localhost"]
-                  ["-p" "--port PORT" "Port to connect to PuppetDB server (HTTP protocol only)" :parse-fn #(Integer. %) :default 8080]]
+                  ["-p" "--port PORT" "Port to connect to PuppetDB server (HTTP protocol only)" :parse-fn #(Integer. %) :default 8080]
+                  ["" "--url-prefix PREFIX" "Server prefix (HTTP protocol only)"
```
ajroetker: So I know in config files in PE when we got to the point of having to specify all these things, we decided to just use a single `base-url` item that the user specifies (e.g. `http://localhost:8080/my_prefix`) instead, parsing it into the map when we received it. Might make things cleaner!

rbrw:

rbrw:
The failures seem likely to be transient.

senior:

PR #1207, opened 2015-01-08
(maint) don't change dashed queryable fields to underscores before valid...

wkalt:
...ation

This removes a conversion to underscores before validation of top-level extract fields.
The prexisting conversion had broken querying on dashed fields.


senior:
I hit this when trying to construct a query that extracted producer-timestamp, looks good to me, will merge once the current build is green

PR #1208, opened 2015-01-09
Refactor lazy collapsing of DB results to use common code

senior:
Reports, factsets and catalogs all have similar code, lazily collapsing n DB rows into a single JSON object. This commit refactors that logic to a single place.

I originally did this because I was repeating the logic again. Turned out I didn't need to collapse rows anymore, but the refactor still seemed good, so I'm pushing it up as a maint PR.


wkalt:
+1

PR #1209, opened 2015-01-09
(PDB-800) Move metrics server to its own tk service

ajroetker:
This commit moves the metrics endpoints to their own trapperkeeper
service, using the web-router-service to mount it and the
puppetdb service at the appropriate routes.

Must be merged with https://github.com/puppetlabs/puppet-docs/pull/425.


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -355,6 +356,23 @@
           (get-in (service-context this) [:shared-globals :scf-read-db])
           row-callback-fn)))
 
+(defprotocol MetricsServer)
```
kbarber: Can we break this stuff into its own file somewhere? This would aid with ripping it out later on when we choose to.

kbarber:
This will need some documentation changes, to reflect the fact that the endpoint is now in its own namespace. I would recommend putting it in its own top-level hierachy in docs.puppetlabs.com (this will need puppet-docs changes).

Also, your tests are failing, might want to dig into that.

ajroetker:
So @kbarber under `documentation/api/metrics/v1/index.markdown` is what you mean?

kbarber:
@ajroetker as far as the file, yes. But you also need to update the index as well, which is located elsewhere: https://github.com/puppetlabs/puppet-docs/blob/master/source/_includes/puppetdb_master.html. Give me a call later today, and I'll explain it in more detail if you need it.

test/puppetlabs/puppetdb/http/metrics_test.clj
```diff
@@ -37,22 +33,23 @@
   [req]
   (assoc-in req [:headers "accept"] "text/plain"))
 
-(deftestseq metrics-set-handler
+(let [app (server/build-app)]
+  (deftestseq metrics-set-handler
```
wkalt: this seems strange to me. Any reason not to put the let inside the deftestseq?

ajroetker: This is similar to what we do elsewhere, such as the command_test.clj, but also I just wanted to execute the let once, not each time the deftestseq executes over the list.

wkalt: Okay, I get it. I think the way we're handling it more commonly is with the with-http-app fixture. That seems a bit cleaner to me but might take some modification to get working with the new metrics prefix. I haven't looked too much into it but maybe it'd be a worthwhile update.

ajroetker: I actually had it like that in the original incarnation but decided against it for two reasons:

1. It was verbose, a function and a dynamic variable declaration for use by a single test, with no plans to add more functionality to the app or add more tests, it just seemed unnecessary.

2. We're planning on abstracting the metrics application to its own repo or project(?) so that it can be shared by different applications. So 1. seemed even more applicable.

Totally don't mind changing it, an easy one, but that was my reasoning.

src/puppetlabs/puppetdb/cli/metrics_service.clj
```diff
@@ -0,0 +1,23 @@
+(ns puppetlabs.puppetdb.cli.metrics-service
+  (:require [clojure.tools.logging :as log]
+            [puppetlabs.puppetdb.http.metrics.server :as server]
+            [puppetlabs.trapperkeeper.core :refer [defservice main]]
+            [compojure.core :as compojure]
+            [puppetlabs.puppetdb.cli.services :refer [build-whitelist-authorizer]]))
+
+(defprotocol MetricsServer)
```
senior: Why have an empty protocol?

test/puppetlabs/puppetdb/http/metrics_test.clj
wkalt: I guess that's fine -- I was more wondering if build-app could be folded into the existing function in server.clj to save a function, but if we're splitting metrics keeping them separate may make more sense.

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -191,37 +191,31 @@
...
-                        ;; requested
-                        (doseq [metric-root metric-roots
-                                :let [meter-key [:meters metric-root status]]
-                                :when (not (get-in @storage meter-key))]
-                          (swap! storage assoc-in meter-key (meter [prefix metric-root (str status)] "reqs/s"))
-                          (mark! (get-in @storage meter-key)))
-
-                        response)))))))
+    ;; add metric timers for the uri as we service the request.
+    (let [metric-roots (let [s (normalize-uri (:uri req))]
```
wkalt: I might be miscounting, but I don't think this function is being used anywhere outside of test code after this patch.  Commenting out the line in server/build-app that calls it has no bearing on tests passing. Can wrap-with-metrics and wrap-with-metrics* be removed?

src/puppetlabs/puppetdb/http/metrics/v1.clj
```diff
@@ -0,0 +1,16 @@
...
+            [ring.middleware.params :refer [wrap-params]]))
+
+(def v1-app
+  (moustache/app
+    ["mbeans"]
+    {:get metrics/list-mbeans}
+
+    ["mbean" & names]
+    {:get (moustache/app
+            (metrics/mbean names))}))
```
wkalt: would it make sense to call this http/metrics.clj and use a case statement on version to cut down on future code duplication?

src/puppetlabs/puppetdb/middleware.clj
ajroetker: We still use it here though? https://github.com/ajroetker/puppetdb/blob/master/src/puppetlabs/puppetdb/http/server.clj#L82

wkalt: yeah that's the call I meant, but I take it back -- commenting that out doesn't affect tests but it breaks the dashboard

src/puppetlabs/puppetdb/http/metrics/v1.clj
ajroetker: Yeah we could also just use a case statement and put it in server.clj instead since its a pretty barebones app, i was just following the same convention we're using for the pdb app. Thoughts?

src/puppetlabs/puppetdb/middleware.clj
wkalt: Actually, the dashboard nonfunctional with this patch unchanged. Can you check that it's not?

src/puppetlabs/puppetdb/http/metrics/v1.clj
wkalt: I see where you're coming from. Personally I can't see much reason for those files to be separate either -- wonder what everyone else thinks.

src/puppetlabs/puppetdb/middleware.clj
ajroetker: I can verify that!

src/puppetlabs/puppetdb/cli/metrics_service.clj
ajroetker: Is it not required for the tk service?

senior: If you're only a consumer of other services, no. Here's an example: https://github.com/puppetlabs/puppetdb/blob/stable/src/com/puppetlabs/puppetdb/cli/services.clj#L357

src/puppetlabs/puppetdb/cli/metrics_service.clj
```diff
@@ -0,0 +1,20 @@
+(ns puppetlabs.puppetdb.cli.metrics-service
```
kbarber: So the 'cli' in this namespace is meant to designate its a CLI tool, like service/import/export etc. so I'm not sure this is an appropriate place to put it. I would rather see the service in the com.puppetlabs.http.metrics namespace instead of possible, so all related items are together.

acceptance/tests/security/cert_whitelist.rb
```diff
@@ -24,7 +24,7 @@
                  "--cacert #{ssldir}/certs/ca.pem " +
                  "--cert #{ssldir}/certs/#{dbname}.pem " +
                  "--key #{ssldir}/private_keys/#{dbname}.pem "+
-                 "https://#{dbname}:8081/v3/metrics/mbeans " +
+                 "https://#{dbname}:8081/metrics/v1/mbeans " +
```
kbarber: This is good, but there are two other places in acceptance/helper.rb that will need a similar change also. Grep for 'mbean' (not mbeans) and you'll see what I mean.

src/puppetlabs/puppetdb/http/v4.clj
```diff
@@ -49,15 +48,6 @@
    ["resources" &]
    {:any (resources/resources-app version)}
 
-   ["metrics" &]
-   (moustache/app
-    ["mbeans"]
-    {:get met/list-mbeans}
-
-    ["mbean" & names]
```
kbarber: So here is a thing to consider, there is a ticket open today that expresses the desire to make this all 'mbeans' instead of mbean & mbeans (PDB-200). If its not a hard slog, we could do it now as part of this, but its up to you.

documentation/api/metrics/v1/index.markdown
```diff
@@ -1,15 +1,13 @@
 ---
```
kbarber: Another doc change that I forgot ... should be minor, but the current versioning policy might want to be slightly modified to make sure the metrics versioning is clear: https://docs.puppetlabs.com/puppetdb/2.2/versioning_policy.html ... it should align with the HTTP API policy more or less, so I doubt you need to double up, just mentioning it covers the scope of the metrics stuff also might be enough.

src/puppetlabs/puppetdb/http/metrics/core.clj
```diff
@@ -1,4 +1,4 @@
-(ns puppetlabs.puppetdb.http.metrics
+(ns puppetlabs.puppetdb.http.metrics.core
```
kbarber: I'm almost tempted to say drop the http from the namespace here, and just make it puppetlabs.puppetdb.metrics.* for everything, then put say your service (which is the entry point really) in puppetlabs/puppetdb/metrics.clj ... gives it sort of a nice hierarchy, and since the http namespace has PDB specific stuff in it, kind of separates it from the rest of the query stuff.

kbarber:
@ajroetker all in all looks pretty good mate, a few concerns, putting this back in WIP for now. Mainly the helper.rb thing I wanted to correct since that has now changed. I think the only reason it works now is because its still pointing at v3, but it won't work when we remove that :-). One less thing for @rbrw to worry about I guess.

Also I didn't really like the namespace layout, but thats more cosmetic than a problem per se.

nfagerlund:
+1 to docs changes.

acceptance/tests/security/cert_whitelist.rb
ajroetker: taken care of!

wkalt:
I think these commits should probably get squashed; +1 otherwise

ajroetker:
@wkalt there ye go, also refactored the core_test.clj (formerly metrics_test.clj) take a look!

documentation/api/metrics/v1/index.markdown
ajroetker: Added it to the list we mention in that doc section.

ajroetker:

ajroetker:

resources/public/dashboard/index.html
```diff
@@ -155,7 +155,7 @@
   var height = getParameter("height") || 60;
 
   counterAndSparkline()
-  .url("../v4/metrics/mbean/java.lang:type=Memory")
+  .url("../metrics/v1/mbean/java.lang:type=Memory")
```
kbarber: without even testing it, I'm pretty sure all these urls need to change to 'mbeans'.

kbarber:
Something weird, if I try to look into the old 'http monitoring' jmx end points, when I query /metrics/v1/mbeans I get:

```
  "user:type=/v4,name=200" : "/mbeans/user%3Atype%3D%2Fv4%2Cname%3D200",
  "user:type=/v4,name=service-time" : "/mbeans/user%3Atype%3D%2Fv4%2Cname%3Dservice-time",
  "user:type=/v4/facts,name=200" : "/mbeans/user%3Atype%3D%2Fv4%2Ffacts%2Cname%3D200",
  "user:type=/v4/facts,name=service-time" : "/mbeans/user%3Atype%3D%2Fv4%2Ffacts%2Cname%3Dservice-time"
```

Not the 'user' namespace here? Its changed from having a namespace like puppetlabs.puppetdb.* to just 'user'.

I think it was this commit:

```
commit 56e325c68e6d3ac547206fab7655abc807ac79ff
Author: Andrew Roetker <andrew.roetker@puppetlabs.com>
Date:   Wed Dec 3 14:02:52 2014 -0800

    (PDB-863) Remove com. from jmx and mq endpoints

    This commit removes the com. from the jmx and mq endpoints so that these
    endpoints use the new puppetlabs.puppetdb namespace. In order to change
    all the outstanding references to the old namespace, this commit also
    updates the dashboard index.html to reference the v4 metrics api.
```

Can you see a quick fix, or should we raise a bug separately?

Otherwise, the patch you have here needs a dashboard update to use 'mbeans' instead of 'mbean' I think.

ajroetker:
Fixed up both the issues @kbarber looks like the part where we removed the `prefix# ~(str *ns*)` was necessary after all, I blame my ignorance to clojure macros.

ajroetker:

ajroetker:
Free money for anyone who merges this now!

PR #1211, opened 2015-01-13
(PDB-916) Only support puppet 3.7.3+ and 4.y.z releases

senior:
Puppet 5.0.0 will likely include breaking changes as it is a major
version boundry. Putting this metadata in the package will prevent users
from accidentally having that broken setup.

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -4,9 +4,9 @@ ezbake: {
     debian: { dependencies: ["pe-puppet (>= 4.0.0-1puppetlabs1)"] }
   }
   foss: {
-    redhat: { dependencies: ["puppet >= 3.7.3"],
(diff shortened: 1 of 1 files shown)


resources/ext/ezbake.conf
```diff
@@ -4,9 +4,9 @@ ezbake: {
     debian: { dependencies: ["pe-puppet (>= 4.0.0-1puppetlabs1)"] }
```
senior: @kbarber @haus the pe-puppet changes above are the ones I'm not sure about, seems like we should just leave it for now?

kbarber: Its probably going to change at some point anyway with AIO work. Today we pin everything within PE anyway, so this kind of thing matters very little, since the installer does the dependency handling outside of the packaging to some degree. I wouldn't stress that much about it, this will all be resolved when we solve the PE build problems for ezbake, and probably change for AIO anyway.

PR #1216, opened 2015-01-14
(bugfix) Only test hyphenated classnames with Puppet 3

ajroetker:
Hyphens are not allowed in classnames in Puppet 4, so we should only
test that they work in Puppet 3. This commit makes this so.


kbarber:
@ajroetker this needs to be rebased against stable. Closing it, so you can re-open it against stable.

PR #1217, opened 2015-01-14
(PDB-693) Remove previously deprecated 'event-query-limit' setting

LaMetterey:
Removed documentation for this setting, and code to check/warn about it's deprecation.

### DIFF
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -317,14 +317,11 @@
 (defn configure-globals
   "Configures the global properties from the user defined config"
   [{:keys [global] :as config}]
-  (let [product-name (normalize-product-name (get global :product-name "puppetdb"))]
-    (when (:event-query-limit global)
-      (log/warn "The configuration item `event-query-limit` in the [global] section is deprecated and now ignored. It will be removed in the future."))
-    (update-in config [:global]
-               (fn [global-config]
-                 (-> global-config
-                     (assoc :product-name product-name)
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/config.clj
```diff
@@ -318,8 +318,6 @@
   "Configures the global properties from the user defined config"
   [{:keys [global] :as config}]
   (let [product-name (normalize-product-name (get global :product-name "puppetdb"))]
-    (when (:event-query-limit global)
```
ajroetker: Since global isn't used by multiple things now it would be more concise to make the params simply `config` and the above `get` a `get-in config [:global :product-name]`

ajroetker: Actually configure-globals could be even simpler
```
(defn configure-globals
  [{:keys [global] :as config}]
  (assoc config :global (-> global
                                         (utils/assoc-when :product-name "puppetdb")
                                         (update-in [:product-name] normalize-product-name)
                                         (utils/assoc-when :update-server "http://updates.puppetlabs.com/check-for-updates"))
```

kbarber:
Changing to WIP, due to pending review items.

src/puppetlabs/puppetdb/config.clj
LaMetterey: Excellent point - sorry it took so long for me to get around to making such a trivial change.

LaMetterey:

senior:
+1 will merge when the the build pipeline is clean

PR #1218, opened 2015-01-15
(PDB-477) Setup access-logging by default

ajroetker:
This commit add the necessary logback config and commented out param in
the jetty.ini to easily enable access-logging for PuppetDB.


senior:
Looks like legit failures, you've added a new file that needs to be packaged. I think all you need to do is add something [here](https://github.com/puppetlabs/puppetdb/blob/master/ext/redhat/puppetdb.spec.erb) and [here](https://github.com/puppetlabs/puppetdb/blob/master/ext/templates/deb/terminus.install.erb)

ajroetker:
Tested the config locally and sets up access-logging!

PR #1220, opened 2015-01-16
WIP: (PDB-1034) Ezbake source based testing

kbarber:
methodology can be retired.


kbarber:
Still technically wip, since I'm waiting for the other PR's to get reviewed/merged, and I'll also need to unpin things etc.

PR #1221, opened 2015-01-19
(PDB-1125) Fetch structured facts from cache instead of stringified

dalen:
Use the v4 API to fetch a factset instead of fetching facts using the v3
API.


kbarber:
The rspec tests here are failing @dalen.

I need to ponder this one also, we had a specific reason for holding off on structured facts for this, I believe it had a lot to do with inventory search from the PE console which was querying PuppetDB via the Puppet master facts search interface. Having said that, I believe this problem is gone to some extent ... I'm just wary I've forgotten something, especially considering this is still potentially a path that is still available with 3.x puppet masters. I could be wrong. Fix the tests and ping me again :-).

kbarber:
Hey @dalen re-open a PR for this with the fixes etc. when you're ready.

PR #1222, opened 2015-01-20
(PDB-959) cluster resource events table on timestamp

wkalt:
This commit introduces a migration to CLUSTER the resource_events table on
idx_resource_events_timestamp. CS has found this necessary in a couple cases
for getting performant queries from the PE event inspector.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -891,6 +891,14 @@
     "ALTER TABLE factsets ADD hash VARCHAR(40)"
     "ALTER TABLE factsets ADD CONSTRAINT factsets_hash_key UNIQUE (hash)"))
 
+(defn cluster-resource-events
+  "CLUSTER resource_events table if using postgres"
+  []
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -891,6 +891,14 @@
...
     "ALTER TABLE factsets ADD CONSTRAINT factsets_hash_key UNIQUE (hash)"))
 
+(defn cluster-resource-events
+  "CLUSTER resource_events table if using postgres"
+  []
+  (when (scf-utils/postgres?)
+    (sql/do-commands
+      "CLUSTER resource_events USING idx_resource_events_timestamp"
+      "ANALYZE")))
+
```
wkalt: putting this up now for discussion. I still want to test the change on a large data set, which I should be able to generate from the one pending from OPS-5140. This has the potential to slow down migrations; I'm not sure by how much. We could also consider moving the analyze into a final post-migration step, but for now I'm leaving as is so it's not forced on each incremental migration.

PR #1223, opened 2015-01-21
(PDB-1067) Warn user with unknown config items

ajroetker:
This commit changes the behavior of the config validation by warning
users in the log when there is an unused config item in the config file
instead of failing hard. This commit also refactors the items in
config.clj so that changes like this are a little easier in the future
by adding a `configure-section` function which removed the duplication
with `configure-puppetdb`, `configure-write-db`, etc.

### DIFF
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -25,34 +25,53 @@
 ;; support defaulting/converting nested maps, these configs can be put
 ;; together in a single schema that defines the config for PuppetDB
 
+(defn warn-unknown-keys
+  [schema data]
+  (doseq [k (pls/unknown-keys schema data)]
+    (log/warn (format "The configuration item `%s` does not exist and should be removed from the config." k))))
+
+(defn warn-and-validate
+  "Warns a user about unknown configurations items, removes them and validates the config."
+  [schema data]
+  (warn-unknown-keys schema data)
+  (->> (pls/strip-unknown-keys schema data)
+       (s/validate schema)))
+
+(defn all-optional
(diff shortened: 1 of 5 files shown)


ajroetker:
The acceptance test failure was transient and the unit test failure should be fixed! Also added some new tests.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -25,34 +25,54 @@
...
+  (doseq [k (pls/unknown-keys schema data)]
+    (log/warn (format "The configuration item `%s` does not exist and should be removed from the config." k))))
+
+(defn warn-and-validate
+  "Warns a user about unknown configurations items, removes them and validates the config."
+  [schema data]
+  (warn-unknown-keys schema data)
+  (let [sanitized-data (pls/strip-unknown-keys schema data)]
+    (s/validate schema sanitized-data)
+    sanitized-data))
```
wkalt: I think this line can be removed -- s/validate will return sanitized-data if it matches the schema.

ajroetker:

test/puppetlabs/puppetdb/schema_test.clj
```diff
@@ -45,6 +45,32 @@
        :foo :foo
        :foo (s/required-key :foo)))
 
+(deftest unknown-keys-test
+  (let [schema {(s/required-key :foo) Number
+                (s/optional-key :bar) Number
+                :baz String}]
+    (is (= '(:foo-1)
```
senior: These can just be [:foo-1], they will still compare as equal

senior:
This is a good change, +1 from me, will merge once the build pipeline is clean

PR #1225, opened 2015-01-23
(maint) fix puppetdb-env.erb so it generates valid bash

wkalt:
There is a bug affecting some OSes where this template will miss a necessary newline in the bash it generates.  An example is here:

https://gist.github.com/wkalt/fe8949ea40eb95c3aa9b

note that JAVA_BIN and INSTALL_DIR are on the same line

### DIFF
--- a/ext/templates/puppetdb-env.erb
+++ b/ext/templates/puppetdb-env.erb
@@ -10,9 +10,9 @@ elif [ `uname` == "OpenBSD" ] ; then
     USER="_puppetdb"
     INSTALL_DIR=<%= @install_dir %>
 else
-    JAVA_BIN=<%= @java_bin || "/usr/bin/java"  -%>
-    INSTALL_DIR="<%= @install_dir || "/usr/share/puppetdb" -%>"
-    JAVA_ARGS="<%= @java_args || @default_java_args -%>"
-    USER="<%= @name -%>"
-    CONFIG="<%= @config_dir -%>"
+    JAVA_BIN=<%= @java_bin || "/usr/bin/java"  %>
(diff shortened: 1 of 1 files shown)


ajroetker:
I don't think we use that file in our packaging any more?

wkalt:
@ajroetker yeah affects stable. A customer raised it on PE 3.3.1.

ext/templates/puppetdb-env.erb
```diff
@@ -10,9 +10,9 @@ elif [ `uname` == "OpenBSD" ] ; then
...
     INSTALL_DIR=<%= @install_dir %>
 else
-    JAVA_BIN=<%= @java_bin || "/usr/bin/java"  -%>
-    INSTALL_DIR="<%= @install_dir || "/usr/share/puppetdb" -%>"
-    JAVA_ARGS="<%= @java_args || @default_java_args -%>"
-    USER="<%= @name -%>"
-    CONFIG="<%= @config_dir -%>"
+    JAVA_BIN=<%= @java_bin || "/usr/bin/java"  %>
+    INSTALL_DIR=<%= @install_dir || "/usr/share/puppetdb" %>
+    JAVA_ARGS=<%= @java_args || @default_java_args %>
```
kbarber: Hrm ... not sure this is going to be a problem or not, I think we should keep quoting stuff where we were before. Should be only the unquoted elements that were a problem before.

wkalt:
@kbarber fixed

rbrw:

rbrw:

rbrw:

rbrw:

PR #1234, opened 2015-01-29
(PDB-1198) Added support for submitting commands in-process

senior:
Added another function to the PuppetDBServer TK service to allow
submission of commands in-process, similar to the way we allow querying
in-process. This allows other modules running in the same JVM instance
to submit commands directly to PuppetDB without having to use an HTTP call.


senior:

senior:

senior:
BTW there was a successful Travis run here, I'm rerunning it to try and reproduce the memory errors we've been seeing

kbarber:
@senior rerunning travis again, your memory fix is in master now, I can never remember if a rebase is required here, but might be an option to remove this thrashing.

PR #1235, opened 2015-01-29
(PDB-1199) Remove timestamps from factset hash calculation

senior:
Submitting the same factset more than once causes the factset hash to
change even though it's payload is still the same. This patch removes
those fields from the hash calculation.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -1,23 +1,23 @@
 (ns puppetlabs.puppetdb.scf.storage
   "Catalog persistence
 
-   Catalogs are persisted in a relational database. Roughly speaking,
-   the schema looks like this:
+  Catalogs are persisted in a relational database. Roughly speaking,
+  the schema looks like this:
 
-   * resource_parameters are associated 0 to N catalog_resources (they are
-   deduped across catalogs). It's possible for a resource_param to exist in the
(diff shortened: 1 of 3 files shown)


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1,23 +1,23 @@
 (ns puppetlabs.puppetdb.scf.storage
   "Catalog persistence
 
-   Catalogs are persisted in a relational database. Roughly speaking,
-   the schema looks like this:
+  Catalogs are persisted in a relational database. Roughly speaking,
```
senior: I think all of this whitespace change might be coming in via merges from stable. We auto formatted master but not stable.

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -972,7 +972,7 @@
                                         "f" nil, "b" 3.14, "a" 1}},
          "environment" "DEV"
          "certname" "foo1"
-         "hash" "2148456e95cb3c513ebe80ffe10dd3c74991734b"}
+         "hash" "1ac62c7f4d290d8f064575b0ac5453a5a860d127"}
```
senior: Not a fan of this...

senior:
@wkalt this was a fix that you originally sent over with some stuff I needed to add on

PR #1236, opened 2015-01-29
(PDB-908) Improve debugging info of in-process launched PDB

senior:
This patch dynamically creates a temp logback and temp log file that an
in-process launched PuppetDB instances uses. Previously when an error
occurred in that Jetty instance, it was difficult to determine the
problem. Sometimes errors were lost, others were difficult to sort
through since it wrote to the same System.err that the test thread was
running in. This patch writes all in-process PuppetDB log information to
an actually file, which is scraped and dumped into the test thread's
console, make it much easier to diagnose errors

### DIFF
--- a/test/puppetlabs/puppetdb/testutils/jetty.clj
+++ b/test/puppetlabs/puppetdb/testutils/jetty.clj
@@ -1,5 +1,5 @@
 (ns puppetlabs.puppetdb.testutils.jetty
-  (:require [puppetlabs.puppetdb.testutils :refer [temp-dir]]
+  (:require [puppetlabs.puppetdb.testutils :refer [temp-dir temp-file]]
             [puppetlabs.puppetdb.fixtures :as fixt]
             [puppetlabs.trapperkeeper.app :as tka]
             [puppetlabs.trapperkeeper.testutils.bootstrap :as tkbs]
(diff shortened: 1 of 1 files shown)


senior:
One easy way to test this out (and pretty typical of problems that would really hang me up when creating integration tests) is to go to this (line)[https://github.com/puppetlabs/puppetdb/blob/master/test/puppetlabs/puppetdb/cli/import_export_roundtrip_test.clj#L104] and do something like dissoc :name (a required catalog field). The test will fail (block-on-node will say it didn't find anything) but you should also see a dump of the PuppetDB log from the in-process launched Jetty thread.

rbrw:
In the commmit message, perhaps s/actually/actual/.

senior:
@rbrw I think you've got some good feedback here, but I can't figure out which line(s) of code the comments are associated with. I think you clicked on "Commits" and added them there, rather than the "Files changed".

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -44,22 +74,31 @@
 
 (defn puppetdb-instance
   "Stands up a puppetdb instance with `config`, tears down once `f` returns.
-   Adjusts *server* and *base-url* to refer to the instance."
+  If the port is assigned by Jetty, use *port* to get the currently running port."
```
rbrw: Suspect you might have accidentally clobbered the server/base-url docs with the older port docs here.

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -44,22 +74,31 @@
...
   ([f] (puppetdb-instance (create-config) f))
-  ([config f]
-   (let [config (conf/adjust-tk-config config)
+  ([config f] (puppetdb-instance config [] f))
+  ([config services f]
```
rbrw: Is this new services arg for future use? (And if so, wonder if we should mention it in the docstring and/or commit message.)

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -44,22 +74,31 @@
...
+                                      :host "localhost"
+                                      :port (current-port server)}
+                                     (when prefix {:prefix prefix}))]
+           (f)))
+       (finally
+         (let [log-contents (slurp log-file)]
+           (when-not (str/blank? log-contents)
+             (println "-------Begin PuppetDB Instance Log--------------------\n"
+                      log-contents
+                      "\n-------End PuppetDB Instance Log----------------------"))))))))
```
rbrw: Any chance the log should go to *err* rather than *out*?

rbrw:
OK, hopefully that puts the comments in the right places.

test/puppetlabs/puppetdb/testutils/jetty.clj
senior: I am using this in a separate project. I found myself duplicating the list of services (line 89 below) several times, so I added another arity that will add any passed in services. I've added the param to the doc.

senior:
@rbrw Good catch on the squashed docstring. It was a messy merge, must of missed that one. I made the switch from System/out to System/err. I think the end result will be the same, as it's intended for a REPL/console invocation usage (which should have both streams going to the same place). After I switched it, I check CIDER (still goes to the REPL window) and I checked the command line invocation and it looks good there too.

senior:

wkalt:
+100 this is awesome

PR #1238, opened 2015-01-30
Pin against Ruby 1.9.x compatible version of retriable

kbarber:
This pins our version of retriable so we keep Ruby 1.9.x compatibility.


kbarber:

kbarber:

PR #1239, opened 2015-01-30
(PDB-1200) Added pinning to Puppet 3.7.3 on Fedora

senior:
Puppet 3.7.4 introduced a bug with Fedora (more info here
https://tickets.puppetlabs.com/browse/PUP-3927). This patch pins to
Puppet 3.7.3 only on Fedora, until a fix has been released.


senior:
This needed to be targeted at stable: https://github.com/puppetlabs/puppetdb/pull/1240

PR #1241, opened 2015-01-30
(PDB-1034) Ezbake source based testing

kbarber:
methodology can be retired.


rbrw:

kbarber:
maven transient ^^

acceptance/README.md
```diff
@@ -116,10 +114,6 @@ in your hash; the environment variable names are the same but uppercased
   from source. This is primarily so we can test against alternate versions of
   Puppet, so if the Puppet repo is not specified we fall back to using packages.
 
-* `:puppetdb_repo_puppetdb` (`PUPPETDB_REPO_PUPPETDB`) :
```
senior: Do we just use the beaker arg for this functionality now?

kbarber: It now does a local tar.gz build, transfers that to the remote host, then unpacks it there etc. it not longer needs to have the repo, as in no longer needs to do a secondary pull of that repo. It uses the local repo for all its needs.

PR #1242, opened 2015-02-02
(PDB-1202) Fix Travis OOM native thread error

senior:
Travis is frequently failing with the following message:
java.lang.OutOfMemoryError: unable to create new native thread, this
patch fixes that by bumping the ulimit before running the tests.


senior:
I had intended this to go against stable, closing and reopening

PR #1244, opened 2015-02-02
(re-3879)(packaging) Install terminus files to puppet 4 load path

shrug:
Currently we install puppetdb terminus files to the system
ruby load path. In the Puppet 4 Unified Agent, the terminus
needs to be installed in the vendored ruby load path.

This updates the rake tasks and packaging artifacts to
install terminus files to both the system ruby and
the Puppet Agent looad path, to support both existing
Puppet 3.x and Puppet 4.x environments.


haus:
:+1 million:

melissa:
:+1: I suppose @kbarber might also want to take a look at this?

shrug:
or @senior

kbarber:
Thanks very much @shrug.

PR #1245, opened 2015-02-03
(PDB-1034) Remove old packaging system

kbarber:
DO NOT MERGE: this is simply for review now, we'll need to get the ezbake source stuff working before this can be merged.

This removes the old packaging system and its files.


kbarber:
I'm going to stop the PR testing on this one, because I know it will fail.

kbarber:
I'm going to bundle this into the main patch, as a separate commit for visibility.

PR #1246, opened 2015-02-03
WIP - (PDB-1031) Reproduce the fact paths constraint issue

senior:
This PR doesn't fix the issue. I have tested out a fix that just changes the transaction to serializable and it appears to work. That may or it may not be the best solution to our problem. I'm pushing up this PR so we can discuss the problem more in depth along with other potential solutions.

We'll probably want to target any fix (and this test) at stable later, leaving it on master for now since I'll need to be rebased anyway.


kbarber:
Testing for hipchat integration

senior:
This is being folded in [here](https://github.com/puppetlabs/puppetdb/pull/1296)

PR #1248, opened 2015-02-05
(PDB-1034) Upgrade lein-ezbake to support :local-repo

kbarber:
Previously lein-ezbake 0.2.0 wasn't working when we overrode local-repo
during beaker source based installation. The 0.2.1 revision passes the local-repo
through to the correct places so this works.


kbarber:
@pljenkinsro retest this please ... you annoying plugin that fails on me!

PR #1252, opened 2015-02-06
(PDB-698) Switch dashes to underscores in API inputs and responses

wkalt:
This changes our dashed API inputs and outputs to underscore separated both
internally and externally.


acceptance/helper.rb
```diff
@@ -676,7 +676,7 @@ def compare_export_data(export_file1, export_file2, opts={})
 
   def get_export_entry_type(path)
     case path
-      when "puppetdb-bak/export-metadata.json"
+      when "puppetdb-bak/export_metadata.json"
```
kbarber: Do we really want to change to underscores for the filename?

documentation/api/query/v4/facts.markdown
```diff
@@ -80,7 +80,7 @@ Subquery against `/fact-contents` to get all remotely-authenticated trusted fact
 
     curl -X GET http://localhost:8080/v4/facts --data-urlencode 'query=["in", ["name","certname"],
       ["extract",["name","certname"],
-        ["select-fact-contents", ["~>", "path", [".*", "authenticated"]]]]]'
+        ["select_fact_contents", ["~>", "path", [".*", "authenticated"]]]]]'
```
kbarber: Looking at this, it obviously points out we need to redesign this operator a bit better so the keyword, and the thing its operated on is separate. More importantly, this points out something to me ... if we don't change our endpoints to use underscores, this now doesn't map correctly, as it uses underscores whereas the endpoint uses hyphens.

wkalt: fixed the bit about the export filenames. I don't really like the idea of underscored endpoints but I'll go along with it if that's what we decide we want. You're definitely right about the need to generalize the operator.

senior: UX would be good to talk to about this, I'm trying to figure out now who's leading that effort now

src/puppetlabs/puppetdb/command.clj
```diff
@@ -64,6 +64,8 @@
             [clj-time.coerce :refer [to-timestamp]]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.jdbc :as jdbc]
+            [clojure.walk :as walk]
+            [clojure.set :refer [rename-keys]]
```
senior: This and the previous line can be removed

src/puppetlabs/puppetdb/query/nodes.clj
```diff
@@ -33,9 +33,7 @@
 (defn munge-result-rows
   [version _]
   (fn [rows]
-    (map
-     #(kitchensink/mapkeys jdbc/underscores->dashes %)
-     rows)))
+    rows))
```
senior: Is this still needed? Maybe just use identity instead?

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -8,9 +8,10 @@
             [clojure.string :as string]
             [puppetlabs.puppetdb.http :as http]
             [puppetlabs.kitchensink.core :refer [keyset seq-contains? parse-int order-by-expr?]]
-            [puppetlabs.puppetdb.jdbc :refer [underscores->dashes]]))
+            [puppetlabs.puppetdb.jdbc :refer [underscores->dashes]]
```
senior: Is this still needed?

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -8,9 +8,10 @@
...
+            [clojure.walk :refer [keywordize-keys]]))
```
senior: I think this can be removed

src/puppetlabs/puppetdb/reports.clj
```diff
@@ -57,32 +57,19 @@
...
+    (if (not-every? string? (resource-event :containment_path))
       (throw (IllegalArgumentException.
               (format "Containment path should only contain strings: '%s'"
-                      (resource-event :containment-path))))))
+                      (resource-event :containment_path))))))
   (when (nil? (:environment report))
-    (throw (IllegalArgumentException. "Version 3 of reports must contain an environment")))
-  report)
-
-(defmethod validate! 4
```
senior: Was this something we missed earlier?

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,15 +886,43 @@
     (sql/do-commands
       "DROP INDEX fact_values_string_trgm")))
 
-(defn insert-factset-hash-column
-  "Insert a column in factsets to be populated by a hash."
+(defn collapse-fact
+  [fact]
+  {(:name fact) (or (json/parse-string (:value_json fact)) (:value fact))})
+
+(defn replace-factset
```
senior: Should this be in this patch?

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -112,12 +112,7 @@
 (defn current-queue-depth
   "Return the queue depth currently running PuppetDB instance (see `puppetdb-instance` for launching PuppetDB)"
   []
-  (let [base-metrics-url (assoc *base-url* :prefix "/metrics" :version :v1)
-        _ (println (str (utils/base-url->str base-metrics-url)
```
senior: Good catch

senior:
This looks good to me. I'm a +1 after the above comments.

documentation/api/query/v4/facts.markdown
kbarber: Its an interesting problem, since we've really focused on what the API returns more than what is passed in in regards to underscore/hyphens. I mean, if we focus on the user problems, its related to JS usage of returned data. In the future we may see more usage of passing in the endpoint names in a query however, not just in the endpoint that is ... if we wanted to start doing single endpoint style queries. This even affects PQL to some extent. I dunno, the OCD person in me says underscores in endpoints, but I didn't expect we'd want to do this just yet.

In a way I'd be fine sticking with select-foo-bar ... for the sake of getting this patch in, and coming back to this discussion.

senior: I talked with @jdwelch and he said what we're trying to unify on in PE is dashes in endpoint names, underscores in JSON and form params

kbarber: Well then. I say we just go back to select-resources and such. Honestly, the real fix here is separation of the 'select' from the item I think. Any step in between is possibly distracting :-).

src/puppetlabs/puppetdb/query/catalogs.clj
```diff
@@ -82,7 +82,7 @@
 (pls/defn-validated collapse-catalog :- (catalog-response-schema :v4)
   [version :- s/Keyword
    catalog-rows :- [row-schema]]
-  (let [first-row (kitchensink/mapkeys jdbc/underscores->dashes (first catalog-rows))
+  (let [first-row (first catalog-rows)
```
rbrw: Trailing space here.

wkalt:
@senior @rlb addressed your comments

documentation/api/query/v4/facts.markdown
wkalt: spoke to @jdwelch about this, consensus is select_fact_contents and fact-contents

rbrw:
Finished reviewing the previous version, and nothing else jumped out at me, so presumably the current one may be ready to go.

src/puppetlabs/puppetdb/reports.clj
wkalt: yes

wkalt:

ajroetker:

ajroetker:

PR #1255, opened 2015-02-11
(PDB-1003) Better message on whitelist failure

mullr:
Perhaps this is too ambitious? But simply changing the error message would mean middleware/wrap-with-authorization only really works for the whitelist, when an abstraction is already implemented via the `authorizer` parameter.

### DIFF
--- a/acceptance/tests/security/cert_whitelist.rb
+++ b/acceptance/tests/security/cert_whitelist.rb
@@ -38,6 +38,19 @@ test_name "certificate whitelisting" do
     curl_against_whitelist.call "", 403
   end
 
+  step "forbidden requests should return a relevant error message in the body" do
+    create_remote_file database, "#{confd}/whitelist", ""
+    on database, "chmod 644 #{confd}/whitelist"
+    restart_puppetdb database
+    on database, "curl --tlsv1 -sL " +
+                 "--cacert #{ssldir}/certs/ca.pem " +
+                 "--cert #{ssldir}/certs/#{dbname}.pem " +
+                 "--key #{ssldir}/private_keys/#{dbname}.pem "+
+                 "https://#{dbname}:8081/metrics/v1/mbeans"
+    actual_body = stdout.chomp
+    assert_match "whitelist", actual_body, "The response body should contain a message about the whitelist; actual is: #{actual_body}"
+  end
+
   step "restore original config.ini" do
     on database, "mv #{confd}/config.ini.bak #{confd}/config.ini"
     on database, "chmod 644 #{confd}/config.ini"
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -184,17 +184,22 @@
 (defn build-whitelist-authorizer
   "Build a function that will authorize requests based on the supplied
   certificate whitelist (see `cn-whitelist->authorizer` for more
-  details). Rejected requests are logged."
+  details). Returns :authorized if the request is allowed, otherwise a
+  string describing the reason not."
   [whitelist]
   {:pre  [(string? whitelist)]
    :post [(fn? %)]}
   (let [allowed? (kitchensink/cn-whitelist->authorizer whitelist)]
     (fn [{:keys [ssl-client-cn] :as req}]
       (if (allowed? req)
-        true
+        :authorized
         (do
           (log/warnf "%s rejected by certificate whitelist %s" ssl-client-cn whitelist)
-          false)))))
+          (format (str "The client certificate name (%s) doesn't "
+                       "appear in the certificate whitelist. Is your "
+                       "master's (or other PuppetDB client's) certname "
+                       "listed in PuppetDB's certificate-whitelist file?")
+                  ssl-client-cn))))))
 
 (defn shutdown-mq-broker
   "Explicitly shut down the queue `broker`"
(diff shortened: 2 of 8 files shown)


wkalt:
@pljenkinsro okay to test

wkalt:

mullr:

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -184,17 +184,24 @@
...
       (if (allowed? req)
-        true
+        :authorized
         (do
           (log/warnf "%s rejected by certificate whitelist %s" ssl-client-cn whitelist)
-          false)))))
+          (format (str "The client certificate name (%s) doesn't "
+                       "appear on the certificate whitelist. Is your "
+                       "master's certname listed in PuppetDB's "
+                       "cetificate-whitelist file? If you recently "
```
kbarber: @pizzaops can you remind me again, why would a user get this far, if there certs are bad? What I mean is, why would puppetdb ssl-setup help them in this scenario? Surely if there certs were bad, they would get a completely different error before this one was ever hit, more to do with bad CA, or the pdb cert is not signed by a trusted CA etc. etc.

mullr:

pizzaops:
@kbarber I think you're right. If there's a reason why you'd get to this point with bad certs, it's been so long since I was thinking about the issue that I no longer remember it.

mullr:

kbarber:
@pizzaops yeah, I can't recall either. @mullr maybe remove the mention of puppetdb ssl-setup for now in the error message, its easy enough to change if we need it later on.

mullr:
@kbarber sounds reasonable to me. I've changed the message and made yet another pass at the acceptance tests, we'll see if sam the american eagle approves of this one.

mullr:

acceptance/tests/security/cert_whitelist.rb
```diff
@@ -38,6 +38,19 @@
     curl_against_whitelist.call "", 403
   end
 
+  step "forbidden requests should return a relevant error message in the body" do
+    create_remote_file database, "#{confd}/whitelist", ""
+    on database, "chmod 644 #{confd}/whitelist"
+    restart_puppetdb database
+    on database, "curl --tlsv1 -sL " +
```
wkalt: need to assign this to body

mullr:

mullr:
Well that only took a week :sweat:

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -184,17 +184,22 @@
...
       (if (allowed? req)
-        true
+        :authorized
         (do
           (log/warnf "%s rejected by certificate whitelist %s" ssl-client-cn whitelist)
-          false)))))
+          (format (str "The client certificate name (%s) doesn't "
+                       "appear on the certificate whitelist. Is your "
+                       "master's certname listed in PuppetDB's "
+                       "cetificate-whitelist file?")
```
wkalt: certificate

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -184,17 +184,22 @@
...
   (let [allowed? (kitchensink/cn-whitelist->authorizer whitelist)]
     (fn [{:keys [ssl-client-cn] :as req}]
       (if (allowed? req)
-        true
+        :authorized
         (do
           (log/warnf "%s rejected by certificate whitelist %s" ssl-client-cn whitelist)
-          false)))))
+          (format (str "The client certificate name (%s) doesn't "
+                       "appear on the certificate whitelist. Is your "
```
wkalt: in the...

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -30,16 +30,18 @@
...
-      (app req)
-      (-> "You shall not pass!"
-          (rr/response)
-          (rr/status http/status-forbidden)))))
+    (let [auth-result (authorizer req)]
+     (if (= :authorized auth-result)
+       (app req)
+       (-> (str "Permission denied: " auth-result)
+         (rr/response)
+         (rr/status http/status-forbidden))))))
```
wkalt: indentation

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -184,17 +184,22 @@
...
+                       "master's certname listed in PuppetDB's "
```
ajroetker: This won't always be a master, in PE we have multiple services that connect to pdb.

src/puppetlabs/puppetdb/http/server.clj
```diff
@@ -52,15 +52,15 @@
...
+  * `authorizer` - a function that takes a request and returns a
+    :authorized if the request is authorized, or a user-visible reason if not.
+    If not supplied, we default to authorizing all requests."
   [& options]
   (let [opts (apply hash-map options)]
     (-> (routes (get-in opts [:globals :url-prefix]))
         (wrap-resource "public")
         (wrap-params)
-        (wrap-with-authorization (opts :authorized? (constantly true)))
+        (wrap-with-authorization (opts :authorizer (constantly :authorized)))
```
ajroetker: This may be out of the scope of this PR, but it seems like we could remove some duplication here by having this default in `wrap-with-authorization` itself. Also since we already have defaults here we don't really need them at the build app level.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -296,10 +301,10 @@
           context (if updater
                     (assoc context :updater updater)
                     context)
-          _       (let [authorized? (if-let [wl (:certificate-whitelist puppetdb)]
-                                      (build-whitelist-authorizer wl)
-                                      (constantly true))
-                        app (server/build-app :globals globals :authorized? authorized?)]
+          _       (let [authorizer (if-let [wl (:certificate-whitelist puppetdb)]
```
ajroetker: Like I suggested in my comment below, this could really just be ` app (server/build-app :globals globals :authorizer (when-let [wl (:certificate-whitelist puppetdb)] (build-whitelist-authorizer wl)))` since we're defaulting at a lower level already.

src/puppetlabs/puppetdb/http/server.clj
mullr: Seems reasonable, but like something we should handle separately. This patch really, really needs to go to bed.

kbarber:
@pljenkinsro retest this please ... you silly bot.

src/puppetlabs/puppetdb/http/server.clj
kbarber: Agreed, we can refactor this later. The goal of this patch has been achieved as it stands.

PR #1256, opened 2015-02-12
(PDB-332) Add tags to catalog hash

mullr:
If the user submits a new catalog with no changes other than tags, 
the catalog should still be stored. It previously was not, since the
tags were not part of the catalog hash computation. 

This does not update existing hashes stored in the db, so the next 
catalog update for each node will cause a db write.

### DIFF
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -1,8 +1,30 @@
 ---
-title: "PuppetDB 2.2 » Release Notes"
+title: "PuppetDB 3.0 » Release Notes"
 layout: default
 canonical: "/puppetdb/latest/release_notes.html"
 ---
+
+3.0.0
+-----
+
+### Upgrading
+
+* You may notice some additional system load for the first 30 to 60
+  minutes after upgrading. This is expected, and is due to a change in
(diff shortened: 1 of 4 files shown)


documentation/release_notes.markdown
```diff
@@ -1,9 +1,31 @@
 ---
-title: "PuppetDB 2.2 » Release Notes"
+title: "PuppetDB 2.3 » Release Notes"
```
kbarber: wouldn't bother doing this now ... this will be done across the board later anyway.

documentation/release_notes.markdown
```diff
@@ -1,9 +1,31 @@
...
 layout: default
 canonical: "/puppetdb/latest/release_notes.html"
 ---
 
+2.3.0
```
kbarber: next release for master is '3.0.0' fwiw. But this could also be just 'master branch' or some place holder that could change later. We don't always know what version things are going to be, but this time we definitely are cutting a major, so its cool to leave this as 3.0.0 now.

mullr:
@kbarber fixed those two doc issues

mullr:

PR #1259, opened 2015-02-17
(PDB-1177) add noop flag to reports

wkalt:
This adds a "noop" field to the reports object, which is a boolean flag
indicating whether the run producing the report was a --noop.

### DIFF
--- a/acceptance/tests/reports/report_storage.rb
+++ b/acceptance/tests/reports/report_storage.rb
@@ -2,14 +2,14 @@ require 'json'
 
 test_name "basic validation of puppet report submission" do
 
-  step "setup a test manifest for the master and perform agent runs" do
-    manifest = <<-MANIFEST
+  manifest = <<-MANIFEST
       notify { "hi":
         message => "Hi ${::clientcert}"
       }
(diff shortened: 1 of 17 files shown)


wkalt:

wkalt:

wkalt:

acceptance/tests/reports/report_storage.rb
```diff
@@ -2,14 +2,14 @@
 
 test_name "basic validation of puppet report submission" do
 
-  step "setup a test manifest for the master and perform agent runs" do
-    manifest = <<-MANIFEST
+  manifest = <<-MANIFEST
```
kbarber: I'm not sure we should be overloading noop related testing with an existing test here. I think a new test is in order.

kbarber: Although I can see what you've done at the end of this file. Somehow I think this should be inverted though, we've gone and modified an existing test with noop stuff, and then added back a non-noop test. I'd at least invert the way we're doing this here, the noop test should be the bit we've added at the end, not the other way around.

PR #1262, opened 2015-02-18
(PDB-1241) Resource Event timestamps should be normalized before stored

senior:
There are many ways to represent an ISO 8601 timestamp in string
form. The end result is the posibility of having to two timestamps that
are the same, but represented different which causes the report to hash
differently. Although Puppet should be sending PuppetDB UTC dates, our
tests were using non-UTC dates, which would cause the dates to be stored
in another timezone, then converted to UTC on the way out of the
database which is really confusing.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -963,20 +963,20 @@
   "Given a certname and a map of fact names to values, store records for those
    facts associated with the certname."
   ([fact-data]
-     (add-facts! fact-data true))
+   (add-facts! fact-data true))
   ([{:keys [name values environment timestamp producer_timestamp] :as fact-data} :- facts-schema
     include-hash? :- s/Bool]
-     (let [factset {:certname name
-                    :timestamp (to-timestamp timestamp)
-                    :environment_id (ensure-environment environment)
-                    :producer_timestamp (to-timestamp producer_timestamp)}
-           fact-data-hash (shash/generic-identity-hash (dissoc fact-data :timestamp :producer_timestamp))]
-       (sql/insert-record :factsets
-                          (if include-hash?
-                            (assoc factset :hash fact-data-hash) factset))
-       (insert-facts!
-        (certname-to-factset-id name)
-        (set (new-fact-value-ids values))))))
+   (let [factset {:certname name
+                  :timestamp (to-timestamp timestamp)
+                  :environment_id (ensure-environment environment)
+                  :producer_timestamp (to-timestamp producer_timestamp)}
+         fact-data-hash (shash/generic-identity-hash (dissoc fact-data :timestamp :producer_timestamp))]
+     (sql/insert-record :factsets
+                        (if include-hash?
+                          (assoc factset :hash fact-data-hash) factset))
+     (insert-facts!
+      (certname-to-factset-id name)
+      (set (new-fact-value-ids values))))))
 
 (pls/defn-validated delete-facts!
   "Delete all the facts (1 arg) or just the fact-ids (2 args) for the given certname."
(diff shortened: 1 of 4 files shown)


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1103,31 +1103,48 @@
...
     row-map))
 
+(defn normalize-resource-event
+  "Prep `event` for comparison/computation of a hash"
+  [event]
+  (-> event
+      (utils/update-when [:timestamp] to-timestamp)
+      (utils/update-when [:old_value] sutils/db-serialize)
+      (utils/update-when [:new_value] sutils/db-serialize)
+      (assoc :containing_class (find-containing-class (get event :containment_path)))))
```
ajroetker: How come you use `(get event :containment_path)` instead of `(:containment_path event)`?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1103,31 +1103,48 @@
...
+      (update-in [:resource_events] #(map normalize-resource-event %))))
+
+
+(defn convert-containment-path
+  "Convert the contain path from a collection to the jdbc array type"
+  [event]
+  (utils/update-when event
+                     [:containment_path]
+                     (fn [cp]
+                       (when cp
```
ajroetker: This when seems unnecessary unless you expect `cp` to be `false`. `update-when` should already only update if its non-nil.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1137,15 +1154,15 @@
...
+                                 :start_time             start_time
+                                 :end_time               end_time
                                  :receive_time           (to-timestamp timestamp)
                                  :transaction_uuid       transaction_uuid
                                  :environment_id         (ensure-environment environment)
                                  :status_id              (ensure-status status)}))
 
-            (apply sql/insert-records :resource_events resource-event-rows)
-            (if update-latest-report?
+            (apply sql/insert-records :resource_events (map (comp convert-containment-path #(assoc % :report report-hash)) resource_events))
```
ajroetker: This might be a tad more readable in a `(->> ...)`

test/puppetlabs/puppetdb/http/events_test.clj
```diff
@@ -29,12 +29,12 @@
...
   ([endpoint query]
-     (get-response endpoint query {}))
+   (get-response endpoint query {}))
   ([endpoint query extra-query-params]
-     (let [resp (*app* (get-request endpoint query extra-query-params))]
-       (if (string? (:body resp))
-         resp
-         (update-in resp [:body] slurp)))))
+   (let [resp (*app* (get-request endpoint query extra-query-params))]
+     (if (string? (:body resp))
```
ajroetker: Maybe this could just be `(update-in resp [:body] #(if (string? %) % (slurp %)))` since you're digging into the map in the conditional anyways.

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -1130,6 +1130,14 @@
        (assoc report
          :puppet_version "3.2.1 (Puppet Enterprise 3.0.0-preview0-168-g32c839e)") timestamp)))
 
+  (deftest report-with-event-timestamp
+    (let [z-report (update-in report [:resource_events] (fn [events]
```
ajroetker: This `update-in` could probably go in a function in the let with a name that makes it a little more clear what it's doing.
```clojure
[add-event-timetstamps (fn [report timestamp]
                                        (update-in report [:resource_events] (partial map #(assoc % :timestamp timestamp))))]
```

src/puppetlabs/puppetdb/scf/storage.clj
senior: Not true, update-when checks for the presence of the key. I'm not sure if it's possible to have a nil value for that key. This was intended to be a straight refactor and not change the existing functionality. I've got no problems removing it if we can prove it's not needed.

src/puppetlabs/puppetdb/scf/storage.clj
senior: It should be the latter, I'll switch it

src/puppetlabs/puppetdb/scf/storage.clj
senior: Probably right, that line is getting a bit long

test/puppetlabs/puppetdb/http/events_test.clj
senior: Makes sense to me, I didn't change that code but I can rework it

senior:
@ajroetker make the changes you suggested

kbarber:
So there is a problem with 'existing data' in regards to this patch, in that existing data will have the wrong format & thus the wrong hash already. So new data will have the correct patch, old data will not.

You can see the output from just dropping the report JSON here in UTC timezone:

```
{"certname":"kb.local","puppet_version":"3.7.4","report_format":4,"configuration_version":"1424356172","start_time":"2015-02-19T14:29:29.268556000+00:00","end_time":"2015-02-19T14:29:30.009997000+00:00","resource_events":[],"environment":"production","transaction_uuid":"17fe5041-494d-4b8b-8cdf-35b249211cf4","status":"unchanged"}
```

And now with AEST:

```
{"certname":"kb.local","puppet_version":"3.7.4","report_format":4,"configuration_version":"1424356416","start_time":"2015-02-20T00:33:33.538599000+10:00","end_time":"2015-02-20T00:33:34.276198000+10:00","resource_events":[],"environment":"production","transaction_uuid":"097d5661-1b49-472a-b1be-791d9e103521","status":"unchanged"}
```

So its not just as simple as making it the +00:00 format, we also have timezone to contend with.

PR #1263, opened 2015-02-18
(PDB-908) Adding a new test until-consumed function

senior:
These changes are to the tests only. This function will invoke a
function and wait for the messages put on the queue to be consumed. Will
throw an exception if the message takes too long, ends up in the
DLO. There is also code to ensure that the JMX beans have been
initialized after startup before the checking for consumed messages.

### DIFF
--- a/test/puppetlabs/puppetdb/testutils/jetty.clj
+++ b/test/puppetlabs/puppetdb/testutils/jetty.clj
@@ -5,7 +5,7 @@
             [puppetlabs.trapperkeeper.testutils.bootstrap :as tkbs]
             [puppetlabs.trapperkeeper.services.webserver.jetty9-service :refer [jetty9-service]]
             [puppetlabs.trapperkeeper.services.webrouting.webrouting-service :refer [webrouting-service]]
-            [puppetlabs.puppetdb.cli.services :refer [puppetdb-service]]
+            [puppetlabs.puppetdb.cli.services :refer [puppetdb-service mq-endpoint]]
             [puppetlabs.puppetdb.metrics :refer [metrics-service]]
             [puppetlabs.puppetdb.mq-listener :refer [message-listener-service]]
             [puppetlabs.puppetdb.command :refer [command-service]]
@@ -15,7 +15,8 @@
             [clj-http.util :refer [url-encode]]
             [clj-http.client :as client]
             [clojure.string :as str]
-            [fs.core :as fs]))
+            [fs.core :as fs]
+            [slingshot.slingshot :refer [throw+]]))
 
 ;; See utils.clj for more information about base-urls.
 (def ^:dynamic *base-url* nil) ; Will not have a :version.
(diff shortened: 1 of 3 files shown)


test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -109,16 +110,121 @@
...
+  provided. Returns the result of `f` if successful. Requires JMX to
+  be enabled in ActiveMQ (the default, but `without-jmx` will cause
+  this to fail).
+
+  Exceptions thrown in the following cases:
+
+  timeout - if the message isn't consumed in approximately 5 seconds
+  new message in the DLO - if any message is discarded"
+  ([f] (until-consumed 1 f))
+  ([num-messages f]
```
ajroetker: Could we use the `& args` pattern here so every `f` doesn't need to be wrapped in a `(fn [] ...)`?

senior: Do you mean rewrite this as a macro?

ajroetker: I'm thinking something like
```clojure
(defn until-consumed
  ([f &args] (apply until-consumed 1 f args))
  ([num-messages f &args]
     (let [....
     results (apply f args))...))
```

senior: Oh, I get what you're saying now, sure if you think that would be useful, it's easy to add.

senior: Oh, but you can't have varargs and multiple arities, we can support it on the second arity though.

ajroetker: Ah that's good to know, sometimes I overestimate pattern matching (ie forget that clojure has no types and isn't haskell).

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -109,16 +110,130 @@
...
+  ;; period of time that the server is up, the broker has been
+  ;; started, but the JMX beans have not been initialized, so querying
+  ;; for queue metrics fails, this check ensures it's started
   (let [base-metrics-url (assoc *base-url* :prefix "/metrics" :version :v1)]
     (-> (str (utils/base-url->str base-metrics-url)
              "/mbeans/org.apache.activemq:BrokerName="
              (url-encode (:host base-metrics-url))
-             ",Type=Queue,Destination=puppetlabs.puppetdb.commands")
+             ",Type=Queue,Destination="
+             mq-endpoint)
```
kbarber: There are 2 other places in this patch you are using the static string for puppetlabs.puppetdb.command, do we want to fix them now also?

kbarber:
As a general comment that doesn't need solving now, testutils.jetty should almost be renamed now to testutils.service or something more encompassing ... I mean, its not really just about jetty any more is it, esp. with all this mq stuff now :-).

test/puppetlabs/puppetdb/testutils/jetty.clj
senior: Unfortunately no. ActiveMQ doesn't allow two instances of itself with JMX enabled. To stand up two MQ instances, you have to shut of JMX, if you shut off JMX you can't use the above.

senior:
Agreed on the eventual Jetty ns refactor

PR #1264, opened 2015-02-18
Remove some unused legacy query code

kbarber:
This removes some of the legacy query code since we've dropped v3.

This doesn't put much of a dent in the file query.clj, but it at least
removes the obvious unused functions. We still use this query engine in
some of our weird endpoints, so greater work will be required to remove
the lot.

For now this removes some of the dead code at least.


ajroetker:
:+1:  :+1: :+1:

PR #1265, opened 2015-02-18
(PDB-1218) Add id to reports table

ajroetker:
This commit adds a migration which changes our database schema to use an
integer primary key for reports. This commit merges this migration with
the noop column migration since they are going out in the same release.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -890,11 +890,116 @@
     "ALTER TABLE factsets ADD hash VARCHAR(40)"
     "ALTER TABLE factsets ADD CONSTRAINT factsets_hash_key UNIQUE (hash)"))
 
-(defn insert-noop-column
-  "Insert a column in reports to be populated by boolean noop flag"
+(defn migrate-to-report-id-and-noop-column
+  "Migrate to report id and
+   insert a column in reports to be populated by boolean noop flag"
   []
-  (sql/do-commands
-    "ALTER TABLE reports ADD noop boolean"))
+    (sql/do-commands
+     "CREATE SEQUENCE reports_id_seq CYCLE")
+
+    (sql/create-table :reports_transform
+                      ["id"                    "bigint NOT NULL DEFAULT nextval('reports_id_seq')"]
+                      ["hash"                  "varchar(40) NOT NULL"]
+                      ["certname"              "text NOT NULL"]
+                      ["puppet_version"        "varchar(255) NOT NULL"]
+                      ["report_format"         "smallint NOT NULL"]
+                      ["configuration_version" "varchar(255) NOT NULL"]
+                      ["start_time"            "timestamp with time zone NOT NULL"]
+                      ["end_time"              "timestamp with time zone NOT NULL"]
+                      ["receive_time"          "timestamp with time zone NOT NULL"]
+                      ["transaction_uuid"      "varchar(255) DEFAULT NULL"]
+                      ["noop"                  "boolean"]
+                      ["environment_id"        "bigint"]
+                      ["status_id"             "bigint"])
+
+    (sql/create-table :resource_events_transform
+                      ["report_id"        "bigint NOT NULL"]
+                      ["status"           "varchar(40) NOT NULL"]
+                      ["timestamp"        "timestamp with time zone NOT NULL"]
+                      ["resource_type"    "text NOT NULL"]
+                      ["resource_title"   "text NOT NULL"]
+                      ["property"         "varchar (40)"]
+                      ["new_value"        "text"]
+                      ["old_value"        "text"]
+                      ["message"          "text"]
+                      ["file"             "varchar(1024) DEFAULT NULL"]
+                      ["line"             "integer"]
+                      ["containment_path" (scf-utils/sql-array-type-string "TEXT")]
+                      ["containing_class" "varchar(255)"])
+
+    (sql/create-table :latest_reports_transform
+                      ["certname"  "text NOT NULL"]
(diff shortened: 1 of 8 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -890,11 +890,118 @@
...
   []
-  (sql/do-commands
-    "ALTER TABLE reports ADD noop boolean"))
+    (sql/do-commands
+     "CREATE SEQUENCE reports_id_seq CYCLE")
+
+    (sql/create-table :reports_transform
+                      ["id"                    "bigint NOT NULL DEFAULT nextval('reports_id_seq')"]
+                      ["hash"                  "varchar(40) NOT NULL"]
+                      ["certname"              "text"] ;; TODO: why isn't this not null?
```
kbarber: I never did figure this out. Seems like it should be NOT NULL to me.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -890,11 +890,118 @@
...
+        SELECT lr.certname, rt.id
+          FROM latest_reports AS lr
+          INNER JOIN reports_transform rt on lr.report = rt.hash")
+
+    (sql/do-commands
+     "DROP TABLE resource_events"
+     "DROP TABLE latest_reports"
+     "DROP TABLE reports")
+
+    ;; TODO: do we need to delete other references here, or do they magically disappear?
```
kbarber: Another one of my TODO's :-). Objects related to these tables disappear, I think the main exception is the way we create sequences now. But since these objects don't use these sequences, doesn't matter in this case.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -890,11 +890,118 @@
...
+     "ALTER TABLE latest_reports_transform RENAME to latest_reports"
+     "ALTER TABLE reports_transform RENAME to reports")
+
+    (sql/do-commands
+     "ALTER TABLE reports ADD CONSTRAINT reports_pkey PRIMARY KEY (id)"
+     "CREATE INDEX idx_reports_certname ON reports(certname)"
+     "CREATE INDEX idx_reports_end_time ON reports(end_time)"
+     "CREATE INDEX idx_reports_environment_id ON reports(environment_id)"
+     "CREATE INDEX idx_reports_status_id ON reports(status_id)"
+     "CREATE INDEX idx_reports_transaction_uuid ON reports(transaction_uuid)" ;; TODO: is this really necessary today?
```
kbarber: Just leave it for now.

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -57,11 +57,11 @@
 
         (testing "should attempt a partial migration if there are migrations missing"
           (clear-db-for-testing!)
-          ;; We are using migration 13 here because it is isolated enough to be able
+          ;; We are using migration 28 here because it is isolated enough to be able
```
kbarber: Umm, why are we modifying an existing test, and using it for our migration test? Shouldn't we go and create a new one?

wkalt: this change makes sense to me because the modified test is about out-of-order migrations, and migration 13 uses the report column in latest_reports, which this migration drops.

The new migration should probably have a test though IMO.

ajroetker: This is modified because the report-id migration relies on latest-reports, so that comment there is no longer valid, in that this new migration relies on 13 having taken place. The migration chosen here really just needs to be independent from the others. I am going to make a new test for this migration though.

kbarber: Oh I think I see what you are all saying. I'm a bit vague on the understanding, but sounds like this test should be tested on 'sample' migrations that have nothing to do with PDB so our own work can't break this. This test isn't about testing our real migrations, rather its about testing the migration code in general - so 'real' data isn't needed I think.

ajroetker:

ajroetker:

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -175,3 +174,52 @@
...
+         :certname               "testing1"})
+
+      (apply-migration-for-testing! 29)
+
+      (let [response
+            (query-to-vec
+              "SELECT r.hash, r.certname, e.name AS environment, rs.status
+               FROM latest_reports lr
+               INNER JOIN reports r          on lr.report_id=r.id AND lr.certname=r.certname
+               INNER JOIN environments e     on r.environment_id=e.id
```
wkalt: this alignment seems a bit strange to me

ajroetker:

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -175,3 +174,52 @@
                  :timestamp (to-timestamp current-time) :value_string "true"}
                 {:path "baz" :environment_id 3 :environment "test_env_3"
                  :timestamp (to-timestamp current-time) :value_string "false"}]))))))
+
+(deftest migration-29
+  (testing "should contain same reports before and after migration")
```
senior: This should slurp the next form

kbarber:
Couple of comments on this one, changing to wip, especially the deftest slurp, thats a biggy :-).

ajroetker:

wkalt:

wkalt:
this looks good to me

ajroetker:

ajroetker:

kbarber:
Merged manually, also added some release notes.

PR #1267, opened 2015-02-19
(PDB-1192) add metrics support

wkalt:
This adds a metrics field to the reports response containing metrics passed
from Puppet in the terminus.


wkalt:
I think we should consider calling the field "report-metrics" instead of metrics to avoid confusion.

wkalt:

wkalt:

src/puppetlabs/puppetdb/query/reports.clj
```diff
@@ -61,6 +63,7 @@
    :report_format s/Int
    :configuration_version String
    :resource_events [resource-event-schema]
+   :metrics (s/maybe s/Any)
```
ajroetker: Isn't `(maybe Any)` redundant since both include `nil`?

wkalt: yeah, but the real issue is that should be a proper schema. nice catch.

src/puppetlabs/puppetdb/reports.clj
```diff
@@ -21,7 +21,7 @@
                               :type :string}
    :status                   {:optional? true
                               :type :string}
-   })
+   :metrics                  :coll})
```
ajroetker: Is this not an optional field?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1142,9 +1149,20 @@
...
-                       status noop]
-                :as report}         (normalize-report orig-report)
-                report-hash         (shash/report-identity-hash report)]
+                       status noop] :as report}         (normalize-report orig-report)
+                report-hash         (shash/report-identity-hash report)
+        containment-path-fn (fn [cp] (if-not (nil? cp) (sutils/to-jdbc-varchar-array cp)))
+        report-metrics (if (scf-utils/postgres?)
+                         (scf-utils/clj->pgobject (:metrics report))
+                         (json/generate-string (:metrics report)))
+        resource-event-rows (map #(-> %
```
ajroetker: Did you mean to add this or was this part of rebasing? I think @senior removed this in a recent patch

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1142,9 +1149,20 @@
...
          (let [{:keys [puppet_version certname report_format configuration_version
                        start_time end_time resource_events transaction_uuid environment
-                       status noop]
-                :as report}         (normalize-report orig-report)
-                report-hash         (shash/report-identity-hash report)]
+                       status noop] :as report}         (normalize-report orig-report)
+                report-hash         (shash/report-identity-hash report)
+        containment-path-fn (fn [cp] (if-not (nil? cp) (sutils/to-jdbc-varchar-array cp)))
+        report-metrics (if (scf-utils/postgres?)
+                         (scf-utils/clj->pgobject (:metrics report))
```
ajroetker: You can probably just add the `metrics` to the destructuring above in the `:keys` list.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -28,6 +29,19 @@
 
 ;; FUNCTIONS
 
+(defn clj->pgobject
```
ajroetker: Maybe two extra helpers here would be nice that do the work of
```
(if (scf-utils/postgres?)
  scf-utils/pgobject->clj
  json/parse-string)
```
and
```
(if (scf-utils/postgres?)
  (scf-utils/clj->pgobject (:metrics report))
  (json/generate-string (:metrics report)))
```
So that the pieces of code needing to know about postgres or not are minimized.

src/puppetlabs/puppetdb/reports.clj
wkalt: not as far as I know.

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: yeah good catch

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: metrics is a global var in this file, so that won't work without changing the name of the metrics variable. I don't care particularly though, happy to make the change.

ajroetker: Does clojure not scope that properly if you try and bind it locally?

wkalt:
@ajroetker addressed your comments except for the optional key part. I think it should be required but possibly null. This will be easy when report validation is switched to schema.

wkalt:

documentation/api/wire_format/report_format_v5.markdown
```diff
@@ -51,8 +55,38 @@ match a report with the catalog that was used for the run.  This field may be `n
 
 `"status"` is a string used to identify the status of the puppet run.
 
+`"resource_events"` is an array of resource events taking the form described in
```
kbarber: Does this line need removal now?

puppet/lib/puppet/reports/puppetdb.rb
```diff
@@ -73,6 +74,20 @@ def build_events_list
     end
   end
 
+  def build_metrics_list
+    profile("Build metrics list (count: #{metrics.count})",
+            [:puppetdb, :metrics_list, :build]) do
+              metrics_list = []
```
kbarber: This indenting is clojure-like, not ruby-like.

src/puppetlabs/puppetdb/query/reports.clj
```diff
@@ -17,6 +18,7 @@
    :puppet_version String
    :report_format s/Int
    :configuration_version String
+   :metrics (s/maybe s/Any)
```
kbarber: Is this because it could be either a PGObject or a string? Can we validate for that?

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -252,3 +253,28 @@ must be supplied as the value to be matched."
        (sql/do-commands
         (str "ALTER TABLE " table " ALTER COLUMN " column
              " RESTART WITH " restartid))))))
+
+(defn clj->pgobject
```
kbarber: I'm not sure, but I feel like this file and jdbc.clj are becoming mixed up in purpose. There are some functions in jdbc.clj that we draw on for storage purposes, and some in this one as well. I guess in my head I think that this file is about the 'storage' side of the business, versus any query side. But looking at the rest of this file, we've probably lost track of any original intent etc.

test-resources/puppetlabs/puppetdb/cli/export/sample-report.json
```diff
@@ -38,5 +38,18 @@
   "transaction_uuid": "68b08e2a-eeb1-4322-b241-bfdf151d294b",
   "environment": "DEV",
   "status" : "unchanged",
+  "metrics" : [ {
```
kbarber: indentation is weird. I know its just a sample :-).

test/puppetlabs/puppetdb/examples/reports.clj
```diff
@@ -12,6 +12,15 @@
     :environment            "DEV"
     :status                 "unchanged"
     :noop                   false
+    :metrics                [{:category  "resources"
+                              :name  "changed"
+                              :value  1}
```
kbarber: Be good to mix some floats in here, if these are used in testing.

test-resources/puppetlabs/puppetdb/cli/export/sample-report.json
```diff
@@ -38,5 +38,18 @@
...
+      "category" : "resources",
+      "name" : "changed",
+      "value" : 1
```
kbarber: Floats in this one would be great also.

src/puppetlabs/puppetdb/scf/storage_utils.clj
wkalt: Yeah, I waffled a bit. I put it where I did because the functions originally used postgres? which made the choice easier. My argument for storage_utils is that these functions are ultimately used around postgres/hsql special-casing which I saw mainly in storage_utils, but I can move it if we want.

wkalt:
@KenBarber addressed your comments. Leaving the label pending PDB-1129.

wkalt:
@kbarber ^^

test/puppetlabs/puppetdb/examples/reports.clj
wkalt: they're in there now, just not that one

wkalt:
This is now rebased and accepts null for metrics.

wkalt:

wkalt:
I just added a second commit to this for the logs work.

senior:
Looks like a legit failure

PR #1268, opened 2015-02-19
(PDB-1243) Accept null noop field in reports

ajroetker:
This commit changes the defmodel for report to have the noop field be
optional as it is in the database which will allow users to use the
store-report! command on old exported data.

### DIFF
--- a/test/puppetlabs/puppetdb/reports_test.clj
+++ b/test/puppetlabs/puppetdb/reports_test.clj
@@ -5,12 +5,16 @@
             [puppetlabs.puppetdb.testutils.reports
              :refer [munge-example-report-for-storage]]))
 
-(let [report (munge-example-report-for-storage (:basic reports))]
+(let [report (munge-example-report-for-storage (:basic reports))
+      report-missing-noop (dissoc report :noop)]
 
   (deftest test-validate!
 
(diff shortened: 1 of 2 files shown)


ajroetker:
@kbarber is this what you had in mind?

kbarber:
@ajroetker pretty much yeah.

wkalt:
isn't the issue that noop will be present but null, rather than absent? It seems like we need the equivalent of (s/maybe s/Bool) rather than making it an optional key.

kbarber:
@wkalt I thought it meant the same thing.

test/puppetlabs/puppetdb/reports_test.clj
```diff
@@ -5,13 +5,17 @@
             [puppetlabs.puppetdb.testutils.reports
              :refer [munge-example-report-for-storage]]))
 
-(let [report (munge-example-report-for-storage (:basic reports))]
+(let [report (munge-example-report-for-storage (:basic reports))
+      report-missing-noop (dissoc report :noop)]
```
kbarber: Oh I see @wkalt's point now, in this test, we're testing its removal, not that its null. I think we could test for both cases, it equates to the same thing I think.

PR #1270, opened 2015-02-23
(maint) benchmark improvements

wkalt:
This adds two optional flags to benchmark:

-A/--archive is passed a path to a PDB export, and will use whatever
catalogs/reports/facts are inside.

-D/--default is used without a parameter and will use the sample data in
resources/puppetlabs/puppetdb/benchmark/samples


src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -292,36 +295,92 @@
...
+  [path]
+  (let [tar-reader (archive/tarball-reader path)
+        entries (archive/all-entries tar-reader)]
+    (reduce (process-tar-entry path tar-reader)
+            {:catalogs [] :reports [] :facts []} entries)))
 
 (defn -main
   [& args]
-  (let [[options _]     (validate-cli! args)
+  (let [options         (validate-cli! args)
```
ajroetker: you could probably use some destructuring here with catalogs, facts, reports, archive, config, numhosts, to clean up this let

wkalt: I've cleaned it up a bit with load-data-from-options, should do the trick I think.

wkalt:

kbarber:
@wkalt there is a docs page where this tool's flags are documented in load_testing_tool fyi.

kbarber:
@wkalt oh yeah, and release notes :-).

wkalt:
@kbarber dealt with docs and release notes

wkalt:

kbarber:
This seems pretty cool.

Found a bug: If you pass in a tarball with -A, that has no reports, this will break. In fact I bet this happens for any missing object type. I only found this because I had disabled/broken reports in my test env, so I was lucky :-).

I think also, leading from this, this is something worth fixing. Think about -D ... it probably also suffers the same issue, I presume this is why you skipped out the other objects in the defaults there.

I think we should become resilient to missing object types, it helps both places.

Totally future stuff beyond this though:

There is another side to this, we could potentially drop the use of the 3 object types switches for paths, but just expecting a root dir (like an export root with facts/reports/catalogs in it). This way, we reduce the amount of switches, and force people to pass in 'export' format objects only, either a tarball, or an unpacked tarball directory basically ..

This would lead to the need for another switch, that allows us to control what objects get sent though possibly :-).

ajroetker:
@kbarber 
------
There is another side to this, we could potentially drop the use of the 3 object types switches for paths, but just expecting a root dir (like an export root with facts/reports/catalogs in it). This way, we reduce the amount of switches, and force people to pass in 'export' format objects only, either a tarball, or an unpacked tarball directory basically ..

This would lead to the need for another switch, that allows us to control what objects get sent though possibly :-).
--------

Do we not have switches for this kind of thing already in import/export? if not it seems useful to have on either end there as well, though export seems better if we had to choose.

wkalt:
@kbarber fixed that archive issue and tightened up the handling on empty directories

wkalt:

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -284,6 +290,8 @@
    ["-F" "--facts FACTS" "Path to a directory containing sample JSON facts (files must end with .json)"]
    ["-C" "--catalogs CATALOGS" "Path to a directory containing sample JSON catalogs (files must end with .json)"]
    ["-R" "--reports REPORTS" "Path to a directory containing sample JSON reports (files must end with .json)"]
+   ["-A" "--archive ARCHIVE" "Path to a PuppetDB export tarball. Incompatible with -C, -F, or -R, or -D"]
```
senior: There's and extra or in there

wkalt: @senior thanks, fixed it

wkalt:
@senior has raised the issue that we don't ship the acceptance directory, so as written package users won't be able to use the -D flag. This shouldn't be merged until we figure out what to do there.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -284,6 +290,8 @@
...
+   ["-D" "--default" "Use the sample data in acceptance/benchmark. Incompatible with -C, -F, -R, or -A"]
```
senior: Depending on the outcome of "where do we store it", could probably just change it to "Use sample seed data" or something similar. This looks to me like it only supports facts. We should probably state that, and out of curiosity, why doesn't it work with reports or catalogs since it only applies to facts?

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -292,36 +300,90 @@
 (def required-cli-options
   [:config])
 
+(defn validate-options
```
senior: We definitely need more of this in benchmark.clj, it's good to see you adding that for the new params. When I was trying out the patch, I found NumberFormatExceptions when run interval and numhosts was missing. Wrote it up as a separate ticket: https://tickets.puppetlabs.com/browse/PDB-1268

src/puppetlabs/puppetdb/cli/benchmark.clj
wkalt: it will, the data just doesn't exist yet.

I held off for now because the data for reports will change when the metrics and logs patches land, and we already have some sample data from that work that will be perfect.

kbarber:
Still blocked on a decision about what to do with -D. Right now we're proposing storing it in the resource area, so its shipped out with our jar, and works the same way in dev mode.

wkalt:
I've augmented the sample data with reports and catalogs and the -D flag now loads the data from the classpath.

wkalt:

acceptance/performance/facts/fact_load.rb
```diff
@@ -2,7 +2,8 @@
   clear_and_restart_puppetdb(database)
 
   step "run benchmark" do
-    on(database, "#{LeinCommandPrefix} lein run benchmark -F acceptance/benchmark/facts/ -n 1000 -c acceptance/benchmark/config.ini -N 5 --rand-perc 10")
+    sample_data_path = "resources/puppetlabs/puppetdb/benchmark"
+    on(database, "#{LeinCommandPrefix} lein run benchmark -F #{sample_data_path}/samples/facts/ -n 1000 -c #{sample_data_path}/config.ini -N 5 --rand-perc 10")
```
senior: I'm not sure I understand how this works. For the performance tests we're using this sample data and it's looking in resources/puppetlabs/puppetdb/benchmark. Does this mean we're perf testing against source, not a package? Not a problem, just wondering how that works /cc @kbarber

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -292,36 +306,93 @@
...
+      (println (:message m))
+      (case (:type m)
+        :puppetlabs.kitchensink.core/cli-error (System/exit 1)
+        :puppetlabs.kitchensink.core/cli-help (System/exit 0)))))
+
+(defn process-tar-entry
+  [archive-path tar-reader]
+  (fn [acc entry]
+    (let [path (.getName entry)
+          catalog-pattern "catalogs.*\\.json$"
```
senior: Seems like you could let bind this outside of the fn and push the re-pattern creation there too, then you wouldn't need to recreate these regex patterns on each file in the tarball.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -292,36 +306,93 @@
...
+          report-pattern "reports.*\\.json$"
+          facts-pattern "facts.*\\.json$"
+          parsed-entry (json/parse-string (archive/read-entry-content tar-reader))]
+      (cond
+        (re-find (re-pattern catalog-pattern) path) (update-in acc [:catalogs] #(cons parsed-entry %))
```
senior: If you switch you conj here you can get rid of your anonymous fns

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -60,13 +62,23 @@
...
-  [dir]
-  (->> (fs/glob (fs/file dir "*.json"))
-       (map try-load-file)
-       (remove nil?)
-       (vec)))
+  ([dir]
+   (load-sample-data dir false))
+  ([dir from-classpath?]
+   (let [target-files (if from-classpath?
+                        (rest (file-seq (io/file (io/resource dir))))
```
senior: Looking at the code in file-seq, it returns the root directory as the first item, though the docstring etc doesn't claim that it will. Is it worth switching the rest call for a (remove #(.isDirectory %) ...)?

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -60,13 +62,23 @@
...
+                        (fs/glob (fs/file dir "*.json")))]
+     (->> target-files
+          (map try-load-file)
+          (remove nil?)
+          (vec)
+          (error-if empty? (format "Supplied directory %s contains no usable data!" dir))))))
```
senior: This is writing the error to the log but below it's being written to the console via println etc. Was that intentional?

senior:
@wkalt I think you're missing facts from this PR, you removed it from acceptance, but I don't see it in test. For me this shows up as an NPE when I run: 
```
lein run benchmark -D -c postgres.ini -r 10 -i 10 -n 10
```

acceptance/performance/facts/fact_load.rb
wkalt: from the Rakefile it looks like :performance may be a separate task than :acceptance. We might not be running this outside the dedicated performance testing job.

src/puppetlabs/puppetdb/cli/benchmark.clj
wkalt: switched it to using log everywhere

wkalt:
@senior addressed your comments

acceptance/performance/facts/fact_load.rb
wkalt: as an addendum to that, now that this is being shipped we should have some real tests around it

wkalt:
I realize that having a flag for --default was a little weird. Now default is inferred if none of -F, -C, -R, or -A are supplied.

documentation/load_testing_tool.markdown
```diff
@@ -61,6 +61,9 @@ are below.
 - **`--catalogs / -C`** --- Directory containing catalogs to use for testing (probably from a previous PuppetDB export)
 - **`--reports / -R`** --- Directory containing reports to use for testing (probably from a previous PuppetDB export)
 - **`--facts / -F`** -- Directory containing facts to use for testing (probably from a previous PuppetDB export)
+- **`--archive / -A`** -- Tarball archive obtained via a PuppetDB export. This option is incompatible with the preceding four.
 - **`--runinterval / -i`** --- Integer indicating the amount of time in minutes between puppet runs for each simulated node, 30 or 60 would be a typical value for this
 - **`--numhosts / -n`** --- Number of separate hosts that the tool should simulate
 - **`--rand-perc / -r`** --- What percentage of catalogs submissions should be changed (i.e. simulating normal/typical catalog changes, as in adding a resource, edge or something similar). More changes to catalogs will cause a higher load on PuppetDB, 10 is a pretty typical change percentage.
+
+**NOTE** If --facts, --catalogs, --reports, and --archive are unspecified, the PuppetDB sample data will be used. This data includes catalogs, facts, and reports.
```
senior: This seems more intuitive

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -292,36 +305,94 @@
 (def required-cli-options
   [:config])
 
+(defn activate-logging!
+  [options]
+  (-> (:config options)
+      (ks/inis-to-map)
```
senior: No need for these parens, but it looks like this is really just moved from down below

senior:
This is a nice change, I don't know how many times I've had to go fishing for an export tarball for benchmark.

PR #1271, opened 2015-02-23
(PDB-1229) Use schema for reports validation

ajroetker:
Prior to this commit we were using validation.clj, a home-baked version
of prismatic/schema  to validate report maps. This commit removes
validation.clj and changes reports to use prismatic/schema for
validation instead.


wkalt:
aw crap @ajroetker I just left a bunch of comments in here but they're under the commit instead of the file changes so they don't show up in the channel and they're hard to read

ajroetker:
@wkalt Addressed your comments!

src/puppetlabs/puppetdb/reports.clj
```diff
@@ -3,93 +3,45 @@
...
-    (throw (IllegalArgumentException. "Version 5 of reports must contain an environment")))
-  (when (nil? (:status report))
-    (throw (IllegalArgumentException. "Version 5 of reports must contain a status")))
-  report)
-
-(defn sanitize-events
+  (:require [schema.core :as s]
+            [puppetlabs.puppetdb.schema :as pls]))
+
+(def full-resource-event
```
wkalt: why full?

src/puppetlabs/puppetdb/command.clj
```diff
@@ -235,8 +235,9 @@
 (defn store-report*
   [version db {:keys [payload annotations]}]
   (let [id          (:id annotations)
-        report      (upon-error-throw-fatality
-                     (report/validate! version payload))
+        report      (->> payload
+                         (s/validate report/full-report)
```
wkalt: this is getting validated twice on the way in, once in command and immediately afterward in storage. I can see that we're also doing that with some of the other commands, but it seems unnecessary. I'm not sure that it matters where the validation goes; all things equal I think storage is probably better because then it gets hit by a lot more test code.

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -31,11 +32,10 @@
...
    (contains? event "resource_type")))
 
-(defn report?
+(pls/defn-validated report? :- report/full-report
   "Returns true if it looks like a report"
-  [report version]
-  ;; Utilise our existing report validation
-  (report/validate! version (keywordize-keys report)))
+  [report _]
+  (keywordize-keys report))
```
wkalt: sorry, what I meant was that the need for this function could be eliminated by just using schema on anonymize-report instead of pre/postchecks

src/puppetlabs/puppetdb/reports.clj
ajroetker: That's what we do with catalogs, I just figured i would follow the same convention

src/puppetlabs/puppetdb/command.clj
wkalt: on further discussion in the channel looks like this isn't a big deal

wkalt:

ajroetker:

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -31,11 +32,10 @@
    (contains? event "old_value")
    (contains? event "resource_type")))
 
-(defn report?
+(pls/defn-validated report? :- report/report-schema
```
kbarber: Should we just switch anonymize-report to use prismatic schema instead and completely drop this function?

src/puppetlabs/puppetdb/reports.clj
```diff
@@ -3,93 +3,45 @@
...
+  {:certname                 s/Str
+   :puppet_version           s/Str
+   :report_format            s/Int
+   :configuration_version    s/Str
+   :start_time               pls/Timestamp
+   :end_time                 pls/Timestamp
+   :resource_events          [resource-event-schema]
+   :noop                     (s/maybe s/Bool)
+   :transaction_uuid         (s/maybe s/Str)
+   :environment              s/Str
```
kbarber: Have we made this & status now mandatory on purpose?

src/puppetlabs/puppetdb/command.clj
```diff
@@ -235,16 +235,16 @@
 (defn store-report*
   [version db {:keys [payload annotations]}]
   (let [id          (:id annotations)
-        report      (upon-error-throw-fatality
-                     (report/validate! version payload))
-        certname    (:certname report)
+        {:keys [certname puppet_version] :as report} (->> payload
+                                                          (s/validate report/report-schema)
```
kbarber: Is there a better way to avoid an explicit call to s/validate? For example, we could switch store-report* or its upstream to be a defn-validated.

src/puppetlabs/puppetdb/schema.clj
```diff
@@ -57,6 +58,9 @@
 (def Function
   (s/pred fn? 'fn?))
 
+(def JSONable
```
kbarber: A quick docstring would be nice here. Looks like Function wasn't doc'd either, wanna add that while you're in there?

src/puppetlabs/puppetdb/anonymizer.clj
ajroetker: So @wkalt had also mentioned we should do this and I changed it but I think this would require us to use `string` keys instead of `keyword` keys everywhere inside the anonymizer code which is a pretty significant change and would be inconsistent with catalogs for instance.

wkalt: you can stringify the keys of the schema inline

src/puppetlabs/puppetdb/reports.clj
ajroetker: They were mandatory for `v5` I thought based on some of the code I removed below.

src/puppetlabs/puppetdb/anonymizer.clj
wkalt: actually this function doesn't do what it says anymore, it's just a wrapper around keywordize-keys

wkalt: oh sorry I'm blind ^^^

wkalt: I think this is fine with me now, I see where AJ is coming from. Using report? is consistent at least with how we're treating the other data types. I don't think stringifying inline is notably cleaner.

kbarber: I don't think we should use the other functions in this file for testing as good examples of how to do this, they are pre-prismatic schema also. And yeah, this isn't a test any more - its keywordizes as well ... which is kind of redundant right?

src/puppetlabs/puppetdb/command.clj
ajroetker: I could think of a view choices here, for instance we're already validating at `add-report`. My main reason for keeping it was I'm not quite sure what `upon-error-throw-fatality` does, and my other reason is the we basically do the same thing for catalogs via `parse-catalog`.

src/puppetlabs/puppetdb/anonymizer.clj
kbarber: Maybe we need to remove the anonymisation out of the equation, if we require more keywordize work to make this work consistenly, perhaps thats for another day. Right now the fact? report? crud, is using manual validation, maybe just make report? work that way, its consistently terrible, but at least it avoids the need to fix the keywordize stuff straight away.

kbarber: Nevermind, keyword is right yeah ... sorry. Umm. This still feels wrong, hang on let me think about this.

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -389,8 +389,9 @@
 (defn anonymize-report
   "Anonymize a report"
   [config version report]
-  {:pre  [(report? report version)]
-   :post [(report? % version)]}
+  {:pre  [(report? report)]}
+  {:post [#(report? %)]}
```
wkalt: I don't think # is right here

wkalt: the :pre and :post should be in the same map, then you can use (report? %). As written this will validate true on output no matter what.

ajroetker: whoops nice catch

wkalt:

wkalt:

ajroetker:

ajroetker:

wkalt:

ajroetker:

wkalt:

ajroetker:

wkalt:

wkalt:

ajroetker:

src/puppetlabs/puppetdb/anonymizer.clj
kbarber: Nope I get it, it sucks but yeah, I see why we need this. Not that important, but it no longer returns false when it doesn't validate - I guess it just throws a validation exception instead. I figure this won't be a problem?

ajroetker:

ajroetker:

ajroetker:

ajroetker:

wkalt:
+ 1

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -17,25 +18,15 @@
...
-   (contains? event "timestamp")
-   (contains? event "resource_title")
-   (contains? event "property")
-   (contains? event "message")
-   (contains? event "new_value")
-   (contains? event "old_value")
-   (contains? event "resource_type")))
-
-(defn report?
+  (keywordize-keys event))
```
senior: So is this a predicate that munges the keys of an event and returns it? Seems like the function name no longer matches it's body.

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -17,25 +18,15 @@
...
+
+(pls/defn-validated report? :- report/report-schema
   "Returns true if it looks like a report"
-  [report version]
-  ;; Utilise our existing report validation
-  (report/validate! version (keywordize-keys report)))
+  [report]
+  (keywordize-keys report))
```
senior: This looks like the same problem as above

src/puppetlabs/puppetdb/anonymizer.clj
ajroetker: @senior it does that same kind of validation as before though, to a consumer of the function via pre/post conditions, it is the same. If we had consumers outside of pre and post conditions I suppose it wouldn't make much sense.

senior:
It's great to get rid of the older validation stuff, just 2 minor changes above and I'm a +1

wkalt:

PR #1278, opened 2015-02-25
(PDB-1254) move latest report to certnames table

wkalt:
Prior to this commit we were storing two columns of certnames, one in latest_reports
and one in certnames. This moves the report_id column of latest_reports to a new
latest_report column in certnames, then dropso the latest_reports table to eliminate
the redundancy.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -890,7 +890,7 @@
     "ALTER TABLE factsets ADD hash VARCHAR(40)"
     "ALTER TABLE factsets ADD CONSTRAINT factsets_hash_key UNIQUE (hash)"))
 
-(defn migrate-to-report-id-and-noop-column
+(defn migrate-to-report-id-and-noop-column-and-drop-latest-reports
   "Migrate to report id and
    insert a column in reports to be populated by boolean noop flag"
   []
@@ -927,10 +927,10 @@
                       ["containment_path" (scf-utils/sql-array-type-string "TEXT")]
                       ["containing_class" "varchar(255)"])
 
-    (sql/create-table :latest_reports_transform
-                      ["certname"  "text NOT NULL"]
-                      ["report_id" "bigint NOT NULL"])
-
+    (sql/create-table :certnames_transform
+                      ["name"  "text NOT NULL"]
+                      ["latest_report_id" "bigint"]
+                      ["deactivated" "timestamp with time zone"])
 
     (sql/do-commands
      "INSERT INTO reports_transform (
@@ -958,19 +958,26 @@
           INNER JOIN reports_transform rt on re.report = rt.hash")
 
     (sql/do-commands
-     "INSERT INTO latest_reports_transform(certname, report_id)
-        SELECT lr.certname, rt.id
-          FROM latest_reports AS lr
-          INNER JOIN reports_transform rt on lr.report = rt.hash")
+
+      "INSERT INTO certnames_transform(name,latest_report_id,deactivated)
+       SELECT c.name, rt.id as latest_report_id, c.deactivated FROM
+       certnames c left outer join latest_reports lr on c.name=lr.certname
+       inner join reports_transform rt on lr.report=rt.hash"
+
+      "ALTER TABLE edges DROP CONSTRAINT edges_certname_fkey"
+      "ALTER TABLE catalogs DROP CONSTRAINT catalogs_certname_fkey"
+      "ALTER TABLE factsets DROP CONSTRAINT factsets_certname_fk"
+
+      "DROP TABLE latest_reports"
+      "DROP TABLE certnames CASCADE"
+      "ALTER TABLE certnames_transform RENAME TO certnames")
 
     (sql/do-commands
      "DROP TABLE resource_events"
-     "DROP TABLE latest_reports"
      "DROP TABLE reports")
 
     (sql/do-commands
      "ALTER TABLE resource_events_transform RENAME to resource_events"
-     "ALTER TABLE latest_reports_transform RENAME to latest_reports"
      "ALTER TABLE reports_transform RENAME to reports")
 
     (sql/do-commands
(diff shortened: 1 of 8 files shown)


wkalt:
I started to write release notes for this but decided it wasn't user-facing or interesting enough. Happy to do whatever people want.

kbarber:
@wkalt We already have a warning in the release notes now about migrations, I'd add a new sub-bullet point about this there. Its user facing, in that we need to warn users a migration is happening (which we already have for this release) and studious users will want to know why this is useful perhaps also, to justify their waiting. I don't think a lot of detail is required, just a brief mention of what and a why.

kbarber:
@wkalt actually, let me restate this, I was totally looking at the wrong thing.

We should mention it in the release notes, as an improvement I think. I think its interesting, and users want to know we are making improvements in the storage side I think. We already list the report_id fix, this is just like that, maybe not as big but close.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1001,6 +1001,30 @@
      "CREATE INDEX idx_resource_events_timestamp ON resource_events(timestamp)"
      "ALTER TABLE resource_events ADD CONSTRAINT resource_events_report_id_fkey FOREIGN KEY (report_id) REFERENCES reports(id) ON DELETE CASCADE"))
 
+(defn drop-latest-reports
```
kbarber: Yeah, I'd merge this up with the last migration if possible. We are already dropping and recreating latest_report in the last migration, seems silly to just then go and remove it.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1001,6 +1001,30 @@
...
+  (sql/do-commands
+    (if (scf-utils/postgres?)
+      "SELECT c.name,c.deactivated,lr.report_id as latest_report INTO certnames_transform FROM certnames c LEFT OUTER JOIN latest_reports lr ON c.name=lr.certname"
+      "CREATE TABLE certnames_transform (name, deactivated, latest_report) AS (SELECT c.name,c.deactivated,lr.report_id as latest_report FROM certnames c LEFT OUTER JOIN latest_reports lr ON c.name=lr.certname) WITH DATA")
+    "ALTER TABLE edges DROP CONSTRAINT edges_certname_fkey"
+    "ALTER TABLE catalogs DROP CONSTRAINT catalogs_certname_fkey"
+    "ALTER TABLE factsets DROP CONSTRAINT factsets_certname_fk"
+    "ALTER TABLE reports DROP CONSTRAINT reports_certname_fkey"
+    "ALTER TABLE latest_reports DROP CONSTRAINT latest_reports_certname_fkey"
+    "DROP TABLE certnames"
```
kbarber: Add an id to certnames is so tempting here, but I'd avoid it for now, we don't want to boil the ocean.

wkalt:
@kbarber added a note and combined the migrations

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1172,6 +1171,8 @@
   the specified date/time."
   [time]
   {:pre [(kitchensink/datetime? time)]}
+  (sql/update-values :certnames ["latest_report in (select id from reports where end_time < ?)" (to-timestamp time)]
+                     {:latest_report nil})
```
wkalt: This handles the nullification of latest_report in the certnames table when the corresponding report is deleted. I had hoped to handle this with a foreign key constraint with set null on delete, but this created a circular relationship between certnames and reports which hsqldb could not handle. I think this should be adequate because I think this is the only place we're deleting reports, but it's something to watch out for.

wkalt:

wkalt:
just rebased this

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -999,7 +1004,15 @@
      "CREATE INDEX idx_resource_events_resource_title ON resource_events(resource_title)"
      "CREATE INDEX idx_resource_events_status ON resource_events(status)"
      "CREATE INDEX idx_resource_events_timestamp ON resource_events(timestamp)"
-     "ALTER TABLE resource_events ADD CONSTRAINT resource_events_report_id_fkey FOREIGN KEY (report_id) REFERENCES reports(id) ON DELETE CASCADE"))
+     "ALTER TABLE resource_events ADD CONSTRAINT resource_events_report_id_fkey FOREIGN KEY (report_id) REFERENCES reports(id) ON DELETE CASCADE")
+
+(sql/do-commands
+      ;; recreate indices
+      "ALTER TABLE certnames ADD CONSTRAINT certnames_pkey PRIMARY KEY (name)"
```
kbarber: We've missing the index for report_id ... https://gist.github.com/kbarber/2cfb762adcf87a1b243c

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -999,7 +1004,15 @@
      "CREATE INDEX idx_resource_events_resource_title ON resource_events(resource_title)"
      "CREATE INDEX idx_resource_events_status ON resource_events(status)"
      "CREATE INDEX idx_resource_events_timestamp ON resource_events(timestamp)"
-     "ALTER TABLE resource_events ADD CONSTRAINT resource_events_report_id_fkey FOREIGN KEY (report_id) REFERENCES reports(id) ON DELETE CASCADE"))
+     "ALTER TABLE resource_events ADD CONSTRAINT resource_events_report_id_fkey FOREIGN KEY (report_id) REFERENCES reports(id) ON DELETE CASCADE")
+
+(sql/do-commands
```
kbarber: Indent

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1073,10 +1073,9 @@
                                             ORDER BY end_time DESC
                                             LIMIT 1" node])))]
     (sql/update-or-insert-values
```
kbarber: Should we just switch to sql/update-values ?

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -927,10 +927,10 @@
                       ["containment_path" (scf-utils/sql-array-type-string "TEXT")]
                       ["containing_class" "varchar(255)"])
 
-    (sql/create-table :latest_reports_transform
-                      ["certname"  "text NOT NULL"]
-                      ["report_id" "bigint NOT NULL"])
-
+    (sql/create-table :certnames_transform
+                      ["name"  "text NOT NULL"]
+                      ["latest_report" "bigint"]
```
kbarber: For consistency with other FK column names: latest_report_id

src/puppetlabs/puppetdb/scf/storage.clj
kbarber: Gotcha, is it possible to do the cascade for postgresql, and the nullification here for hsqldb only? Would mean if we retire hsqldb it would be more obvious for us (since grep would get this) + postgresql gets the better behaviour. Also a comment or update to the docstring as to the 'why' for this update would be useful.

wkalt: yeah, I see no issue with that

src/puppetlabs/puppetdb/scf/migrate.clj
wkalt: good catch.

wkalt:
@kbarber addressed your comments

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -999,7 +999,19 @@
      "CREATE INDEX idx_resource_events_resource_title ON resource_events(resource_title)"
      "CREATE INDEX idx_resource_events_status ON resource_events(status)"
      "CREATE INDEX idx_resource_events_timestamp ON resource_events(timestamp)"
-     "ALTER TABLE resource_events ADD CONSTRAINT resource_events_report_id_fkey FOREIGN KEY (report_id) REFERENCES reports(id) ON DELETE CASCADE"))
+     "ALTER TABLE resource_events ADD CONSTRAINT resource_events_report_id_fkey FOREIGN KEY (report_id) REFERENCES reports(id) ON DELETE CASCADE")
+
+    (sql/do-commands
+      "ALTER TABLE certnames ADD CONSTRAINT certnames_pkey PRIMARY KEY (name)"
+      "CREATE INDEX idx_certnames_latest_report_id ON certnames(latest_report_id)"
```
kbarber: We should be switching to the certnames_latest_report_id_idx format from now on. I just noticed we let this slip in the report_id changes though ...

senior:
+1

kbarber:
Merged manually with naming convention fixes.

PR #1279, opened 2015-02-25
(PDB-908) Remove dyanmic from mq-endpoint, test function refactors

senior:
Making mq-endpoint ended up not being useful in running multiple
PuppetDB command queues, this commit removes it. This also moves a
couple of now shared functions to a more reusable place.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -76,8 +76,12 @@
 ;; PuppetDB components.
 
 (def mq-addr "vm://localhost?jms.prefetchPolicy.all=1&create=false")
-(def ^:dynamic mq-endpoint "puppetlabs.puppetdb.commands")
-(def send-command! (partial command/enqueue-command! mq-addr mq-endpoint))
+(def mq-endpoint "puppetlabs.puppetdb.commands")
+
+(defn send-command!
+  "Send the command using the default MQ address and endpoint"
+  [command version payload]
+  (command/enqueue-command! mq-addr mq-endpoint command version payload))
 
 (defn auto-deactivate-nodes!
   "Deactivate nodes which haven't had any activity (catalog/fact submission)
@@ -302,8 +306,8 @@
                     (assoc context :updater updater)
                     context)
           _       (let [authorizer (if-let [wl (:certificate-whitelist puppetdb)]
-                                          (build-whitelist-authorizer wl)
-                                          (constantly :authorized))
+                                     (build-whitelist-authorizer wl)
+                                     (constantly :authorized))
                         app (server/build-app :globals globals :authorizer authorizer)]
                     (log/info "Starting query server")
                     (add-ring-handler service (compojure/context url-prefix [] app)))
--- a/src/puppetlabs/puppetdb/query/factsets.clj
+++ b/src/puppetlabs/puppetdb/query/factsets.clj
@@ -43,14 +43,6 @@
 
 ;; FUNCS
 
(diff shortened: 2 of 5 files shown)


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -76,8 +76,12 @@
 ;; PuppetDB components.
 
 (def mq-addr "vm://localhost?jms.prefetchPolicy.all=1&create=false")
-(def ^:dynamic mq-endpoint "puppetlabs.puppetdb.commands")
-(def send-command! (partial command/enqueue-command! mq-addr mq-endpoint))
+(def mq-endpoint "puppetlabs.puppetdb.commands")
+
+(defn send-command!
```
senior: There's a subtle difference between this code and what was there before. Subtle enough to lose a couple of days you could say :-). One issue in setting up two PuppetDB instances in-process is creating separate command queues for each instance. The previous code closed over the value of mq-endpoint when the instance started up. Even if the mq-endpoint was dynamically bound (which didn't work due to the various threads involved), it didn't matter, because it was closed over at startup. The above code uses the value at the time the command is invoked. This gives enough wiggle room to change the queue name after startup, but before the commands are posted.

test/puppetlabs/puppetdb/cli/import_export_roundtrip_test.clj
```diff
@@ -196,14 +185,3 @@
   (check-invalid-url-handling
    #(#'import/main "--host" "local:host" "--infile" "/dev/null" "--port" 10000)
    #"^Invalid destination .*"))
-
-(deftest changing-queue-destination-name
```
senior: Just binding a different destination name is no good. There are threads listening to the queue that need to have the correct value and HTTP requests coming in (also on a different thread) that need to get this value. A global with-redefs works better in this case, so I removed the dynamic tag and removed the test.

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -263,8 +264,7 @@
                                                      (.setUseJmx broker# false))]
      (do ~@body)))
 
-(defmacro with-command-endpoint
-  "Invoke `body` with a different command endpoint (destination) name"
-  [new-endpoint-name & body]
-  `(binding [puppetlabs.puppetdb.cli.services/mq-endpoint ~new-endpoint-name]
-     (do ~@body)))
+(defn sync-command-post [base-url cmd version payload]
```
senior: Created a ticket for refactoring [jetty.clj](https://tickets.puppetlabs.com/browse/PDB-1255)

wkalt: @senior this is the function I meant

wkalt:

src/puppetlabs/puppetdb/query/factsets.clj
```diff
@@ -43,14 +43,6 @@
...
 ;; FUNCS
 
-(defn create-certname-pred
-  "Create a function to compare the certnames in a list of
-  rows with that of the first row."
-  [rows]
-  (let [certname (:certname (first rows))]
-    (fn [row]
-      (= certname (:certname row)))))
-
```
rbrw: Presumably this move is just incidental?

rbrw:
Looks fine to me.

src/puppetlabs/puppetdb/query/factsets.clj
senior: This was moved to utils because it's needed by the new sync code.

PR #1280, opened 2015-02-25
(maint) Switch us to use c4.large instances for acceptance

kbarber:
The c3.large instances are starting to hit capacity in Oregon, this switches
us to c4.large which are ever-so-slightly faster, but lack the SSD disks (we
don't use these anyway).


kbarber:
This needs other fixes in place for beaker to support hvm, closing for now.

PR #1282, opened 2015-02-26
(PDB-1247) Upgrade TK, move certificate-whitelist

mullr:
This is a cherry-pick of commit ea8f696, whose commit message is as 
follows:

This patch upgrades trapperkeeper, and the jetty component to the latest
revision. All lines of this patch are attributable to getting this to
work.

The major change was the change to certificate-whitelist, which is
not supported in trapperkeeper-webserver-jetty9 so we had to intercept
the configuration item, deprecate it and move it to a new [puppetdb] section
so we could do this work ourselves.

We avoid using [global] in this instance, since we have identified
a potential to clash if we want to bundle PDB with other clojure applications,
this is a general problem anyway - but our first move to try and reduce
this potential.

### DIFF
--- a/acceptance/tests/security/cert_whitelist_legacy.rb
+++ b/acceptance/tests/security/cert_whitelist_legacy.rb
@@ -1,4 +1,4 @@
-test_name "certificate whitelisting" do
+test_name "certificate whitelisting using legacy location in [jetty]" do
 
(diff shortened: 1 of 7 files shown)


mullr:
closing this, created it with the wrong branch name.

PR #1283, opened 2015-02-26
(PDB-1247) Upgrade TK to get max-threads support

mullr:
This includes cherry-picks of ea8f696:

    This patch upgrades trapperkeeper, and the jetty component to the latest
    revision. All lines of this patch are attributable to getting this to
    work.

    The major change was the change to certificate-whitelist, which is
    not supported in trapperkeeper-webserver-jetty9 so we had to intercept
    the configuration item, deprecate it and move it to a new [puppetdb] section
    so we could do this work ourselves.

    We avoid using [global] in this instance, since we have identified
    a potential to clash if we want to bundle PDB with other clojure applications,
    this is a general problem anyway - but our first move to try and reduce
    this potential.

and 1476b28:

    Correct the ssl-protocols rewrite for the master branch

    Right now tk-webserver-jetty9 0.7.5 requires a list to be passed to it for
    the setting `ssl-protocols` but our rewrite still provides a comma separated
    string.

    While this is a bug in-and-of-itself, its causing our tests to fail. This
    patch gets us back to passing. We'll look at the configuration issue as a
    part of the fixes going into TK that might resolve this.

### DIFF
--- a/acceptance/tests/security/cert_whitelist_legacy.rb
+++ b/acceptance/tests/security/cert_whitelist_legacy.rb
@@ -1,4 +1,4 @@
-test_name "certificate whitelisting" do
+test_name "certificate whitelisting using legacy location in [jetty]" do
 
   confd = "#{puppetdb_confdir(database)}/conf.d"
   dbname = database.node_name
@@ -7,10 +7,10 @@ test_name "certificate whitelisting" do
   ssldir = stdout.chomp
 
   step "reconfigure PuppetDB to use certificate whitelist" do
-    on database, "cp #{confd}/config.ini #{confd}/config.ini.bak"
-    on database, "grep -v ^certificate-whitelist #{confd}/config.ini > #{confd}/config.ini.tmp"
-    on database, "mv -f #{confd}/config.ini.tmp #{confd}/config.ini"
-    modify_config_setting(database, "config.ini", "puppetdb",
+    on database, "cp #{confd}/jetty.ini #{confd}/jetty.ini.bak"
+    on database, "grep -v ^certificate-whitelist #{confd}/jetty.ini > #{confd}/jetty.ini.tmp"
+    on database, "mv -f #{confd}/jetty.ini.tmp #{confd}/jetty.ini"
+    modify_config_setting(database, "jetty.ini", "jetty",
                           "certificate-whitelist", "#{confd}/whitelist")
   end
 
@@ -20,7 +20,7 @@ test_name "certificate whitelisting" do
     create_remote_file database, "#{confd}/whitelist", whitelist
     on database, "chmod 644 #{confd}/whitelist"
     restart_puppetdb database
-    on database, "curl --tlsv1 -sL -w '%{http_code}\\n' " +
+    on database, "curl -sL -w '%{http_code}\\n' " +
                  "--cacert #{ssldir}/certs/ca.pem " +
                  "--cert #{ssldir}/certs/#{dbname}.pem " +
                  "--key #{ssldir}/private_keys/#{dbname}.pem "+
@@ -38,9 +38,9 @@ test_name "certificate whitelisting" do
     curl_against_whitelist.call "", 403
   end
 
-  step "restore original config.ini" do
-    on database, "mv #{confd}/config.ini.bak #{confd}/config.ini"
-    on database, "chmod 644 #{confd}/config.ini"
+  step "restore original jetty.ini" do
+    on database, "mv #{confd}/jetty.ini.bak #{confd}/jetty.ini"
+    on database, "chmod 644 #{confd}/jetty.ini"
     on database, "rm #{confd}/whitelist"
     restart_puppetdb database
   end
--- a/acceptance/tests/security/cert_whitelist_legacy.rb
+++ b/acceptance/tests/security/cert_whitelist_legacy.rb
@@ -20,7 +20,7 @@ test_name "certificate whitelisting using legacy location in [jetty]" do
     create_remote_file database, "#{confd}/whitelist", whitelist
     on database, "chmod 644 #{confd}/whitelist"
     restart_puppetdb database
-    on database, "curl -sL -w '%{http_code}\\n' " +
+    on database, "curl --tlsv1 -sL -w '%{http_code}\\n' " +
                  "--cacert #{ssldir}/certs/ca.pem " +
                  "--cert #{ssldir}/certs/#{dbname}.pem " +
                  "--key #{ssldir}/private_keys/#{dbname}.pem "+
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -184,7 +184,7 @@ this setting, you must also set the corresponding setting in your Puppet Master'
 `[puppetdb]` Settings
 -----
 
-The `[puppetdb]` section is used to configure PuppetDB application specific behavior.
+The `[puppetdb]` section is used to configure PuppetDB application-specific behavior.
 
 ### `certificate-whitelist`
 
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -180,6 +180,19 @@ This optional setting may be used to mount the PuppetDB web application at a URL
 unless you intend to run additional web applications in the same server with your PuppetDB instance.  **NOTE:** if you change
 this setting, you must also set the corresponding setting in your Puppet Master's [puppetdb.conf][puppetdb.conf] file.
 
+
+`[puppetdb]` Settings
+-----
+
+The `[puppetdb]` section is used to configure PuppetDB application specific behavior.
+
+### `certificate-whitelist`
+
+Optional. This describes the path to a file that contains a list of certificate names, one per line.  Incoming HTTPS requests will have their certificates validated against this list of names and only those with an _exact_ matching entry will be allowed through. (For a puppet master, this compares against the value of the `certname` setting, rather than the `dns_alt_names` setting.)
+
+If not supplied, PuppetDB uses standard HTTPS without any additional authorization. All HTTPS clients must still supply valid, verifiable SSL client certificates.
+
+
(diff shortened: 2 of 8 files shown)


test/com/puppetlabs/puppetdb/test/config.clj
```diff
@@ -210,13 +210,13 @@
 (deftest sslv3-warn-test
   (testing "output to log"
     (tu-log/with-log-output log-output
-      (let [bad-config {:jetty {:ssl-protocols "SSLv3, TLSv1, TLSv1.1, TLSv1.2"}}]
+      (let [bad-config {:jetty {:ssl-protocols ["SSLv3" "TLSv1" "TLSv1.1" "TLSv1.2"]}}]
```
senior: @cprice404 @camlow325 Looks like [TK-133](https://tickets.puppetlabs.com/browse/TK-133) is now more important. We're wanting to upgrade TK so that we can get [max-threads](https://tickets.puppetlabs.com/browse/TK-130) but since this is wanted in stable which is a y release, we can't break [ssl-protocols](https://docs.puppetlabs.com/puppetdb/2.2/configure.html#ssl-protocols) like we did on master.

cprice404: @senior PRs welcome :)

mullr:
This depends on tk-jetty9/master; leaving it here so we don't forget about it once the new version has been released.

mullr:
Updated to the released tk-jetty9, this should be good to go now. I would like to do a deliberate test on this once it gets merged, just to make sure everything's kosher; any thoughts on how to go about that?

documentation/configure.markdown
```diff
@@ -180,6 +180,19 @@ This optional setting may be used to mount the PuppetDB web application at a URL
 unless you intend to run additional web applications in the same server with your PuppetDB instance.  **NOTE:** if you change
 this setting, you must also set the corresponding setting in your Puppet Master's [puppetdb.conf][puppetdb.conf] file.
 
+
+`[puppetdb]` Settings
+-----
+
+The `[puppetdb]` section is used to configure PuppetDB application specific behavior.
```
wkalt: application-specific

src/com/puppetlabs/puppetdb/config.clj
```diff
@@ -208,13 +211,20 @@
     (s/validate command-processing-out converted-config)
     (assoc config :command-processing converted-config)))
 
+(defn configure-puppetdb
+  "Validates the [puppetdb] section of the config"
+  [{:keys [puppetdb] :as config :or {puppetdb {}}}]
```
wkalt: Would it make sense to use s/maybe in puppetdb-config-in rather than defaulting to an empty map here? Defaulting like this seems weird to me but I'm not sure of the downstream implications.

src/com/puppetlabs/puppetdb/config.clj
```diff
@@ -320,25 +330,21 @@
...
                      (utils/assoc-when :update-server "http://updates.puppetlabs.com/check-for-updates"))))))
 
-(defn warn-if-sslv3
-  "If the ssl-protocols config is present and contains sslv3, warn the user. Logging
-   at this point may or may not be setup, so warning via *err* and logging is done."
+(defn fix-tk-config
+  "Fix configuration before passing it to trapperkeeper.
+
+   In particular:
+   * Move certificate-whitelist from [jetty] to [global]"
```
wkalt: I think if we rename the function we can use just this line of the docstring.

src/com/puppetlabs/puppetdb/config.clj
```diff
@@ -320,25 +330,21 @@
...
-  config-data)
-
-(defn default-ssl-protocols
-  "Provide a default for ssl-protocols (that does NOT include sslv3). If sslv3
-   is present in the user provided config, warn the user."
-  [config-data]
-  (if (get-in config-data [:jetty :ssl-protocols])
-    (warn-if-sslv3 config-data)
-    (assoc-in config-data [:jetty :ssl-protocols] "TLSv1, TLSv1.1, TLSv1.2")))
+          (println "Option `certificate-whitelist` in [jetty] is now deprecated, the option must now be placed in [puppetdb]"))
```
wkalt: I think this reads better with the comma swapped for a period.

src/com/puppetlabs/puppetdb/config.clj
```diff
@@ -320,25 +330,21 @@
...
+  "Fix configuration before passing it to trapperkeeper.
+
+   In particular:
+   * Move certificate-whitelist from [jetty] to [global]"
   [config-data]
-  (when-let [protocol-str (get-in config-data [:jetty :ssl-protocols])]
-    (when (re-matches #"(?i).*sslv3.*" protocol-str)
+  (if-let [cw (get-in config-data [:jetty :certificate-whitelist])]
+    (do
+      ;; Log to stderr, logging is not yet initialized (and may never be).
```
wkalt: probably not within scope here, but I wonder if we need a better pattern for this

wkalt: the issue being that it logs to console only, rather than respecting the user's config

acceptance/tests/security/cert_whitelist_legacy.rb
```diff
@@ -0,0 +1,47 @@
...
+    on database, "chmod 644 #{confd}/whitelist"
+    restart_puppetdb database
+    on database, "curl -sL -w '%{http_code}\\n' " +
+                 "--cacert #{ssldir}/certs/ca.pem " +
+                 "--cert #{ssldir}/certs/#{dbname}.pem " +
+                 "--key #{ssldir}/private_keys/#{dbname}.pem "+
+                 "https://#{dbname}:8081/v3/metrics/mbeans " +
+                 "-o /dev/null"
+    actual_status_code = stdout.chomp
+    assert_equal expected_status_code.to_s, actual_status_code, "Reqs from #{dbname} with whitelist '#{whitelist}' should return #{expected_status_code}, not #{actual_status_code}"
```
wkalt: Request

src/com/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -369,6 +371,10 @@
         (stop-puppetdb context)))
 
 (defn -main
+  "Calls the trapperkeeper main argument to initialize tk.
+
+   For configuration customization, we intercept the call to parse-config-data
+   within TK."
```
wkalt: I think this is self-evident unless it explains what is happening.

src/com/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -57,7 +57,9 @@
             [com.puppetlabs.puppetdb.config :as conf]
             [com.puppetlabs.puppetdb.utils :as utils]
             [puppetlabs.kitchensink.core :as kitchensink]
+            [robert.hooke :as rh]
```
senior: This is already declared in this ns

src/com/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -57,7 +57,9 @@
...
             [puppetlabs.trapperkeeper.core :refer [defservice main]]
+            [puppetlabs.trapperkeeper.services :refer [service-id]]
```
senior: Oh funny, so you didn't add any new references to service-id other than here, but the code before worked. We must be getting it from the anaphora in  the service macro.

src/com/puppetlabs/puppetdb/config.clj
senior: You can swap the binding for [this] (https://github.com/puppetlabs/puppetdb/blob/stable/src/com/puppetlabs/puppetdb/utils.clj#L32)

src/com/puppetlabs/puppetdb/config.clj
senior: Yeah, we've had bugs related to that too, but we'd either need to make significant changes to config or get fancy with our logback usage at start time

src/com/puppetlabs/puppetdb/config.clj
senior: I think you mean [puppetdb] rather than global here? The code isn't moving it to global

src/com/puppetlabs/puppetdb/cli/services.clj
mullr: I believe you, but I can't for the life of me find the code in trapperkeeper that does this. It must been several macrons deep. In any case, I'll remove the import since it's misleading.

src/com/puppetlabs/puppetdb/config.clj
mullr: I think defaulting to an empty map is pretty good, since it removes the (irrelevant, to us) distinction between an absent config section and an empty section. In any case, this is how it's done elsewhere in stable, and its refactored into `convert-config` in master, with similar defaulting behavior.

src/com/puppetlabs/puppetdb/cli/services.clj
mullr: service-id, interestingly, is not automatically introduced by defservice. Perhaps an older TK did it that way.

src/com/puppetlabs/puppetdb/config.clj
wkalt: fair enough, I see what you mean

mullr:
Updated to address the above comments. I split this into a number of commits to make it easier to merge these into master.

mullr:

wkalt:

senior:
Had a mostly successful run, el5 failed, el5 has passed but now getting gem errors (from upstream). Travis has had some infra issues and jobs are way backed up. I ran the unit tests locally, they're good. I think the combo of the last two acceptance runs are green, merging in and I'll verify there.

PR #1287, opened 2015-02-27
(re-4060)(packaging) Update terminus install paths for puppet 4

shrug:
Update the terminus installation path for puppet 4 to match the paths specified
at https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md,
changing agent/ to puppet/


haus:
:+1:

PR #1288, opened 2015-03-01
(PDB-1259) Faster conversion of strings to timestamps

senior:
Previously we were using clj-time.format/to-timestamps's default
ordering of date formatters to attempt to parse the incoming
string. This just used the values in the formatters map in whatever
order they are stored. These formats are used in the order they come out
of the hash-map until one successfully parses the time. This means we
could potentially have 10+ failures to parse a date before a successful
parse, and we do this each time in each report/catalog/factset etc.

This commit switches to use an ordered list of formatters, using the
most likely parsers to work (i.e. ISO8601), then falls through to the
existing behavior.

### DIFF
--- a/src/puppetlabs/puppetdb/http/events.clj
+++ b/src/puppetlabs/puppetdb/http/events.clj
@@ -3,10 +3,10 @@
   (:require [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.query.paging :as paging]
             [puppetlabs.puppetdb.query-eng :refer [produce-streaming-body]]
-            [clj-time.coerce :refer [to-timestamp]]
             [puppetlabs.puppetdb.middleware :as middleware]
             [net.cgrand.moustache :refer [app]]
-            [puppetlabs.puppetdb.http :as http]))
+            [puppetlabs.puppetdb.http :as http]
+            [puppetlabs.puppetdb.time :refer [to-timestamp]]))
 
 (defn validate-distinct-options!
   "Validate the HTTP query params related to a `distinct_resources` query.  Return a
--- a/src/puppetlabs/puppetdb/query.clj
+++ b/src/puppetlabs/puppetdb/query.clj
@@ -63,7 +63,7 @@
   (:require [clojure.string :as str]
             [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.jdbc :as jdbc]
-            [clj-time.coerce :refer [to-timestamp]]
+            [puppetlabs.puppetdb.time :refer [to-timestamp]]
             [puppetlabs.kitchensink.core :refer [parse-number keyset valset order-by-expr?]]
             [puppetlabs.puppetdb.scf.storage-utils :refer [db-serialize sql-as-numeric sql-array-query-string sql-regexp-match sql-regexp-array-match]]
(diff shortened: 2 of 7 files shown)


src/puppetlabs/puppetdb/http/events.clj
```diff
@@ -3,10 +3,11 @@
   (:require [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.query.paging :as paging]
             [puppetlabs.puppetdb.query-eng :refer [produce-streaming-body]]
-            [clj-time.coerce :refer [to-timestamp]]
             [puppetlabs.puppetdb.middleware :as middleware]
             [net.cgrand.moustache :refer [app]]
-            [puppetlabs.puppetdb.http :as http]))
+            [puppetlabs.puppetdb.http :as http]
+            [puppetlabs.puppetdb.utils :as utils]
```
rbrw: Was utils accidentally included here?

src/puppetlabs/puppetdb/query.clj
```diff
@@ -120,14 +120,14 @@
...
-         (execute-paged-query* fail-limit sql-and-params paging-options)
-         (execute-query* fail-limit sql-and-params)))))
+   {:pre [((some-fn string? sequential?) query)]
+    :post [(map? %)
+           (vector? (:result %))
+           ((some-fn nil? integer?) (:count %))]}
+   (let [sql-and-params (if (string? query) [query] query)]
+     (if (requires-paging? paging-options)
+       (execute-paged-query* fail-limit sql-and-params paging-options)
+       (execute-query* fail-limit sql-and-params)))))
```
rbrw: Not sure where we ended up with respect to whitespace, but this appears incidental (if you were trying to avoid it now).  And it looks like there's more in the rest of the patch.

src/puppetlabs/puppetdb/time.clj
```diff
@@ -201,3 +204,36 @@
...
+                               (remove (set likely-formats) (keys tf/formatters))))))
+
+(s/defn ^:always-validate attempt-date-time-parse
+  "Parses `timestamp-str` using `formatter`. Returns nil if an
+  Exception is thrown, otherwise the parsed DateTime object"
+  [formatter :- DateTimeFormatter
+   timestamp-str :- String]
+  (try
+    (tf/parse formatter timestamp-str)
+    (catch Exception _ nil)))
```
rbrw: Wondered if it could make any sense to try to catch a more specific set of exceptions here, unless we really do want to keep going no matter what (and maybe we do), or unless we don't really know/care what tf/parse might throw.

I'm just thinking (generally) about situations where something might be more seriously wrong, but we have to wade through all the parsers before we hear about it.

senior: I took this code from clj-time, but reading the JodaTime API, it looks like we can expect IllegalArgumentException. I've updated what's there to be more specific.

src/puppetlabs/puppetdb/query.clj
senior: Good catch. I removed my format the whole file hook, but didn't restart emacs after, this has been removed.

src/puppetlabs/puppetdb/http/events.clj
senior: Yes, I had that to-timestamp fn in utils, then moved it. This like is no longer needed.

wkalt:

PR #1289, opened 2015-03-01
(PDB-1255) Switch tests to pick open port before bootstrapping TK

senior:
Previously tests would set Jetty's port to 0, which automatically picks
an open port on startup. The downside of that was the PuppetDB config
wouldn't be updated with the actual port chosen, so that info would have
to be pulled from the Jetty server. This patch switches to determine an
open port before starting TK, then use that port in the TK config. This
makes our test more like what users would run (specifying a port rather
than letting Jetty choose).

### DIFF
--- a/test/puppetlabs/puppetdb/testutils/jetty.clj
+++ b/test/puppetlabs/puppetdb/testutils/jetty.clj
@@ -6,7 +6,7 @@
             [puppetlabs.trapperkeeper.services.webserver.jetty9-service :refer [jetty9-service]]
             [puppetlabs.trapperkeeper.services.webrouting.webrouting-service :refer [webrouting-service]]
             [puppetlabs.puppetdb.client :as pdb-client]
-            [puppetlabs.puppetdb.cli.services :refer [puppetdb-service mq-endpoint]]
+            [puppetlabs.puppetdb.cli.services :as svcs]
             [puppetlabs.puppetdb.metrics :refer [metrics-service]]
             [puppetlabs.puppetdb.mq-listener :refer [message-listener-service]]
             [puppetlabs.puppetdb.command :refer [command-service]]
@@ -17,7 +17,8 @@
             [clj-http.client :as client]
             [clojure.string :as str]
             [fs.core :as fs]
-            [slingshot.slingshot :refer [throw+]]))
+            [slingshot.slingshot :refer [throw+]]
+            [clojure.tools.logging :as log]))
 
 ;; See utils.clj for more information about base-urls.
 (def ^:dynamic *base-url* nil) ; Will not have a :version.
@@ -49,7 +50,7 @@
    :jetty {:port 0}
    :database (fixt/create-db-map)
    :command-processing {}
-   :web-router-service {:puppetlabs.puppetdb.cli.services/puppetdb-service ""
+   :web-router-service {::svcs/puppetdb-service ""
                         :puppetlabs.puppetdb.metrics/metrics-service "/metrics"}})
 
 (defn assoc-logging-config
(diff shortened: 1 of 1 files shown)


test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -60,16 +60,18 @@
...
-      .getConnectors
-      first
-      .getLocalPort))
+(defn open-port-num
+  "Returns a currently open port number"
+  []
+  (let [s (java.net.ServerSocket. 0)
+        local-port (.getLocalPort s)]
+    (.close s)
+    local-port))
```
rbrw: Perhaps (with-open [s ...] (.getLocalPort s))?

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -81,19 +83,22 @@
...
-         (concat [jetty9-service puppetdb-service message-listener-service command-service webrouting-service metrics-service]
+         (concat [jetty9-service svcs/puppetdb-service message-listener-service command-service webrouting-service metrics-service]
                  services)
          config
          (binding [*server* server
                    *base-url* (merge {:protocol "http"
                                       :host "localhost"
-                                      :port (current-port server)}
+                                      :port (get-in config [:jetty :port])}
                                      (when prefix {:prefix prefix}))]
```
rbrw: What happens if the port is taken between (open-port-num) and the server's binding?  As long that's an error, and we don't end up just choosing some other port, then I think we're fine.

senior: Jetty will throw an error if you try to bind to a port that is already bound.

test/puppetlabs/puppetdb/testutils/jetty.clj
senior: Good suggestion. Just pushed up the change.

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -60,16 +60,16 @@
...
-  [server]
-  (-> @(tka/app-context server)
-      (get-in [:WebserverService :jetty9-servers :default :server])
-      .getConnectors
-      first
-      .getLocalPort))
+(defn open-port-num
+  "Returns a currently open port number"
+  []
+  (with-open [s (java.net.ServerSocket. 0)]
```
mullr: There's a race between this process and any others on the same system to use whatever port this is. Can we run jetty with port 0 then interrogate it to find what port it's actually running on?

mullr: This is the same as Rob's concern below, I see now. I suppose such things are unlikely, and that we should probably just go with it. We can re-evaluate if we start seeing port evaluation problems in test runs.

senior: I talked with @mullr about this. The issue is the source of a PuppetDB replication needs to know it's own port number so it can send it's URL to the remote (destination) of the replication. If the port isn't determined until startup, you can't go back and change the config (config has already started). Switching to this model lets the startup happen normally with the port already set.

Also there seems to be some state kept by the server socket, i.e. opening one port, then closing and opening it again picks a different port.

rbrw: @senior might be related to SO_REUSEADDR and/or SO_REUSEPORT (i.e. TTL/linger).

senior:
@mullr @kbarber @rbrw Pushed up some logic to catch problems with binding to the random port and retrying

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -61,42 +62,55 @@
...
+  `services` is a seq of additional services that should be started in
+  addition to the core PuppetDB services. Binds *server* and
+  *base-url* to refer to the instance. `attempts` tracks how many
+  times an attempts were made to start Jetty (defaults to 1)"
   ([f] (puppetdb-instance (create-config) f))
   ([config f] (puppetdb-instance config [] f))
   ([config services f]
-   (let [[log-file config] (-> config conf/adjust-tk-config assoc-logging-config)
+   (puppetdb-instance config services 1 f))
+  ([config services attempts f]
```
rbrw: Minor quibble - I might be tempted to adjust this so that attempts counts down rather than up, and defaults to 10.  That way the (publically visible arg) is useful.  And then I might also be tempted to just hide the implementation via an internal recur (since you'll still need to track the count for the log message).  Other than that, looks good to me.

test/puppetlabs/puppetdb/testutils/jetty.clj
kbarber: Fwiw the only other port interogation technique I can think of is via JMX: http://www.tiikoni.com/tis/view/?id=6f89f04

mullr: We're unfortunately disabling JMX in the replication tests, due to name conflicts when running to pdbs at once. Which is where we need this change in the first place.

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -61,42 +62,55 @@
...
-      .getConnectors
-      first
-      .getLocalPort))
+(defn open-port-num
+  "Returns a currently open port number"
+  []
+  (with-open [s (java.net.ServerSocket. 0)]
+    (.getLocalPort s)))
+
+(defn assoc-open-port
```
ajroetker: Should this have a `!` at the end to imply side-effects?

senior:
@rbrw Switched to decrement instead of increment

test/puppetlabs/puppetdb/testutils/jetty.clj
senior: I don't think so. The convention I follow is functions that directly modify mutable variables have a !, like swap! or reset! in clojure core. Things that have side affects like DB calls or I/O, I don't use that (i.e. spit in clojure core)

senior:

test/puppetlabs/puppetdb/testutils/jetty.clj
```diff
@@ -61,42 +62,56 @@
...
+  failed attempts should be made to bind to an HTTP port before giving
+  up, defaults to 10"
   ([f] (puppetdb-instance (create-config) f))
   ([config f] (puppetdb-instance config [] f))
   ([config services f]
-   (let [[log-file config] (-> config conf/adjust-tk-config assoc-logging-config)
+   (puppetdb-instance config services 10 f))
+  ([config services attempts f]
+   (when (zero? attempts)
+     (throw (RuntimeException. (format "Attempted bind of port %s times and failed, giving up" attempts))))
```
rbrw: Won't this always print "0 times" now?

senior: Fixed

wkalt:

PR #1290, opened 2015-03-03
(PDB-1225) Accommodate move of fact_paths to facts [WIP]

rbrw:
This series (3 patches) migrates the database to use the new facts structure:

    (factset_id fact_path_id fact_value_id)

and changes the code to accommodate that.  When reviewing, it may be easier to look at the changes to storage.clj and facts.clj directly, as compared to the diff.  Or, if we like, I can probably take some time to break up the tip patch into more easily digested pieces.

There are still a few FIXMEs (questions) to consider, but nothing significant (that I know of).


src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -51,6 +51,7 @@
   (:require [clojure.java.jdbc :as sql]
             [clojure.tools.logging :as log]
             [clojure.string :as string]
+            [com.puppetlabs.puppetdb.scf.migration-baggage :as baggage]
```
wkalt: I think legacy sounds better than baggage

rbrw: Right, I'd figured we might change that.  Legacy sounds good -- done.

rbrw:

src/com/puppetlabs/puppetdb/facts.clj
```diff
@@ -171,21 +183,29 @@
...
+  [path :- fact-path]
+  {:path (factpath-to-string path)
+   :name (first path)
+   :depth (dec (count path))})
 
-(pls/defn-validated factmap-to-paths :- [fact-path-map]
-  "Converts a map of facts to a list of `fact-path-map`s."
-  [hash :- fact-set]
-  (factmap-to-paths* hash))
+;; FIXME: doesn't look like prismatic can handle [[x y]]
```
senior: I'm not understanding what you're wanting to handle here with Prismatic, is it just a list of pairs?

rbrw:
Exactly, and perhaps I was doing it wrong, but [[s/Int s/Int]] won't work, and I believe I found discussion about it in a bug report.

src/com/puppetlabs/puppetdb/facts.clj
```diff
@@ -11,17 +11,37 @@
 
 ;; SCHEMA
 
+;; Terminology:
+;;   facts: {"kernel" "Linux" ...}
+;;   factmap: {:factset_id ... :fact_path_id ... fact_value_id ...}
+;;   fact-path path: ["foo" "bar"]
+;;   pathstr: "foo~#bar..."
+;;   pathmap: {... :path ... :name ... :depth ...}
+;;   valuemap: {... :value_type_id ... :value_boolean ...}
```
wkalt: I'm not sure if we want this comment in the file, at least without more detailed written descriptions. I've been thinking today though that we don't do a very good job of explaining our facts model in a centralized place in docs; it's sort of spread around the fact-paths and fact-contents endpoint documentation. It'd be good if we had a blurb somewhere to flesh out the model, maybe with a diagram or something. Separate patch from this though obviously.

senior:
Is something like this what you're looking for?
``` clojure
(s/validate [[(s/one s/Int "path-segment-1")
              (s/one s/Int "path-segment-2")]]
            [[1 2]])
```

src/com/puppetlabs/puppetdb/facts.clj
```diff
@@ -11,17 +11,37 @@
...
+
+(def fact-set
+  {s/Str s/Any})
+
+(def facts-schema
+  {:name String
+   :values fact-set
+   :timestamp pls/Timestamp
+   :environment (s/maybe s/Str)
+   :producer-timestamp (s/either (s/maybe s/Str) pls/Timestamp)})
```
wkalt: why is this dashed and not underscored?

wkalt: oh I get it this is stable

src/com/puppetlabs/puppetdb/facts.clj
rbrw: Right, this is one of the bits I figured I might delete from the final version -- I just added it while I was working so I could keep track.  And I left it in for now to possibly spark this discussion.

Another couple of things I mentioned to @senior are that it might be nice to have somewhat more consistent naming of types in the code, and that it might be nice to have some approach to naming schemas in particular, wether via Capitalization (which he mentioned some people do), or a naming convention like valuemap-scm, or putting them all in separate namespaces.  Having (for example) both factpath and fact-path scattered around seems a bit muddled, and I think I might prefer to save the really obvious names like fact, path, or value, for local-var/global-var/function names, rather than schemas.

src/com/puppetlabs/puppetdb/facts.clj
rbrw: @senior that looks like it might be what I need.  Thanks.

rbrw: @wkalt, @senior well that might be more luck than anything else.  If you mean the var names too, I hadn't realized/remembered that we'd carried it that far.  i.e. I'd thought we'd kept dashes for function args, and other "public", top-level bits.  In any case, glad you pointed it out.  I (or whoever does it) will need to keep that in mind during the merge to master.

src/com/puppetlabs/puppetdb/scf/migration_legacy.clj
```diff
@@ -0,0 +1,281 @@
+;;;
+;;; Vestigial bits we have to keep around for various migrations.
```
senior: This can go in the ns block and then be available as a clojure doc, i.e. 
```
(ns com...
  "some docs here"...) 
```

src/com/puppetlabs/puppetdb/facts.clj
rbrw: @senior oh, or perhaps: [(s/pair s/Int "foo" s/Int "bar")]

rbrw:
I think I've addressed all the comments in the 1226 PR which should be a strict superset of this one: https://github.com/puppetlabs/puppetdb/pull/1296

PR #1291, opened 2015-03-03
(maint) Change version policy to account for upgrades

senior:


documentation/versioning_policy.markdown
```diff
@@ -88,3 +89,7 @@ Some examples of changes that will require a new command version:
 * Change to serialization for wire formats inside payload.
 
 The [API Commands][commands] documentation contains more concrete information about the existing commands, versions and statuses for this version of PuppetDB.
+
+## Upgrades
+
+PuppetDB supports upgrading from prior releases. Upgrading ensures data and configuration information is preserved across releases. Upgrades are supported from any previous release in the same major version or any release in the prior major version only. As an example, it's safe to upgrade from 2.0.0 to 2.2.2, or from 1.6.0 to 2.2.2. We don't support support upgrading from 1.6.0 straight to 3.0.0. Users in this situation will want to first upgrade from 1.6.0 to 2.2.2, then from 2.2.2 to 3.0.0.
```
rbrw: Perhaps wrap the text, and wonder about changing it to read "are only supported", and "straight from" or "directly from"?  Otherwise looks good.

kbarber: Actually - technically documentation is meant to be written using soft wrap as a docs team convention like Ryan has done here, I've just been terrible at enforcing it on PR's lately. There's a long explanation for this, its been a while since I've thought about it though :-).

And yeah "support support" @senior :-).

senior: fixed support support

PR #1292, opened 2015-03-09
Fix "Overview" link

lwindolf:
The "Overview" link is broken. I guess it should point to index.markdown in the "documentation/api" instead of "documentation/api/query"


wkalt:
@lwindolf thanks!

PR #1295, opened 2015-03-10
(PDB-1197) anonymize logs and metrics

wkalt:
This implements anonymization for logs and metrics.

### DIFF
--- a/documentation/anonymization.markdown
+++ b/documentation/anonymization.markdown
@@ -45,7 +45,7 @@ At the moment there are some different built-in profiles you can choose from, an
 
 ### Profile: full
 
-The `full` profile will anonymize everything, while keeping the shape of data as previously mentioned. This includes: node names, resource types, resource titles, parameter names, values, any log messages, file names and file lines. The result should be a completely anonymized data set.
+The `full` profile will anonymize everything, while keeping the shape of data as previously mentioned. This includes: node names, resource types, resource titles, parameter names, values, any log messages, file names and file lines. The result should be a completely anonymized data set. Report metrics under the `resources` and `events` categories are left intact, as these can be inferred from the rest of the data, but names of metrics under the `time` category are anonymized as resource types.
 
 This is useful if you really paranoid about what data you expose, but provides the least amount of usefulness for the consumer of such data depending on the activity they are trying to test.
 
@@ -63,6 +63,9 @@ The `moderate` profile attempts to be a bit smarter about what it anonymizes and
 * report log messages: are always anonymized
 * file names: are always anonymized
 * file numbers: are left as they are
+* log messages: are always anonymized
+* metrics: metric names in the `time` category are anonymized as resource
+  types.
 
 ### Profile: low
 
@@ -70,6 +73,7 @@ This profile is aimed at hiding security information specifically, but leaving m
 
 * node name: is always anonymized
 * parameter values: we specifically anonymize any values and messages for any parameter name containing the strings: password, pwd, secret, key, private. Everything else is left alone.
+* log messages: are always anonymized
 
 Verifying your Anonymized Data
 -----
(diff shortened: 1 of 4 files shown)


src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -350,6 +359,38 @@
         (update-in-nil ["containment_path"]
                        anonymize-containment-path newcontext config))))
 
+(pls/defn-validated anonymize-metric :- metric-schema-str
+  [metric :- metric-schema-str
+   context
+   config]
+  (-> metric
+      (update-in ["name"] anonymize-leaf :name context config)
+      (update-in ["category"] anonymize-leaf :name context config)
```
kbarber: Does category even need anonymization?

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -350,6 +359,38 @@
         (update-in-nil ["containment_path"]
                        anonymize-containment-path newcontext config))))
 
+(pls/defn-validated anonymize-metric :- metric-schema-str
+  [metric :- metric-schema-str
+   context
+   config]
+  (-> metric
+      (update-in ["name"] anonymize-leaf :name context config)
```
kbarber: Sometimes the 'name' presents a resource type ... this should follow the same anonymization rules as resource type elsewhere to keep the shape the same.

kbarber:
I notice there is no policy additions to handle for example, logs ... seems like we've added a couple of things that should be switchable. For example the metric 'values' aren't always important to anonymized.

wkalt:

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -350,6 +366,37 @@
         (update-in-nil ["containment_path"]
                        anonymize-containment-path newcontext config))))
 
+(pls/defn-validated anonymize-metric
+  [metric :- metric-schema-str
+   context
+   config]
+  (if (= "time" (get metric "category"))
+    (update-in metric ["name"] anonymize-leaf :type context config)
```
kbarber: So the metric resource name under `time` is actually in lower case, while the :type stuff expects the first letter to be upper case. This is creating a difference between the events and metrics area, ie. the random strings don't match - not even if you make the first letter lower case later on. I think you need to upcase the first letter of the input before getting something back, then lower it again or something like that - haven't really thought it through, perhaps something a bit more elegant :-).

documentation/anonymization.markdown
```diff
@@ -45,7 +45,7 @@ At the moment there are some different built-in profiles you can choose from, an
 
 ### Profile: full
 
-The `full` profile will anonymize everything, while keeping the shape of data as previously mentioned. This includes: node names, resource types, resource titles, parameter names, values, any log messages, file names and file lines. The result should be a completely anonymized data set.
+The `full` profile will anonymize everything, while keeping the shape of data as previously mentioned. This includes: node names, resource types, resource titles, parameter names, values, any log messages, file names and file lines. The result should be a completely anonymized data set. Report metrics under the resources and events categories are left intact, as these can be inferred from the rest of the data, but names of metrics under the 'time' category are anonymized.
```
kbarber: Backtick 'time'. And maybe resources and events also?

src/puppetlabs/puppetdb/anonymizer.clj
kbarber: I think tags had a similar issue, take a look at what its doing.

documentation/anonymization.markdown
```diff
@@ -63,6 +63,8 @@ The `moderate` profile attempts to be a bit smarter about what it anonymizes and
 * report log messages: are always anonymized
 * file names: are always anonymized
 * file numbers: are left as they are
+* metrics: are left intact
```
kbarber: Well this isn't true, the `time` category gets anonymized based on the resource type policy above.

wkalt:
@kbarber thanks, fixed

wkalt:

wkalt:

PR #1296, opened 2015-03-11
(PDB-1031) Lift fact paths into facts (also PDB-1224, PDB-1225, PDB-1226, PDB-1227)

rbrw:
This series migrates the database to use the new facts structure:

(factset_id fact_path_id fact_value_id)

and changes the code to accommodate that. When reviewing, it may be easier to look at the changes to storage.clj and facts.clj directly, as compared to the diff. Or, if we like, I can probably take some time to break up the PDB-1225 patch into more easily digested pieces.

Rob Browning (3):
  (PDB-1031) Move paths from values to facts
  (PDB-1031) Accommodate move of fact paths to facts
  (PDB-1031) Verify gc behavior in update race test

Ryan Senior (3):
  (PDB-1031) Fix query engine, v3 endpoints for fact_path move
  (PDB-1031) Speed migration with unique facts table
  (PDB-1031) Reproduce the fact paths constraint issue

 documentation/repl.markdown                        |   3 +-
 src/com/puppetlabs/puppetdb/cli/services.clj       |   3 +-
 src/com/puppetlabs/puppetdb/facts.clj              |  84 ++--
 src/com/puppetlabs/puppetdb/query.clj              |   8 +-
 src/com/puppetlabs/puppetdb/query/facts.clj        |   2 +-
 src/com/puppetlabs/puppetdb/query_eng.clj          |  22 +-
 src/com/puppetlabs/puppetdb/scf/migrate.clj        |  99 +++-
 .../puppetlabs/puppetdb/scf/migration_legacy.clj   | 278 +++++++++++
 src/com/puppetlabs/puppetdb/scf/storage.clj        | 556 ++++++++++-----------
 test/com/puppetlabs/puppetdb/test/command.clj      | 155 +++++-
 test/com/puppetlabs/puppetdb/test/facts.clj        |  72 ---
 test/com/puppetlabs/puppetdb/test/scf/migrate.clj  |  95 +++-
 .../puppetdb/test/scf/migration_legacy.clj         |  76 +++
 test/com/puppetlabs/puppetdb/test/scf/storage.clj  | 159 +++++-
 14 files changed, 1162 insertions(+), 450 deletions(-)
 create mode 100644 src/com/puppetlabs/puppetdb/scf/migration_legacy.clj
 create mode 100644 test/com/puppetlabs/puppetdb/test/scf/migration_legacy.clj


src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,106 @@
...
+   ;; to support this.  Have not determined exactly when support was
+   ;; introduced yet.
+   (let [opt-deferrable (if (and (scf-utils/postgres?)
+                                 (scf-utils/db-version-newer-than? [9 0]))
+                          "DEFERRABLE" "")]
+     (format "ALTER TABLE fact_paths
+                  ADD CONSTRAINT fact_paths_path_key UNIQUE (path)
+                    %s" opt-deferrable)
+     (format "ALTER TABLE fact_values
+                  ADD CONSTRAINT fact_values_value_hash_key UNIQUE (value_hash)
```
kbarber: This creates a duplicate index, you already have a non-unique index - line 904.

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,106 @@
...
+         fact_path_id bigint NOT NULL,
+         fact_value_id bigint NOT NULL)"
+
+   ;; Needed for the insert if nothing else.
+   "CREATE INDEX fact_paths_path_idx on fact_paths(path)"
+   "CREATE INDEX fact_values_value_hash_idx on fact_values(value_hash)"
+
+   ;; Patch up facts refrences to refer to the min id path/value
+   ;; wherever there's more than one option.
+   "INSERT INTO facts_transform (factset_id, fact_path_id, fact_value_id)
```
senior: @rbrw I'm having difficulty getting this to finish. I loaded up a database using Wyatt's test functions in benchmark.clj. It had about 430k fact paths and fact values, 720k facts. From the params, it was 2000 nodes, 100 facts per node, 50% duplicate. Thus far the migration has been running for about 20 minutes, not yet finished. Querying PG, it looks like it's running this query. Running explain on it I see (cost=61731.98..1371353952.47 rows=728159 width=85).

Am I doing something wrong? It doesn't look like this will finish for me.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+          in-pids (jdbc/in-clause pids)
+          in-fv-id-pairs (jdbc/in-clause-multi fv-id-pairs 2)]
+      (sql/do-prepared
+       (format
+        "DELETE FROM fact_paths fp
+           WHERE fp.id %s
+             AND NOT EXISTS (SELECT 1 FROM facts f
+                               WHERE f.fact_path_id %s
+                                 AND f.fact_path_id = fp.id
+                                 AND (f.factset_id, f.fact_value_id) NOT %s)"
```
kbarber: I'm a little worried about this constraint here, is it truly needed? I'm seeing quite a bit of deadlock detection errors now with this patch, whereas I saw none for stable and I'm wary we're making this delete too far reaching, thus causing locks & contention.

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,106 @@
...
+            INNER JOIN fact_values fv on f.fact_value_id = fv.id
+            INNER JOIN fact_paths fp on fv.path_id = fp.id"
+   "DROP TABLE facts"
+
+   "ALTER TABLE facts_transform
+        ADD CONSTRAINT facts_factset_id_fact_path_id_fact_key
+          UNIQUE (factset_id, fact_path_id)"
+   "ALTER TABLE facts_transform
+        ADD CONSTRAINT factset_id_fk
+          FOREIGN KEY (factset_id) REFERENCES factsets(id)
```
kbarber: Can we double check this, and the other 3 additional FK references here have backing indexes? I can't see them in the migration, and PostgreSQL doesn't always create them.

kbarber: There are a couple of ways to check for covering indexes automatically, PgAdmin3 has a facility. Also this was provided by @dalen: http://mlawire.blogspot.se/2009/08/postgresql-indexes-on-foreign-keys.html?m=1

kbarber:
We need to include the tests that @senior had put here: https://github.com/puppetlabs/puppetdb/pull/1246

test/com/puppetlabs/puppetdb/test/scf/storage.clj
```diff
@@ -510,6 +627,8 @@
     (is (= (query-to-vec ["SELECT COUNT(*) as c FROM catalog_resources"])
            [{:c 3}]))))
 
+;; FIXME: should this and delete-certname-facts! go away once the new
```
senior: @rbrw Maybe this is a note to yourself?

test/com/puppetlabs/puppetdb/test/scf/storage.clj
```diff
@@ -259,6 +283,99 @@
                  :environment_id (environment-id "DEV")}]
                (query-to-vec "SELECT certname, environment_id FROM factsets")))))))
 
+(deftest fact-path-value-gc
```
senior: Can you grab the commit from [here](https://github.com/puppetlabs/puppetdb/pull/1246) that originally reproduced the fact path issue and add it here? Would also be easy to then run the gc and check that it "cleans up" after the concurrent tx issue.

src/com/puppetlabs/puppetdb/scf/migrate.clj
senior: To tie up the loose ends in the above comment, I had a way large fact_path table. 430k fact paths is pretty ridiculous. Bumping that down to a reasonable (but still high) number of 500 (with 100k fact_values), I do get the migration to finish. It took about 2 minutes on my laptop.

I thought this was just sensitive to the size of the fact_paths table, but this looks to be sensitive to the size of the fact_values table as well. I increased the fact values to 195k (fact_paths still ~500) and migrated. I eventually killed it after 20 minutes. It looks like it's sensitive to larger fact_values as well. Although 195k fact_values is high, seems likely we have users with that amount (or more). My guess is, it's the select mins from the facts_transform population query that takes so long.

I think I've [found another way](https://github.com/senior/puppetdb/commit/8cda94e6dbf8ab105b4ddbacef8948e1727fd932). I created two tables, one that has the factset_id, path and value_hash. With that populated, I deleted the path/value duplicates (same code as before, just moved up in the migration). From this, the data can be moved from this intermediate table to the facts_transform table with the correct PKs using a simple join, no aggregates. We can drop the facts table once the first intermediate table is loaded, and drop the first intermediate after the second is loaded. There's only a small increase in size from what was there previously (the value_hash and path mainly).

I ran the migration on the same dataset as the one I killed after 20 minutes and it ran in just under 10 seconds.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1264,3 +1243,21 @@
...
+
+(def ^:dynamic *orphaned-path-gc-limit* 200)
+(def ^:dynamic *orphaned-value-gc-limit* 200)
+
+(defn garbage-collect!
+  "Delete any lingering, unassociated data in the database"
+  [db]
+  (time!
+   (:gc metrics)
+   (jdbc/with-transacted-connection db
```
senior: I think you changed the semantics of this here. I think that when you use sql/transaction inside another sql/transaction (which is what with-transacted-connection creates) it doesn't actually create a separate tx from a database perspective. Maybe with-connection here is what you want instead?

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1264,3 +1243,21 @@
...
+     (sql/transaction (delete-unassociated-params!))
+     (sql/transaction (delete-unassociated-environments!))
+     (sql/transaction (delete-unassociated-statuses!)))
+   ;; These require serializable because they make the decision to
+   ;; delete based on row counts in another table.
+   (jdbc/with-transacted-connection' db :serializable
```
senior: I'd separate these too, like the above and make each serializable

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1264,3 +1243,21 @@
   [action-for-unsupported-fn]
   (fail-on-unsupported action-for-unsupported-fn)
   (warn-on-db-deprecation))
+
+(def ^:dynamic *orphaned-path-gc-limit* 200)
+(def ^:dynamic *orphaned-value-gc-limit* 200)
```
senior: @kbarber do you think we should have an escape hatch here? On one hand, I can't imagine a scenario where we would even have 10 to remove, much less 200.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+                     value-hashes)]
+          [value_hash id])))
+
+(defn insert-new-values!
+  "Inserts the valuemaps into the database and returns a map from
+  value hash to valuemap.  The returned valuemaps will have an :id."
+  [valuemaps]
+  (let [vhashes (map :value_hash valuemaps)]
+    (zipmap vhashes
+            (let [vmaps valuemaps]
```
senior: why let bind to vmaps here?

test/com/puppetlabs/puppetdb/test/scf/storage.clj
rbrw: @senior  Yes, and to anyone reviewing.  I'll deal with it one way or another before the final version, but am happy to hear other opinions.

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior Likely vestigial.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+              (map (fn [id vmap] (assoc vmap :id id))
+                   (map :id (apply sql/insert-records :fact_values valuemaps))
```
senior: This seems like a useful chunk of generic functionality on it's own. Given a list of records, insert them and assoc into the record the id.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+    {}))
+
+(defn realize-values!
+  [valuemaps]
+  "Ensures that all valuemaps exist in the database and returns a
+  map of value hashes to valuemaps with ids."
+  (if-let [valuemaps (seq valuemaps)]
+    (let [value-hashes (map :value_hash valuemaps)
+          existing-vhash-ids (existing-value-hashes->ids value-hashes)
+          vhashes-to-valuemaps (into {} (for [vmap valuemaps]
```
senior: I'm a little confused by this whole let block. Seems like we're doing a lot of extra work. Fundamentally we've a list of valuemaps, and we want a map of hash->valuemap, including an existing or newly created id. Here we're adding that id to existing ones, passing through ones without. Even though we know which ones are existing already, on 941 below, we compute it again. Then we take the recomputed value and get the valuemap that we have above on line 943. After that we are overwriting the valuemaps for new ones we passed through here, with the new ids on 945/946.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+                       :environment_id (ensure-environment environment)
+                       :producer_timestamp (to-timestamp producer-timestamp)})
+   ;; Ensure that all the required paths and values exist, and then
+   ;; insert the new facts.
+   (let [paths-and-valuemaps (facts/facts->paths-and-valuemaps values)
+         pathstrs (map (comp facts/factpath-to-string first) paths-and-valuemaps)
+         valuemaps (map second paths-and-valuemaps)
+         vhashes (map :value_hash valuemaps)
+         path-ids (map (realize-paths! pathstrs) pathstrs)
+         vhash-ids (map :id (map (realize-values! valuemaps) vhashes))
```
senior: I think we should be able to just use something like
```
(map :id (realize-values! valuemaps))
```
Then we won't need vhashes and it won't traverse valuemaps again

senior: Looks like maybe the same could be said with pathstrs

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+         pv-pairs (map vector path-ids vhash-ids)]
+     (insert-facts-pv-pairs! (certname-to-factset-id name) pv-pairs))))
```
senior: Once those are collapsed above, it can probably be pushed down out of the let and put directly in pv-pairs

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior well, for one thing, that missing-valuemaps is also vestigial.  I broke some of these functions out from a much larger one, so add-facts! and update-facts! could share code, but I missed that during the process.  Though that may not be your only concern here.

senior: Also looks like the valuemaps are just being discarded after this is called anyway, so this function could have a similar contract to realize-paths! and not return the maps at all.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+         pathstrs (map (comp facts/factpath-to-string first) paths-and-valuemaps)
+         path-ids (map (realize-paths! pathstrs) pathstrs)
+         valuemaps (map second paths-and-valuemaps)
+         vhashes (map :value_hash valuemaps)
+         vhash-ids (map :id (map (realize-values! valuemaps) vhashes))]
+
+     ;; Add new facts and remove obsolete facts.
+     (let [replacement-pv-pairs (set (map vector path-ids vhash-ids))
+           current-pairs (set (select-pid-vid-pairs-for-factset factset-id))
+           new-pairs (set/difference replacement-pv-pairs current-pairs)
```
senior: Seems like this line and the one below it are emulating the diffing functions that we use in catalogs, any reason not to use that here?

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior It could, though at least as an internal function, I figured it didn't matter, and might turn out to be useful.  I also felt that both the paths and values fuctions were doing the same thing - returning maps of "the item" to an id.  For paths, the path is fully descriptive, for values, you need the map.  But I'm certainly happy to change it.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+  (sql/transaction
+   (let [factset-id (certname-to-factset-id name)
+         initial-factset-paths-vhashes
+         (sql/with-query-results result-set
+           ["SELECT fp.path, fv.value_hash FROM facts f
+               INNER JOIN fact_paths fp ON f.fact_path_id = fp.id
+               INNER JOIN fact_values fv ON f.fact_value_id = fv.id
+               WHERE factset_id = ?"
+            factset-id]
+           (doall result-set))
```
senior: Why not use query-to-vec here like you did above?

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior I'll have to take a look; if those are using data/diff, I imagined that set/diff might be at least as efficient, if not more efficient when you don't care about the extra functionality.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+           new-pairs (set/difference replacement-pv-pairs current-pairs)
+           rm-pairs (set/difference current-pairs replacement-pv-pairs)]
+
+       (when-let [rm-pairs (seq rm-pairs)]
+         (sql/do-prepared
+          (format "DELETE FROM facts
+                   WHERE factset_id = ? AND (fact_path_id, fact_value_id) %s"
+                  (jdbc/in-clause-multi rm-pairs 2))
+          ;; CAUTION: seq required - currently: (flatten #{[1 2] [3 4]}) => ()
+          (flatten [factset-id rm-pairs])))
```
senior: Another option here is (conj (apply concat rm-pairs) factset-id)

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior I think you're right about the semantics, but want to make sure it's for the same reason.  Previously, the with-transacted-connection here was in the caller (garbage-collect! in services.clj).  I just pushed it down into storage/garbage-collect! so that we could put the original functions inside the same transaction they were in before, and add a new serialized transaction for the facts/paths/ gc.

Though that *is* a change to the semantics of storage/garbage-collect!; it didn't establish any of its own transactions before.  However, storage/garbage-collect! isn't called by anyone other than services.clj (aside from tests -- though those would indeed see different behavior now).  Thoughts?  I could break this up into two functions, so that we can leave the transaction wrappers in services.clj, but I wondered if we really wanted servcies.clj to have to know about the differing transactional requirements.

rbrw: @senior hmm, or were you saying that the original code wasn't quite right either?

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+
+       (insert-facts-pv-pairs! factset-id new-pairs)
+
+       (when-let [rm-pairs (seq rm-pairs)]
+         (delete-pending-path-id-orphans! factset-id
+                                          (set/difference (set rm-pairs)
```
senior: Because of how rm-pairs and new-pairs were created, won't this always just be rm-pairs?

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,269 @@
...
+                                                          (set new-pairs)))
+         (delete-pending-value-id-orphans! factset-id
+                                           (set/difference (set rm-pairs)
```
senior: Same as above

test/com/puppetlabs/puppetdb/test/scf/storage.clj
senior: is fact-path-value-gc test what local-gc below refers to?

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior I already changed this during the rework to take advantage of factset-id/path uniqueness.

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior umm, yeah, of course.

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior ahh, right.  Though this is even simpler in the new version where we don't need the pairs anymore.  There it's now just (cons factset-id rm-pids), with the simpler query.

test/com/puppetlabs/puppetdb/test/scf/storage.clj
rbrw: Yes, that's what I had in mind when I wrote the comment at least (before I wrote the tests).

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @kbarber We didn't need the factset/value id pairs for fact_paths (since paths are unique per factset), but we do need the similar constraint for fact_values to protect different paths in the same factset that point to the same value, and that aren't being deleted in the current transaction.

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior at least in the current code, the reason those extra maps are there is that they reestablish the ordering of paths and values, so that they will correspond properly when we make the vector pairs just below.

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @kbarber I believe @senior has handled this.

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @kbarber thanks - fixed in the upcoming push.

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior no reason, that was just before I'd gotten used to query-to-vec (fixed in upcoming push).

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @senior nice - added to the upcoming series.

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior (oversight) fixed.

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior I've separated the new orphaned facts and paths gc transactions as suggested, but I left the delete-unassociated-* calls together since that's how they were before.

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @senior right - switched to data/diff instead of the dual set/difference.  I'd originally misunderstood the data/diff docs, and thought it was saying that the [path-id value-id] pairs would have to be sets for it to work correctly, not that the *parent* collection (of the pairs) just had to be a set.

After some trivial benchmarking I found that data/diff appeared be quite a bit faster from a cold start, and still notably faster on a second pass.  Thanks for pursuing this.

test/com/puppetlabs/puppetdb/test/scf/storage.clj
rbrw: @senior done, and the clean up check will be in the next revision.

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,108 @@
     (sql/do-commands
      "DROP INDEX fact_values_string_trgm")))
 
+(defn- lift-fact-paths-into-facts
+  "Pairs paths and values directly in facts, i.e. change facts from (id
+  value) to (id path value)."
+  []
+  (sql/do-commands
+   (if (scf-utils/postgres?)
+     "SET CONSTRAINTS ALL DEFERRED" ;; In case it allows additional optimization?
```
wkalt: was this determined?

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,108 @@
...
+        WHERE t1.path = t2.path AND t1.id > t2.id"
+     "DELETE FROM fact_paths t1
+        WHERE t1.id <> (SELECT MIN(t2.id) FROM fact_paths t2
+                          WHERE t1.path = t2.path)")
+   (if (scf-utils/postgres?)
+     "DELETE FROM fact_values t1 USING fact_values t2
+        WHERE t1.value_hash = t2.value_hash AND t1.id > t2.id"
+     "DELETE FROM fact_values t1
+        WHERE t1.id <> (SELECT MIN(t2.id) FROM fact_values t2
+                          WHERE t1.value_hash = t2.value_hash)")
```
wkalt: What's the reason for the divergence between the pg and hsqldb queries here?

rbrw: @wkalt well "t1.id > t2.id" was the query I wanted to use, but hsqldb couldn't handle it.

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -911,7 +1014,11 @@
    24 add-producer-timestamps
    25 structured-facts
    26 structured-facts-deferrable-constraints
-   27 switch-value-string-index-to-gin})
+   27 switch-value-string-index-to-gin
+   28 #(let [time-msg (with-out-str
+                        (time (apply lift-fact-paths-into-facts %&)))]
```
wkalt: is there no test for this migration?

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,108 @@
...
+                          WHERE t1.path = t2.path)")
+   (if (scf-utils/postgres?)
+     "DELETE FROM fact_values t1 USING fact_values t2
+        WHERE t1.value_hash = t2.value_hash AND t1.id > t2.id"
+     "DELETE FROM fact_values t1
+        WHERE t1.id <> (SELECT MIN(t2.id) FROM fact_values t2
+                          WHERE t1.value_hash = t2.value_hash)")
+
+   (let [opt-deferrable (if (and (scf-utils/postgres?)
+                                 (scf-utils/db-version-newer-than? [9 0]))
```
wkalt: what's the significance of 9.0?

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,108 @@
...
+   "ALTER TABLE facts_transform
+      ADD CONSTRAINT facts_factset_id_fact_path_id_fact_key
+        UNIQUE (factset_id, fact_path_id)"
+
+   "ALTER TABLE facts_transform
+      ADD CONSTRAINT factset_id_fk
+        FOREIGN KEY (factset_id) REFERENCES factsets(id)
+        ON UPDATE CASCADE
+        ON DELETE CASCADE"
+   ;; CHECK: Do we still want/need these to be deferrable, after the GC changes?
```
wkalt: marking for visibility

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @wkalt not yet.  But I'll see about adding one.

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @wkalt versions of postgresql apparently can't make that deferrable.

wkalt: oh I see, unique constraints specifically

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -911,7 +1014,11 @@
    24 add-producer-timestamps
    25 structured-facts
    26 structured-facts-deferrable-constraints
-   27 switch-value-string-index-to-gin})
+   27 switch-value-string-index-to-gin
+   28 #(let [time-msg (with-out-str
+                        (time (apply lift-fact-paths-into-facts %&)))]
+         (log/info (str "lift: " time-msg)))
```
wkalt: "lift" won't make sense without context

test/com/puppetlabs/puppetdb/test/scf/migration_legacy.clj
```diff
@@ -0,0 +1,76 @@
+(ns com.puppetlabs.puppetdb.test.scf.migration_legacy
```
wkalt: swap the _ to -

test/com/puppetlabs/puppetdb/test/command.clj
```diff
@@ -972,6 +972,174 @@
...
+
+        ;; I'm modifying delete-pending-path-id-orphans! so that I can
+        ;; coordinate access between the two threads, I'm storing the
+        ;; reference to the original delete-pending-path-id-orphans!
+        ;; here, so that I can delegate to it once I'm done
+        ;; coordinating
+        storage-delete-pending-path-id-orphans!
+        scf-store/delete-pending-path-id-orphans!]
+
+    ;; (println "Before submitting anything (" (thread-id) "):")
```
wkalt: I think we can get rid of the commented debugging stuff

src/com/puppetlabs/puppetdb/scf/migration_legacy.clj
```diff
@@ -0,0 +1,279 @@
+(ns com.puppetlabs.puppetdb.scf.migration-legacy
+  "Vestigial code required by various migrations.  Note that every
+  external call here increases the risk that future changes might
+  break a migration."
+  (:require
+   [com.puppetlabs.puppetdb.facts :as facts]
```
wkalt: doesn't need a new line

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,108 @@
...
+      "ALTER TABLE facts_transform
+         ADD CONSTRAINT fact_value_id_fk
+           FOREIGN KEY (fact_value_id) REFERENCES fact_values(id)
+           %s" opt-deferrable))
+
+   "ALTER TABLE facts_transform RENAME TO facts"
+   "CREATE INDEX facts_fact_path_id_idx ON facts(fact_path_id)"
+   "CREATE INDEX facts_fact_value_id_idx ON facts(fact_value_id)"
+
+   (if (scf-utils/postgres?) "ANALYZE facts" "SELECT 1")
```
wkalt: can we just do (when (scf-postgres?) "ANALYZE facts")? I think there are a couple of places here this applies to

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,108 @@
...
+
+   ;; These are for the more pedantic HSQLDB.
+   "ALTER TABLE fact_paths DROP CONSTRAINT fact_paths_path_type_id_key"
+   "ALTER TABLE fact_paths DROP CONSTRAINT fact_paths_value_type_id"
+   "DROP INDEX fact_paths_value_type_id"
```
wkalt: this index isn't recreated on the facts table -- should it be?

wkalt: fact_values table*

src/com/puppetlabs/puppetdb/scf/migrate.clj
wkalt: I'd say we strip it out here and implement in on every migration in master. Definitely a cool touch.

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -883,6 +884,108 @@
...
+           FOREIGN KEY (fact_value_id) REFERENCES fact_values(id)
+           %s" opt-deferrable))
+
+   "ALTER TABLE facts_transform RENAME TO facts"
+   "CREATE INDEX facts_fact_path_id_idx ON facts(fact_path_id)"
+   "CREATE INDEX facts_fact_value_id_idx ON facts(fact_value_id)"
+
+   (if (scf-utils/postgres?) "ANALYZE facts" "SELECT 1")
+
+   ;; These are for the more pedantic HSQLDB.
```
wkalt: I don't understand this comment

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @wyatt sure wrt "lift:", I was planning on removing it altogether in the final version, until @senior suggested we add it everwhere: https://tickets.puppetlabs.com/browse/PDB-1320

src/com/puppetlabs/puppetdb/scf/migration_legacy.clj
rbrw: @wkalt ok, I've sometimes just done that to provide more width in the future, and it lines up the imports and requires, but I have no strong feelings about it.

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @wkalt I found that HSQLDB required that we do that explicitly where PostgreSQL doesn't.

src/com/puppetlabs/puppetdb/scf/migration_legacy.clj
wkalt: we don't do it anywhere else in the codebase -- better to be consistent IMO

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,245 @@
...
+  "Inserts the records (maps) into the named database and returns them
+  with their new :id values."
+  [database records]
+  (map #(assoc %2 :id %1)
+       (map :id (apply sql/insert-records database records))
+       records))
+
+(defn realize-paths!
+  [pathstrs]
+  "Ensures that all paths exist in the database and returns a map of
```
wkalt: the docstring should be above the function arguments

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,245 @@
...
+                   (map #(vector (:path %) (:id %))
+                        (realize-records!
+                         :fact_paths
+                         (map (comp facts/path->pathmap facts/string-to-factpath)
+                              missing-db-paths))))))
+    {}))
+
+(defn realize-values!
+  [valuemaps]
+  "Ensures that all valuemaps exist in the database and returns a
```
wkalt: here as well

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -928,119 +757,245 @@
...
+                  (jdbc/in-clause rm-pids))
+          (cons factset-id rm-pids))))
+
+     (insert-facts-pv-pairs! factset-id new-pairs)
+
+     (when rm-pairs
+       (delete-pending-path-id-orphans! factset-id
+                                        (set/difference
+                                         (set (map first current-pairs))
+                                         (set (map first replacement-pv-pairs))))
```
wkalt: I haven't tested, but is this expressible in terms of rm-pairs? (set (map first rm-pairs)) or something maybe? it seems like we might be able to avoid diffing both here and above somehow.

test/com/puppetlabs/puppetdb/test/scf/storage.clj
```diff
@@ -259,6 +283,99 @@
...
+                                                cast(fv.value_float as text),
+                                                '') as value
+                                  FROM fact_values fv"))))]
+    (testing "during add/replace (generally)"
+      ;; Keep resetting the db facts to match the current state of
+      ;; @fact-x and @facts-y, and then verify that fact_paths always
+      ;; contains exactly the set of keys across both, and fact_values
+      ;; contains exactly the set of values across both.
+      (let [facts-x (atom  {"a" "1" "b" "2" "c" "3"})
+            facts-y (atom  {})]
```
wkalt: I like the way you're doing this

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @wkalt as per other discussion, I'll be removing deferrals when possible in the next revision.

wkalt: :+1:

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @wkalt gone entirely now (removed the deferral).

test/com/puppetlabs/puppetdb/test/scf/migration_legacy.clj
rbrw: @wkalt umm, yeah - copy/paste fail.

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @wkalt unfortunately not; unlike str, do-commands can't handle nil (or even "").

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @wkalt I think we're going to have a diff regardless.  We have to compute all of the pids that won't be referred to by the current factset once this update is complete.  So we need something like either the current diff, or (I suspect) something like a diff of rm-pairs vs new-pairs instead.  Think the latter might be preferable/clearer?

src/com/puppetlabs/puppetdb/scf/migrate.clj
rbrw: @wkalt test in next revision

src/com/puppetlabs/puppetdb/scf/storage.clj
wkalt: @rbrw I see what you mean about needing two diffs. It looks to me like diffing rm-pairs and new-pairs would generally mean much smaller sets, since replacement-pv-pairs and current-pairs both correspond to entire factsets and new-pairs and rm-pairs only represent the delta. It also seems like a safe change to me, since I think everything that should incur a GC is represented in those smaller sets. What do you think?

rbrw:
@wkalt @senior hopefully this version addresses the outstanding concerns (and adds a preliminary migration test).

src/com/puppetlabs/puppetdb/scf/migrate.clj
wkalt: the column existed unindexed already. After discussing this yesterday I think it's fine. If we have related performance issues we can consider an index.

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -950,28 +1049,30 @@
         (select-keys migrations pending))))
 
 (defn migrate!
-  "Migrates database to the latest schema version. Does nothing if database is
```
rbrw: I just noticed while reviewing another PR that we already have a fast-forward-to-migration! in master.  Given that, we may want to back out this change to migrate!.

src/com/puppetlabs/puppetdb/scf/storage.clj
rbrw: @wkalt I wondered about the same thing.  I'd also tested it, and it passed all the unit tests, so I'll plan to change it in the next version.

rbrw:
@pljenkinsro retest this please.

wkalt:

wkalt:
I tested the performance of this patch relative to stable and I am +1 on that front. Infomation here: https://tickets.puppetlabs.com/browse/PDB-1227

src/com/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -86,10 +86,13 @@
                :source-table "fact_paths"
                :alias "fact_paths"
                :subquery? false
-               :source "SELECT path, type
+               ;; FIXME: is this plausible?
```
wkalt: marking this

src/com/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -911,7 +1005,12 @@
    24 add-producer-timestamps
    25 structured-facts
    26 structured-facts-deferrable-constraints
-   27 switch-value-string-index-to-gin})
+   27 switch-value-string-index-to-gin
+   ;; FIXME: remove from final version
```
wkalt: fixme

rbrw:
Cleaned up the series a bit, and pushed - should address the outstanding comments.

wkalt:

wkalt:

wkalt:

PR #1297, opened 2015-03-11
(PDB-1179) Add test classifier for pdb testutils

ajroetker:
This commit adds a test classifier and a corresponding profile so that
when we build and deploy the puppetdb jar we get a `-test`
jar so that downstream consumers of puppetdb can use the pdb-testutils code.


mullr:
:+1:

PR #1298, opened 2015-03-12
(PDB-1288) AOT compile main so the uberjar works

rbrw:
Pay particular attention to the test.  i.e. is it possibly unreasonable?

And note that (for whoever merges this to master) the fix for master is simiar, but it's also missing the (:gen-class) in the puppetlabs.puppetdb.core ns declaration.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -94,7 +94,8 @@
                                   [puppetlabs/kitchensink ~ks-version :classifier "test"]
                                   [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version :classifier "test"]
                                   [org.flatland/ordered "1.5.2"]
-                                  [org.clojure/test.check "0.5.8"]]}}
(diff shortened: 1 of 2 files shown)


project.clj
```diff
@@ -94,7 +94,8 @@
                                   [puppetlabs/kitchensink ~ks-version :classifier "test"]
                                   [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version :classifier "test"]
                                   [org.flatland/ordered "1.5.2"]
-                                  [org.clojure/test.check "0.5.8"]]}}
+                                  [org.clojure/test.check "0.5.8"]]}
+             :uberjar {:aot [com.puppetlabs.puppetdb.core]}}
```
ajroetker: will this conflict with the `^:skip-aot` below?

rbrw:
Wrongheaded - closing (see ext/resources/cli/...).

PR #1300, opened 2015-03-12
(PDB-1277) Fix puppetdb-ssl-setup/postinsts for AIO

rbrw:
Convert puppetdb-ssl-setup to puppetdb-ssl-setup.erb, and prepend the
AIO path to PATH.

Remove extraneous PATH settings from the Debian and Red Hat postinsts
since puppetlabs-ssl-setup sets the PATH internally, and nothing else in
the postinsts is relevant.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -134,7 +134,9 @@ case @osfamily
 end
 
 # The Puppet 4 load path
-@p4libdir = "/opt/puppetlabs/puppet/lib/ruby/vendor_ruby"
+@labsdir = '/opt/puppetlabs'
+@labsbindir = "#{@labsdir}/bin"
+@p4libdir = "#{@labsdir}/puppet/lib/ruby/vendor_ruby"
 
 @heap_dump_path = "#{@log_dir}/puppetdb-oom.hprof"
 @default_java_args = "-Xmx192m -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=#{@heap_dump_path} -Djava.security.egd=file:/dev/urandom"
--- a/ext/templates/deb/postinst.erb
+++ b/ext/templates/deb/postinst.erb
@@ -1,10 +1,6 @@
 #!/bin/bash
 
-<% if @pe -%>
-export PATH=/opt/puppet/bin:$PATH
-<% end -%>
-
-# Setup the SSL stuff
+# Set up the SSL stuff
 
 <%= @sbin_dir -%>/puppetdb ssl-setup
 
--- a/ext/templates/puppetdb-ssl-setup.erb
+++ b/ext/templates/puppetdb-ssl-setup.erb
@@ -222,7 +222,7 @@ then
   done
 else
   # This should be run on the host with PuppetDB
-  PATH=/opt/puppet/bin:$PATH
+  PATH="<%= @labsbindir -%>:/opt/puppet/bin:$PATH"
   agent_confdir=`puppet agent --configprint confdir`
   agent_vardir=`puppet agent --configprint vardir`
 
@@ -392,7 +392,7 @@ else
   echo "Error: Unable to find PuppetDB Jetty configuration at ${jettyfile} so unable to provide automatic configuration for that file."
   echo
   echo "   Confirm the file exists in the path specified before running the"
-  echo "   tool again. The file should have been created automatically when" 
+  echo "   tool again. The file should have been created automatically when"
   echo "   the package was installed."
 fi
 
--- a/ext/templates/dev/redhat/redhat_dev_postinst.erb
+++ b/ext/templates/dev/redhat/redhat_dev_postinst.erb
@@ -1,9 +1,5 @@
 #!/bin/bash
 
-<% if @pe -%>
-export PATH=/opt/puppet/bin:$PATH
-<% end -%>
-
 # Setup the SSL stuff
 <%= @sbin_dir -%>/puppetdb ssl-setup
 
--- a/tasks/template.rake
+++ b/tasks/template.rake
@@ -16,7 +16,8 @@ task :template => [ :clean ] do
     "ext/templates/puppetdb.erb"            => "ext/files/puppetdb",
     "ext/templates/puppetdb-legacy.erb"     => "ext/files/puppetdb-legacy",
     "ext/templates/init_debian.erb"         => "ext/files/#{@name}.debian.init",
-    "ext/templates/puppetdb-env.erb"        => "ext/files/puppetdb.env"
+    "ext/templates/puppetdb-env.erb"        => "ext/files/puppetdb.env",
+    "ext/templates/puppetdb-ssl-setup.erb"  => "ext/files/puppetdb-ssl-setup"
   }
 
   deb_templates = {
@@ -56,7 +57,6 @@ task :template => [ :clean ] do
   chmod 0700, "ext/files/puppetdb-import"
   chmod 0700, "ext/files/puppetdb-export"
   chmod 0700, "ext/files/puppetdb-anonymize"
-  cp_p "ext/templates/puppetdb-ssl-setup", "ext/files"
   chmod 0700, "ext/files/puppetdb-ssl-setup"
   chmod 0700, "ext/files/puppetdb"
   chmod 0700, "ext/files/puppetdb-legacy"


Rakefile
```diff
@@ -134,7 +134,9 @@ case @osfamily
 end
 
 # The Puppet 4 load path
-@p4libdir = "/opt/puppetlabs/puppet/lib/ruby/vendor_ruby"
+@p4home = "/opt/puppetlabs/puppet"
```
johnduarte: There are two bindirs
1. Public (for our bins e.g. puppet, facter) /opt/puppetlabs/bin
2. Private (for vendored deps e.g. ruby) /opt/puppetlabs/puppet/bin

So, suggested values for these would be

    @p4home = "/opt/puppetlabs"
    @p4bindir = "#{@p4home}/bin"
    @p4puppetdir = "#{@p4home}/puppet"
    @p4libdir = ""#{@p4puppetdir}/lib/ruby/vendor_ruby"


@joshcooper does this look sane to you?

kbarber:
@haus @melissa any comments on this one?

haus:
@kbarber nothing crazy here. looks sane to me.

haus:
I have some further questions about the ssl-setup stuff with respect to puppetserver support. Going to try to comment in the diff if i can.

haus:
Can't seem to comment in the diff view. There are a few calls to `puppet master` in ssl-setup and I'm wondering why those are using the master runmode and if they will work as expected if puppetserver is installed on the same machine instead of ruby puppetmaster.

joshcooper:
@haus I don't understand why the script is using [master run mode](https://github.com/puppetlabs/puppetdb/blob/stable/ext/templates/puppetdb-ssl-setup#L226-L244), and why it's going through the trouble to get the agent's `confdir` and `vardir`, and then running `puppet master --configprint` using the agent's directories?

Seems like it could just do `puppet agent --configprint certname`, etc? Perhaps it was an issue with older puppet versions, e.g. http://projects.puppetlabs.com/issues/16637, or an issue in PE?

joshcooper:
@rbrw I have a few path related questions:

https://github.com/puppetlabs/puppetdb/blob/stable/acceptance/helper.rb#L212: if your acceptance tests only need to execute `puppet`, `facter`, etc it should be using `/opt/puppetlabs/bin` (for AIO). If it needs to execute `ruby` or `openssl` it should be using `/opt/puppetlabs/puppet/bin`.

https://github.com/puppetlabs/puppetdb/blob/stable/acceptance/helper.rb#L393: this is PE specific, so should not need to change

https://github.com/puppetlabs/puppetdb/blob/stable/acceptance/helper.rb#L766-L770: FYI, the `puppetpath` setting is essentially deprecated in beaker, and non-existent when running `aio` type acceptance tests, see https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/unix.rb#L65-L72. It is preferrable to execute `host.puppet('confdir')`. That way the host will tell you where it's `confdir` is, and your test will work against puppet 3.x and 4.0 versions. Similarly comment in https://github.com/puppetlabs/puppetdb/blob/stable/acceptance/helper.rb#L969-L971

https://github.com/puppetlabs/puppetdb/blob/stable/acceptance/helper.rb#L978: the master's pidfile has changed in puppet 4/puppetserver 2.0. If you are using webrick, it will be `/var/run/puppetlabs/master.pid`. If you are using puppetserver, it will be `/var/run/puppetlabs/puppetserver/puppetserver.pid`. See https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md. For webrick, you can use beaker's method `host.puppet('master')['pidfile']` to ask the host where it's pidfile is instead.

https://github.com/puppetlabs/puppetdb/blob/stable/ext/redhat/puppetdb.spec.erb#L176-L177: may need to prepend `/opt/puppetlabs/bin` and/or `/opt/puppetlabs/puppet/bin`?

https://github.com/puppetlabs/puppetdb/blob/stable/ext/templates/deb/postinst.erb#L4: there should be an else case that prepends `/opt/puppetlabs/bin`

https://github.com/puppetlabs/puppetdb/blob/stable/ext/templates/deb/rules.erb#L13-L14: need an else case that does `@rake = "/opt/puppetlabs/puppet/bin/rake"`

https://github.com/puppetlabs/puppetdb/blob/stable/ext/templates/dev/redhat/redhat_dev_postinst.erb#L4: need an else case that prepends `/opt/puppetlabs/bin`

https://github.com/puppetlabs/puppetdb/blob/stable/ext/templates/init_redhat.erb#L38: may need an else case that includes `/opt/puppetlabs/bin`?

Rakefile
joshcooper: @rbrw @johnduarte the values look sane. It looks like this script only needs to run `puppet` so using `/opt/puppetlabs/bin` should be fine.

kbarber:
So @joshcooper 

> haus I don't understand why the script is using master run mode, and why it's going through the trouble to get the agent's confdir and vardir, and then running puppet master --configprint using the agent's directories?

Its an older squirrelly issue we're yet to get back to basically:

https://tickets.puppetlabs.com/browse/PDB-143

> https://github.com/puppetlabs/puppetdb/blob/stable/acceptance/helper.rb#L212: if your acceptance tests only need to execute puppet, facter, etc it should be using /opt/puppetlabs/bin (for AIO). If it needs to execute ruby or openssl it should be using /opt/puppetlabs/puppet/bin.

Thanks, we'll need to fix that when we start testing against AIO. Seems like this is a separate spike on acceptance that needs to be raised. I'm pretty sure right now we're just testing 3.x. Since we're dribbling out 4 releases + AIO we'll need to figure out exactly how to do this. @rbrw I'd raise a ticket and put it in say next sprint for getting AIO coverage in our acceptance tests with all of @joshcooper's acceptance related comments included in the ticket.

The rest of your non-acceptance comments we'll need to dig through. Cheers for this.

rbrw:
OK, hopefully I've addressed all the bits that are within the scope of this ticket, and I've opened https://tickets.puppetlabs.com/browse/PDB-1297 for the rest.

Rakefile
```diff
@@ -134,7 +134,11 @@ case @osfamily
 end
 
 # The Puppet 4 load path
-@p4libdir = "/opt/puppetlabs/puppet/lib/ruby/vendor_ruby"
+@labsdir = "/opt/puppetlabs"
+@labsbindir = "#{@labsdir}/bin"
+@p4dir = "#{@labsdir}/puppet"
+@p4bindir = "#{@p4puppetdir}/bin"
+@p4libdir = ""#{@p4puppetdir}/lib/ruby/vendor_ruby"
```
kbarber: double quotes here

Rakefile
```diff
@@ -134,7 +134,11 @@ case @osfamily
 end
 
 # The Puppet 4 load path
-@p4libdir = "/opt/puppetlabs/puppet/lib/ruby/vendor_ruby"
+@labsdir = "/opt/puppetlabs"
+@labsbindir = "#{@labsdir}/bin"
+@p4dir = "#{@labsdir}/puppet"
+@p4bindir = "#{@p4puppetdir}/bin"
```
kbarber: its @p4dir not @p4puppetdir ...

Rakefile
```diff
@@ -134,7 +134,11 @@ case @osfamily
 end
 
 # The Puppet 4 load path
-@p4libdir = "/opt/puppetlabs/puppet/lib/ruby/vendor_ruby"
+@labsdir = "/opt/puppetlabs"
+@labsbindir = "#{@labsdir}/bin"
```
kbarber: is this var yagni? Doesn't look like we are using it in this patch.

ext/templates/puppetdb-ssl-setup.erb
```diff
@@ -222,7 +222,7 @@ then
   done
 else
   # This should be run on the host with PuppetDB
-  PATH=/opt/puppet/bin:$PATH
+  PATH="<%= @labsbindir -%>:/opt/puppet/bin:$PATH"
```
kbarber: Wait, this should be p4bindir shouldn't it?

Rakefile
kbarber: err ... I mean double double quotes

ext/templates/puppetdb-ssl-setup.erb
kbarber: My mistake, whatever it is, it should equate to /opt/puppetlabs/bin ...

Rakefile
kbarber: My mistake, we are using it.

ext/templates/deb/postinst.erb
```diff
@@ -1,10 +1,6 @@
 #!/bin/bash
 
-<% if @pe -%>
-export PATH=/opt/puppet/bin:$PATH
-<% end -%>
```
joshcooper: Is this ok to remove because there's a pe-puppet fork? If so, could you update the commit message?

Rakefile
kbarber: This one is yagni, unless we start needing there ruby. Since we still pull in system ruby for build purposes, its probably yagni for now.

rbrw: Right, already removed p4bindir.

ext/templates/deb/postinst.erb
rbrw: @joshcooper My understanding was that I should remove these because puppetlabs-ssl-setup already handles the path internally, and nothing else in the postinst(s) is relevant. (@kenbarber)

rbrw:
@kbarber hopefully addressed all your concerns now.

johnduarte:
:+1:

PR #1303, opened 2015-03-13
(maint) Remove facter requirement from rpm specfile

haus:
Nothing in puppetdb's runtime requires facter itself. Facter is a
transitive dependency of puppet, but does not need to be called out by
itself in the specfile. Puppet is required for the ssl setup.


stahnma:
:+1:

PR #1306, opened 2015-03-18
(PDB-1255) Refactor jetty testutils

wkalt:
This renames testutils/jetty to testutils/services, which captures the scope of
its use and content a bit better. Some minimal cleanup as well.

### DIFF
--- a/test/puppetlabs/puppetdb/testutils/services.clj
+++ b/test/puppetlabs/puppetdb/testutils/services.clj
@@ -1,7 +1,6 @@
-(ns puppetlabs.puppetdb.testutils.jetty
+(ns puppetlabs.puppetdb.testutils.services
   (:require [puppetlabs.puppetdb.testutils :refer [temp-dir temp-file]]
             [puppetlabs.puppetdb.fixtures :as fixt]
-            [puppetlabs.trapperkeeper.app :as tka]
             [puppetlabs.trapperkeeper.testutils.bootstrap :as tkbs]
             [puppetlabs.trapperkeeper.services.webserver.jetty9-service :refer [jetty9-service]]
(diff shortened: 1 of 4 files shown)


wkalt:
As a refactor this feels a bit uninspired, but I didn't see a clear dividing line between the "jetty" test functions and those related to our various TK services. The new name fits the file and the tests that use it better. We can split it out more if anyone feels strongly about it.

This will require an update to the puppetdb-sync repo.

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -131,10 +129,13 @@
...
   [base-url]
-  (str "org.apache.activemq:BrokerName="
-       (url-encode (:host base-url))
-       ",Type=Queue,Destination="
-       svcs/mq-endpoint))
+  (format "org.apache.activemq:BrokerName=%s,Type=Queue,Destination=%s"
+          (url-encode (:host base-url))
+          svcs/mq-endpoint))
+
+(url-encode "foo")
```
ajroetker: Is this supposed to be here?

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -131,10 +129,13 @@
...
+
+(command-mbean-name {:host "foo"})
```
ajroetker: Same here.

ajroetker:
I like `service-utils` better since it does seem to better capture the trapperkeeper side of what's going on.

wkalt:
@ajroetker thanks for the comments

senior:
+1 will merge with the pipeline is clear

PR #1307, opened 2015-03-18
(PDB-1268) Add validation for benchmark parameters

wkalt:
This adds validation that nummsgs or runinterval is used, and adds numhosts to
the list or required parameters.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/benchmark.clj
+++ b/src/puppetlabs/puppetdb/cli/benchmark.clj
@@ -303,7 +303,7 @@
    ["-N" "--nummsgs NUMMSGS" "Number of commands and/or reports to send for each host"]])
 
 (def required-cli-options
-  [:config])
+  [:config :numhosts])
 
 (defn activate-logging!
   [options]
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -323,6 +323,11 @@
         "Error: -N/--nummsgs runs immediately and is not compatable with -i/--runinterval")
       (action-on-error-fn))
 
+    (not (some #{:nummsgs :runinterval} (keys options)))
```
senior: Maybe not-any? instead, it's just (comp not some)

ajroetker:
:+1:

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -323,6 +323,11 @@
         "Error: -N/--nummsgs runs immediately and is not compatable with -i/--runinterval")
       (action-on-error-fn))
 
+    (not-any? #{:nummsgs :runinterval} (keys options))
+    (do
+      (log/error "Error: Either -N/--nummsgs or -i/--runinterval is required.")
```
senior: Aren't we required to have both?
```
lein run benchmark -F /Users/ryan/work/test-data/more-fact-paths -n 100 -c postgres.ini
2015-03-19 13:01:27,459 ERROR [main] [p.p.c.benchmark] Error: Either -N/--nummsgs or -i/--runinterval is required.
```

wkalt: No, they're incompatible with each other.

PR #1308, opened 2015-03-19
(PDB-1305) Update install from source docs for AIO

johnduarte:
This commit updates the install from source documentation to reflect
steps needed when using the Puppet all-in-one agent.

It also updates the system calls to by init style agnostic.

Specific changes:

* If AIO is used the rake command should be the puppet supplied rake.
* The puppetdb user and group need to be created, since the service runs as
this user by default.
* System files need to be owned by the puppetdb user.
*  If AIO is used, the terminus code should be copied to the puppet vendored
ruby location.
* Change sysvinit system calls to be init system agnostic. Indicate where the
 init files go for both sysvinit and systemd

### DIFF
--- a/documentation/connect_puppet_master.markdown
+++ b/documentation/connect_puppet_master.markdown
@@ -43,7 +43,7 @@ Currently, puppet masters need additional Ruby plugins in order to use PuppetDB.
 If your puppet master isn't running Puppet from a supported package, you will need to install the plugins manually:
 
 * [Download the PuppetDB source code][puppetdb_download], unzip it and navigate into the resulting directory in your terminal.
-* Run `sudo cp -R ext/master/lib/puppet /usr/lib/ruby/site_ruby/1.8/puppet`. Replace the second path with the path to your Puppet installation if you have installed it somewhere other than `/usr/lib/ruby/site_ruby`. If you are using the Puppet all-in-one agent, replace the second path with the path to the Puppet supplied ruby at `/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet`.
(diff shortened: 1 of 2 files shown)


nfagerlund:
This looks good enough to me.

documentation/connect_puppet_master.markdown
```diff
@@ -43,7 +43,7 @@ Currently, puppet masters need additional Ruby plugins in order to use PuppetDB.
 If your puppet master isn't running Puppet from a supported package, you will need to install the plugins manually:
 
 * [Download the PuppetDB source code][puppetdb_download], unzip it and navigate into the resulting directory in your terminal.
-* Run `sudo cp -R ext/master/lib/puppet /usr/lib/ruby/site_ruby/1.8/puppet`. Replace the second path with the path to your Puppet installation if you have installed it somewhere other than `/usr/lib/ruby/site_ruby`.
+* Run `sudo cp -R ext/master/lib/puppet /usr/lib/ruby/site_ruby/1.8/puppet`. Replace the second path with the path to your Puppet installation if you have installed it somewhere other than `/usr/lib/ruby/site_ruby`. If you are using the Puppet all-in-one agent, replace the second path with the path to the Puppet supplied ruby at `/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet`.
```
joshcooper: I'm wondering if users will know what all-in-one is? Perhaps just use the version, e.g. "using Puppet 4 or greater"?

nfagerlund: Oh yeah, good call.

johnduarte: Thanks @joshcooper , you are completely correct. The AIO term will be meaningless to users.

PR #1310, opened 2015-03-19
(docs) Add docs sidebar nav (TOC) file to PuppetDB repo, as _puppetdb_na...

nfagerlund:
...v.html

This is similar to commit 7739f6d5baceba94356712926c8572ec3a9c5d00 on stable,
but it pulls in the source/_includes/puppetdb_master.html file from puppet-docs,
rather than the 2.3 version of the nav.

New process:

- When you create or remove docs pages, please edit this TOC file to match.
- When you branch to make a new PuppetDB version, please edit the version number
  in the TOC.
- We'll still need to wire up new PuppetDB versions in the puppet-docs repo,
  but we won't need to host the TOC file in our repo; we can use it in place.

### DIFF
--- a/documentation/_puppetdb_nav.html
+++ b/documentation/_puppetdb_nav.html
@@ -1,0 +1,95 @@
+{% capture puppetdb_version %}master{% endcapture %}
+
+<h2 id="puppetdb">PuppetDB NEXT</h2>
+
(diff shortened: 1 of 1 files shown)


wkalt:
@nfagerlund we rolled this change up to master this morning, so I'll close this PR. Thanks for the help though, this will make things smoother going forward.

nfagerlund:
@wkalt Okay... but did you notice that the two nav snippets are *significantly* different because the docs on master have diverged quite a bit from those on stable? For example, the v3 API docs are supposed to be gone in master.

wkalt:
@nfagurlund yes, I'm looking into that

nfagerlund:
@wkalt OK, cool.

PR #1311, opened 2015-03-20
(PDB-1264) Ensure producer_timestamp has a value

mullr:
Other features would like to rely on the value of producer_timestamp. 
Newer versions of puppet populate this field, but it may be null
in upgrade scenarios. In that case, we can use the receive time 
('timestamp') with no ill effects.

### DIFF
--- a/acceptance/tests/inventory/basic_fact_retrieval.rb
+++ b/acceptance/tests/inventory/basic_fact_retrieval.rb
@@ -40,10 +40,13 @@ test_name "structured and trusted facts should be available through facts termin
     end
 
     step "create a custom structured fact" do
+      time = Time.now.iso8601
       payload = <<-EOM
       -H "Accept: application/json" -H "Content-Type: application/json" \
       -d '{"command":"replace facts","version":4, \
       "payload":{"environment":"DEV","certname":"#{master}", \
+      "timestamp": "#{time}", \
+      "producer_timestamp": "#{time}", \
       "values":{"my_structured_fact":#{JSON.generate(structured_data)}}}}' http://localhost:8080/v4/commands
       EOM
       on database, %Q|curl -X POST #{payload}|
--- a/src/puppetlabs/puppetdb/catalogs.clj
+++ b/src/puppetlabs/puppetdb/catalogs.clj
@@ -92,7 +92,8 @@
             [schema.core :as s]
             [puppetlabs.puppetdb.schema :as pls]
             [puppetlabs.puppetdb.utils :as utils]
-            [clojure.core.match :refer [match]]))
+            [clojure.core.match :refer [match]]
+            [puppetlabs.puppetdb.time :refer [to-timestamp]]))
 
 (def ^:const catalog-version
   "Constant representing the version number of the PuppetDB
@@ -114,7 +115,8 @@
    :version s/Str
    :environment (s/maybe s/Str)
    :transaction_uuid (s/maybe s/Str)
-   :producer_timestamp (s/maybe pls/Timestamp)
+   :producer_timestamp pls/Timestamp
+
 
    ;; This is a crutch. We use sets for easier searching and avoid
    ;; reliance on ordering. We should pick one of the below (probably
@@ -149,7 +151,6 @@
   (utils/assoc-when catalog
                     :transaction_uuid nil
                     :environment nil
-                    :producer_timestamp nil
                     :api_version 1))
 
 (pls/defn-validated canonical-catalog
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -727,7 +727,7 @@
              :values facts
              :timestamp timestamp
              :environment environment
-             :producer_timestamp nil}))))))
+             :producer_timestamp timestamp}))))))
 
 (defn structured-facts []
   ;; -----------
@@ -1125,6 +1125,18 @@
    "ALTER TABLE fact_paths DROP COLUMN value_type_id"
    "ALTER TABLE fact_values DROP COLUMN path_id"))
 
+(defn fill-in-null-producer-timestamp
+  "The producer_timestamp column can sometimes contain null values,
+  especially when updating from older versions. New use cases require
+  a value in that position, so we'll use the value from the
+  'timestamp' column if it's not there. "
+  []
+  (sql/do-commands
+   "UPDATE catalogs SET producer_timestamp=timestamp WHERE producer_timestamp IS NULL"
+   "UPDATE factsets SET producer_timestamp=timestamp WHERE producer_timestamp IS NULL"
+   "ALTER TABLE catalogs ALTER COLUMN producer_timestamp SET NOT NULL"
+   "ALTER TABLE factsets ALTER COLUMN producer_timestamp SET NOT NULL"))
+
 (def migrations
   "The available migrations, as a map from migration version to migration function."
   {1 initialize-store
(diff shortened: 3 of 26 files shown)


senior:
@mullr unit and acceptance test failures here, I think I know why. One thing the patch doesn't account for is that in our current wire format, producer timestamp can be null https://docs.puppetlabs.com/puppetdb/2.3/api/wire_format/catalog_format_v5.html same with v6 to be released in 3.0.0.

mullr:
Sorry about the false start before, I must have spaced my local unit test run after adding the NOT NULL constraint. This turned into a bigger change test-wise, but I think it's for the best.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -727,7 +727,7 @@
              :values facts
              :timestamp timestamp
              :environment environment
-             :producer_timestamp nil}
+             :producer_timestamp timestamp}
```
wkalt: I don't think this will work because we need to head off people who already got through this migration with nils

wkalt: didn't read far enough. We have another migration down at the bottom that will handle anyone who already made it past this one. Should work fine.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1033,6 +1033,18 @@
       "ALTER TABLE reports ADD metrics text"
       "ALTER TABLE reports ADD logs text")))
 
+(defn fill-in-null-producer-timestamp
+  "The produer_timestamp column can sometimes contain null values,
```
wkalt: producer

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -85,7 +85,7 @@
    :values facts/fact-set
    :timestamp pls/Timestamp
    :environment (s/maybe s/Str)
-   :producer_timestamp (s/either (s/maybe s/Str) pls/Timestamp)})
+   :producer_timestamp (s/either s/Str pls/Timestamp)})
```
wkalt: prexisting obviously, but I'm not sure we actually have a reason to accept strings here

mullr:

mullr:

wkalt:

mullr:

test/puppetlabs/puppetdb/query/nodes_test.clj
```diff
@@ -109,9 +109,11 @@
...
                                              [4 "node_d" 2 3]
                                              [5 "node_e" 5 2]]]
-      (sql/insert-record :certnames {:certname node})
-      (sql/insert-record :factsets {:certname node :timestamp (to-timestamp (-> facts-age days ago))})
-      (sql/insert-record :catalogs {:id id :hash node :api_version 0 :catalog_version 0 :certname node :timestamp (to-timestamp (minus right-now (-> catalog-age days)))})))
+      (let [factset-timestamp (to-timestamp (-> facts-age days ago))
+            catalog-timestamp (to-timestamp (minus right-now (-> catalog-age days)))]
+        (sql/insert-record :certnames {:certname node})
+        (sql/insert-record :factsets {:certname node :timestamp factset-timestamp :producer_timestamp factset-timestamp})
+        (sql/insert-record :catalogs {:id id :hash node :api_version 0 :catalog_version 0 :certname node :timestamp catalog-timestamp :producer_timestamp catalog-timestamp}))))
```
rbrw: FWIW, since you're already changing the code here (and anywhere else relevant), I'd be tempted to wrap the long lines.

acceptance/tests/inventory/basic_fact_retrieval.rb
```diff
@@ -40,10 +40,13 @@
     end
 
     step "create a custom structured fact" do
+      time = Time.now.iso8601
       payload = <<-EOM
       -H "Accept: application/json" -H "Content-Type: application/json" \
       -d '{"command":"replace facts","version":4, \
       "payload":{"environment":"DEV","certname":"#{master}", \
+      "timestamp": "#{time}"\
+      "producer_timestamp": "#{time}"\
```
wkalt: @mullr there's a missing comma after timestamp here -- I think that's contributing to your woe :)

mullr:

mullr:

mullr:

mullr:

wkalt:

src/puppetlabs/puppetdb/catalogs.clj
```diff
@@ -106,6 +107,9 @@
 ;;; edges/resources in here and more of the codebase switched over to
 ;;; using the new schemas
 
+(def timestamp-str
+  (s/both s/Str (s/pred #(not (nil? (to-timestamp %))) 'coercable-to-timestamp)))
```
wkalt: used?

test/puppetlabs/puppetdb/query/nodes_test.clj
wkalt: seconded

src/puppetlabs/puppetdb/catalogs.clj
```diff
@@ -114,7 +115,8 @@
    :version s/Str
    :environment (s/maybe s/Str)
    :transaction_uuid (s/maybe s/Str)
-   :producer_timestamp (s/maybe pls/Timestamp)
+   :producer_timestamp (s/pred #(not (nil? (to-timestamp %))) 'coercable-to-timestamp)
```
wkalt: only question I have is whether we want/need this. My understanding is it's not needed and could just be pls/Timestamp, in which case it seems pretty heavy to keep around.

mullr: Well, it gives a more useful error message on bad input, which is the entire point of having a schema around.

wkalt: I disagree that that's the point -- the point in my mind is to very specifically specify valid inputs and throw an error when the input is unexpected. By specifying pls/Timestamp we get the benefit of *knowing* that it's a timestamp. When we do stuff like this we need to look at to-timestamp to figure out what's valid.

As it happens, this schema will accept an integer, which should not be the case.

wkalt: Actually, it looks like pls/Timestamp will also validate an integer so I retract the last bit there. I still feel pretty strongly about keeping these things as simple as possible though.

mullr: We have at least three different internal representations of them in this
codebase: strings, java.sql.Date, and joda time. Whatever we do here should
(1) document what kind of data we expect and (2) kick out an error when we
get data that we can't deal with. I think this code gives that in the most
concrete possible way, by attempting the very conversion whose result we
care about.

That said: I assumed that pls/Timestamp was something stock, but it's
actually our own thing that does almost the exactly what I wrote down here.
(except with joda time, which is fine, since that's what clj-time converts
to internally in to-timestamp) So I suppose I'm ok just using that :)

mullr: Updated to be `pls/Timestamp`, and checked that it gives a good message on invalid input. 

```
Value does not match schema: {:producer_timestamp (not (datetime? a-java.lang.String))}
```

wkalt:

wkalt:

wkalt:
+1

kbarber:
@mullr needs a rebase.

mullr:

mullr:

kbarber:
Transients ^^

PR #1312, opened 2015-03-20
(PDB-1286) backport resource timestamp normalization to stable

wkalt:
In master, we've changed our report hashing so that timestamps are converted to
a consistent representation before being hashed.

This backports that change to stable so that users will receive the update in 2.3.1,
and by the time they make it to 3.0 their reports should already be right.

Additional handling will be required in master for various edge cases.

### DIFF
--- a/src/com/puppetlabs/puppetdb/scf/hash.clj
+++ b/src/com/puppetlabs/puppetdb/scf/hash.clj
@@ -132,16 +132,16 @@
   [{:keys [resource-type resource-title property timestamp status old-value
            new-value message file line] :as resource-event}]
   (generic-identity-string
-    { :resource-type resource-type
-      :resource-title resource-title
-      :property property
-      :timestamp timestamp
-      :status status
-      :old-value old-value
-      :new-value new-value
-      :message message
-      :file file
-      :line line}))
+    {:resource_type resource-type
+     :resource_title resource-title
+     :property property
+     :timestamp timestamp
+     :status status
+     :old_value old-value
+     :new_value new-value
+     :message message
+     :file file
+     :line line}))
 
 (defn report-identity-hash
   "Compute a hash for a report's content
@@ -154,11 +154,11 @@
            start-time end-time resource-events transaction-uuid] :as report}]
   (generic-identity-hash
     {:certname certname
-     :puppet-version puppet-version
-     :report-format report-format
-     :configuration-version configuration-version
-     :start-time start-time
-     :end-time end-time
-     :resource-events (sort (map resource-event-identity-string resource-events))
-     :transaction-uuid transaction-uuid}))
+     :puppet_version puppet-version
+     :report_format report-format
+     :configuration_version configuration-version
+     :start_time start-time
+     :end_time end-time
+     :resource_events (sort (map resource-event-identity-string resource-events))
+     :transaction_uuid transaction-uuid}))
 
--- a/src/com/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/com/puppetlabs/puppetdb/scf/storage.clj
@@ -1094,48 +1094,70 @@
     (dissoc row-map :environment_id)
     row-map))
 
+(defn normalize-resource-event
+  "Prep `event` for comparison/computation of a hash"
+  [event]
+  (-> event
+      (utils/update-when [:timestamp] to-timestamp)
+      (utils/update-when [:old-value] sutils/db-serialize)
+      (utils/update-when [:new-value] sutils/db-serialize)
+      (assoc :containing-class (find-containing-class (:containment-path event)))))
+
+(defn normalize-report
+  "Prep the report for comparison/computation of a hash"
+  [report]
+  (-> report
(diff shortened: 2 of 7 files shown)


wkalt:
This is a kind of hard patch to test. To see that it works, what you need to do is

1. Clear your database and check out *stable*
2. Do a puppet run and stop puppetdb / master
3. Checkout master and start PDB so your report gets migrated
4. Curl reports and write down the hash
5. Export your data, clear your database, and reimport the data. Curl reports again and note that the hash does not match.

Repeat the above with this patch instead of stable, and the hashes should match. If anyone has thoughts on a good way to test for this I'd be interested.

test/com/puppetlabs/puppetdb/test/scf/storage.clj
```diff
@@ -1094,11 +1094,18 @@
     (is (= (map :name (query-to-vec "SELECT name FROM certnames ORDER BY name ASC"))
            ["node1" "node3"]))))
 
+(defn update-event-timestamps
+  "Changes each timestamp in the `report`'s resource-events to `new-timestamp`"
+  [report new-timestamp]
+  (update-in report [:resource-events]
+             (fn [events]
```
ajroetker: Should this just be `(partial map #(assoc % :timestamp new-timestamp))`?

src/com/puppetlabs/puppetdb/scf/hash.clj
```diff
@@ -132,16 +132,16 @@
...
-      :resource-title resource-title
-      :property property
-      :timestamp timestamp
-      :status status
-      :old-value old-value
-      :new-value new-value
-      :message message
-      :file file
-      :line line}))
+    {:resource_type resource-type
```
senior: Why underscores?

wkalt: @senior we need to hash underscores because master uses underscores. This function is only used in hash generation though, so as long as we don't start using it elsewhere in stable I think we should be good.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1094,48 +1094,70 @@
     (dissoc row-map :environment_id)
     row-map))
 
+(defn normalize-resource-event
```
mullr: To me, 'normalize' means that there are many different shapes of input that we're converting to be one shape of output... but I don't think that's what's happening here.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1094,48 +1094,70 @@
...
+      (update-in [:end-time] to-timestamp)
+      (update-in [:resource-events] #(map normalize-resource-event %))))
+
+(defn convert-containment-path
+  "Convert the contain path from a collection to the jdbc array type"
+  [event]
+  (utils/update-when event
+                     [:containment-path]
+                     (fn [cp]
+                       (when cp
```
mullr: This should already be caught by update-when, yes?

mullr: n/m, I see now that update-when will call the function if the path ends in `false` or `nil`.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1094,48 +1094,70 @@
...
+      (utils/update-when [:new-value] sutils/db-serialize)
+      (assoc :containing-class (find-containing-class (:containment-path event)))))
+
+(defn normalize-report
+  "Prep the report for comparison/computation of a hash"
+  [report]
+  (-> report
+      (update-in [:start-time] to-timestamp)
+      (update-in [:end-time] to-timestamp)
+      (update-in [:resource-events] #(map normalize-resource-event %))))
```
mullr: Could be `(partial map normalize-resource-event)`, if we like point-free style. Me, I like what you have here.

senior: I also prefer the anon function

src/com/puppetlabs/puppetdb/scf/storage.clj
senior: Maybe it's differing definitions or splitting hairs (or both?) but what I was intending by this function name was normalizing the values. Timestamps as strings can come in with different formats, old/new values can have maps that are sorted arbitrarily etc, but we need some baseline from whatever way it came in to something that is repeatable before we compute the hash.

Is there something that reveals that intention better than normalize?

mullr: Ah, I didn't think through to the different inputs that to-timestamp accepts. That makes sense.

test/com/puppetlabs/puppetdb/test/scf/storage.clj
wkalt: this is just a straight backport of what we have in master, so I think any style concerns should be addressed there to ease the rollups.

PR #1313, opened 2015-03-20
(PDB-1290) fix inner join in migration 29

wkalt:
This fixes an issue with migrations from stable->master if facts exist in the
database without reports. Migration 29 was using an inner join
where a left outer join was required, which was causing the new certnames
table to leave behind everything without an associated report.

To reproduce the original problem,
git checkout stable
* start puppet, puppetdb
git checkout 1aafe0637f59bebf39ae61a1125907f5655b1e59
lein run benchmark -c postgres.ini -F acceptance/benchmark/facts -n 5 -N 20 -r 10
* stop puppetdb, puppet
git checkout master
* restart puppet, puppetdb
<migration fails>


senior:
+1 will merge once acceptance tests come back

PR #1315, opened 2015-03-23
(PDB-1127) remove the experimental API

wkalt:
This removes the 'experimental' API and the associated tests. The API used to host
the endpoints like reports, events, nodes, and resources that we currently serve from the versioned APIs. As far as I can tell it's been dead since 1.5.


wkalt:

wkalt:

senior:
@wkalt what about [query](https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/query/population.clj) and [dashboard](https://github.com/puppetlabs/puppetdb/blob/master/resources/public/dashboard/index.html#L171)?

PR #1316, opened 2015-03-23
(BKR-147) add Gemfile setting for BEAKER_VERSION for puppet, ...

anodelman:
... puppetdb, etc

- add support for BEAKER_VERSION env var, simplifies configuring beaker
  across jenkins


kbarber:
I think the PR into stable is enough, we'll roll up the change ourselves to master.

anodelman:
Thanks for the quick turn around!

PR #1318, opened 2015-03-24
(PDB-1228) Expansion support with honeysql

kbarber:
This series of patches solves the problem of many rows to one record, by implementing
json_agg for PostgreSQL only. Due to this only working for PostgreSQL we've switched to
using honeysql in the core engine to allow selecting fields more convenient.


kbarber:
Docs are coming.

src/puppetlabs/puppetdb/catalogs.clj
```diff
@@ -341,3 +384,54 @@
...
+  [edges :- edges-expanded-query-schema]
+  (map edge-query->wire-v6
+       (:data edges)))
+
+(pls/defn-validated dissoc-if-nil :- {s/Keyword s/Any}
+  [resource :- {s/Keyword s/Any}
+   key :- s/Keyword]
+  (if (get resource key)
+    resource
+    (dissoc resource key)))
```
wkalt: this function exists in kitchensink too

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -56,6 +54,36 @@
                    ["=" ["node" "active"] true]]
                   req))
 
+(defn add-extract [projections query]
```
ajroetker: This is just a note that this won't work for nested extracts.

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -63,31 +66,32 @@
...
+          (http/add-headers resp {:count (jdbc/get-result-count count-query)})
           resp)))))
 
 (defn produce-streaming-body
   "Given a query, and database connection, return a Ring response with the query
   results.
+
   If the query can't be parsed, a 400 is returned."
-  [entity version query paging-options db]
+  [entity version query paging-options db base-url]
```
ajroetker: It looks like this is just url-prefix when it gets passed through.

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -24,33 +25,35 @@
...
+  (fn [_ _ _ _]
     f))
 
 (defn stream-query-result
   "Given a query, and database connection, return a Ring response with the query
   results.
+
   If the query can't be parsed, a 400 is returned."
-  [entity version query paging-options db output-fn]
+  [entity version query paging-options db base-url output-fn]
```
ajroetker: So any reason for using base-url here instead of just passing url-prefix through? Just passing url-prefix and version through would save a lot of duplication.

src/puppetlabs/puppetdb/http/reports.clj
```diff
@@ -2,31 +2,35 @@
...
+               (:scf-read-db globals)
+               (str (:url-prefix globals) "/" (name version))))}))
 
 (defn reports-app
   [version]
-  (build-reports-app
-   (-> (query-app version)
-       (validate-query-params
+  (-> (query-app version)
+      (validate-query-params
```
ajroetker: Could we factor out `validate-query-params, wrap-with-paging-options, and verify-accepts-json` to a single piece of middleware?

src/puppetlabs/puppetdb/query/edges.clj
```diff
@@ -0,0 +1,74 @@
...
+;; MUNGE
+
+(pls/defn-validated munge-result-rows
+  "Convert empty rows to []"
+  [_
+   projected-fields :- [s/Keyword]
+   _
+   _]
+  (fn [rows]
+    (if (empty? rows)
```
ajroetker: Is this if necessary? won't map on an empty list return an empty list?

src/puppetlabs/puppetdb/query/factsets.clj
```diff
@@ -1,106 +1,64 @@
...
-(pls/defn-validated convert-types :- [converted-row-schema]
-  [rows :- [row-schema]]
-  (map (partial facts/convert-row-type [:type :value_integer :value_float]) rows))
-
-(defn int-map->vector
-  "Convert a map of form {1 'a' 0 'b' ...} to vector ['b' 'a' ...]"
-  [node]
-  (when (and
-         (map? node)
-         (not (empty? node)))
```
ajroetker: I think you can combine these two with a `(map? (seq node))`

ajroetker: Whoops sorry this code is gone!

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -738,75 +920,81 @@
...
-          (assoc qr :project project-map)
-          (assoc qr :late-project project-map)))
-      (let [[subquery-name & subquery-expression] expr]
-        (assoc (user-query->logical-obj subquery-name)
-          :project project-map
-          :where (when (seq subquery-expression)
-                   (user-node->plan-node (user-query->logical-obj subquery-name)
-                                         (first subquery-expression))))))))
+  (if (or (nil? expr)
+          (not (subquery-expression? expr)))
```
ajroetker: So this was a problem before as well but when I changed everything to use `late-project?` I didn't add `late-project? false` to the appropriate query-recs because `late-project? nil` would also be fine for the if statement below. The problem is that now those query-rec's aren't instances of the Query protocol because they don't have a `late-project? false` field. That means that the queries which actually do support projections to the sql won't. I think we just want to add `late-project? true` to all the query-recs.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -846,27 +1035,30 @@
...
-    (if (instance? Query (user-node->plan-node query-rec user-query))
-      where
-      (assoc query-rec :where where))))
+  [query-rec paging-options user-query]
+  (let [plan-node (user-node->plan-node query-rec user-query)
+        projections (projectable-fields query-rec)]
+    (if (instance? Query plan-node)
+      plan-node
+      (-> query-rec
+          (assoc :where plan-node)
```
ajroetker: Any reason to space these out over multiple assoc's?

src/puppetlabs/puppetdb/scf/hash_debug.clj
```diff
@@ -35,18 +38,56 @@
...
+   version]
+  (if (:data edges)
+    edges
+    (assoc edges :data (query-edges certname version))))
+
+(defn expand-catalog-resources
+  [resources
+   certname
+   version]
+  (if (:data resources)
```
ajroetker: could you use an update-when here?

test/puppetlabs/puppetdb/query/reports_test.clj
```diff
@@ -41,30 +86,28 @@
...
 (deftest paging-results
-  (let [hash1        (:hash (store-example-report! (:basic  my-reports) (now)))
-        hash2        (:hash (store-example-report! (:basic2 my-reports) (now)))
-        hash3        (:hash (store-example-report! (:basic3 my-reports) (now)))
-        hash4        (:hash (store-example-report! (:basic4 my-reports) (now)))
-        report1      (assoc (:basic  my-reports) :hash hash1)
-        report2      (assoc (:basic2 my-reports) :hash hash2)
-        report3      (assoc (:basic3 my-reports) :hash hash3)
-        report4      (assoc (:basic4 my-reports) :hash hash4)
+  (let [_ (store-example-report! (:basic  my-reports) (now))
```
ajroetker: Can these be moved to below this let now?

src/puppetlabs/puppetdb/catalogs.clj
kbarber: aah, meant to make a note to look this up later, thanks will fix.

src/puppetlabs/puppetdb/http/reports.clj
kbarber: Out of scope for this patch IMHO. I see we've raised another patch for this already.

test/puppetlabs/puppetdb/query/reports_test.clj
kbarber: yep

src/puppetlabs/puppetdb/query/edges.clj
kbarber: Yep it will, and this pattern is everywhere, let me fix this.

src/puppetlabs/puppetdb/query_eng/engine.clj
kbarber: fixed

src/puppetlabs/puppetdb/scf/hash_debug.clj
kbarber: well, really we want the inverse of update-when ...

ajroetker: oops yep my bad! I think maybe it's called `update-non-nil` elsewhere, @mullr has it named `update-maybe` in the sync repo.

mullr: Hey I just called you
And this seq's laa-zy
But here's my reference
So `update-maybe`

project.clj
```diff
@@ -68,11 +68,12 @@
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version]
-                 [prismatic/schema "0.2.2"]
+                 [prismatic/schema "0.2.6"]
```
wkalt: did we require this for some feature? No objection just curious.

kbarber: no, I bumped it because I found what I thought were bugs, but left it like this because its the highest version bump we can make without breaking API and there was mention of performance improvements around the parsing of schemas related to other schemas. I'm not beholden to it, it could stay or go it doesn't matter.

documentation/release_notes.markdown
```diff
@@ -181,15 +181,19 @@ derive data from a PuppetDB export tarball.
...
+    This allows import and export to work against a PuppetDB configured with a url prefix.
 
 * Add the `[puppetdb] disable-update-checking` configuration key ([PDB-158](https://tickets.puppetlabs.com/browse/PDB-158))
 
-    The value defaults to false; see the [configuration](./configure.html) documentation for more details. 
+    The value defaults to false; see the [configuration](./configure.html) documentation for more details.
+
+* Added ability to expand child data for reports/facts & catalogs endpoints.
+
+* The `reports` database table was using the hash string as its primary key, we know have switched to using a smaller bigint primary key for that table, which should result in faster joins in most cases, and hopefully smaller foreign key indexes sizes for tables relating to `reports`. ([PDB-1218](https://tickets.puppetlabs.com/browse/PDB-1218))
```
wkalt: "index sizes"

wkalt: also I think we can drop the hopefully since we know they'll be smaller

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -38,54 +43,114 @@
...
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Fact Exporting
 
-(defn parse-response
-  "The parsed JSON response body"
-  [{:keys [status body]}]
-  (when (= status 200)
-    (seq (json/parse-string body true))))
+(pls/defn-validated factsets-expansion :- {s/Keyword s/Any}
+  "Expands known expandeable fields."
```
wkalt: expandable

documentation/release_notes.markdown
mullr: "now have switched"

src/puppetlabs/puppetdb/http/reports.clj
```diff
@@ -2,31 +2,35 @@
...
+            [puppetlabs.puppetdb.http.query :as http-q]
+            [puppetlabs.puppetdb.http.events :as e]
             [puppetlabs.puppetdb.middleware :refer [verify-accepts-json validate-query-params
                                                     wrap-with-paging-options]]))
 
 (defn query-app
   [version]
   (app
-    [&]
+    [hash "events" &]
```
wkalt: I think we should consider changing the resource_events field name of reports to events if we're calling this one events.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -160,88 +147,33 @@
...
-                 start_time, end_time, transaction_uuid,
-                 COALESCE(distinct_names.status, ''),
-                 COALESCE(distinct_names.environment,''),
-                 configuration_version, certname FROM (%s) distinct_names %s%s%s) %s"
-                 sql sql inner-order-by limit-clause offset-clause order-by-clause)
-
-         (format "SELECT paged_results.* FROM (%s) paged_results%s%s%s"
-                 sql order-by-clause limit-clause offset-clause)))))
-
-(defn count-sql
```
wkalt: awesome to see all that junk finally gone

documentation/release_notes.markdown
kbarber: Lol, umm, yeah this point doesn't belong in this patch. But thanks for the review :-/.

src/puppetlabs/puppetdb/http/reports.clj
kbarber: I have another ticket to create for doing this, again, didn't want to add any more into it.

src/puppetlabs/puppetdb/cli/export.clj
kbarber: fixed

src/puppetlabs/puppetdb/query_eng.clj
kbarber: fixed

src/puppetlabs/puppetdb/query_eng.clj
kbarber: fixed

wkalt:
There's a substantial performance regression here in catalog querying. See
https://gist.github.com/wkalt/2ad3ff5e2bd0d99ff20b

src/puppetlabs/puppetdb/http/reports.clj
wkalt: cool

wkalt:
we seem to lose pretty printed SQL in the debug output with this patch -- would that be possible to get back?

test/puppetlabs/puppetdb/http/catalogs_test.clj
```diff
@@ -103,7 +106,7 @@
             expected (get paging-options p)]
         (is (= (map :certname expected) (map :certname response-body)))))))
 
-  (testing "/v4 endpoint is still responsive to old-style node queries"
-    (let [{:keys [body]} (get-response "/v4/catalogs" "myhost.localdomain")
+  (testing "endpoint is still responsive to old-style node queries"
+    (let [{:keys [body]} (get-response endpoint "myhost.localdomain")
```
wkalt: this functionality (curling /v4/catalogs/hostname) seems completely broken on postgres. Not sure how this test passes.

wkalt: This is the response I get:
}[~/Documents/puppetdb] (ticket/master/pdb-1228) $ curl -X GET http://localhost:8080/v4/catalogs/mbp.local
<html>
<head>
<title>Error 500 </title>
</head>
<body>
<h2>HTTP ERROR: 500</h2>
<p>Problem accessing /v4/catalogs/mbp.local. Reason:
<pre>    Value does not match schema: {:edges {:data (not (sequential? a-clojure.lang.PersistentHashSet))}}</pre></p>
</body>
</html>

wkalt: it actually appears not to pass this test on postgres

wkalt: sorry, this was due to a change I made to the code. This is fine.

src/puppetlabs/puppetdb/query/catalogs.clj
```diff
@@ -1,106 +1,111 @@
...
-    (into acc [edge])))
+(pls/defn-validated rtj->edge :- catalogs/edge-query-schema
+  "Convert the row_to_json PG output to real data."
+  [edge :- {s/Str s/Any}]
+  (-> edge
+      (set/rename-keys {"f1" :source_type
+                        "f2" :source_title
+                        "f3" :target_type
+                        "f4" :target_title
+                        "f5" :relationship})))
```
wkalt: this rename-keys call is one of the biggest time-consumers in catalog processing. I was able to cut the query time for catalogs by more than half by changing this code to use set/rename in edges->expansion instead of mapping this. The change would necessitate some schema alterations too since it outputs a set.

src/puppetlabs/puppetdb/query/catalogs.clj
```diff
@@ -1,106 +1,111 @@
...
+  (-> resource
+      (set/rename-keys {"f1" :resource
+                        "f2" :type
+                        "f3" :title
+                        "f4" :tags
+                        "f5" :exported
+                        "f6" :file
+                        "f7" :line
+                        "f8" :parameters})
+      (update-in [:parameters] #(json/parse-strict-string % true))))
```
wkalt: rename-keys here is not nearly as costly as for edges, but using set/rename instead of mapping this function may still be a good idea.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -33,430 +39,606 @@
...
+                                                         (h/row
+                                                          :sources.type :sources.title :targets.type :targets.title
+                                                          :edges.type)))]
+                                              :from [:edges]
+                                              :join [[:catalog_resources :sources]
+                                                     [:= :edges.source :sources.resource]
+
+                                                     [:catalog_resources :targets]
+                                                     [:= :edges.target :targets.resource]]
+                                              :where [:= :edges.certname :c.certname]}}}
```
wkalt: this opens us up to arbitrarily large columns

kbarber:
@wkalt now the perf issue should be behaving better, it was as I said a cross-join problem with edges, it needed more constraints in the subquery basically.

For pretty formatted sql, we have solutions for this we've tried before (in particular org.hibernate.engine.jdbc.internal.BasicFormatterImpl) - we can reimplement those later on if required, perhaps dropping formatted sql only in a debug block so we don't do that heavy lifting unless necessary. I think though for now, we should leave it and come back to that problem.

src/puppetlabs/puppetdb/scf/hash_debug.clj
kbarber: update-when-not perhaps. Umm, I've got two use-cases here only, can I flag this as a tiny issue and not bother for now or do we feel strongly about this?

kbarber:
Docs are in. Current status:

* Decide what to do about nested extracts. Might create a new bug/ticket?
* Create tickets for 'parts not done'.

src/puppetlabs/puppetdb/http/query.clj
kbarber: I'm probably going to fork this into another ticket, it might be a fiddly fix, and we've done enough this patch.

wkalt:
@kbarber yeah, perf looks great now. Nice work.
https://gist.github.com/d66bac7b4cdd79537f23 
I'm okay with leaving the pretty formatting for later.

documentation/api/query/v4/edges.markdown
```diff
@@ -0,0 +1,102 @@
...
+### Response format
+
+The response is a JSON array of hashes, where each hash has the form:
+
+    {"certname": <string>,
+     "relationship": <string>,
+     "source_title": <string>,
+     "source_type": <string>,
+     "source_title": <string>,
+     "source_type": <string>}
```
wkalt: we should do the json indentation for this one too (like we do right below)

documentation/api/query/v4/upgrading-from-v3.markdown
```diff
@@ -39,7 +42,11 @@ Each change below is marked with the corresponding release version.
 
 #### /v4/factsets
 
-* (3.0) The v4 factsets endpoint will contain a new field `hash`.
+* (2.2) The v4 factsets endpoint was added to facilate the grouping of facts per node. For more information see the [/v4/factsets documentation](./api/query/v4/factsets.html).
```
wkalt: facilitate

documentation/release_notes.markdown
```diff
@@ -64,6 +64,30 @@ stable release of PuppetDB, which included an experimental v4 API.
     responses from the /v4/catalogs endpoint, queries to it, and the
     "replace catalog" and "replace facts" commands.
 
+* Expanded data format is now more formalized. ([PDB-1228](https://tickets.puppetlabs.com/browse/PDB-1228))
+
+    In the past in the experimental /v4 endpoint we used to return related data for endpoints directly
+    in the JSON response with very little formality. However, due to an internal change that only work for
```
wkalt: internal changes that only work

documentation/release_notes.markdown
```diff
@@ -64,6 +64,30 @@ stable release of PuppetDB, which included an experimental v4 API.
...
+    PostgreSQL we've decided to formalize the format of returning expanded/related data so that some of
+    it may be marked as optionally from a contractual perspective.
```
wkalt: optional

documentation/release_notes.markdown
```diff
@@ -64,6 +64,30 @@ stable release of PuppetDB, which included an experimental v4 API.
...
+
+    In the past for example, report events would return an array of events, now it returns that data inside
+    the `data` key, and a link to the original data in a `href` key.
+
+    Note the `data` key is only supported on PostgreSQL, users of HSQLDB will be forced to retrieve the related
```
wkalt: "will be forced" sounds overly coercive

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -127,12 +180,19 @@
                                                node)))
                       {:accept :json}))]
       (map
-       #(dissoc % :receive_time)
-       (map
-        #(merge % {:resource_events (events-for-report-hash base-url (get % :hash))})
-        body)))))
+       reports-expansion
+       body
+       (repeatedly #(identity base-url))))))
```
wkalt: #(identity x) is (constantly x)

src/puppetlabs/puppetdb/factsets.clj
```diff
@@ -0,0 +1,63 @@
...
+
+(pls/defn-validated fact-query->wire-v4
+  [fact :- fact-query-schema]
+  (-> fact
+      (dissoc :environment :certname)))
+
+(pls/defn-validated facts-list-to-map :- {s/Keyword s/Any}
+  [facts :- [fact-query-schema]]
+  (zipmap (map (comp keyword
+                     :name)
```
wkalt: this newline seems inconsistent with our general style

src/puppetlabs/puppetdb/factsets.clj
```diff
@@ -0,0 +1,63 @@
...
+   (map fact-query->wire-v4
+        (:data facts))))
+
+(pls/defn-validated factsets-query->wire-v4
+  [factsets :- [factset-query-schema]]
+  (map
+   #(-> %
+        (dissoc :hash :timestamp)
+        (update-in [:facts] facts-expanded->wire-v4)
+        (set/rename-keys {:facts :values}))
```
wkalt: this is out of scope for this patch, but should we consider renaming the "values" key of the facts wireformat to "facts"?

src/puppetlabs/puppetdb/http/edges.clj
```diff
@@ -0,0 +1,36 @@
...
+  (comp
+   (fn [{:keys [params globals paging-options]}]
+     (produce-streaming-body
+      entity
+      version
+      (params "query")
+      paging-options
+      (:scf-read-db globals)
+      (:url-prefix globals)))
+   http-q/restrict-query-to-active-nodes))
```
wkalt: we don't restrict on the catalogs app, so I think deactivated nodes will produce misleading hrefs for hsqldb users

wkalt: to follow up on that, I think it's better not to restrict

src/puppetlabs/puppetdb/http/resources.clj
```diff
@@ -16,7 +16,8 @@
                   version
                   (params "query")
                   paging-options
-                  (:scf-read-db globals)))
+                  (:scf-read-db globals)
+                  (:url-prefix globals)))
                http-q/restrict-query-to-active-nodes)}))
```
wkalt: also here -- I think this will cause people querying catalogs to get invalid hrefs if their nodes are deactivated.

src/puppetlabs/puppetdb/http/reports.clj
```diff
@@ -2,31 +2,35 @@
...
+    (let [projs ["new_value" "property" "file" "old_value"
+                 "containing_class" "line" "resource_type"
+                 "status" "resource_title" "timestamp"
+                 "containment_path" "message" ]]
+      (comp (e/events-app version) (partial http-q/project-query projs) (partial http-q/restrict-query-to-report hash)))
```
wkalt: line breaks would be nice here

src/puppetlabs/puppetdb/query/catalogs.clj
wkalt: this has been addressed

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -33,430 +39,617 @@
...
+      [field name]
+      (if expandable?
+        [:null name]
+        [field name]))))
+
+(defn honeysql-from-query
+  [{:keys [selection projections paging-options] :as query}]
+  "Convert a query to honeysql format"
+  (let [expand? (su/postgres?)]
+    (log/spy (-> selection
```
wkalt: log spy

wkalt: sorry, mixed that up with spyscope. as you see fit

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -520,27 +701,27 @@
...
   NotExpression
   (-plan->sql [expr]
-    (format "NOT ( %s )" (-plan->sql (:clause expr))))
+    [:not (-plan->sql (:clause expr))])
 
   Object
   (-plan->sql [obj]
-    (str obj)))
+    obj
+    #_(str obj)))
```
wkalt: left over?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -580,12 +762,12 @@
...
   appropriate plan node"
-  {"select_nodes" (assoc nodes-query :subquery? true)
-   "select_resources" (assoc resources-query :subquery? true)
-   "select_params" (assoc resource-params-query :subquery? true)
-   "select_facts" (assoc facts-query :subquery? true)
+  {"select_facts" (assoc facts-query :subquery? true)
+   "select_fact_contents" (assoc fact-contents-query :subquery? true)
+   "select_nodes" (assoc nodes-query :subquery? true)
    "select-latest-report" (assoc latest-report-query :subquery? true)
-   "select-fact-contents" (assoc fact-contents-query :subquery? true)})
```
wkalt: good catch

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -580,12 +762,12 @@
...
   "Keypairs of the stringified subquery keyword (found in user defined queries) to the
   appropriate plan node"
-  {"select_nodes" (assoc nodes-query :subquery? true)
-   "select_resources" (assoc resources-query :subquery? true)
-   "select_params" (assoc resource-params-query :subquery? true)
-   "select_facts" (assoc facts-query :subquery? true)
+  {"select_facts" (assoc facts-query :subquery? true)
+   "select_fact_contents" (assoc fact-contents-query :subquery? true)
+   "select_nodes" (assoc nodes-query :subquery? true)
    "select-latest-report" (assoc latest-report-query :subquery? true)
```
wkalt: we probably need to underscore this too

wkalt: actually whatever "select-latest-report" is it's completely undocumented and untested. Is that dead code?

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -56,6 +54,36 @@
...
+  (cm/match [query]
+            [["extract" columns expr :guard nil?]]
```
wkalt: I think this needs to be (expr :guard nil?). I'm also not sure what it does -- probably needs test coverage, I think it must never be getting hit right now because :guard is just another keyword.

wkalt:
this introduces some inconsistent behavior around extracts without subqueries. See that this
```
'query=["extract",["environment","certname"]]'
```
goes through on catalogs (but doesn't do what you want) and fails on facts

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -978,31 +972,68 @@
...
+    (update-in record ["facts" "data"] set)
+    (assoc record "facts" (dissoc facts "data"))))
+
+(defn munge-actual-factset
+  [factset]
+  (if (sutils/postgres?)
+    (update-in factset ["facts" "data"] set)
+    factset))
+
+(defn munge-actual-factsets
```
wkalt: "actual" is a bit confusing (but I get it). Maybe munge-factset-response or munge-factset-test-response?

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -1105,86 +1136,115 @@
...
-                         "e" "1"}
-                        "domain" "testing.com"
-                        "uptime_seconds" "6000"}
-               "timestamp" (to-string (to-timestamp "2013-01-01"))
-               "producer_timestamp" (to-string (to-timestamp "2013-01-01"))
-               "environment" "DEV"
-               "certname" "foo2"
-               "hash" "28ea981ebb992fa97a1ba509790fd213d0f98411"}]))
+               "hash" "b966980c39a141ab3c82b51951bb51a2e3787ac7"})))
+      (is (= (munge-actual-factsets (into [] (nth responses 1)))
```
wkalt: (first responses) ?

wkalt: second rather

test/puppetlabs/puppetdb/http/reports_test.clj
```diff
@@ -43,32 +55,42 @@
...
   [version reports]
-  (set (map report-response reports)))
+  (set (map report-response
+            reports)))
+
+(defn munge-actual-report
+  [report]
+  (-> report
+      (update-in [:resource_events] (comp keywordize-keys munge-resource-events))
+      strip-expanded))
```
wkalt: same comment with "actual" -- any action taken with factsets should be taken here

wkalt:
finally made it through -- found a few minor bugs and nitpicks. I'll take another pass on Monday.

src/puppetlabs/puppetdb/cli/export.clj
kbarber: k

documentation/release_notes.markdown
kbarber: agreed

src/puppetlabs/puppetdb/factsets.clj
kbarber: Yeah quite possibly, I'll make a note of it.

src/puppetlabs/puppetdb/http/reports.clj
kbarber: k

src/puppetlabs/puppetdb/query_eng/engine.clj
kbarber: well, I left this here (and the other) on purpose, its a debug log while honeysql is new, I thought it might be a good idea.

src/puppetlabs/puppetdb/query_eng/engine.clj
kbarber: yep

src/puppetlabs/puppetdb/query_eng/engine.clj
kbarber: Nope, its used to achieve the 'latest_report?' field which is how its tested as well, its a little weird I agree but its how it is today. At some point we should remove the specialisation.

kbarber: I'll fix the hyphens anyway though.

src/puppetlabs/puppetdb/http/query.clj
kbarber: Okay, I have a better answer to all this, but for now I think its best we drop this specialisation and just remove this function completely. I think the extract business here specifically for reports/<hash>/events can wait, the root /v4/events doesn't have it, for now I'm keeping it that way. The long answer, is sorting out the events endpoint properly, but thats going to require more work, in short I think the better answer is better expressiveness and querying for parent data instead of expressing it in the child. I'll talk to you more about it on Monday.

kbarber:
> this introduces some inconsistent behavior around extracts without subqueries. See that this
>
> 'query=["extract",["environment","certname"]]'
> goes through on catalogs (but doesn't do what you want) and fails on facts

@wkalt actually, that situation was pre-existing. Extracts with no query doesn't work on master either, I don't think it ever worked?

src/puppetlabs/puppetdb/http/edges.clj
kbarber: Actually, this is not the function that the /catalogs/node/edges endpoint hits ... so no it won't create missing href's during an export or a href follow, but yeah its weird I agree.

src/puppetlabs/puppetdb/http/resources.clj
kbarber: hehehe, well it won't during a follow as I've said, /catalogs/nodes/resources doesn't have an active node restrict. So for this endpoint more than anything, its dangerous to change this default since exported resources would expect this to be restricted. That is, it would need a change to the terminus and potentially we'd need to let people know we're breaking this for puppetdbquery queries and such that exist in the wild. I think ... we need to circle around to the active-nodes problem in another ticket. Its a squirrely one, and I'd like to put removing the concept on that table as a potential solution, so I'd rather punt the discussion if thats okay. For now, catalogs/* doesn't restrict, but some of the other endpoints do: nodes, resources, facts, factsets.

kbarber:

src/puppetlabs/puppetdb/jdbc.clj
kbarber: Agreed.

kbarber:

kbarber:
Transient maven failure on 1 cell, so its pretty much green ^^

kbarber:
Rebase needed.

src/puppetlabs/puppetdb/scf/hash_debug.clj
ajroetker: For sure not really that important

wkalt:

kbarber:

senior:
Looks like it's transient failures above, removing WIP

src/puppetlabs/puppetdb/honeysql.clj
```diff
@@ -0,0 +1,73 @@
...
+(pls/defn-validated coalesce :- SqlCall
+  [& args :- [(s/one key-or-sql "Keyword") key-or-sql]]
+  "coalesce(arg, ...) sql function"
+  (apply hcore/call :coalesce args))
+
+(pls/defn-validated scast :- SqlCall
+  [source :- s/Keyword
+   target :- s/Keyword]
+  "cast(source AS target) sql function"
+  (hcore/call :cast (hcore/raw (str (hfmt/to-sql source)
```
senior: Looks like 0.5.2 has been released and includes my cast patch upstream https://github.com/jkk/honeysql/blob/0.5.2/src/honeysql/format.clj#L110

src/puppetlabs/puppetdb/catalogs.clj
```diff
@@ -139,6 +139,49 @@
...
+   :type s/Str})
+
+(def resources-expanded-query-schema
+  "Resources expanded format schema."
+  {:href s/Str
+   (s/optional-key :data) [resource-query-schema]})
+
+(def catalog-query-schema
+  "Full catalog query result schema."
+  {(s/optional-key :certname) s/Str
```
senior: @ajroetker don't you have a function for this repeated optional thing?

src/puppetlabs/puppetdb/honeysql.clj
senior: Looks like it's in anything 0.5.0+

wkalt:
we have a minor inconsistency in that this query:
```
curl -X GET http://localhost:8080/v4/catalogs/fakenode
```
returns
```
{
  "error" : "Could not find catalog for fakenode"
}
```

but
```
curl -X GET http://localhost:8080/v4/catalogs/fakenode/edges
```
returns an empty list.

I actually wouldn't object if we changed all endpoints that currently return singleton maps to return arrays of length one, and had both these responses return [].

kbarber:

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -38,54 +43,114 @@
...
+                     (client/get
+                      (str (utils/base-url->str base-url)
+                           "/factsets?query="
+                           (url-encode
+                            (format "[\"=\",\"certname\",\"%s\"]" node)))
+                      {:accept :json}))]
+      (map
+       factsets-expansion
+       body
+       (repeatedly #(identity base-url))))))
```
senior: I think you just want (repeat base-url) here

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -38,54 +43,114 @@
...
+  (if data
+    value
+    (assoc value :data (retrieve-related-data base-url href))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Catalog Exporting
 
-(defn-validated catalog-for-node :- (s/maybe s/Str)
-  "Given a node name, retrieve the catalog for the node."
+(pls/defn-validated catalogs-for-node-query :- (s/maybe (s/pred seq? 'seq?))
```
senior: Why seq? Will something downstream fail if it's a vector?

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -127,12 +180,19 @@
                                                node)))
                       {:accept :json}))]
       (map
-       #(dissoc % :receive_time)
-       (map
-        #(merge % {:resource_events (events-for-report-hash base-url (get % :hash))})
-        body)))))
+       reports-expansion
+       body
+       (repeatedly (constantly base-url))))))
```
senior: Similar (repeat base-url) comment as above

src/puppetlabs/puppetdb/factsets.clj
```diff
@@ -0,0 +1,61 @@
...
+          (map :value facts)))
+
+(pls/defn-validated facts-expanded->wire-v4
+  [facts :- facts-expanded-query-schema]
+  (facts-list-to-map
+   (map fact-query->wire-v4
+        (:data facts))))
+
+(pls/defn-validated factsets-query->wire-v4
+  [factsets :- [factset-query-schema]]
```
senior: I'm curious as to why strict validation on input but not output on most of these conversion functions

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -24,33 +25,35 @@
...
           :facts [facts/query->sql facts/munge-result-rows]
           :event-counts [event-counts/query->sql (ignore-engine-params (event-counts/munge-result-rows (first paging-options)))]
           :aggregate-event-counts [aggregate-event-counts/query->sql (ignore-engine-params (comp (partial kitchensink/mapvals #(if (nil? %) 0 %)) first))]
           :fact-contents [fact-contents/query->sql fact-contents/munge-result-rows]
-          :fact-paths [facts/fact-paths-query->sql (ignore-engine-params facts/munge-path-result-rows)]
+          :fact-paths [facts/fact-paths-query->sql facts/munge-path-result-rows]
           :events [events/query->sql events/munge-result-rows]
           :nodes [nodes/query->sql nodes/munge-result-rows]
-          :environments [environments/query->sql (fn [_ _] identity)]
+          :environments [environments/query->sql (fn [_ _ _ _] identity)]
```
senior: ignore-engine-params from above

kbarber:
> we have a minor inconsistency in that this query:

This is pre-existing @wkalt, I think its a little late in the patch to be changing this now.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -466,47 +663,35 @@
...
-    (str/join " OR "
-              (map
-               #(format "%s %s %s"
-                        (-plan->sql %1)
-                        (:operator expr)
-                        (-plan->sql %2))
-               (maybe-vectorize-string (:column expr))
-               (maybe-vectorize-string (:value expr)))))
+    (concat [:or] (map
+                    #(identity [(:operator expr)
```
senior: Seems like
```
#(vector (:operator expr) (-plan->sql %1) (-plan->sql %2))
```
or
```
(fn [col value] [(:operator expr) (-plan->sql col) (-plan->sql value)])
```
would be better than identity here

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -33,432 +39,623 @@
...
+(defn projectable-fields
+  "Returns a list of projectable fields from a query record.
+
+   Fields marked as :query-only? true are unable to be projected and thus are
+   excluded."
+  [{:keys [projections]}]
+  (->> projections
+       (remove (comp :query-only? val))
+       keys
+       (into [])
```
senior: I don't think you need this here. Sort doesn't need a vec and doesn't return a vec even if you give it one.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -33,432 +39,623 @@
...
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Conversion from plan to SQL
 
+(defn queryable-fields
+  "Returns a list of queryable fields from a query record.
+
+  These are fields with the setting :queryable? set to true."
+  [{:keys [projections]}]
+  (->> projections
+       (remove (comp (complement :queryable?) val))
```
senior: I think
```
(filter (comp queryable? val)) will give you the same results

**Comment by kbarber** [comment]

**Comment by kbarber** [comment]
Here's my bucket list of issues/future design items/future fixes I made as part of this patch, it covers things that will be fixed later. I'll be converting these into tickets when I find the time: https://gist.github.com/kbarber/c0a07dc7ef1577b0264b

**Comment by kbarber** [comment]

File: documentation/api/query/v4/upgrading-from-v3.markdown [hunk 50]
```diff
@@ -24,6 +24,9 @@ Each change below is marked with the corresponding release version.
 
 #### /v4/catalogs
 
+* (3.0) The v4 catalogs endpoint has changed the response of the `edges` and `resources` fields
+  to be expanded. For more information see [/v4/catalogs documentation](./catalogs.html).
```
wkalt: we should also note that the catalogs endpoint will no longer return data for deactivated nodes

kbarber:

documentation/api/query/v4/upgrading-from-v3.markdown
kbarber: Done.

PR #1320, opened 2015-03-24
(maint) applied-migrations: don't ignore exceptions

rbrw:
Check to see if there is a schema_migrations table before querying it.
This eliminates these error messages from the postgresql log:

  STATEMENT:  SELECT version FROM schema_migrations ORDER BY version
  ERROR:  relation "schema_migrations" does not exist at character 21

A very large number of these messages are produced during the unit
tests.

 src/com/puppetlabs/puppetdb/scf/migrate.clj | 14 ++++++++------
 1 file changed, 8 insertions(+), 6 deletions(-)


mullr:
This looks good. :+1:

PR #1321, opened 2015-03-24
(PDB-1182) Factor out default puppetdb middleware

ajroetker:
This commit factors out the wrapping of the default puppetdb middleware
to a `wrap-with-puppetdb-middleware` that is to be used to for all the
puppetdb webservers.


ajroetker:
@senior @mullr This what you guys had in mind?

mullr:
Yeah, I like that.

ajroetker:

PR #1326, opened 2015-03-26
(PDB-1342) Fix trusted facts fallback

mullr:
When adding trusted facts, a fallback mechanism is used to pull them
from the node object if there aren’t any there already. This code is
broken in puppet 4; it’s unclear whether it happened to work in puppet 3
because there was a global request object available or because this 
code was never called. In any case, the change works in both.

### DIFF
--- a/puppet/spec/unit/indirector/facts/puppetdb_spec.rb
+++ b/puppet/spec/unit/indirector/facts/puppetdb_spec.rb
@@ -109,15 +109,30 @@ describe Puppet::Node::Facts::Puppetdb do
   end
 
   describe "#get_trusted_info" do
-
     it 'should return trusted data' do
+      node = Puppet::Node.new('my_certname')
+      trusted = subject.get_trusted_info(node)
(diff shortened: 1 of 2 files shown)


puppet/spec/unit/indirector/facts/puppetdb_spec.rb
```diff
@@ -109,15 +109,24 @@ def save
...
+      trusted = subject.get_trusted_info(node)
+      expect(trusted).to eq({'authenticated'=>"local", 'certname'=>'testing', 'extensions'=>{}})
+    end
 
-      if Puppet::Util::Puppetdb.puppet3compat?
-        Puppet[:trusted_node_data] = true
-      end
+    it 'should return trusted data when falling back to the node' do
+      # This removes :trusted_information from the global context, triggering our fallback code.
+      Puppet.rollback_context('initial testing state')
```
kbarber: This won't pass on Puppet 3.5.1.

mullr: Updated and tested against 3.7.1 and latest.

PR #1332, opened 2015-04-01
(PDB-1362) Drop values constraint earlier in fact paths migration

rbrw:
(PDB-1362) Drop values constraint earlier in fact paths migration

```
 src/com/puppetlabs/puppetdb/scf/migrate.clj       |  2 +-
 test/com/puppetlabs/puppetdb/test/scf/migrate.clj | 22 +++++++++++++++++++++-
 2 files changed, 22 insertions(+), 2 deletions(-)
```

### DIFF
--- a/src/com/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/com/puppetlabs/puppetdb/scf/migrate.clj
@@ -889,7 +889,6 @@
   value) to (id path value)."
   []
   (sql/do-commands
-   ;; Build complete facts table as of migration 28.
 
    "CREATE TABLE facts_unique_transform
       (factset_id bigint NOT NULL,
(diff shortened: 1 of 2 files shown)


wkalt:
+1

senior:
Jus to clarify, we can get away with changing the migration because users that had this issue will have had their transaction rolled back (so never made it to 28) and users that already have committed 28 won't care about running this since their migration already worked?

rbrw:
@senior I believe that's correct.  Since the constraint violation happened within the migration's  transaction, everything should have rolled back for those affected, and others should have ended up with the same result that (the forthcoming) 2.3.2 would have produced for them.

senior:
talked with @wkalt and he verified the above is correct

PR #1336, opened 2015-04-01
(maint) Test migration in a different schema

rbrw:


wkalt:
When I check out the 2.3.1 tag and cherry-pick this commit, my tests just stop at the migrate test without giving a reason for the failure:
```
lein test com.puppetlabs.puppetdb.test.scf.hash-debug

lein test com.puppetlabs.puppetdb.test.scf.migrate
Tests failed.
[~/work/puppetdb] ((detached from 2.3.1)) $
```

would it be possible to get this to fail more verbosely?

rbrw:
@wkalt don't think there's anything we can do, unless we're going to change migrate!'s error handling, since it currently calls (System/exit 1).

wkalt:
@rbrw That makes sense. I guess we have the same issue with some other tests as well already. I'm +1 on this.

rbrw:
@wkalt and I put up a ticket a while back that might or might not be relevant (depends on whether we decide we absolutely have to have a hard stop here): PDB-1118

test/com/puppetlabs/puppetdb/test/scf/migrate.clj
```diff
@@ -3,9 +3,10 @@
             [com.puppetlabs.puppetdb.scf.migrate :as migrate]
             [com.puppetlabs.puppetdb.scf.migration-legacy :as legacy]
             [com.puppetlabs.puppetdb.scf.storage :as store]
-            [com.puppetlabs.puppetdb.scf.storage-utils
+            [com.puppetlabs.puppetdb.scf.storage-utils :as storeutil
              :refer [db-serialize postgres?]]
             [cheshire.core :as json]
+            [clojure.string :as string]
```
ajroetker: Is this dep used?

rbrw: @ajroetker vestigial, and removed; thanks.

PR #1337, opened 2015-04-01
(maint) Support iterative acceptance test development

mullr:
Add support for iterating on acceptance tests more quickly by allowing a
developer to re-run them against an already created VM.

- Add PUPPETDB_SKIP_INSTALL and BEAKER_NO_PROVISION environment variables
  to the Rakefile
- Add a beaker configuration designated for running these tests 
  (ec2-west-dev)
- Add the beaker:first_run and beaker:rerun rake tasks
- Use a patched version of beaker that works on OS X hosts
- Document how this all works

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,8 +57,12 @@ group :acceptance do
     #use the specified version
     gem 'beaker', *location_for(beaker_version)
   else
-    #use the pinned version
-    gem 'beaker', '~>2.4'
+    # use the pinned version
+
+    # We're pinning to a a prerelease version of beaker that fixes errors when
+    # running with an OS X host. We only need to do this until a new version is
+    # released with the fix included.
+    gem 'beaker', :git => 'git://github.com/puppetlabs/beaker.git', :ref => '3e2e1e6d3e0234612c3a6af7dbfac06c797676f4'
   end
   # This forces google-api-client to not download retirable 2.0.0 which lacks
   # ruby 1.9.x support.
--- a/Rakefile
+++ b/Rakefile
@@ -3,15 +3,16 @@ require 'rake'
 RAKE_ROOT = File.dirname(__FILE__)
 
 def run_beaker(test_files)
-  config = ENV["BEAKER_CONFIG"] || "vbox-el6-64mda"
+  config = ENV["BEAKER_CONFIG"] || "ec2-west-dev"
   options = ENV["BEAKER_OPTIONS"] || "postgres"
   preserve_hosts = ENV["BEAKER_PRESERVE_HOSTS"] || "never"
+  no_provision = ENV["BEAKER_NO_PROVISION"] == "true" ? true : false
   color = ENV["BEAKER_COLOR"] == "false" ? false : true
   xml = ENV["BEAKER_XML"] == "true" ? true : false
   type = ENV["BEAKER_TYPE"] || "git"
-  rake_root = File.dirname(__FILE__)
+  keyfile = ENV["BEAKER_KEYFILE"] || nil
 
-  beaker = "beaker " +
+  beaker = "bundle exec beaker " +
      "-c '#{RAKE_ROOT}/acceptance/config/#{config}.cfg' " +
      "--type #{type} " +
      "--debug " +
@@ -22,6 +23,7 @@ def run_beaker(test_files)
 
   beaker += " --no-color" unless color
   beaker += " --xml" if xml
+  beaker += " --no-provision" if no_provision
 
   sh beaker
 end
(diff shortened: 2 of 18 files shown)


Rakefile
```diff
@@ -38,4 +40,25 @@ namespace :beaker do
...
+    args.with_defaults(:test_files => 'acceptance/tests/')
+    ENV["BEAKER_PRESERVE_HOSTS"] = "always"
+    run_beaker(args[:test_files])
+  end
+
+  desc "Re-run beaker based acceptance tests, using previously provisioned VMs"
+  task :rerun, [:test_files] do |t, args|
+    args.with_defaults(:test_files => 'acceptance/tests/')
+    ENV["PUPPETDB_SKIP_INSTALL"] = "true"
+    ENV["BEAKER_NO_PROVISION"] = "true"
```
ajroetker: Is there a case where we would want BEAKER_NO_PROVISION to be true and PUPPETDB_SKIP_INSTALL false or vice versa?

mullr: sure, if you have some raw VMs you want to set them up for testing.

mullr: In any case, the distinction here is because they're controlling different parts of the system. I think it's useful to directly expose what beaker is doing, and the PUPPETDB stuff messes with our own code.

ajroetker: I guess the PUPPETDB_SKIP_INSTALL is misleading to me because it forces us to skip provisioning completely unrelated to PuppetDB.

acceptance/setup/pre_suite/20_install_puppet.rb
```diff
@@ -1,6 +1,8 @@
 test_name "Install Puppet" do
-  step "Install Puppet" do
-    install_puppet
+  unless (test_config[:skip_install])
+    step "Install Puppet" do
```
ajroetker: This is an example to my comment about PUPPETDB_SKIP_INSTALL, I wouldn't expect that option to skip normal provisioning tasks like this.

Rakefile
```diff
@@ -38,4 +40,25 @@ namespace :beaker do
...
+
+  desc "Run beaker based acceptance tests, leaving VMs in place for later runs"
+  task :first_run, [:test_files] do |t, args|
+    args.with_defaults(:test_files => 'acceptance/tests/')
+    ENV["BEAKER_PRESERVE_HOSTS"] = "always"
+    run_beaker(args[:test_files])
+  end
+
+  desc "Re-run beaker based acceptance tests, using previously provisioned VMs"
+  task :rerun, [:test_files] do |t, args|
```
ajroetker: Could these two things (first_run/rerun) be combined to `:run` or something with an arg for `--test-only` or `--no-provisioning`?

Rakefile
mullr: True... what it does is skip provisioning ('installation') steps in our presuite. PUPPETDB_TURBO_PRESUITE?

ajroetker: :) PDB_BOOM_GOES_THE_DYNAMITE

Rakefile
mullr: Given the manual step required in between, I think it's better as two
tasks. Once that goes away, then I can automate away the difference and
this won't matter (I hope).

documentation/acceptance_tests.markdown
```diff
@@ -0,0 +1,52 @@
...
+* For now, you have to modify the host file you're using to include the IP
+  addresses for the VMs that were provisioned the first time. You should be able
+  to find these near the top of the console output. You could also ask the
+  hypervisor for help:
+
+      rake beaker:list_vms
+
+  Take the IP address of the VM that was created for you and put it in the
+  appropriate beaker hosts file. If you didn't specify one, the default is
+  `acceptance/config/ec2-west-dev.cfg`. It has a commented-out IP address field,
```
ajroetker: So to *maybe* solve the issue about this manual step in between, we could have the `first_run` deposit the appropriate beaker hosts file to `acceptance/config/local.cfg` which gets used if it exists, and that could also be used to skip the presuite in the first place if we wanted to right?

mullr: We *could* automate this just how you say, with step in there somewhere to
find the IP address of the VM... but that feature belongs in beaker, and
they said they're going to do it. So I think manual is ok for now.

mullr: https://tickets.puppetlabs.com/browse/BKR-167

ajroetker: ah cool! well mr. 2 steps ahead, the bulk of it lgtm otherwise

wkalt:
this has some travis issue -- might be fixable by pinning to a beaker sha rather than the fork. Also looks like it needs a rebase.

wkalt:
I tested this and aside from the travis issue it seems to work great.

mullr:

Gemfile
```diff
@@ -58,7 +58,7 @@ group :acceptance do
     gem 'beaker', *location_for(beaker_version)
   else
     #use the pinned version
-    gem 'beaker', '~>2.4'
+    gem 'beaker', :git => 'git://github.com/puppetlabs/beaker.git', :ref => '3e2e1e6d3e0234612c3a6af7dbfac06c797676f4'
```
wkalt: @mullr sorry to backtrack, but could we put a note here explaining the reason for the pin? Other than that this looks good.

Rakefile
rbrw: @mullr @ajroetker may not matter, but I wondered if there was any chance we might want to distinguish our testing vars from normal puppetdb vars, i.e PDBDEV_* or whatever.

acceptance/setup/pre_suite/40_install_deps.rb
```diff
@@ -1,93 +1,95 @@
...
+      if use_our_jdk then
+        create_remote_file database, '/etc/apt/sources.list.d/jpkg.list', <<-REPO
+  # Oracle JDK Packages
+  deb http://s3-us-west-2.amazonaws.com/puppetdb-jdk/jpkg/ pljdk main
+        REPO
+        # Import GPG key
+        on database, "gpg --keyserver keys.gnupg.net --recv-keys B8615A77BBBFA17C"
+        on database, "gpg -a --export B8615A77BBBFA17C | apt-key add -"
+        on database, "apt-get update"
+      end
```
rbrw: @mullr I'd be tempted to back out whitespace only changes.

documentation/acceptance_tests.markdown
```diff
@@ -0,0 +1,52 @@
...
+        :aws_secret_access_key: <your AWS secret key>
+
+* The included configuration files in `acceptance/config` refer to resources
+  (security groups, VPCs, etc) that exist in the Puppet Labs AWS account. If
+  you're using your own AWS account, you'll have to create the appropriate
+  resources and modify the appropriate configuration file to refer to them.
+
+Running the Tests
+-----------------
+* If you previously munged your host file, remove the IP address entry.
```
rbrw: @mullr very minor, but might be worth adding something like "as described below" regarding "previously munged".

acceptance/setup/pre_suite/40_install_deps.rb
mullr: It's not whitespace-only; the whole thing is wrapped in a conditional, so it's indented in an extra two spaces.

PR #1340, opened 2015-04-06
(maint) Always stop updater, and wait for finish

rbrw:
This is based on the rm-clamq patch, so should only go in after (being rebased on) that.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -201,9 +201,22 @@
     (.close mq-factory)
     (mq/stop-broker! broker)))
 
+(defn shutdown-updater
+  [context]
+  (when-let [updater (context :updater)]
+    (log/info "Shutting down updater thread.")
+    (future-cancel updater)
+    (let [timeout (Object.)]
+      (try
+        (when (= timeout (deref updater 3000 timeout))
+          (log/error "waited for updater thread for 3s but it didn't finish"))
+        (catch java.util.concurrent.CancellationException ex
+          true)))))
+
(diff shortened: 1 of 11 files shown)


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -204,30 +192,40 @@
...
 
+(defn shutdown-updater
+  [context]
+  (when-let [updater (context :updater)]
+    (log/info "Shutting down updater thread.")
+    (future-cancel updater)
+    (let [timeout (Object.)]
+      (try
+        (when (= timeout (deref updater 3000 timeout))
+          (log/error "waited for updater thread for 3s but it didn't finish"))
```
ajroetker: Like I said in the channel I don't think we necessarily want to log an error if they don't connect since it doesn't affect puppetdb, and it may be the case for some users that they aren't connected to the internet (which we've heard can often be the case for our services).

ajroetker:
Also I don't know how this exactly ties in but we might want to think about using https://github.com/puppetlabs/dujour-version-check as our updater code.

src/puppetlabs/puppetdb/cli/services.clj
rbrw: @ajroetker I believe this will only log an error if the updater thread can't be cancelled.  Though now that you mention it, perhaps it needs a clause like this too:
```
        (catch java.io.IOError ex ;; Covers situation where there's no net.
          true)
```

rbrw:
@ajroetker ok, then I'm fine pulling this PR in favor of considering the broader issue.  Thanks.

PR #1347, opened 2015-04-10
(maint) add upgrading from v3 document to stable branch

wkalt:
This adds the upgrading from v3 document to our stable branch,
for better visibility to users.

### DIFF
--- a/documentation/_puppetdb_nav.html
+++ b/documentation/_puppetdb_nav.html
@@ -56,6 +56,7 @@
   </li>
   <li><strong>Query API Version 4 (experimental)</strong>
     <ul>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/preparing-for-3.0.markdown">Upgrading from Version 3</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/query.html">Query Structure</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/operators.html">Available Operators</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/paging.html">Query Paging</a></li>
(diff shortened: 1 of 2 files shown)


wkalt:
this was a change suggested by @MosesMendoza. The stable branch has much better user visibility than master, so to the extent that we can communicate these changes from the stable docs, we should.

We need to be careful that we are sure that everything mentioned in this document will remain true through release time, and we should be sure to cut out mention of any changes that we may yet revert.

kbarber:
I think my concern is around the message in that document, it concerns itself with people upgrading from v3 to v4 final, not experimental. Plus it also mentions version 3.0 things. I just think we need to be careful about the message we are sending here, I don't want people to look at the document thinking this is how they go from v3 to v4 experimental etc. Maybe the note you've made already about this not being about v4 experimental needs to be put in highlights, like we do for other notes or something? Whatever we do, we have to be absolutely clear :-). Otherwise I'm +1 on the concept.

wkalt:
I added more emphasis on the fact that these changes don't all relate to released code. I left the 3.0 stuff in there though because I think some of it's pretty important and is set in stone, like dashes->underscores. If there are specific things we want to omit or qualify I'm open to it, but nothing pops out at me.

kbarber:
@wkalt just thinking about this some more, I realised that if we add a feature to v4, and we change this document there, we'll need to backport this each time to this document. What if ... we just hyperlink instead? We can talk about this some more when you're alive.

documentation/api/query/v4/upgrading-from-v3-api.markdown
```diff
@@ -0,0 +1,14 @@
+---
+title: "PuppetDB 2.3 » API » v4 » Upgrading to PuppetDB 3.0"
+layout: default
+canonical: "/puppetdb/latest/api/query/v4/upgrading-from-v3.html"
+---
+
+### Preparing for PuppetDB 3.0
```
kbarber: This title, and the document title, and the title in the nav are now different - did we want to fix this before the merge?

PR #1350, opened 2015-04-14
(PDB-1428) Suppress version-check INFO logging

grimradical:
The version-check function uses clj-http, which uses Apache HttpClient
for all the heavy lifting. Recent versions of HttpClient log connection
problems to INFO when using the library's built-in retry facilities.
This is suboptimal, as what's supposed to be a silent background process
(one where it's okay if it fails) to something that may freak out users
with scary looking error messages that are actually harmless.

If you disable automatic retries on the request, HttpClient throws an
exception; this matches the behavior of older versions. Since retries
aren't really required for version-checking, this patch disables them
and thus returns control of logging back to us.

### DIFF
--- a/src/com/puppetlabs/puppetdb/version.clj
+++ b/src/com/puppetlabs/puppetdb/version.clj
@@ -57,6 +57,7 @@
         query-string           (ring-codec/form-encode version-data)
         url                    (format "%s?product=puppetdb&%s" update-server query-string)
         {:keys [status body]}  (client/get url {:throw-exceptions false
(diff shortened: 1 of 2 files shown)


wkalt:
@grimradical I get a 400 error with this patch when I run
```
puppet agent -t
```

wkalt:
@grimradical my mistake, configuration error

kbarber:
Looks like its working to me: https://gist.github.com/kbarber/45b1ae455375581febfb

PR #1351, opened 2015-04-14
(maint) Bump to latest pristmatic/schema

kbarber:
This aligns us with trapperkeeper bumping to latest schema, so we won't have
conflicts later on.


kbarber:
For some reason travis is stating there is a permission issue installing sqlite3 gem. No idea where this came from, but I've seen it twice now (unrelated to this patch).

kbarber:
Okay, this was a transient issue with travis as I suspected. All green now.

project.clj
```diff
@@ -69,7 +69,8 @@
                  [puppetlabs/kitchensink ~ks-version]
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version]
-                 [prismatic/schema "0.2.2"]
+                 [prismatic/schema "0.4.0"]
+                 [prismatic/plumbing "0.4.2"]
```
senior: Does TK and schema 0.4.0 now depend on different versions of plumbing? Is that why we need to declare a direct dependency on plumbing?

kbarber: TK right now (as in the version we use) depends on plumbing 0.2.1, which is not compatible with schema 0.4.0. So if we don't include plumbing this fails. We could I think ... wait for a TK release, bump to that and probably avoid this extra requirement if so desired.

senior: No, I think it's ok to merge, makes sense why we're adding it.

senior:
+1 will merge when the current pipeline clears

PR #1354, opened 2015-04-15
(MAINT) Build pe-puppetdb from this repo.

waynr:


project.clj
```diff
@@ -105,7 +106,7 @@
                        :build-type "foss"
                        :main-namespace "puppetlabs.puppetdb.main"
                        :create-varlib true}
-                :config-dir "ext/config"
+                :config-dir "ext/config/foss"
```
kbarber: oh, was this breaking something before?

kbarber:
@waynr what's the syntax to test this mate? ```lein with-profile pe ezbake build``` etc.? I just want to see if I can make it work end-to-end myself as a bit of a functional test, any other pointers would be useful.

waynr:
@kbarber strange, I didn't see your comment show up in my inbox...gotta muck around with email filters some more I suppose.

To test this, try `lein with-profile ezbake,pe ezbake build`. We could probably eliminate the need for the `ezbake` profile but I have been thinking of the `pe` profile as building on top of that.

project.clj
waynr: No, `ext/config` actually wasn't even being used (that directory doesn't exist).

I mainly added `foss` subdirectory here to demonstrate a further means of differentiation between `foss` and `pe` if necessary. I don't see any problem with both the `foss` and the `pe` build pointing to the same directory, especially since they aren't being used--just let me know what your preference is and I will update the PR.

kbarber:
@waynr hehehe ... new git lesson I've just learnt. Check this out:

```
[ken@kb puppetdb]# git fetch -u waynr +maint:maint
error: there are still refs under 'refs/heads/maint'
error: Cannot lock the ref 'refs/heads/maint'.
From git://github.com/waynr/puppetdb
 ! [new branch]      maint      -> maint  (unable to update local ref)
ruby-2.0.0-p481@puppetdb (git)-[master]
[ken@kb puppetdb]# git branch -d maint
error: branch 'maint' not found.
ruby-2.0.0-p481@puppetdb (git)-[master]
[ken@kb puppetdb]# ls .git/refs/heads/maint
master stable
```

Because I have branches in my clone that use maint/stable/name-of-patch it lays them out in folders in .git, but because your branch is called `maint` it refuses to do anything, it wants to create a file, but there is a directory. The lesson here is, because of our naming conventions you probably should avoiding calling any branch 'maint' as it will conflict with stuff like 'maint/stable/foo' which has been a convention at PL for quite some time. Fun fun fun. The solution for me was obviously:

```
[ken@kb puppetdb]# git fetch -u waynr maint:maint-foo
From git://github.com/waynr/puppetdb
 * [new branch]      maint      -> maint-foo
```

So I just call it `maint-foo` locally just for me.

kbarber:
My VPN isn't working ATM after that renewal we did, so I can't do this right now anyway. If anyone else on the team has the time and access to test this out, feel free - don't wait for me.

project.clj
ajroetker: What's `:config-dir` used for? If were not using it @waynr should it be an optional are to ezbake?

ajroetker:
:+1: package built successfully for me

http://jenkins-release.delivery.puppetlabs.net/job/pe-puppetdb-packaging-2015-04-17-15-13-17-0.1.0.SNAPSHOT.2015.04.17T1513

ajroetker:
@kbarber do the PE packaging templates for puppetdb have parity with the pe-puppetdb[https://github.com/puppetlabs/pe-puppetdb] repo's templates?

kbarber:
@ajroetker well, pe-puppetdb only reflects stable (2.x), and is non-ezbake. This stuff is for master (3.x), and involves ezbake. I would imagine the files involved would be similar I guess, but definitely not the same. There were a lot of changes, it really depends on what templates you are exactly talking about.

PR #1357, opened 2015-04-17
(PDB-1416) Change the postgres type for hashes and uuids

ajroetker:
UUIDs stored as strings are about 36 bytes, and are 16 bytes stored as
the native type. Hashes are 40 bytes as characters, but only 20 has a
bytea field.
This commit changes PuppetDB to store the uuid's as their native type
and hashes as the bytea type (we can easily interact with the bytea type
by using the hexidecimal shorthand, prefixing hashes with '\x'.
This commit also enforces that our uuids be non-nil as all supported
versions of puppet enforce this.


ajroetker:

kbarber:
@ajroetker so now we have uuid & hash, does that mean the regexp operator will no longer work for these fields? Have we taken this into account yet?

test/puppetlabs/puppetdb/http/catalogs_test.clj
```diff
@@ -33,7 +33,7 @@
 (def catalog2 (merge catalog1
                  {:certname "host2.localdomain"
                   :producer_timestamp "2010-07-10T22:33:54.781Z"
-                  :transaction_uuid "000000000000000000000000000"
+                  :transaction_uuid "aaaaaaaa-1111-aaaa-1111-aaaaaaaaaaaa"
```
wkalt: this change breaks the ordering test

wkalt:
the regex operator seems to work for me

kbarber:
@wkalt for both the hash and uuid fields? Interesting.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -265,12 +265,29 @@ must be supplied as the value to be matched."
...
-(pls/defn-validated clj->pgjson :- PGobject
-  "Convert a clojure object to a json PGobject"
-  [value :- (s/maybe (s/either {s/Any s/Any} [{s/Any s/Any}]))]
-  (doto (PGobject.)
-    (.setType "json")
-    (.setValue (json/generate-string value))))
+(defn sql-hash-as-str
+  [column]
+  (if (postgres?)
+    (format "trim(leading '\\x' from %s::text)" column)
```
kbarber: Oh I see, we have a coercement to text here, hence why the regexp works.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -265,12 +265,29 @@ must be supplied as the value to be matched."
...
+    column))
+
+(defn sql-uuid-as-str
+  [column]
+  (if (postgres?)
+    (format "%s::text" column)
```
kbarber: Incidentally, we already have scast in honeysql.clj which does the same thing more or less, but uses cast(col as text) instead.

src/puppetlabs/puppetdb/query.clj
```diff
@@ -65,7 +65,7 @@
             [puppetlabs.puppetdb.jdbc :as jdbc]
             [puppetlabs.puppetdb.time :refer [to-timestamp]]
             [puppetlabs.kitchensink.core :refer [parse-number keyset valset order-by-expr?]]
-            [puppetlabs.puppetdb.scf.storage-utils :refer [db-serialize sql-as-numeric sql-array-query-string sql-regexp-match sql-regexp-array-match]]
+            [puppetlabs.puppetdb.scf.storage-utils :as sutils :refer [db-serialize sql-as-numeric sql-array-query-string sql-regexp-match sql-regexp-array-match]]
```
kbarber: Would it be possible to convert all these refers to just using the namespace? This is slowly what I've been doing as the refer's become too long, the advantage of using namespace all the time is that if we follow a rough convention for short names, we can just pick up functions & move them around to other files without having to do as much work in the require block.

src/puppetlabs/puppetdb/query.clj
```diff
@@ -331,13 +331,14 @@
...
         sql (format "SELECT %s
-                       FROM (SELECT c.hash as catalog, e.name as environment, catalog_id, resource,
+                       FROM (SELECT %s as catalog, e.name as environment, catalog_id, resource,
                                     type, title, tags, exported, file, line
                              FROM catalog_resources cr, catalogs c LEFT OUTER JOIN environments e
                                   on c.environment_id = e.id
                              WHERE c.id = cr.catalog_id) AS catalog_resources
                        JOIN catalogs ON catalog_resources.catalog_id = catalogs.id
                      WHERE %s"
+                    (sutils/sql-hash-as-str "c.hash")
```
kbarber: Shouldn't this be after the resource-columns in this format function call?

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -9,6 +9,7 @@
             [puppetlabs.puppetdb.query-eng.engine :as qe]
             [puppetlabs.puppetdb.schema :as pls]
             [puppetlabs.puppetdb.utils :as utils]
+            [puppetlabs.puppetdb.scf.storage-utils :as scf-utils]
```
kbarber: sutils is used in more files then scf-utils, its a convention schism I know, but I'd use sutils 'cos its shorter and avoids the hyphen :-).

src/puppetlabs/puppetdb/query.clj
rbrw: FWIW, I sometimes prefer refers (in some cases I think it can make the code (formatting) easier to read), but it depends on the situation.  In any case, I'd definitely prefer wrapping the long lines if we do keep the refers.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -12,12 +12,13 @@
             [puppetlabs.puppetdb.jdbc :as jdbc]
             [puppetlabs.puppetdb.query.paging :as paging]
             [puppetlabs.puppetdb.scf.hash :as hash]
+            [puppetlabs.puppetdb.utils :as utils]
             [puppetlabs.puppetdb.scf.storage-utils :as su]
```
kbarber: A third convention I forgot about. Hahaha.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -666,13 +670,21 @@
   (-plan->sql [query]
     (let [has-where? (boolean (:where query))
           has-projections? (not (empty? (:projected-fields query)))
-          update-when (fn [m pred ks f]
-                        (if pred
-                          (update-in m ks f)
-                          m))
+          update-hash-fields
```
kbarber: Huh. If we had this working in the query-rec's, we wouldn't need this, as it would be baked into the 'view'.

kbarber: Would this lower the amount of changes in general for this patch, if we got this working?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -951,11 +963,29 @@
...
   [query-rec
    columns :- [s/Str]]
-  (map #(get-in query-rec [:projections % :field])
+  ; This case expression here could be eliminated if we just used a projections list
+  ; and had the InExpression use that to generate the sql, but as it is the zipper we
+  ; use to walk the plan won't see instances of hashes and uuids among fields that have
+  ; gone through this function
+  (map #(let [projection (get-in query-rec [:projections %])]
+          (case (:type projection)
+            :uuid
```
kbarber: Similar point here actually, ie. if we had the conversion baked into the query-rec etc.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -23,7 +23,7 @@
             [puppetlabs.puppetdb.reports :as reports]
             [puppetlabs.puppetdb.facts :as facts :refer [facts-schema]]
             [puppetlabs.kitchensink.core :as kitchensink]
-            [puppetlabs.puppetdb.scf.storage-utils :as scf-utils]
+            [puppetlabs.puppetdb.scf.storage-utils :as sutils]
```
kbarber: Hahaha, yeah, I see that you started thinking like me at some point in this patch :-).

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -865,20 +875,18 @@
          (for [[pid vid] pairs]
            {:factset_id factset-id :fact_path_id pid :fact_value_id vid})))
 
+(defn foo [x] (prn x) x)
```
kbarber: Left over dev code?

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -278,12 +295,38 @@ must be supplied as the value to be matched."
   (when value
     (json/parse-string (.getValue value))))
 
+(pls/defn-validated str->pgobject :- PGobject
```
kbarber: Should the older pgjson->str (and reverse) conversion methods just wrap this now perhaps? Or maybe we no longer need the pgjson code at all any more?

kbarber: Sorry scratch that idea, I'm wrong I think.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -278,12 +295,38 @@ must be supplied as the value to be matched."
...
   [value]
-  (if (postgres?)
-    (clj->pgjson value)
-    (json/generate-string value)))
+  (let [json-str (json/generate-string value)]
+    (if (postgres?)
+      (str->pgobject "json" json-str)
+      json-str)))
+
+(pls/defn-validated pgjson->clj
```
kbarber: Is this a dup function?

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -126,6 +126,13 @@
       m
       (assoc-in m ks (apply f val args)))))
 
+(defn update-cond
```
kbarber: Could update-when now use this?

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -3,6 +3,7 @@
             [clojure.java.jdbc :as sql]
             [cheshire.core :as json]
             [puppetlabs.puppetdb.scf.storage :as scf-store]
+            [puppetlabs.puppetdb.scf.storage-utils :as scf-utils]
```
kbarber: Aha!

test/puppetlabs/puppetdb/examples.clj
```diff
@@ -5,7 +5,7 @@
    {:certname         "empty.catalogs.com"
     :api_version      1
     :version          "1330463884"
-    :transaction_uuid nil
+    :transaction_uuid "aaaaaaaa-1111-aaaa-1111-aaaaaaaaaaaa"
```
kbarber: Does changing this from nil have implications & coverage concerns for other tests now? I mean, right now we're testing a blank tuuid in other places, and this patch means we aren't any more. nil is still valid right?

test/puppetlabs/puppetdb/testutils/resources.clj
```diff
@@ -5,7 +5,7 @@
             [puppetlabs.puppetdb.fixtures :refer :all]
             [puppetlabs.puppetdb.jdbc :refer [with-transacted-connection]]
             [puppetlabs.puppetdb.scf.storage :refer [add-facts! ensure-environment]]
-            [puppetlabs.puppetdb.scf.storage-utils :refer [db-serialize to-jdbc-varchar-array]]))
+            [puppetlabs.puppetdb.scf.storage-utils :as scf-utils :refer [db-serialize to-jdbc-varchar-array]]))
```
kbarber: Same points as before.

kbarber:
@ajroetker I'm done on my review, and its generally looking pretty good mate, well done. I think my main point is that it feels like for query-eng stuff we can move the work into the query-rec perhaps, saving a lot of hassle in conversion later on potentially. Obviously your heads in the code, so if you can see a problem with this I'm happy to hear it.

src/puppetlabs/puppetdb/query_eng/engine.clj
ajroetker: Yeah I really wasn't sure what to do there, for `testutils` we also have `tutils`, `tu`, sometimes `utils`. I would prefer `sutils` in general because we use `utils` and not `u` for that namespace.

src/puppetlabs/puppetdb/query_eng/engine.clj
ajroetker: So you can't munge the hashes and uuids in the query-recs becaue they are `def`s and not `defn`s. Because they are `def`s clojure tries to evaluate the `sutils/postgres?` function before the db connection is available, at compile time, and gets a null pointer exception.

src/puppetlabs/puppetdb/query.clj
ajroetker: Oh nice catch

src/puppetlabs/puppetdb/scf/storage_utils.clj
ajroetker: Precisely, and test code which also needs the coercement.

src/puppetlabs/puppetdb/scf/storage_utils.clj
ajroetker: Since the `pgobject->str` is really just `(.getValue ...)` I didn't really see the need since it would really just be a renaming.

src/puppetlabs/puppetdb/utils.clj
ajroetker: Oh totally!

src/puppetlabs/puppetdb/query_eng/engine.clj
kbarber: Alright, hmm. This now bothers me perhaps. I wonder how often we call this during these decisions as well, postgres? is not memoized, I don't think it calls the db per se, it just grabs the JDBC meta data. I guess the only way around this is some sort of threading of db type instead ... hmm.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1230,6 +1278,7 @@
                                    (convert-to-plan query-rec paging-options)
                                    extract-all-params)
         sql (plan->sql plan)
+        ;_ (println sql)
```
wkalt: left over

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1124,17 +974,329 @@
...
+(defn version-2yz-to-300-migration
+  ;; This migration includes:
+  ;;   Insertion of the factsets hash column
+  ;;   Using a report_id (surrogate key) instead of the reports.hash (natural key)
+  ;;   Insert a noop column into reports
+  ;;   Drop latest_report table, merge with certnames
+  ;;   Change name to certname where inconsistent
+  ;;   Insert reports metrics and logs
+  ;;   Ensuring producer_timestamp is NOT NULL
+  ;;   Changing the types of hashes and uuids in postgres to bytea and uuid respectively
```
wkalt: I find the fact that we're collapsing six migrations into one a bit scary. Is there really a benefit?

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1124,17 +974,329 @@
...
+
+    (sql/do-commands
+      (str "INSERT INTO resource_params_transform
+            (resource, name, value)
+              SELECT " (munge-hash "resource") ", name, value
+              FROM resource_params"))
+
+    (sql/create-table :edges_transform
+                      ["certname" "TEXT" "NOT NULL"]
+                      ;; FIXME These should have FOREIGN KEY constraints on catalog_resources or resource_params_cache
```
wkalt: does this need to be addressed?

src/puppetlabs/puppetdb/scf/migrate.clj
kbarber: Usually its about reducing the amount of downtime for the upgrade, by simplifying any transformations into as little steps as possible.

wkalt: if there were a difference in timing that would be a benefit, but more steps doesn't necessarily mean an appreciable increase in downtime. I don't have a specific issue with this (seems to at least preserve all the indices and obvious stuff) but generally I'm against collapsing these without a solid reason for it. Makes debugging and understanding more difficult.

src/puppetlabs/puppetdb/scf/migrate.clj
ajroetker: So we don't currently have these on master, and I'm not sure where the actually FK should point to, we join edges on catalog_resources but catalog_resources has a fk constraint on resource_params_cache, what do you think?

src/puppetlabs/puppetdb/scf/migrate.clj
ajroetker: There's a few big benefits to this specific migration code as well @wkalt, because most of the tables from the report_id migration overlap with the uuid/hash migration, if we didn't collapse them we would have a bunch of duplicated migrations on master.

ajroetker: Plus most of these migrations were pretty small in the first place, so collapsing them was a trivial exercise

kbarber: Its true that doing it this way means review needs to be much more carefully done, to ensure we didn't lose an index or constraint especially, stuff a test won't show up.

src/puppetlabs/puppetdb/scf/migrate.clj
ajroetker: cc @kbarber

wkalt: looks like you could put in an fk constraint on resource_params_cache(resource), but if it wasn't prexisting it's out of scope here I'd say. I was just pointing it out because you've marked it as something we need.  If we just remove the FIXME I'm happy.

src/puppetlabs/puppetdb/scf/migrate.clj
wkalt: I don't want to beat a dead horse on this -- the collapsed migration seems alright as far as I can see. In cases like these, absent a real performance benefit, my tendency would be to keep things discrete and duplicate code as necessary to do so. The fact is that code is heavily duplicated across all of our migrations, since generally they happen across releases and refactoring existing ones is usually not an option.

wkalt:

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1124,17 +974,329 @@
...
+                      ["value_boolean" "boolean"]
+                      ["value_json"    "text"])
+
+    (sql/do-commands
+      (str "INSERT INTO fact_values_transform
+            (id, value_hash, value_type_id, value_integer, value_float, value_string, value_boolean, value_json)
+              SELECT id, " (munge-hash "value_hash") ", value_type_id, value_integer, value_float, value_string, value_boolean, value_json
+              FROM fact_values"))
+
+
```
wkalt: extra line

src/puppetlabs/puppetdb/scf/migrate.clj
ajroetker: @wkalt To some degree I would agree, things like Russ's timestamp patch or adding the noop column or renaming name to certname we didn't really get any benefit of collapsing here, it just obscured that fact those things were actually happening. The only reason I collapsed those was because I really wanted to collapse this migration with the report_id migration and since I was reordering things anyway it seemed more reasonable to collapse things. But when we have to drop our table/transform/rebuild our tables I would say there could be real performance concerns if we duplicate that procedure across migrations in the same release, and any benefit of keeping things discrete should be weighed against the complexity of duplicating migration code in a single release. Like @kbarber said though I think the major concerns of things like this is lack of testing around it.

ajroetker:

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -319,44 +321,44 @@
...
-                 certnames c INNER JOIN reports r on c.latest_report_id=r.id AND c.name=r.certname
-                 INNER JOIN environments e on r.environment_id=e.id
-                 INNER JOIN report_statuses rs on r.status_id=rs.id
-                 order by c.name")]
+                (format "SELECT %s AS hash, r.certname, e.name AS environment, rs.status
+                         FROM
+                         certnames c INNER JOIN reports r on c.latest_report_id=r.id AND c.certname=r.certname
+                         INNER JOIN environments e on r.environment_id=e.id
+                         INNER JOIN report_statuses rs on r.status_id=rs.id
+                         order by c.certname" (sutils/sql-hash-as-str "r.hash")))]
```
wkalt: do we need to be testing the transformation of transaction_uuid here as well?

wkalt:
I'm +1 on this aside from outstanding comments.

test/puppetlabs/puppetdb/scf/migrate_test.clj
ajroetker: I'll add that here

ajroetker:
@pljenkinsro restest this please

src/puppetlabs/puppetdb/scf/migrate.clj
kbarber: Another option or middle ground here is support fast forward migrations. That is, we implement a system that keeps the discrete migrations while also allow us to provide a fast forward migration for major upgrades, that reduces the upgrade, but means we keep the history of updates (like in this case we could make a 29...33 migration). Has the added benefit of eventually allowing us to collapse all migrations to. I guess for us though, it means taking on more complication.

src/puppetlabs/puppetdb/query_eng/engine.clj
kbarber: Do you want to adjust this one as well?

ajroetker:
@wkalt @senior @mullr  I think everyone's comments have been addressed and tests are passing if we want to close this ticket before the sprint ends

PR #1358, opened 2015-04-19
(maint) Move late-project? from queries to columns

ajroetker:
This commit moves the late-project? field from the query-recs to
specific columns in the query-recs so that the specific fields that
queries can't do without are called out specifcally. The field is also
renamed to must-have? to align the name with the purpose.

### DIFF
--- a/src/puppetlabs/puppetdb/query/fact_contents.clj
+++ b/src/puppetlabs/puppetdb/query/fact_contents.clj
@@ -2,6 +2,7 @@
   (:require [puppetlabs.puppetdb.facts :as f]
             [puppetlabs.puppetdb.query-eng.engine :as qe]
             [puppetlabs.puppetdb.schema :as pls]
+            [puppetlabs.puppetdb.utils :as utils]
             [schema.core :as s]))
 
 (def row-schema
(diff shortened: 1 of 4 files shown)


ajroetker:
ping @wkalt so here's a start at least

ajroetker:
I imagine that we could could move from this to doing something like a `late-join` (keep the `fact_value_id` as a `must-have` field, do the `group-by`, and then a `join` of that whole thing with `type`/`path`.

ajroetker:

wkalt:
I'm +1 on this -- makes things a bit more explicit. Ultimately we'll need to handle the underlying issue of late-project (which is relevant to group-by) by pushing all projections to the database and getting rid of the must-have flag. I've created https://tickets.puppetlabs.com/browse/PDB-1447 to cover this work.

src/puppetlabs/puppetdb/query/fact_contents.clj
```diff
@@ -26,9 +27,9 @@
    an array structure."
   [row :- row-schema]
   (-> row
-      (update-in [:value] #(or (:value_integer row) (:value_float row)
+      (utils/update-when [:value] #(or (:value_integer row) (:value_float row)
                                (f/unstringify-value (:type row) %)))
```
mullr: Indentation here

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -941,8 +941,9 @@
   (if (or (nil? expr)
           (not (subquery-expression? expr)))
     (let [qr (assoc query-rec :where (user-node->plan-node query-rec expr))]
-      (if (:late-project? query-rec)
-        (assoc qr :late-projected-fields column-list)
+      (if-let [must-haves (seq (keys (filter (comp :must-have? val) (:projections query-rec))))]
```
mullr: This line is bugging me... I think it's probably because I don't understand the structure of `query-rec`, so I couldn't really tell you what it's doing. This could be a good place for an internal schema, if indeed they all share a schema. 

(I'm not asking for a change here, just thinking aloud)

mullr:
Generally lgtm, just some small comments.

ajroetker:

ajroetker:
@mullr updated per your comments

PR #1359, opened 2015-04-20
(PDB-1263) Use beaker EC2 subnet rotation

rbrw:
Use the beaker version at the current tip of
rbrw:tmp/rotate-ec2-subnets, and change all of the acceptance/config
files to use CONFIG level vpc_id and subnet_ids settings.

(Don't merge this.)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,11 +58,7 @@ group :acceptance do
     gem 'beaker', *location_for(beaker_version)
   else
     # use the pinned version
-
-    # We're pinning to a a prerelease version of beaker that fixes errors when
-    # running with an OS X host. We only need to do this until a new version is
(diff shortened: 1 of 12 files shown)


kbarber:
@rbrw looks like this change is available in Beaker 2.11.0 gem now, shall we update this PR to reflect that and remove its WIP status?

kbarber:
+1 if tests pass.

@doug-rosser I'm not sure if you're still using the aws VPC stuff for your testing, if so you might be interested in this also - the idea is that subnet-id's pin you to an availability zone, however sometimes they are busy, this patch @rbrw has done is about making it round-robin a series of subnet-id's instead to avoid this (ie. if one is busy it moves to the next etc.).

PR #1360, opened 2015-04-20
(PDB-1431) Remove PE packages from el-7 AMI

rbrw:
Switch to a hand-modified el-7 image -- the only change was to purge the
pe-* packages that had been left behind by a bug in the puppetdb-packer
scripts.


rbrw:

PR #1361, opened 2015-04-22
(PDB-1310) Add producer_timestamp to deactivate node command

mullr:
This is a partial fix to PDB-1310. It doesn't address the self-issued deactivate command, that will come in a different PR.


documentation/api/commands.markdown
```diff
@@ -90,9 +90,23 @@ payload of this command.
...
 
-The payload is expected to be the certname of a node, as a raw JSON
-string, which will be deactivated effective as of the time the command
-is *processed*.
+The payload is a JSON map, formatted as follows:
+
+     {
+      "certname": <string>,
+      "producer_timestamp": <datetime>
+     }
```
wkalt: I think we need to bump the command version

mullr: Ah, so we do.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1179,22 +1180,41 @@
...
-  [certname :- String]
-  (sql/do-prepared "UPDATE certnames SET deactivated = ?
+  [certname :- String & [timestamp :- pls/Timestamp]]
+  (let [timestamp (to-timestamp (or timestamp (now)))]
+   (sql/do-prepared "UPDATE certnames SET deactivated = ?
                     WHERE certname=? AND deactivated IS NULL"
-                   [(to-timestamp (now)) certname]))
+                    [timestamp certname])))
+
+
```
wkalt: extra line

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1179,22 +1180,41 @@
...
+               :limit 1}]
+    (sql/with-query-results result-set (hcore/format query)
+      (time-field (first result-set)))))
+
+(pls/defn-validated have-record-produced-after?
+  [entity :- s/Keyword
+   certname :- String
+   time :- pls/Timestamp]
+  (let [time (to-timestamp time)
+        time-field (if (= entity :reports) :start_time :producer_timestamp)]
```
wkalt: there may be some subtlety here around using start_time vs end_time. If the puppet run takes a very long time for some reason, the two could be quite different. I'm not sure what the right choice is exactly, but I think there's a possibility here that a puppet run starts at t=0, deactivate node issued at t=1, puppet run completes at t=2, report processed at t=3, and node deactivate is processed at t=4. I think my expectation under this situation would be for the deactivate to fail (because a report has come in), but as written here I believe it would would succeed. receive_time would be yet another candidate timestamp. I'm not sure that it's wrong as written, I think I just need a bit more clarity on the intent.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -971,43 +971,54 @@
...
-          (is (empty? (fs/list-dir discard-dir)))))))
+       (let [one-day   (* 24 60 60 1000)
+             yesterday (to-timestamp (- (System/currentTimeMillis) one-day))]
+         (sql/insert-record :certnames {:certname certname :deactivated yesterday})
+         (test-msg-handler command publish discard-dir
+           (is (= (query-to-vec "SELECT certname,deactivated FROM certnames")
+                  [{:certname certname :deactivated yesterday}]))
+           (is (= 0 (times-called publish)))
+           (is (empty? (fs/list-dir discard-dir)))
+           (sql/do-prepared "delete from certnames"))))))
```
wkalt: are these deletions handled by the with-test-db fixture at the beginning of each test? not sure why they're needed now but not before

wkalt: oh, because you're iterating. makes sense.

src/puppetlabs/puppetdb/scf/storage.clj
mullr: A few things:
- receive_time is definitely out, as it's from from the local clock.

- No matter what we do, it needs to converge in the same way no matter what
order the messages are received and processed in. start_time and end_time
are the choices.

- If an admin is logged into the master and tries to deactivate a node
while, at the same time, a puppet run is happening on that node, what
should happen? We can either prefer the deactivate or prefer the puppet
run. I think it makes sense to prefer the deactivate since it was
explicitly indicated by the user (i.e. use start_time), but I can't say I'm
really invested in that viewpoint.

wkalt: makes sense. I'm happy with it as is.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -214,12 +215,17 @@
...
 (defmethod process-command! [(command-names :deactivate-node) 2]
-  [{certname :payload {:keys [id]} :annotations} {:keys [db]}]
-  (jdbc/with-transacted-connection db
-    (when-not (scf-storage/certname-exists? certname)
-      (scf-storage/add-certname! certname))
-    (scf-storage/deactivate-node! certname))
-  (log/info (format "[%s] [%s] %s" id (command-names :deactivate-node) certname)))
+  [{:keys [payload annotations]} {:keys [db]}]
+  (let [certname (:certname payload)
+        producer_timestamp (to-timestamp (or (:producer_timestamp payload) (now)))
```
senior: why not use the (now) as a default value and avoid the or?

mullr: as in `(get payload :producer_timestamp (now))`? I always forget that's a
thing you can do, will fix.

ajroetker:
looks like you've got a terminus spec test failure

documentation/api/commands.markdown
kbarber: Should we fork out a new document to contain these details like we have for the other wire formats? I think up until now there has been minimal complexity so we haven't needed it.

mullr: I'm not sure. I'm mostly ok leaving this as is, as it's still pretty small, but could probably be convinced otherwise.

ajroetker:
Looks like a legitimate GC error in the acceptance tests

documentation/api/commands.markdown
kbarber: Well, its all about creating obvious navigation I guess, right now our other commands have their own documents and this one doesn't (largely for historical reasons, wire formats use to mean something else, ie. they were related to queries also). Also, the formats are different, 

But beyond just creating this new document, lots of other things bother me about this document, the separation of data related to the command seems to be in two places, this higher level doc, and the individual wire format docs. I'd like to see a more concerted effort to push the per command data into their own docs over time, in particular the historical blurb we add here about differences to the last revision, and even examples (which we have at the bottom of the document). I also want to see the term "wire format" replaced with something more related to commands, since I feel like wire format is more of an internal term these days, and it might not make sense to consumers. Anyway.

mullr:

ajroetker:
^ transient

src/puppetlabs/puppetdb/command.clj
```diff
@@ -213,13 +214,18 @@
...
+(defmethod process-command! [(command-names :deactivate-node) 3]
+  [{:keys [payload annotations]} {:keys [db]}]
+  (let [certname (:certname payload)
+        producer_timestamp (to-timestamp (:producer_timestamp payload (now)))
+        id  (:id annotations)
+        have-newer? (fn [entity] (scf-storage/have-record-produced-after? entity certname producer_timestamp))]
+    (jdbc/with-transacted-connection db
+      (when-not (scf-storage/certname-exists? certname)
+        (scf-storage/add-certname! certname))
+      (when-not (or (have-newer? :catalogs) (have-newer? :factsets) (have-newer? :reports))
```
ajroetker: What would you think about `(some have-newer? [:catalogs :factsets :reports])` here?

mullr: I like it!

mullr:
@kbarber @ajroetker I think this is all set now, if you want to take one more glance.

documentation/api/commands.markdown
```diff
@@ -88,11 +89,14 @@ or missing fields are an error.
...
 payload of this command.
 
-### "deactivate node", version 2
+### "deactivate node", version 3
 
-The payload is expected to be the certname of a node, as a raw JSON
-string, which will be deactivated effective as of the time the command
-is *processed*.
+* Previous versions of deactivate node required only the certname, as a raw JSON
+  string. It is now formatted as a JSON map, as an additional property has been
```
wkalt: might be worth mentioning what the property is

documentation/api/wire_format/deactivate_node_format_v3.markdown
```diff
@@ -0,0 +1,60 @@
...
+---
+
+PuppetDB receives deactivate node commands from puppet masters in the following wire format. 
+
+Deactivate Node Command Format
+-----
+
+### Version
+
+This is **version 3** of the decativate node command.
```
wkalt: decativate

documentation/api/wire_format/deactivate_node_format_v3.markdown
```diff
@@ -0,0 +1,60 @@
+---
+title: "PuppetDB 3.0 » API » Deactivate Node Wire Format, Version 3"
+layout: default
+canonical: "/puppetdb/latest/api/wire_format/deactivate_node_format_v3.html"
+---
+
+PuppetDB receives deactivate node commands from puppet masters in the following wire format.
```
wkalt: I think "Puppet masters" is our preferred capitalization per https://github.com/puppetlabs/puppet-docs/blob/master/style_and_usage.markdown#other-capitalization-questions

documentation/api/wire_format/deactivate_node_format_v3.markdown
```diff
@@ -0,0 +1,60 @@
...
+This is **version 3** of the decativate node command.
+
+### Encoding
+
+The command is serialized as JSON, which requires strict UTF-8 encoding. 
+
+### Upgrade notes
+
+Previous versions of this command required only the certname, as a raw JSON
+string. It is now formatted as a JSON map, as an additional property has been
```
wkalt: s/additional property/timestamp field

mullr:
@wkalt fixed up according to your comments, rebased

senior:

PR #1363, opened 2015-04-23
(re-4451) Updates for pe-puppetdb

shrug:
This adds the lein-ezbake plugin to the pe profile and changes
the PE artifact versioning to match FOSS, at least until the
pe-pdb component has a version >= the version of pe-puppetdb
already included in PE.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -128,7 +128,7 @@
              :ezbake {:dependencies ^:replace [[puppetlabs/puppetdb ~pdb-version]
                                                [org.clojure/tools.nrepl "0.2.3"]]
                       :name "puppetdb"
-                      :plugins [[puppetlabs/lein-ezbake "0.2.2"
+                      :plugins [[puppetlabs/lein-ezbake "0.2.9"
                                  :exclusions [org.clojure/clojure]]]}
              :pe {:dependencies ^:replace [[puppetlabs/puppetdb ~pdb-version]
                                            [org.clojure/tools.nrepl "0.2.3"]
@@ -139,7 +139,7 @@
                                        :main-namespace "puppetlabs.puppetdb.core"
                                        :create-varlib true}
                                 :config-dir "ext/config/pe"}
-                  :version ~pe-pdb-version
+                  :version ~pdb-version
                   :name "pe-puppetdb"}
              :testutils {:source-paths ^:replace ["test"]}
              :ci {:plugins [[lein-pprint "1.1.1"]]}}


project.clj
```diff
@@ -139,7 +139,9 @@
                                        :main-namespace "puppetlabs.puppetdb.core"
                                        :create-varlib true}
                                 :config-dir "ext/config/pe"}
-                  :version ~pe-pdb-version
+                  :plugins [[puppetlabs/lein-ezbake "0.2.8"
+                              :exclusions [org.clojure/clojure]]]
```
ajroetker: I'd propose consolidating the `:ezbake` and `:pe` versions of this dep to the top level. Would there be a good reason to keep them separate @kbarber @shrug @waynr?

shrug: We definitely should, but I'd suggest we do it in a later PR, so that we can move FOSS and PE to the 0.3 ezbake series independently, at least until the work to update the PE filesystem paths to match FOSS is complete.

kbarber: ezbake requires VPN access doesn't it? I'd rather not make it a hard dep globally if it means having to have private access each time, or consumers of the source have to comment it out to make things work.

project.clj
```diff
@@ -139,7 +139,9 @@
                                        :main-namespace "puppetlabs.puppetdb.core"
                                        :create-varlib true}
                                 :config-dir "ext/config/pe"}
-                  :version ~pe-pdb-version
+                  :plugins [[puppetlabs/lein-ezbake "0.2.8"
```
kbarber: Why are we special casing PE here? Why don't we just bump the revision for FOSS also? This particular change would have zero CI pipeline testing coverage for starters today, I'd at least feel more comfortable if FOSS was being tested with the new rev.

project.clj
```diff
@@ -139,7 +139,9 @@
...
+                              :exclusions [org.clojure/clojure]]]
+                  :version ~pdb-version
```
kbarber: good catch, and yet again, no CI pipeline - thats why these mistakes bleed through :-).

project.clj
shrug: Updated.

project.clj
shrug: It does require VPN access to use the build functionality (or if a dependency needs to be pulled from the internal repo), it could be used to generate the packaging artifacts that could be used externally, with some effort on the part of the user, however you're right it might be better to leave it tied to the profiles that require it.

kbarber: To be clear, my point about VPN is specifically about trying to retrieve the jar, not so much running the build after that fact. That is, non-staff would hit problems doing anything (even non build stuff) with our project if its at the top level, and they'd have to comment out the dep to go any further.

project.clj
kbarber: So the original way @waynr proposed running the build was in this PR when it was first introduced: https://github.com/puppetlabs/puppetdb/pull/1354

So it was `lein with-profile ezbake,pe ezbake build` which merged the profiles. This meant that you didn't need to repeat yourself by providing a second `:plugins` part. Perhaps we just need to bump this in the ezbake profile only, if this is how we plan on running the build for PE.

kbarber: Oh yeah, we now have 0.2.9 fwiw, probably worth bumping again.

PR #1366, opened 2015-04-24
(maint) Speed up storage.clj `<hash> in (?,?,?...)` queries

ajroetker:
This commit addresses an issue where the above query for both
`resources` and `facts` was very slow on Postgres because we were
trimming the bytea hash to a string which was causing Postgres to do a
table scan. This commit fixes this issue by changing the parameters in
the `(?,?...)` to be munged bytea hashes so we don't do a table scan.

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -86,8 +86,8 @@
       sql-query-and-params
       (let [limited-result-set (limit-result-set! limit result-set)]
         (-> limited-result-set
-            (convert-result-arrays)
-            (vec))))))
+            convert-result-arrays
+            vec)))))
 
 (defn query-to-vec
   "Take an SQL query and parameters, and return the result of the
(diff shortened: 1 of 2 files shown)


ajroetker:
@mullr Preformed the factchurn.sh test on this a few times and it looks like the minimum gap I could get between the hash query and the GC query was about 2x, ie the hash query is still 2x slower than the GC on average.

mullr:
How does it compare to the pre-patch perf?

ajroetker:

ajroetker:
@mullr I feel like I'm going crazy, it actually looks about the same (2x as long) but the query plans are definitely different...

ajroetker:
^transient

wkalt:
I'm perf testing this patch.

wkalt:
I think this patch is great! I've tested it with a really big structured fact to generate a lot of path/value churn. Here are the query plans and timings compared:

master: https://gist.github.com/c1b8e5004e23d0816631
this patch: https://gist.github.com/634aea9202de24057708

PR #1367, opened 2015-04-27
(maint) Bump to clojure 1.7.0-beta2

senior:
Found some failures in the events_test that was relying on the order of
entries in a map that has changed in 1.7.0. Created a version of
clojure.core/select-values that will preserve the order of selected keys
and switched to an ordered-map to fix the issue.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -27,7 +27,7 @@
   ;; requires lein 2.2.0+.
   :pedantic? :abort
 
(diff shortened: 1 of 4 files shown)


senior:
Creating this PR to see the acceptance tests green. Once that's green we may want to either roll back the version bump (just fix the tests) in preparation for the move, or merge it in as-is and follow the release cycle. RC is expected this week.

PR #1368, opened 2015-04-27
(PDB-1181) Add group_by and aggregate functions (count only)

wkalt:
This implements a group_by operator and the "count" aggregate function. It
should be functional on all endpoints except facts and fact-contents, which
will be addressed by PDB-1447.

### DIFF
--- a/documentation/api/query/v4/events.markdown
+++ b/documentation/api/query/v4/events.markdown
@@ -187,19 +187,25 @@ For all events in the report with hash
 To retrieve all of the events within a given time period:
 
     ["and", ["<", "timestamp", "2011-01-01T12:01:00-03:00"],
-            [">", "timestamp", "2011-01-01T12:00:00-03:00"]]
+      [">", "timestamp", "2011-01-01T12:00:00-03:00"]]
 
-To retrieve all of the 'failure' events for nodes named 'foo.*' and resources of
-type 'Service':
+To retrieve all of the 'failure' events for nodes with name matching 'foo.\*'
+and resources of type 'Service':
 
     ["and", ["=", "status", "failure"],
-            ["~", "certname", "^foo\\."],
-            ["=", "resource_type", "Service"]]
+      ["~", "certname", "foo.\*"],
+      ["=", "resource_type", "Service"]]
 
 To retrieve latest events that are tied to the class found in your update.pp file:
 
     ["and", ["=", "latest_report?", true],
-            ["~", "file", "update.pp"]]
+      ["~", "file", "update.pp"]]
+
+To retrieve counts by event status for a particular node:
+
+    ["extract", [["function", "count"], "status"],
+      ["=", "certname", "foo.com"],
+      ["group_by", "status"]]
 
 ## Paging
 
--- a/documentation/api/query/v4/operators.markdown
+++ b/documentation/api/query/v4/operators.markdown
@@ -125,15 +125,40 @@ Every argument of these operators should be a **complete query string** in its o
 
 To reduce the keypairs returned for each result in the response, you can use **extract**. Using extract outside of a subquery (which is discussed below) has the form below:
 
-    ["extract" ["hash" "certname" "transaction_uuid"]
-      ["=" "certname" "foo.com"]]
+    ["extract", ["hash", "certname", "transaction_uuid"]
+      ["=", "certname", "foo.com"]]
 
 When only extracting a single column, the [] are optional
 
-    ["extract" "transaction_uuid"
-      ["=" "certname" "foo.com"]]
+    ["extract", "transaction_uuid"
+      ["=", "certname", "foo.com"]]
+
+When applying an aggregate function over a group-by clause, an extract
+statement takes the form
+
+    ["extract", [["function", "count"], "status"],
+      ["=", "certname", "foo.com"],
+      ["group_by", "status"]]
+
+Extract can also be used with a standalone function application:
+
+    ["extract", [["function", "count"]], ["~", "certname", ".\*.com"]]
 
 At this time extract must always have an expression to extract from, like `["=" "certname" "foo.com"]` above.
+
+### `function`
+
+The **function** operator is used to call a function on the result of a
+subquery. Currently the only supported function is `count`. The function operator
+is applied within the first argument of an extract, as in the examples above.
+
+### `group_by`
+
+The **group_by** operator must be applied as the last argument of an extract,
+and takes one or more column names as arguments.
+
+**Note** Currently function and `group_by` are not supported on the facts or
+fact-contents endpoints.
 
 ## Subquery Operators
 
--- a/documentation/api/query/v4/upgrading-from-v3.markdown
+++ b/documentation/api/query/v4/upgrading-from-v3.markdown
@@ -129,3 +129,6 @@ Each change below is marked with the corresponding release version. Changes mark
   the `fact-contents` or `fact-paths endpoints`. This is documented with the other [operators](https://docs.puppetlabs.com/puppetdb/master/api/query/v4/operators.html#regexp-array-match).
   An example of usage is given at the bottom of the subquery-examples page
   linked just above.
+
+* (3.0) We have added the `group_by` and `function` operators, as well as
+  support for the `count` function. For more information, see the [operators documentation](https://docs.puppetlabs.com/puppetdb/master/api/query/v4/operators.html#function).
--- a/src/puppetlabs/puppetdb/query_eng.clj
+++ b/src/puppetlabs/puppetdb/query_eng.clj
@@ -48,11 +48,13 @@
           :resources [resources/query->sql resources/munge-result-rows]
           :edges [edges/query->sql edges/munge-result-rows]
           :catalogs [catalogs/query->sql catalogs/munge-result-rows])]
+
     (jdbc/with-transacted-connection db
       (let [{[sql & params] :results-query
              count-query :count-query
              projected-fields :projected-fields} (query->sql version query
                                                              paging-options)
+
              query-error (promise)
              resp (output-fn
                    (fn [f]
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -1,6 +1,6 @@
 (ns puppetlabs.puppetdb.query-eng.engine
   (:require [clojure.core.match :as cm]
-            [clojure.string :as str]
+            [clojure.string :as string]
             [clojure.tools.logging :as log]
             [honeysql.core :as hcore]
             [honeysql.helpers :as hsql]
@@ -24,7 +24,7 @@
 ;;; Plan - functions/transformations of the internal query plan
 
 (defrecord Query [projections selection source-table alias where
-                  subquery? entity])
+                  subquery? entity call group-by])
 (defrecord BinaryExpression [operator column value])
 (defrecord RegexExpression [column value])
 (defrecord ArrayRegexExpression [table alias column value])
@@ -667,13 +667,28 @@
         [:null name]
         [field name]))))
 
+(defn merge-function-options
+  "Optionally merge call and grouping into an existing query map.
+   Alias function calls with escape-quoted function name for hsqldb compatability.
+   For instance, (merge-function-options {} ['count' :*] ['status'])"
+  [selection call grouping]
+  (cond-> selection
+    call (hsql/merge-select [(apply hcore/call call) (format "\"%s\"" (first call))])
+    grouping (assoc :group-by (map keyword grouping))))
+
 (defn honeysql-from-query
-  [{:keys [selection projections paging-options] :as query}]
+  [{:keys [projected-fields group-by call selection projections]}]
   "Convert a query to honeysql format"
-  (let [expand? (su/postgres?)]
+  (let [expand? (su/postgres?)
+        call (when-let [[f & args] (when call (utils/vector-maybe call))]
+               (apply vector f (map keyword (or args [:*]))))
+        new-select (if (and call (empty? projected-fields))
+                     []
+                     (vec (remove nil? (map #(extract-fields % expand?)
+                                            (sort projections)))))]
     (log/spy (-> selection
-                 (assoc :select (vec (remove nil? (map #(extract-fields % expand?)
-                                                       (sort projections)))))))))
+                 (assoc :select new-select)
+                 (merge-function-options call group-by)))))
 
 (pls/defn-validated sql-from-query :- String
   [query]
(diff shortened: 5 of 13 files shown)


src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -16,6 +16,7 @@
             [puppetlabs.puppetdb.query.fact-contents :as fact-contents]
             [puppetlabs.puppetdb.query.nodes :as nodes]
             [puppetlabs.puppetdb.query.reports :as reports]
+            [clojure.pprint :refer [pprint]]
```
ajroetker: this isn't use anymore looks like

wkalt:

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1107,19 +1147,25 @@
 
 (declare push-down-context)
 
+(defn unsupported-fields
+  [field allowed-fields]
+  (let [supported-fns ["count"]]
+    (remove (set (concat allowed-fields (map #(vector "function" %) supported-fns)))
```
ajroetker: So we already have a function that does this in scf/storage.clj called basic-diff, any way we could move that to utils and use that?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1213,13 +1259,19 @@
              :state state}
             :else {:node node :state state}))
 
+(defn op-to-string
+  [op]
+  (if (vector? op)
+    (into [] (map string/lower-case op))
```
ajroetker: I think vec is supposed to be the convention here as of 1.7.0

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1213,13 +1259,19 @@
              :state state}
             :else {:node node :state state}))
 
+(defn op-to-string
+  [op]
```
ajroetker: Do we want a docstring or maybe a schema here?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -667,13 +667,31 @@
         [:null name]
         [field name]))))
 
+(defn merge-function-options
+  "Optionally merge call and grouping into an existing query map.
+   Alias function calls with escape-quoted function name for hsqldb compatability."
+  [m call grouping]
```
ajroetker: Could we have a schema in here or something in the docstring to show what a call might look like?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -667,13 +667,31 @@
...
+
+  (if-not call
+    m
+    (let [fn-call (apply hcore/call call)
+          fn-alias (format "\"%s\"" (name (first call)))]
```
ajroetker: I don't think this is necessary, you can just put `(first call)` and the `fn-alias` and honeysql will know what to do.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -667,13 +667,31 @@
...
+      (-> m
+          (update-in [:select] conj [fn-call fn-alias])
+          (#(if grouping (assoc % :group-by (map keyword grouping)) %))))))
```
ajroetker: If you used `cond->` this might be a little cleaner
```
(cond-> m
true (update-in [:select] conj [fn-call fn-alias])
grouping (assoc % :group-by (map keyword grouping))
```

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -667,13 +667,31 @@
...
+
 (defn honeysql-from-query
-  [{:keys [selection projections paging-options] :as query}]
+  [{:keys [projected-fields group-by call selection projections
```
ajroetker: Is there any way to make call a little more general, that we might more easily support multiple functions in a query at a later date?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1087,10 +1112,25 @@
             (map->InExpression {:column (columns->fields query-rec (maybe-vectorize column))
                                 :subquery (user-node->plan-node query-rec subquery-expression)})
 
+            [["extract" [["function" & fargs]] expr]]
+            (let [[f & args] fargs
+                  args' (if args args :*)]
```
ajroetker: `(or args :*)`

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1087,10 +1112,25 @@
...
+                  (assoc :call [f args'])
+                  (create-extract-node [] expr)))
+
             [["extract" column expr]]
             (create-extract-node query-rec (maybe-vectorize column) expr)
 
-            :else nil))
+            [["extract" columns expr ["group_by" & clauses]]]
+            (let [[[f & args] cols] (strip-function-calls columns)
+                  args' (if args args :*)]
```
ajroetker: Similarly `(or args :*)`

src/puppetlabs/puppetdb/query_eng/engine.clj
ajroetker: Also I might this about moving the default behavior to the honeysql-from-query function so you can dedup the logic here.

src/puppetlabs/puppetdb/query_eng/engine.clj
wkalt: this is not correct when it's a honeysql keyword, like count. Give it a shot.

wkalt:
@ajroetker rebased

ajroetker:

documentation/api/query/v4/events.markdown
```diff
@@ -187,19 +187,25 @@ For all events in the report with hash
 To retrieve all of the events within a given time period:
 
     ["and", ["<", "timestamp", "2011-01-01T12:01:00-03:00"],
-            [">", "timestamp", "2011-01-01T12:00:00-03:00"]]
+      [">", "timestamp", "2011-01-01T12:00:00-03:00"]]
```
ajroetker: Did you want to adjust the alignment this way?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -24,7 +24,7 @@
 ;;; Plan - functions/transformations of the internal query plan
 
 (defrecord Query [projections selection source-table alias where
-                  subquery? entity])
+                  subquery? entity late-project? call group-by])
```
ajroetker: Left over `late-project?` field after rebase

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -667,13 +667,32 @@
...
+(defn merge-function-options
+  "Optionally merge call and grouping into an existing query map.
+   Alias function calls with escape-quoted function name for hsqldb compatability.
+   For instance, (merge-function-options {} ['count' :*] ['status'])"
+  [selection call grouping]
+  (cond-> selection
+    call (update-in [:select] conj [(apply hcore/call call) (format "\"%s\"" (first call))])
+    grouping (assoc :group-by (map keyword grouping))))
+
+(defn maybe-vectorize
```
ajroetker: This is already a function in utils.clj, might have got left over after rebase

ajroetker: The function is called vectorize-maybe fwiw

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -667,13 +667,32 @@
         [:null name]
         [field name]))))
 
+(defn merge-function-options
+  "Optionally merge call and grouping into an existing query map.
+   Alias function calls with escape-quoted function name for hsqldb compatability.
+   For instance, (merge-function-options {} ['count' :*] ['status'])"
+  [selection call grouping]
+  (cond-> selection
+    call (update-in [:select] conj [(apply hcore/call call) (format "\"%s\"" (first call))])
```
ajroetker: There's a honeysqy from called `merge-select` that you should probably use instead of conj.

ajroetker: Also I'm a little confused about `(format "\"%s\"" (first call))`, why can't you just do `(keyword (first call))`

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -667,13 +667,32 @@
...
+  (if (vector? arg) arg [arg]))
+
 (defn honeysql-from-query
-  [{:keys [selection projections paging-options] :as query}]
+  [{:keys [projected-fields group-by call selection projections]}]
   "Convert a query to honeysql format"
-  (let [expand? (su/postgres?)]
+  (let [expand? (su/postgres?)
+        call (when-let [[f & args] (when call (maybe-vectorize call))]
+               (apply vector f (map keyword (or args [:*]))))
```
ajroetker: You could just kewordize the function name as well, would save you some trouble above as well.

src/puppetlabs/puppetdb/query_eng/engine.clj
wkalt: @AJ hsqldb does not support keywords as aliases unless they're escape quoted. count is the keyword in this case, but any function name has the same issue. If you try making the change you're suggesting the tests will fail.

wkalt: sorry AJ, meant @ajroetker

ajroetker: Oh wow that's annoying

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -991,6 +1007,12 @@
   ; gone through this function
   (map #(get-in query-rec [:projections % :field]) (sort columns)))
 
+(defn strip-function-calls
+  [column]
+  (let [split (group-by #(= "function" (first %)) column)]
```
ajroetker: Could you destructure the split here instead of doing the gets below?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1214,13 +1255,20 @@
              :state state}
             :else {:node node :state state}))
 
+(pls/defn-validated op-to-string
```
ajroetker: Looks like this function is just `(comp (partial map string/lower-case) utils/vectorize-maybe)`

src/puppetlabs/puppetdb/query_eng/engine.clj
wkalt: no, because the order isn't guaranteed if true and false are not both represented.

src/puppetlabs/puppetdb/query_eng/engine.clj
wkalt: yep, thanks

documentation/api/query/v4/events.markdown
wkalt: yes. It makes it line up with "extract" below. Indenting on the array makes it look weird when the operators aren't all "and".

src/puppetlabs/puppetdb/query_eng/engine.clj
wkalt: sorry, I got you now. This is probably possible.

ajroetker: Or what I means is can't you just do `{functions true args false}` instead of `split`, functions and args could by nil, its the same as a get

wkalt:

ajroetker:
:+1: tested with the pe-state-overview endpoint work and with a few sample queries with malformed data/correct data, everything looked good

wkalt:
updated upgrading notes

wkalt:
^ transient failure there

documentation/api/query/v4/events.markdown
```diff
@@ -187,19 +187,25 @@ For all events in the report with hash
...
 
-To retrieve all of the 'failure' events for nodes named 'foo.*' and resources of
+To retrieve all of the 'failure' events for nodes named 'foo.\*' and resources of
```
kbarber: I think this statement didn't previously match the example, foo.* versus foo\\.

documentation/api/query/v4/events.markdown
```diff
@@ -187,19 +187,25 @@ For all events in the report with hash
...
 
 To retrieve latest events that are tied to the class found in your update.pp file:
 
     ["and", ["=", "latest_report?", true],
-            ["~", "file", "update.pp"]]
+      ["~", "file", "update.pp"]]
+
+To retrieve counts by event status for a particular node:
+
+    ["extract", [["function","count"],"status"],
```
kbarber: spacing between elements changes in this example between two styles.

documentation/api/query/v4/operators.markdown
```diff
@@ -133,8 +133,33 @@ When only extracting a single column, the [] are optional
     ["extract" "transaction_uuid"
       ["=" "certname" "foo.com"]]
 
+When applying an aggregate function over a group-by clause, an extract
+statement takes the form
+
+    ["extract" [["function" "count"] "status"]
```
kbarber: We need commas in JSON :-). In fact the example above is lacks them also, and the null? example, yeah we're lousy with it :-).

documentation/api/query/v4/upgrading-from-v3.markdown
```diff
@@ -129,3 +129,6 @@ Each change below is marked with the corresponding release version. Changes mark
   the `fact-contents` or `fact-paths endpoints`. This is documented with the other [operators](https://docs.puppetlabs.com/puppetdb/master/api/query/v4/operators.html#regexp-array-match).
   An example of usage is given at the bottom of the subquery-examples page
   linked just above.
+
+* (3.0) We have added the `group_by` and `function` operators, as well as
+  support for the `count` function. For more information, see the [operators documentation](https://docs.puppetlabs.com/puppetdb/master/api/query/v4/operators.html#function).
```
kbarber: This document is full of absolute URL's which we really should be avoiding. I know its not just this example, but yeah we should avoid this. Probably out of scope for this to fix it per se.

ajroetker:

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1214,13 +1251,20 @@
...
+  "Convert an operator to a lowercase string or vector of such."
+  [op]
+  (if (vector? op)
+    (vec (map string/lower-case op))
```
senior: mapv

PR #1371, opened 2015-04-30
(PDB-1447) remove late project

wkalt:
This commit adds a migration to change the value_json column of facts to a TEXT
"value" column, which stores a json-serialized copy of the properly typed fact.
This field is returned by facts queries and deserialized in code to avoid
dragging type information out of the database. Doing this allows us to remove
the concept of late-project.

### DIFF
--- a/src/puppetlabs/puppetdb/query/edges.clj
+++ b/src/puppetlabs/puppetdb/query/edges.clj
@@ -23,14 +23,9 @@
 
 (pls/defn-validated munge-result-rows
   "Reassemble rows from the database into the final expected format."
-  [_
-   projected-fields :- [s/Keyword]
-   _
-   _]
+  [_ _]
   (fn [rows]
-    (map (comp (qe/basic-project projected-fields)
-               #(s/validate edge-schema %))
-         rows)))
+    (map #(s/validate edge-schema %) rows)))
 
 ;; QUERY
 
@@ -62,11 +57,10 @@
   [version query-sql url-prefix]
   {:pre [[(map? query-sql)]]}
   (let [{[sql & params] :results-query
-         count-query    :count-query
-         projections    :projections} query-sql
+         count-query    :count-query} query-sql
          result {:result (query/streamed-query-result
                           version sql params
-                          (comp doall (munge-result-rows version projections {} url-prefix)))}]
+                          (comp doall (munge-result-rows version url-prefix)))}]
     (if count-query
       (assoc result :count (jdbc/get-result-count count-query))
       result)))
--- a/src/puppetlabs/puppetdb/query/events.clj
+++ b/src/puppetlabs/puppetdb/query/events.clj
@@ -16,14 +16,10 @@
 
 (pls/defn-validated munge-result-rows
   "Returns a function that munges the resulting rows ready for final
-  presentation."
-  [_
-   projected-fields :- [s/Keyword]
-   _
-   _]
+   presentation."
+  [_ _]
   (fn [rows]
-    (map (comp (qe/basic-project projected-fields)
-               #(utils/update-when % [:old_value] json/parse-string)
+    (map (comp #(utils/update-when % [:old_value] json/parse-string)
                #(utils/update-when % [:new_value] json/parse-string))
          rows)))
 
@@ -140,42 +136,3 @@
     ;; fall back to the old
     (legacy-query->sql version query-options query paging-options)
     (qe/compile-user-query->sql qe/report-events-query query paging-options)))
-
-;; QUERY + MUNGE
-
-(defn query-resource-events
-  "Queries resource events and unstreams, used mainly for testing.
-
-  This wraps the existing streaming query code but returns results
-  and count (if supplied)."
-  [version query-sql]
-  {:pre [(map? query-sql)]}
-  (let [{[sql & params] :results-query
-         count-query    :count-query
-         projections    :projections} query-sql
-         result {:result (query/streamed-query-result
-                          version sql params
-                          ;; The doall simply forces the seq to be traversed
-                          ;; fully.
-                          (comp doall (munge-result-rows version projections nil nil)))}]
-    (if count-query
-      (assoc result :count (jdbc/get-result-count count-query))
-      result)))
-
-(defn events-for-report-hash
-  "Given a particular report hash, this function returns all events for that
-   given hash."
-  [version report-hash]
-  {:pre [(string? report-hash)]
-   :post [(vector? %)]}
-  (let [query          ["=" "report" report-hash]
-        ;; we aren't actually supporting paging through this code path for now
-        paging-options {}]
-    (->> (query->sql version query [nil paging-options])
-         (query-resource-events version)
-         :result
-         (mapv #(dissoc %
-                        :run_start_time
-                        :run_end_time
-                        :report_receive_time
-                        :environment)))))
--- a/src/puppetlabs/puppetdb/query/fact_contents.clj
+++ b/src/puppetlabs/puppetdb/query/fact_contents.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.puppetdb.query.fact-contents
   (:require [puppetlabs.puppetdb.facts :as f]
             [puppetlabs.puppetdb.query-eng.engine :as qe]
+            [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.schema :as pls]
             [puppetlabs.puppetdb.utils :as utils]
             [schema.core :as s]))
@@ -10,10 +11,7 @@
    (s/optional-key :environment) (s/maybe s/Str)
    (s/optional-key :path) s/Str
    (s/optional-key :name) s/Str
-   (s/optional-key :value) (s/maybe s/Str)
-   (s/optional-key :value_integer) (s/maybe s/Int)
-   (s/optional-key :value_float) (s/maybe s/Num)
-   (s/optional-key :type) s/Str})
+   (s/optional-key :value) (s/maybe s/Str)})
 
 (def converted-row-schema
   {(s/optional-key :certname) s/Str
@@ -25,20 +23,16 @@
 (pls/defn-validated munge-result-row :- converted-row-schema
   "Coerce the value of a row to the proper type, and convert the path back to
    an array structure."
-  [{:keys [value_integer value_float type] :as row} :- row-schema]
+  [row :- row-schema]
   (-> row
-      (utils/update-when [:value] #(or value_integer value_float (f/unstringify-value type %)))
-      (utils/update-when [:path] f/string-to-factpath)
-      (dissoc :type :value_integer :value_float)))
+      (utils/update-when [:value] json/parse-string)
+      (utils/update-when [:path] f/string-to-factpath)))
 
 (pls/defn-validated munge-result-rows
   "Munge resulting rows for fact-contents endpoint."
-  [_
-   projected-fields :- [s/Keyword]
-   _
-   _]
+  [_ _]
   (fn [rows]
-    (map (comp (qe/basic-project projected-fields) munge-result-row) rows)))
+    (map munge-result-row rows)))
 
 (defn query->sql
   "Compile a query into an SQL expression."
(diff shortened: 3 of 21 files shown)


src/puppetlabs/puppetdb/query/facts.clj
```diff
@@ -31,26 +28,22 @@
...
+  [_ _ _]
   (fn [rows]
     (if (empty? rows)
       []
       (->> rows
-        convert-types
-        (map #(select-keys % (or (seq projected-fields)
-                                 [:certname :environment :timestamp :name :value])))))))
+           deserialize-fact-value
+           (map identity)))))
```
wkalt: @senior this is the line in question. I don't know why removing it causes the services test to fail. Validating on the whole collection instead of element by element in the deserialization function seems like a potential issue too -- I want to touch that up before this gets merged.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -52,6 +52,15 @@
...
 ;;; Queryable Entities
 
+(defn augment-for-subquery
+  [{:keys [entity] :as query-rec}]
+  (case entity
+    :facts (assoc-in query-rec
+             [:projections "value" :field]
+             (h/coalesce :fv.value_string
+                         (h/scast :fv.value_boolean :text)))
+    query-rec))
```
wkalt: I'm not wild about this, but the reason for it is that we allow queries like this:
```
curl -X GET http://localhost:8080/v4/events -d 'query=["in","new_value",["extract","value",["select_facts",["~","certname",".*"]]]]'
```

This presents a problem when the strings in the value column are serialized as json, since it becomes a comparison an escaped and nonescaped string.  This only comes into play when the facts query rec is a subquery, so I think this function leaves us on solid ground, but it's worth understanding the reasoning and thinking about alternatives. This makes no attempt to handle the case where new_value is numeric. That's not something we currently handle either.

wkalt:
rebased

wkalt:

src/puppetlabs/puppetdb/query/edges.clj
```diff
@@ -23,14 +23,9 @@
 
 (pls/defn-validated munge-result-rows
   "Reassemble rows from the database into the final expected format."
-  [_
-   projected-fields :- [s/Keyword]
-   _
-   _]
+  [_ _ _]
```
ajroetker: looks like this could just be changed to a `(def munge-result-rows (constantly (fn [rows] ....`

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -16,14 +16,10 @@
 
 (pls/defn-validated munge-result-rows
   "Returns a function that munges the resulting rows ready for final
-  presentation."
-  [_
-   projected-fields :- [s/Keyword]
-   _
-   _]
+   presentation."
+  [_ _ _]
```
ajroetker: similar note here

src/puppetlabs/puppetdb/query/fact_contents.clj
```diff
@@ -27,18 +28,14 @@
...
+      (utils/update-when [:value] json/parse-string)
+      (utils/update-when [:path] f/string-to-factpath)))
 
 (pls/defn-validated munge-result-rows
   "Munge resulting rows for fact-contents endpoint."
-  [_
-   projected-fields :- [s/Keyword]
-   _
-   _]
+  [_ _ _]
```
ajroetker: same note about the `def...constantly` here

src/puppetlabs/puppetdb/query/facts.clj
```diff
@@ -31,26 +28,21 @@
...
+  [row]
+  (utils/update-when row [:value] json/parse-string))
 
-(pls/defn-validated munge-result-rows
-  [_
-   projected-fields :- [s/Keyword]
-   _
-   _]
+(defn munge-result-rows
+  [_ _ _]
```
ajroetker: and here

src/puppetlabs/puppetdb/query/facts.clj
```diff
@@ -31,26 +28,21 @@
...
       []
       (->> rows
-        convert-types
-        (map #(select-keys % (or (seq projected-fields)
-                                 [:certname :environment :timestamp :name :value])))))))
+           (map deserialize-fact-value)))))
 
 (defn munge-path-result-rows
-  [_ _ _ _]
+  [_ _ _]
```
ajroetker: and here

src/puppetlabs/puppetdb/query/resources.clj
```diff
@@ -62,16 +62,11 @@
 (pls/defn-validated munge-result-rows
   "Munge the result rows so that they will be compatible with the version
   specified API specification"
-  [_
-   projected-fields :- [s/Keyword]
-   _
-   _]
+  [_ _ _]
```
ajroetker: and here

src/puppetlabs/puppetdb/query/facts.clj
ajroetker: which functions will actually still need any of these after this patch?

ajroetker:
really nice to see a lot of this stuff disappearing

src/puppetlabs/puppetdb/query/facts.clj
wkalt: the need isn't currently apparent, but it'll come back as soon as we cut more API versions after the 3.0 release. The main reason it looks strange now is because we recently dropped v2 and v3, which previously required special casing. I'm a sort of ambivalent about optimizing for our current situation for this reason, although I definitely see where you're coming from.

wkalt: to answer your question directly though, it's still required for every endpoint with a collapsing field

wkalt: eh, thought about it some more and I don't really mind moving to constantly. It may be some time before parameterization is required for a given endpoint.

wkalt:
rebased and added a migration test

wkalt:

wkalt:

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -973,6 +973,12 @@
                              :producer_timestamp test-time})
       (scf-store/deactivate-node! "foo4"))))
 
+(def db {:classname "org.postgresql.Driver"
```
ajroetker: left over?

ajroetker:
:+1: codewise mod the `db` map comment

wkalt:
good catch @AJ

wkalt:

wkalt:

wkalt:

ajroetker:

src/puppetlabs/puppetdb/client.clj
```diff
@@ -14,28 +14,28 @@
...
-                              :accept             :json}))))
+   (let [message (json/generate-string command-map)
+         checksum (kitchensink/utf8-string->sha1 message)
+         url (str (utils/base-url->str base-url)
+                  (format "/commands?checksum=%s" checksum))]
+     (http-client/post url {:body               message
+                            :throw-exceptions   false
+                            :content-type       :json
+                            :character-encoding "UTF-8"
+                            :accept             :json}))))
```
rbrw: @wkalt I'm not sure where we ended up on the topic (regarding merge conflicts, etc.), so ignore me if we decided we didn't mind, but if I'm reading this right, and all the changes to this file are indentation, I'd be tempted to omit them, here, and for other relevant files/sections in the diff.

rbrw: (...and I say that, even though some of the really long lines further down definitely make my reformatting finger twitch.)

test/puppetlabs/puppetdb/cli/import_export_roundtrip_test.clj
```diff
@@ -6,6 +6,7 @@
             [puppetlabs.puppetdb.fixtures :as fixt]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.command.constants :refer [command-names]]
+            [clojure.walk :refer [stringify-keys]]
```
rbrw: @wkalt might be extraneous?

test/puppetlabs/puppetdb/testutils/events.clj
```diff
@@ -71,6 +73,25 @@
...
+         result {:result (query/streamed-query-result
+                          version sql params
+                          ;; The doall simply forces the seq to be traversed
+                          ;; fully.
+                          (comp doall (events/munge-result-rows version nil)))}]
+    (if count-query
+      (assoc result :count (jdbc/get-result-count count-query))
+      result)))
+
 (defn resource-events-query-result
```
rbrw: @wkalt also maybe misunderstood, but thought we might be moving away from aligned let values.

test/puppetlabs/puppetdb/testutils/events.clj
```diff
@@ -71,6 +73,25 @@
...
+(defn query-resource-events
+  "Queries resource events and unstreams, used mainly for testing.
+
+  This wraps the existing streaming query code but returns results
+  and count (if supplied)."
+  [version query-sql]
+  {:pre [(map? query-sql)]}
+  (let [{[sql & params] :results-query
+         count-query    :count-query
+         projections    :projections} query-sql
```
ajroetker: This `:projections` doesn't get used i believe.

test/puppetlabs/puppetdb/testutils/events.clj
ajroetker: I imagine it was just copy/pasting leftovers

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -52,6 +52,15 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Queryable Entities
 
+(defn augment-for-subquery
+  [{:keys [entity] :as query-rec}]
+  (case entity
```
ajroetker: This is equivalent to `(cond-> query-rec (= :facts entity) (assoc-in [:projections ....))`, which I personally like a little more because it calls out what's being transformed/returned over the conditional

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1326,7 +1402,8 @@
    26 structured-facts-deferrable-constraints
    27 switch-value-string-index-to-gin
    28 lift-fact-paths-into-facts
-   29 version-2yz-to-300-migration})
+   29 version-2yz-to-300-migration
+   30 coalesce-fact-values})
```
ajroetker: I'm curious about your thoughts now @wkalt, I'm resurrecting the beaten dead horse. What about merging this migration with the previous one? We've seen some problems in the 2.3.z series with migrations taking a long time and we've seen here that migrations are indeed a big performance concern.

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -223,6 +223,7 @@
                 ["extract" ["certname" "nothing" "nothing2"] ["~" "certname" ".*"]]
                 #"Can't extract unknown 'facts' fields: 'nothing', 'nothing2'.*Acceptable fields are.*")))
 
+
```
ajroetker: extra line here?

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -788,7 +789,6 @@
                                                     {:order_by
                                                      (json/generate-string
                                                       [{"field" "value" "order" "ASC"}])}))))))))))
-
```
ajroetker: line got removed here?

rbrw:
@wkalt finished reviewing.

wkalt:
@ajroetker @rbrw thanks, addressed your comments

wkalt:

rbrw:
@wkalt might need a rebase?  Otherwise looks good.

rbrw:

ajroetker:
@rbrw !!

PR #1372, opened 2015-05-01
(maint) Detangle query-eng database and http streaming

ajroetker:
This commit addresses the "callback hell" of the query_eng.clj. Prior to
this commit the query-eng stream-query-result function had an output-fn
arguement whose interface was as follows, "output-fn takes a function
(f) as an argument and applies that funtion to another function (g),
(g) must realize the values of a lazy sequence of database rows."
This was primarily do to the fact the we need to stream results from the
database as well as stream results of http, which had to be done in a
separate thread. Another layer of complexity was that the http streaming
that we did also needed access to the count-query vector only available
int the stream-query result vector. All this complexity makes it very
confusing developing callbacks for querying puppetdb in process
directly, such as from PE extensions. This commit moves the http
streaming up a functional layer into produce-streaming body, passing
stream-query-result a promise to retrieve the count-query.

### DIFF
--- a/src/puppetlabs/puppetdb/query/event_counts.clj
+++ b/src/puppetlabs/puppetdb/query/event_counts.clj
@@ -116,10 +116,9 @@
   "Helper function to transform the event count subject data from the raw format that we get back from the
   database into the more structured format that the API specifies."
   [summarize_by]
-  (fn [rows]
-    (mapv
-     (partial munge-subject summarize_by)
-     rows)))
+  (fn [_ _] 
+   (fn [rows]
+     (map (partial munge-subject summarize_by) rows))))
 
 (defn query->sql
   "Convert an event-counts `query` and a value to `summarize_by` into a SQL string.
@@ -163,11 +162,12 @@
   {:pre  [(map? query-sql)]}
   (let [{[sql & params] :results-query
          count-query    :count-query} query-sql
-         result {:result (query/streamed-query-result
-                          version sql params
-                          ;; The doall simply forces the seq to be traversed
-                          ;; fully.
-                          (comp doall (munge-result-rows summarize_by)))}]
+         result {:result
+                 (query/streamed-query-result
+                  version sql params
+                  ;; The doall simply forces the seq to be traversed fully.
+                  (comp doall
+                        ((munge-result-rows summarize_by) nil nil)))}]
     (if count-query
       (assoc result :count (jdbc/get-result-count count-query))
       result)))
--- a/src/puppetlabs/puppetdb/query_eng.clj
+++ b/src/puppetlabs/puppetdb/query_eng.clj
@@ -19,62 +19,35 @@
             [puppetlabs.puppetdb.query.report-data :as report-data]
             [puppetlabs.puppetdb.query.resources :as resources]))
 
-(defn ignore-engine-params
-  "Query engine munge functions should take two arguments, a version
-   and a list of columns to project. Some of the munge funcitons don't
-   currently adhere to that contract, so this function will wrap the
-   given function `f` and ignore those arguments"
-  [f]
-  (constantly f))
+(defn entity->sql-fns
+  [entity version paging-options url-prefix]
+  (let [[query->sql munge-fn]
+        (case entity
+          :aggregate-event-counts [aggregate-event-counts/query->sql aggregate-event-counts/munge-result-rows]
+          :event-counts [event-counts/query->sql (event-counts/munge-result-rows (first paging-options))]
+          :facts [facts/query->sql facts/munge-result-rows]
+          :fact-contents [fact-contents/query->sql fact-contents/munge-result-rows]
+          :fact-paths [facts/fact-paths-query->sql facts/munge-path-result-rows]
+          :factsets [factsets/query->sql factsets/munge-result-rows]
+          :catalogs [catalogs/query->sql catalogs/munge-result-rows]
+          :nodes [nodes/query->sql (constantly identity)]
+          :environments [environments/query->sql (constantly identity)]
+          :events [events/query->sql events/munge-result-rows]
+          :edges [edges/query->sql edges/munge-result-rows]
+          :reports [reports/query->sql reports/munge-result-rows]
+          :report-metrics [report-data/metrics-query->sql (report-data/munge-result-rows :metrics)]
+          :report-logs [report-data/logs-query->sql (report-data/munge-result-rows :logs)]
+          :resources [resources/query->sql resources/munge-result-rows])]
+    [#(query->sql version % paging-options) (munge-fn version url-prefix)]))
 
 (defn stream-query-result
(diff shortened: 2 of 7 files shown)


ajroetker:
@wkalt @mullr @senior it looks scary but its actually less scary!! Basically stream-query-results has been fixed such that the callback that it takes is what we apply directly to the result-sets we stream from the database, so no more complicated callbacks and wrappers are necessary in the sync repo yada yada yada.

src/puppetlabs/puppetdb/query/event_counts.clj
```diff
@@ -116,10 +116,11 @@
   "Helper function to transform the event count subject data from the raw format that we get back from the
   database into the more structured format that the API specifies."
   [summarize_by]
-  (fn [rows]
-    (mapv
-     (partial munge-subject summarize_by)
-     rows)))
+  (constantly
```
wkalt: what is the reason for introducing constantly here?

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -19,65 +19,41 @@
...
           :resources [resources/query->sql resources/munge-result-rows]
           :edges [edges/query->sql edges/munge-result-rows]
-          :catalogs [catalogs/query->sql catalogs/munge-result-rows])]
+          :catalogs [catalogs/query->sql catalogs/munge-result-rows]))
 
+(defn stream-query-result
+  "Given a query, and database connection, return a Ring response with the query
+  results."
+  [entity version query paging-options db url-prefix row-callback & [count-query-promise]]
+  (let [[query->sql munge-fn] (entity->sql-fns entity (first paging-options))]
```
wkalt: I think it's a bit confusing to use (first paging-options), given that it's only applicable to event-counts

ajroetker:

senior:
I'm taking a look at this

src/puppetlabs/puppetdb/query/event_counts.clj
```diff
@@ -116,10 +116,11 @@
   "Helper function to transform the event count subject data from the raw format that we get back from the
   database into the more structured format that the API specifies."
   [summarize_by]
-  (fn [rows]
-    (mapv
-     (partial munge-subject summarize_by)
-     rows)))
+  (fn [_ _ _ _]
```
senior: Seems like this should be constantly since that's what you're using everywhere else

ajroetker: Actually @wkalt and I agreed it should probably be `(fn [_ _...] ...)` since eventually we'll probably use the `version` param again, i'll change them to be consistent.

src/puppetlabs/puppetdb/query/event_counts.clj
```diff
@@ -116,10 +116,11 @@
   "Helper function to transform the event count subject data from the raw format that we get back from the
   database into the more structured format that the API specifies."
   [summarize_by]
-  (fn [rows]
-    (mapv
-     (partial munge-subject summarize_by)
-     rows)))
+  (fn [_ _ _ _] 
+   (fn [rows]
+     (mapv
```
senior: This should just be map

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -86,8 +63,35 @@
   If the query can't be parsed, a 400 is returned."
   [entity version query paging-options db url-prefix]
   (try
-    (let [parsed-query (json/parse-strict-string query true)]
-      (stream-query-result entity version parsed-query paging-options db url-prefix http/stream-json-response))
+    (let [parsed-query (json/parse-strict-string query true)
+          query-error (promise)
+          count-query (promise)
```
senior: I find this count-query promise pretty confusing. Why do we need to get the query in such an indirect manner?

senior:
I tested the patch, works fine for me. Mostly minor comments, but I think we need to revisit the count-query stuff.

src/puppetlabs/puppetdb/query_eng.clj
ajroetker: I wanted to keep the api with `streamed-query-results` semi consistent so that we could make small changes to the extensions repo. The problem is essentially that we try to do too much in streamed-query-results, we do a query->sql, which then returns possibly two "queries", one which we then proceed to execute inside the streamed-query-results function, and the other which we really only need if we're doing a http query, the count-query. In a future world, I think we need a different scheme for all this, separating streamed-query-results from query->sql (which i have taken steps towards here by moving the entity-map stuff out of streamed-query-results) and then having streamed-query-results simply take a query arg and munge-fn. But this would require some pretty substantial changes to the extensions repo. Alternatively we could have two different `streamed-query-results` one which is used specifically for http stuff, and knows about http headers and executes the count query, and another which throws out the count query altogether for internal querying, like we need for sync. I wanted to take this intermediate step first to simplify things a little, if you think its just too complicated, I'm open to suggestions!

wkalt: another thing worth considering is that row counts are now available directly through the query language. I'm not exactly sure what people are using --include-total for, but if just calling the count function would suit their needs we could consider deprecating it.

kbarber: "Include total" is used for pagination, its a guide marker to see how many total records there are, so UI devs can produce a maximum page link in their UI.

wkalt: makes sense

senior: I do think we need to find another solution here, doesn't have to be a major shift or anything, but I think this is a complexity we can avoid.

ajroetker:
Refactored but relies on @wkalt's patch to remove late project

ajroetker:

ajroetker:

ajroetker:

ajroetker:

ajroetker:

ajroetker:
@senior that was indeed the problem, looks good now!

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -83,15 +56,30 @@
   If the query can't be parsed, a 400 is returned."
   [entity version query paging-options db url-prefix]
   (try
-    (let [parsed-query (json/parse-strict-string query true)]
-      (stream-query-result entity version parsed-query paging-options db url-prefix http/stream-json-response))
+    (let [[query->sql munge-fn] (entity->sql-fns entity version paging-options url-prefix)
+          {:keys [results-query count-query]} (jdbc/with-transacted-connection db
```
senior: Previously we would use one connection for this query and the one on line 75. Is there a reason to use two now?

ajroetker:
@senior fixed for your comment

ajroetker:

PR #1373, opened 2015-05-01
(PDB-1448) GC values that swap paths correctly

rbrw:
This is broken up into two commits so that it should be easier to test.  i.e. check out the first commit, which should fail "lein test" with a constraint violation, and then check out the tip, which should work.

### DIFF
--- a/src/com/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/com/puppetlabs/puppetdb/scf/storage.clj
@@ -779,11 +779,9 @@
   [factset-id removed-pid-vid-pairs]
   (when-let [removed-pid-vid-pairs (seq removed-pid-vid-pairs)]
     (let [vids (map second removed-pid-vid-pairs)
-          fp-id-pairs (map vector
-                           (repeat factset-id )
-                           (map first removed-pid-vid-pairs))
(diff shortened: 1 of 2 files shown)


wkalt:
@rbrw this looks good to me. Could you change the commit message to something describing the issue?

rbrw:
@wkalt pushed an elaborated commit message.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -779,11 +779,9 @@
   [factset-id removed-pid-vid-pairs]
   (when-let [removed-pid-vid-pairs (seq removed-pid-vid-pairs)]
     (let [vids (map second removed-pid-vid-pairs)
-          fp-id-pairs (map vector
-                           (repeat factset-id )
-                           (map first removed-pid-vid-pairs))
+          rm-facts (map cons (repeat factset-id) removed-pid-vid-pairs)
```
senior: I think you want interleave here

senior: Actually I think either is fine, looks like you're just throwing the structure around it away on line 796

PR #1374, opened 2015-05-01
(maint) make facts response streaming

wkalt:
Previously we were mistakenly realizing the value of our facts response by
validating it as a collection. This changes the validation so that it is applied
on a per-element basis and doesn't disrupt the laziness of the response.


senior:
+1 will merge when the pipeline is clear

PR #1376, opened 2015-05-05
(re-4451) Do not override main-namespace for pe builds

shrug:
The ezbake pe profile is currently overriding the main-namespace
for the project.

This removes that variable definition, taking the default for the
project.


haus:

PR #1378, opened 2015-05-05
(maint) allow ordering on non-extracted fields

wkalt:
This changes our paged sql function to do
```
select paged_results.* from (select blah from facts order by baz)
```
instead of
```
select paged_results.* from (select blah from facts) order by baz
```

the latter of which will not work without a late projection.

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -162,7 +162,7 @@
   (let [limit-clause     (if limit (format " LIMIT %s" limit) "")
         offset-clause    (if offset (format " OFFSET %s" offset) "")
         order-by-clause  (order-by->sql order_by)]
-    (format "SELECT paged_results.* FROM (%s) paged_results%s%s%s"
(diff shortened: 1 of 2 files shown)


kbarber:
Looking.

kbarber:
Is this just me? I'm not 100% but this should work now right?

```
# curl -X GET 'http://localhost:8080/v4/facts' -d 'query=["extract","certname",["=","certname","kb.local"]]&order_by=[{"field":"environment"}]'
<html>
<head>
<title>Error 500 </title>
</head>
<body>
<h2>HTTP ERROR: 500</h2>
<p>Problem accessing /v4/facts. Reason:
<pre>    org.postgresql.util.PSQLException: ERROR: column "environment" does not exist
  Position: 1042</pre></p>
</body>
</html>
```

wkalt:
@kbarber yeah, good catch. The issue here is that there are a number of cases where we alias a column when we extract it (in this case fact_environment to environment). When that aliasing doesn't happen the ordering fails. I'm closing the patch for now.

kbarber:
@wkalt do we have a ticket to track this bug otherwise? The whole order-by not working for this case seems bad.

PR #1379, opened 2015-05-06
(PDB-1310) Don't check reports in deactivate

mullr:
Report start_time comes from the agent, so it doesn't make sense to
compare it to the producer_timestamp on the deactivate command. The
real fix to this is to add a producer_timestamp to the store report
command, but as a stopgap we'll just revert to the old behavior for
these records.


mullr:

mullr:
this travis failure is bogus

mullr:
(testing again, since the failure is intermittent)

mullr:

mullr:
^^ transients

mullr:

PR #1380, opened 2015-05-06
(maint) Remove redirect to PuppetDB dashboard

ajroetker:
This commit changes the PuppetDB dashboard to be served as a file
response on "/" explicitly instead of a redirect to
"/dashboard/index.html". This allows us to refrain from passing
url-prefix to the main PuppetDB app `routes` function, as well as
modularize the dashboard to its own handler so it can be served
separately from the PuppetDB query app.

### DIFF
--- a/resources/public/index.html
+++ b/resources/public/index.html
@@ -89,8 +89,8 @@ body {
 }
 </style>
 
-<script src="d3.v2.js"></script>
-<script src="charts.js"></script>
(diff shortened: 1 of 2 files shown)


ajroetker:
Just something that occurred to me might be nice for our dashboard while going through an `om` tutorial.

ajroetker:
Closing as it complicates existing work

PR #1382, opened 2015-05-06
(PDB-1485) Deprecate url-prefix setting

ajroetker:
This commit deprecates the url-prefix setting under the `[global]`
header. This configuration is not necessary and done differently
for other trapperkeeper services.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -177,6 +177,8 @@ When this is set to true, debugging information will be written to `<vardir>/deb
 
 ### `url-prefix`
 
+> **Deprecated:** This setting should not be necessary in most cases and will not be supported in a future release.
+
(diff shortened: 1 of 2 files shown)


documentation/configure.markdown
```diff
@@ -177,6 +177,8 @@ When this is set to true, debugging information will be written to `<vardir>/deb
 
 ### `url-prefix`
 
+> **Deprecated:** This setting should not be necessary in most cases and this kind of configuration is done different by most other Puppet Labs services.
```
kbarber: `done differently`? Anyway ... do we need to say all this? Maybe just say its going and leave it at that.

kbarber:

kbarber:
Actually I'm just going to merge this in without waiting for tests, since @wkalt is waiting on this for release.

PR #1383, opened 2015-05-07
(PDB-1479) deprecate event counts and aggregate event counts

wkalt:
This adds a deprecation message in the docs for event-counts and
aggregate-event-counts, and also introduces a warning when the endpoints are used.

### DIFF
--- a/documentation/api/query/v3/aggregate-event-counts.markdown
+++ b/documentation/api/query/v3/aggregate-event-counts.markdown
@@ -9,6 +9,9 @@ canonical: "/puppetdb/latest/api/query/v3/aggregate-event-counts.html"
 [curl]: ../curl.html
 [query]: ./query.html
 
+> **Experimental Endpoint**: The aggregate-event-counts endpoint is designated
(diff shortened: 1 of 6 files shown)


kbarber:
I'm not 100% about leaving the deprecation message in code, since PE uses these endpoints. I say we either not do that, or just not drop that deprecation message for PE users. We've made this decision in the past by looking at the `product-name` configuration option, if pe-puppetdb, don't drop a message etc.

This is mainly about warning users of PE, when they can't do anything about it since it's a UI thing.

Having said that, there might be a class of consumers that are outside of this, that are trying to integrate against PE and we want to warn _them_. What do you think?

kbarber:
Oh yeah, probably warrants a mention in the v3/v4 upgrade doc that we're doing this.

kbarber:
Oh, didn't realise this was on stable, I'm not sure but this might just be useful for master only?

wkalt:
updated to reflect the decisions made this morning

kbarber:
+1

PR #1386, opened 2015-05-11
(PDB-1513) Clean up log-format calls

mullr:
(log/error (format ...)) can be better written as (log/errorf ...).
The same goes for other levels as well.


ajroetker:
da :bomb:

rbrw:
Looks good.

PR #1387, opened 2015-05-11
(PDB-1514) Better clj-time style in tests

mullr:


rbrw:
Looks fine to me.  Ryan suggested we get the late-project patch in first, and then I'll merge this.

PR #1389, opened 2015-05-11
(PDB-1289) Use producer_timestamp for up-to-date checks

mullr:
Replace usage of timestamp with producer_timestamp for checking if catalogs and
factsets are up to date. Reports currently lack an appropriate field; this will
be added in a later commit.

- Use producer_timestamp in stale_nodes query
- Change catalog-newer-than? and replace-facts to use producer_timestamp
- Rename timestamp symbols to received-timestamp or producer_timestamp, as
  appropriate


wkalt:
@mullr needs a rebase

wkalt:
taking a look

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1212,7 +1221,7 @@
                :order-by [[time-field :desc]]
                :limit 1}]
     (sql/with-query-results result-set (hcore/format query)
-      (time-field (first result-set)))))
+      (some-> result-set first time-field))))
```
wkalt: why this change?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1220,19 +1229,16 @@
    time :- pls/Timestamp]
   (let [time (to-timestamp time)
         time-field (if (= entity :reports) :start_time :producer_timestamp)]
-    (if-let [db-timestamp (timestamp-of-newest-record entity certname time-field)]
-      (.after db-timestamp time)
-      false)))
+    (some-> entity
+     (timestamp-of-newest-record certname time-field)
+     (.after time))))
```
wkalt: I think this changes the function semantics from returning a real bool to returning true or nil. Was that intended?

mullr: Not explicitly intended, but I don't think it hurts.

wkalt: yeah, it doesn't hurt the way we're currently using it.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1035,6 +1035,15 @@
     ["SELECT timestamp FROM factsets WHERE certname=? ORDER BY timestamp DESC" certname]
     (:timestamp (first result-set))))
 
+
+(pls/defn-validated factset-producer-timestamp :- (s/maybe pls/Timestamp)
+  "Return the factset producer-timestamp for the given certname, nil if not found"
+  [certname :- String]
+  (sql/with-query-results result-set
+    ["SELECT producer_timestamp FROM factsets WHERE certname=? ORDER BY producer_timestamp DESC" certname]
```
wkalt: this should only ever return one element -- I don't think the order by is needed

wkalt: also there's an extra line on both sides of the function

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1035,6 +1035,15 @@
     ["SELECT timestamp FROM factsets WHERE certname=? ORDER BY timestamp DESC" certname]
```
wkalt: likewise here as below, though this is obviously prexisting

wkalt:
spoke with @mullr about this in the channel. This will cause a change in behavior when importing datasets with producer_timestamps older than node-ttl (they will now get deactivated immediately). Not sure what the use case for that is outside of development, but it might be worth discussing a flag for imports.

wkalt:
I think we need to change the benchmark tool to update producer_timestamp to current time for catalogs (and for reports when that change goes in) We already do this for factsets and report start time, so this only currently affects catalog imports.

wkalt:
looks good to me aside from outstanding comments

senior:
Agree with @wkalt, +1 otherwise

src/puppetlabs/puppetdb/command.clj
```diff
@@ -217,29 +219,32 @@
...
-        timestamp   (:received annotations)]
+        {:keys [certname puppet_version] :as report}
+        (->> payload
+             (s/validate report/report-wireformat-schema)
+             upon-error-throw-fatality)]
     (jdbc/with-transacted-connection db
-      (scf-storage/maybe-activate-node! certname timestamp)
-      (scf-storage/add-report! report timestamp))
+      ;; This is using received-timestamp because the store-report command
+      ;; doesn't yet have a producer_timestamp field. It will be added soon.
```
ajroetker: This seems like it should be in a ticket rather than in a comment :dancer:

src/puppetlabs/puppetdb/scf/storage.clj
ajroetker: Also why `some->` instead of `->`?
Also also on a separate note, `query-to-vec` and `hcore/format` would be equivalent to `with-query-results` + `hcore/format` here, (another use case for the helper i mentioned in your other PR

src/puppetlabs/puppetdb/scf/storage.clj
ajroetker: You could always wrap it in `boolean` if we really needed that as well.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -264,12 +265,12 @@
                                  :api_version 1
                                  :catalog_version "foo"
                                  :certname certname
-                                 :producer_timestamp (to-timestamp (now))})
+                                 :producer_timestamp (to-timestamp (t/minus (now) (-> 1 days)))})
```
ajroetker: This should probably just be `(-> 1 days ago)` methinks

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -1219,8 +1219,14 @@
...
-      (replace-catalog! (assoc catalog :certname "node1") (-> 2 days ago))
-      (replace-catalog! (assoc catalog :certname "node2") (now))
+      (replace-catalog! (assoc catalog
+                               :certname "node1"
+                               :producer_timestamp (-> 2 days ago))
+                        (now))
+      (replace-catalog! (assoc catalog
+                               :certname "node2"
+                               :producer_timestamp (now))
+                        (now))
```
ajroetker: Maybe we should bind this `now` above? idk if that could cause any problems(?)

src/puppetlabs/puppetdb/scf/storage.clj
mullr: This is a bald imitation of factset-timestamp above. Did some spelunking to see if I could figure out why these queries were ever sorted... they came about in 71c1af56247337e16d680ef357360651af57fd16 in more or less this form. Not sure what that tells us...

src/puppetlabs/puppetdb/command.clj
mullr: It is! It's also in a followup patch that I'm nearly ready to submit.

ajroetker: Scheet!

test/puppetlabs/puppetdb/scf/storage_test.clj
mullr: The most correct thing to do here is just choose a time and compute things relative to it. Using `now` and `ago` is a race, but it's a super coarse grained one and ends up be very readable, so I'm ok with it.

ajroetker: cool cool

src/puppetlabs/puppetdb/scf/storage.clj
mullr: ... I can't remember. I can imagine it being relevant if there are no results, but 'first' and keyword-as-fn both pass through nils. I'll change it back.

src/puppetlabs/puppetdb/scf/storage.clj
mullr: I'm going to leave these here for now. Tracked as PDB-1524

ajroetker:
I'm +1 on this and ready to merge once tests are green

mullr:
^ transient

PR #1391, opened 2015-05-12
(PDB-1269) extractable-only fields

wkalt:
Without this commit, our expanded fields required both the (possibly empty)
data field and either the hash or certname as an identier to form the href.
With this commit, we return essentially the desired data/href structure in the
postgres case, and in the hsqldb case, we overload the field name with the
identifier to retain the ability to create the href.

### DIFF
--- a/src/puppetlabs/puppetdb/query/catalogs.clj
+++ b/src/puppetlabs/puppetdb/query/catalogs.clj
@@ -29,17 +29,6 @@
                         "f7" :line
                         "f8" :parameters})))
 
-(pls/defn-validated resources->expansion :- {:href s/Str (s/optional-key :data) [s/Any]}
-  "Convert the resources data to a href/data style object."
-  [obj :- (s/maybe PGobject)
-   certname :- s/Str
-   base-url :- s/Str]
-  (let [data-obj {:href (str base-url "/catalogs/" certname "/resources")}]
-    (if obj
-      (assoc data-obj :data (map rtj->resource
-                                 (json/parse-string (.getValue obj))))
-      data-obj)))
-
 (pls/defn-validated rtj->edge :- catalogs/edge-query-schema
   "Convert the row_to_json PG output to real data."
   [edge :- {s/Str s/Any}]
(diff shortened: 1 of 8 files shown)


src/puppetlabs/puppetdb/utils.clj
```diff
@@ -232,3 +237,25 @@
...
+   child :- s/Keyword
+   base-url :- s/Str]
+  (if (string? data)
+    ;; if it's a string it's just an identifier
+    (let [href (as-path base-url (name parent) data (name child))]
+      {:href href})
+    (-> (sutils/parse-db-json data)
+        first
+        (set/rename-keys {"f1" :data "f2" :href})
+        (update-in [:href] #(as-path base-url (name parent) % (name child))))))
```
ajroetker: What do you think of
```
(let [to-href #(as-path base-url (name parent) % (name child)]
 (if (string data)
.....
```

src/puppetlabs/puppetdb/query/catalogs.clj
```diff
@@ -2,6 +2,7 @@
   "Catalog retrieval"
   (:require [clojure.set :as set]
             [puppetlabs.kitchensink.core :as kitchensink]
+            [puppetlabs.puppetdb.scf.storage-utils :as sutils]
```
ajroetker: doesn't look necessary anymore

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -232,3 +237,25 @@
...
+   (as-path '/v4' 'facts' "
+  [root & path]
+  (apply str root "/" (string/join "/" path)))
+
+(pls/defn-validated child->expansion
+  "Convert child to the expanded format."
+  [data :- (s/maybe (s/either PGobject s/Str))
+   parent :- s/Keyword
+   child :- s/Keyword
+   base-url :- s/Str]
```
ajroetker: this should probably be `base-url-str` of `base-path` something to distinguish it from the map called `base-url` we use elsewhere

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -700,18 +740,23 @@
...
-
-  Return nil for fields which are query-only? since these can't be projected
-  either."
-  (when-not query-only?
-    (if expand?
-      [field name]
-      (if expandable?
-        [:null name]
-        [field name]))))
+   returns all fields except for expanded ones, which are returned as [:null k].
```
mullr: It doesn't look like the [:null k] thing is true anymore...

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -1,14 +1,19 @@
...
             [puppetlabs.puppetdb.schema :as pls]
+            [clojure.set :as set]
             [puppetlabs.puppetdb.archive :as archive]
             [clojure.java.io :as io]
+            [puppetlabs.puppetdb.scf.storage-utils :as sutils]
             [clojure.walk :as walk]
             [slingshot.slingshot :refer [try+]])
-  (:import [java.net MalformedURLException URISyntaxException URL]))
+  (:import [java.net MalformedURLException URISyntaxException URL]
+  [org.postgresql.util PGobject]))
```
ajroetker: Alignment here, but also we might want to just move this to a proper schema in `pls` since we're using it in a lot of namespaces as such.

mullr:
lgtm, but I know little enough about the underlying query code that most of my time spent reviewing was things like "gee whiz, we have to rename the fields in the json we get from postgres!" and such. So my +1 should count for maybe +0.2.

ajroetker:

senior:
+1 on this @ajroetker is going to do some functional testing and then we can get this in

ajroetker:
lgtm and works for me

PR #1393, opened 2015-05-13
(PDB-1482) allow ordering on non-extracted fields

wkalt:
Previously you couldn't order on a field that you weren't extracting.
This changes our paged sql function to do
select paged_results.* from (select blah from facts order by baz)
instead of select paged_results.* from (select blah from facts) order by baz,
and also ensures that we never order on column aliases -- always on the column
name or function application itself.

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -118,7 +118,6 @@
   return the SQL string representing this term for use in an ORDER BY clause."
   [[field order]]
   {:pre [(keyword? field)
-         (re-find #"^[\w\-]+$" (name field))
          (contains? #{:ascending :descending} order)]
    :post [(string? %)]}
   (let [field (dashes->underscores (name field))]
@@ -162,7 +161,7 @@
   (let [limit-clause     (if limit (format " LIMIT %s" limit) "")
         offset-clause    (if offset (format " OFFSET %s" offset) "")
         order-by-clause  (order-by->sql order_by)]
-    (format "SELECT paged_results.* FROM (%s) paged_results%s%s%s"
+    (format "SELECT paged_results.* FROM (%s %s%s%s) paged_results"
             sql order-by-clause limit-clause offset-clause)))
 
 (pls/defn-validated count-sql :- String
--- a/src/puppetlabs/puppetdb/query/facts.clj
+++ b/src/puppetlabs/puppetdb/query/facts.clj
@@ -70,15 +70,13 @@
   ascending. This includes facts which are known only for deactivated and
   expired nodes."
   ([]
-     (fact-names {}))
+   (fact-names {}))
   ([paging-options]
-     {:post [(map? %)
-             (coll? (:result %))
-             (every? string? (:result %))]}
-     (paging/validate-order-by! [:name] paging-options)
-     (let [facts (query/execute-query
-                  ["SELECT DISTINCT name
-                   FROM fact_paths
-                   ORDER BY name"]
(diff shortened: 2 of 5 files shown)


src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -162,7 +161,7 @@
   (let [limit-clause     (if limit (format " LIMIT %s" limit) "")
         offset-clause    (if offset (format " OFFSET %s" offset) "")
         order-by-clause  (order-by->sql order_by)]
-    (format "SELECT paged_results.* FROM (%s) paged_results%s%s%s"
+    (format "SELECT paged_results.* FROM (%s %s%s%s) paged_results"
             sql order-by-clause limit-clause offset-clause)))
```
ajroetker: So this is actually a candidate that I think would be more readable as honeysql, and we could use a cond-> to observe the limit/offset/order by decision tree /cc @mullr

mullr: Agreed... anywhere we're squishing together strings to form a query, we should think about using honeysql.

wkalt: I think if we feel that way then refactoring jdbc.clj with honeysql probably deserves its own ticket. There's a lot of places you could apply it here, not really related to this patch.

ajroetker: Oh yeah sorry I wasn't suggesting it happen here, just saw that and wanted to point it out after our disc today

kbarber: No doubt, this pagination stuff is a great place we can consider since it could collapse one entire level of nesting, but yeah agreed this is for later :-). It was on my list when I first added honeysql, but fell out of scope near the end.

src/puppetlabs/puppetdb/query/facts.clj
```diff
@@ -70,15 +70,13 @@
...
-     (paging/validate-order-by! [:name] paging-options)
-     (let [facts (query/execute-query
-                  ["SELECT DISTINCT name
-                   FROM fact_paths
-                   ORDER BY name"]
-                  paging-options)]
-       (update-in facts [:result] #(map :name %)))))
+   {:post [(map? %)
+           (coll? (:result %))
+           (every? string? (:result %))]}
```
ajroetker: could we move this to schema?

src/puppetlabs/puppetdb/query/facts.clj
```diff
@@ -70,15 +70,13 @@
...
+   (paging/validate-order-by! [:name] paging-options)
```
ajroetker: Could this somehow be moved to the schema as well?

wkalt: probably yes to both, but this is preexisting code unrelated to the patch and we use more or less the same constructions in the same functions in every query namespace. I think a coordinated change would be a better idea in this case too.

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -7,6 +7,7 @@
   (:require [puppetlabs.puppetdb.cheshire :as json]
             [clojure.string :as string]
             [puppetlabs.puppetdb.http :as http]
+            [clojure.set :as set]
```
ajroetker: don't think this is used

ajroetker:
:+1: mod the unused namespace

wkalt:
@ajroetker fixed

ajroetker:
Tested and works great for me

PR #1394, opened 2015-05-13
(PDB-1453) don't allow numeric comparisons with strings

wkalt:
Previously we were converting strings to numeric on the equality operator when
used on a field other than fact values, and exposing an error when the inequality
operators were used with a non-timestamp string.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -969,6 +969,14 @@
   (let [query-context (:query-context (meta node))]
     (cm/match [node]
 
+              [["=" field value]]
+              (let [col-type (get-in query-context [:projections field :type])]
+                (when (and (= :number col-type) (string? value))
+                  (throw
+                    (IllegalArgumentException.
+                      (format "Argument \"%s\" is incompatible with numeric field \"%s\"."
+                              value (name field))))))
+
               [[(:or ">" ">=" "<" "<=") field _]]
               (let [col-type (get-in query-context [:projections field :type])]
                 (when-not (or (vec? field)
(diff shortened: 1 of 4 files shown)


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1072,11 +1072,15 @@
                                             :value value})
 
                :number
-               (map->BinaryExpression {:operator :=
-                                       :column field
-                                       :value (if (string? value)
-                                                (ks/parse-number (str value))
-                                                value)})
+               (if (string? value)
```
ajroetker: Below you have the `throw` come in the else clause, maybe an `if-not` here would be more consistent?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1094,21 +1100,16 @@
...
-                                                     (map ks/parse-number [value value]))})
-
-                  (map->BinaryExpression {:operator (keyword op)
-                                          :column field
-                                          :value  (if (= :timestamp type)
-                                                    (to-timestamp value)
-                                                    (ks/parse-number (str value)))}))
-                (throw (IllegalArgumentException.
-                        (format "Value %s must be a number for %s comparison." value op)))))
+              (if (or (= :timestamp type) (and (= :number type) (not (string? value))))
```
ajroetker: what does it mean to have something of type :number but have a value :string?

ajroetker: i.e. is that possible? should we be parsing it?

wkalt: type refers to the column type. value refers to the user-supplied argument. This handles queries like ["<", "line", "20"] which we should be rejecting.

ajroetker: also `(if-let [value (or (and (= :timestamp type) (to-timestamp value))

ajroetker: That makes sense, sorry I was think value was the value of the column

ajroetker:
:+1: from me codewise, I'll pull it down once tests pass to give it a quick functional test

PR #1395, opened 2015-05-14
(maint) Lint with `eastwood` and improve syntax in services.clj

ajroetker:
This commit makes some linting corrections found via the `eastwood` lein
plugin. This commit also makes some syntactic improvements to
services.clj, such as checking conditions before executing business
logic and not redefing varibales in let bindings.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -155,13 +155,10 @@
   (let [{:keys [version newer link]} (try
                                        (update-info update-server db)
                                        (catch Throwable e
-                                         (log/debugf e "Could not retrieve update information (%s)" update-server)))
-        link-str                     (if link
-                                       (format " Visit %s for details." link)
-                                       "")
-        update-msg                   (format "Newer version %s is available!%s" version link-str)]
+                                         (log/debugf e "Could not retrieve update information (%s)" update-server)))]
     (when newer
-      (log/info update-msg))))
+      (log/info (cond-> (format "Newer version %s is available!" version)
+                  link (str " Visit " link " for details."))))))
 
 (defn maybe-check-for-updates
   "Check for updates if our `product-name` indicates we should, and skip the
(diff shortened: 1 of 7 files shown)


ajroetker:

src/puppetlabs/puppetdb/scf/hash.clj
```diff
@@ -48,7 +48,7 @@
 
 ;; Size of the cache is based on the number of unique resources in a
 ;; "medium" site persona
-(def resource-identity-hash* (kitchensink/bounded-memoize resource-identity-hash* 40000))
+(def resource-identity-hash* (kitchensink/bounded-memoize resource-identity-hash** 40000))
```
senior: I'd prefer the other one be renamed or closed over, having ** seems odd to me

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -769,26 +766,29 @@
...
 
 (defn honeysql-from-query
-  [{:keys [projected-fields group-by call selection projections entity]}]
   "Convert a query to honeysql format"
+  [{:keys [projected-fields group-by call selection projections entity]}]
   (let [expand? (su/postgres?)
-        call (when-let [[f & args] (when call (utils/vector-maybe call))]
-               (apply vector f (map keyword (or args [:*]))))
+        call (when-let [[f & args] (some-> call utils/vector-maybe)]
+               (apply vector f (or (some->> args (map keyword)) [:*])))
```
senior: this seems unintuitive to me

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -320,24 +309,25 @@
...
-                                  (when (pos? (to-seconds node-purge-ttl))
-                                    (partial purge-nodes! node-purge-ttl))
-                                  (when (pos? (to-seconds report-ttl))
-                                    (partial sweep-reports! report-ttl))
-                                  ;; Order is important here to ensure
-                                  ;; anything referencing an env or resource
-                                  ;; param is purged first
-                                  garbage-collect!]]
+            seconds-pos? (comp pos? to-seconds)
+            db-maintenance-tasks #(do
```
senior: For such a large anonymous fn, I prefer (fn []...) I have to scan quite a few lines here to figure out if there is an argument or not to the fn

senior:
I'm not convinced on the cond-> usages that I see in this patch, they seem less readable than what was there before, would be interested to get other opinions

wkalt:
I somewhat agree. I think when I see

```
(cond-> foo
  true blah
  ...)
```

it usually seems like a bit of a smell. That's only in here once -- I don't have a particular issue with the other cases.

ajroetker:
@senior transients above, fixed for your comments and removed most of the cond->

PR #1397, opened 2015-05-15
(PDB-1520) Use subselect with row-to-json aggregates

ajroetker:
This commit changes the json-aggregates in the query-eng to use
subselects instead of the `row` function, this means that we can "name"
columns directly from the (honey)SQL and get rid of the json
post-processing munging.

### DIFF
--- a/src/puppetlabs/puppetdb/query/catalogs.clj
+++ b/src/puppetlabs/puppetdb/query/catalogs.clj
@@ -15,45 +15,19 @@
 
 ;; MUNGE
 
-(pls/defn-validated rtj->resource :- catalogs/resource-query-schema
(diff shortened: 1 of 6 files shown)


ajroetker:

wkalt:
taking a look

wkalt:
looks good to me.

wkalt:
also tested query response times with 26k reports for this vs master: https://gist.github.com/wkalt/220b7e63d638732334ce

looks like a slight improvement, which makes sense

ajroetker:
@wkalt @mullr builds green.... just saying....

wkalt:
I'm +1 if that wasn't clear

PR #1398, opened 2015-05-15
(PDB-1487) Add producer_timestamp to store-report

mullr:
The store-report command contains start_time and end_time, but those
are generated on the agent. This adds a producer_timestamp field to the
command, which is generated on the master. It can then be reliably
compared with the producer_timestamp on the deactivate-node command.


ajroetker:
lgtm codewise

wkalt:
does this require an update in the wire format doc?

wkalt:
also in the reports api doc

mullr:
updated api docs

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1331,6 +1331,11 @@
         "ALTER TABLE fact_values RENAME COLUMN value_json TO value"
         "ALTER TABLE fact_values ALTER COLUMN value_json RENAME TO value"))))
 
+(defn add-producer-timestamp-to-reports []
+  (sql/do-commands
+   "ALTER TABLE reports ADD producer_timestamp TIMESTAMP WITH TIME ZONE"
+   "CREATE INDEX idx_reports_producer_timestamp ON reports(producer_timestamp)"))
+
```
rbrw: @mullr wondered if just adding the column was sufficient.

rbrw:
@mullr other than that, seems reasonable.

ajroetker:
@mullr needs a rebase btw

mullr:
Rebased, fixed migration to give `reports.producer_timestamp` an initial value from `end_time` and a `NOT NULL` constraint.

senior:

PR #1401, opened 2015-05-20
(PE-9827) Override ezbake version in PE profile

haus:
There are changes in ezbake 0.3.5 that are required to integrate
puppetdb into PE, which includes a bugfix related to systemd and the
defaults file. This commit copies the plugin definition from the ezbake
profile into the pe profile so that the ezbake version can be moved from
0.2.9 to 0.3.5.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -137,7 +137,9 @@
                                        :create-varlib true}
                                 :config-dir "ext/config/pe"}
                   :version ~pdb-version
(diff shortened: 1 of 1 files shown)


haus:
testing this change locally.

haus:
this works for me.

ajroetker:
:+1: from me

PR #1402, opened 2015-05-20
(PDB-1529) Allow newlines in edge titles

mullr:


ajroetker:
A test for this might be nice

PR #1404, opened 2015-05-20
(PDB-1455) Update package dependencies and postinst for Puppet 4

haus:
In Puppet 4, the filesystem paths were updated and package names were
changed. This commit updates both the dependencies and postinst of
puppetdb to accomodate this change. On debian, versioned dependencies
cannot be satisfied with a package that uses Provides tags to enable
those capabilities (this is fixed in bleeding edge deb and ubuntu), so
the debian dependencies have an `| puppet-agent` added to cover this.
The puppetdb binaries will be laid down in /opt/puppetlabs/bin in the
new filesystem world, so to ensure that both new and old paths work, the
path is updated for the postinst so that both PDB 2.x and 3.x scripts
will be correctly found.

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -3,7 +3,7 @@ ezbake: {
   foss: {
     redhat: { dependencies: ["puppet >= 3.7.3", "puppet < 5.0.0"],
               postinst: ["/usr/bin/puppetdb ssl-setup"] },
-    debian: { dependencies: ["puppet (>= 3.7.3-1puppetlabs1)", "puppet (<< 5.0.0-1puppetlabs1)"],
+    debian: { dependencies: ["puppet (>= 3.7.3-1puppetlabs1)  | puppet-agent", "puppet (<< 5.0.0-1puppetlabs1) | puppet-agent"],
               postinst: ["/usr/bin/puppetdb ssl-setup"] }
   }
 }


resources/ext/ezbake.conf
```diff
@@ -2,8 +2,8 @@ ezbake: {
   pe: {}
   foss: {
     redhat: { dependencies: ["puppet >= 3.7.3", "puppet < 5.0.0"],
```
highb: Doesn't redhat also need `puppet-agent` added?

haus: No, puppet-agent provides puppet of 4.0.0 or higher, which redhat handles just fine. Debian can't cope in the same way.

highb: :+1:

haus: Also, redhat doesn't support dependency disjunction, but debian does. So there is no way to say we depend on puppet or puppet-agent on redhat.

kbarber: @haus did you mean "No, _puppet_ provides puppet of 4.0.0 or higher, which redhat handles just fine. Debian can't cope in the same way." ... that is, requiring puppet is all that is needed for rhel for now?

resources/ext/ezbake.conf
```diff
@@ -2,8 +2,8 @@ ezbake: {
...
-              postinst: ["/usr/bin/puppetdb ssl-setup"] },
-    debian: { dependencies: ["puppet (>= 3.7.3-1puppetlabs1)", "puppet (<< 5.0.0-1puppetlabs1)"],
-              postinst: ["/usr/bin/puppetdb ssl-setup"] }
+              postinst: ["PATH=/opt/puppetlabs/bin:/usr/bin:$PATH puppetdb ssl-setup"] },
```
kbarber: When we switch to ezbake 0.3.x, will the /usr/bin even be needed? I mean, won't we have to go back and remove this old path anyway later on, since /usr/bin will be superfluous? I can't see the compatibility we want to keep anyway, maybe I'm missing something :-P.

resources/ext/ezbake.conf
haus: Yes, that exactly.

resources/ext/ezbake.conf
haus: @kbarber My reasoning here was that if the user has FOSS puppet 3.x installed, they will have the puppet binaries installed in /usr/bin. Since 3.x compatibility is desired, this helps enable that (in the super unlikely case that they don't have /usr/bin in PATH). I could be totally mistaken here of course.

kbarber: So you're concern is about the pathing to `puppet` specifically, not puppetdb binaries per se. So this stuff works I'm guessing: https://github.com/puppetlabs/puppetdb/blob/master/resources/ext/cli/ssl-setup#L242-L246

haus: Cool, then I'm fine just adding /opt/puppetlabs/bin to PATH.

kbarber: Maybe the PATH belongs in the sub-command as well anyway? I mean if someone types that sub-command once it's installed, without the proper PATH it's going to break.

haus:
@kbarber updated PATH in postinst.

PR #1416, opened 2015-05-26
(PDB-1454) Handle IOException while starting MQ

rbrw:
In the KahaDB corruption documentation, mention the possibility that MQ
related errors might be caused by downgrading PuppetDB (and by
extension, the ActiveMQ version), and if we hit an IOException while
starting the MQ broker (one of the known symptoms), suggest looking at
that documentation.

This PR pairs with master PR: #1417

### DIFF
--- a/documentation/trouble_kahadb_corruption.markdown
+++ b/documentation/trouble_kahadb_corruption.markdown
@@ -17,10 +17,18 @@ The KahaDB storage for PuppetDB is located in a sub-directory underneath your co
 Why does corruption occur?
 -----
 
-In some cases this database may corrupt. Lots of things may cause this:
+In some cases KahaDB's storage may become corrupt or simply unreadable
+by the version of PuppetDB that you've launched.  There are a number
+of possible causes, for example:
 
 * Your disk may fill up, so writes are not finalised within the journal or database index.
+
 * There might be a bug in the KahaDB code that the developers haven't catered for.
+
+* You might downgrade PuppetDB to a version that uses an incompatible
+  ActiveMQ without clearing the queue directory.  If so, you can
+  simply remove the `mq/localhost` directory inside `<vardir>`, but
+  note that any unprocessed data will be lost.
 
 How do I recover?
 -----
--- a/src/com/puppetlabs/mq.clj
+++ b/src/com/puppetlabs/mq.clj
@@ -132,10 +132,17 @@
     (start-broker! (build-embedded-broker brokername dir config))
     (catch java.io.EOFException e
       (log/warn
-        (str "Caught EOFException on broker startup, trying to restart it "
-             "again to see if that solves it. This is probably due to "
-             "KahaDB corruption."))
-      (start-broker! (build-embedded-broker brokername dir config)))))
+       "Caught EOFException on broker startup, trying again."
+       "This is probably due to KahaDB corruption"
+       "(see \"KahaDB Corruption\" in the PuppetDB manual).")
+      (start-broker! (build-embedded-broker brokername dir config)))
+    (catch java.io.IOException e
+      (throw (java.io.IOException.
+              (str "Unable to start broker in " (str (pr-str dir) ".")
+                   " This is probably due to KahaDB corruption"
+                   " or version incompatibility after a PuppetDB downgrade"
+                   " (see \"KahaDB Corruption\" in the PuppetDB manual)."))
+             e))))
 
 (defn connect-and-publish!
   "Construct an MQ producer and send the indicated message.


rbrw:
@pljenkinsro retest this please.

rbrw:
@pljenkinsro retest this please.

src/com/puppetlabs/mq.clj
```diff
@@ -132,10 +132,16 @@
...
-             "KahaDB corruption."))
-      (start-broker! (build-embedded-broker brokername dir config)))))
+       "Caught EOFException on broker startup, trying again."
+       "This is probably due to KahaDB corruption "
+       "(see \"KahaDB Corruption\" in the PuppetDB manual).")
+      (start-broker! (build-embedded-broker brokername dir config)))
+    (catch java.io.IOException e
+      (log/error
+       e "Unable to start broker in " (pr-str dir) "."
+       "This is probably due to KahaDB corruption or version downgrade "
```
wkalt: the version downgrade part is only applicable to one case here. I can't remember where we landed this morning on returning that for all IOExceptions. I wonder if getting rid of the "or version downgrade" and simply mentioning the issue in the upgrading notes would be sufficient.

rbrw: @wkalt that'd be fine with me.  (I might misremember, but I thought @kbarber suggested something that looked like the above.  Either way, happy to adjust when we know what we want.)

kbarber: I guess the wording in my mind was something like "or due to an incompatible KahaDB database after a downgrade.". Its usually corruption, but its sometimes this, so its worth saying both (or none, and sending them to the docs) otherwise people will always think it's corruption.

I also don't think mentioning the issue only in upgrade notes is enough, we should attach it to the existing KahaDB troubleshooting guide.

wkalt: I meant in addition to the edits in the kahadb doc in this PR. I don't mind including more in the error, just seems like "corruption" captures both cases.

kbarber: Hmm, its not really 'corrupted' per se, indicating a bug or error of some kind, it's just the inability for old software to read the new 5.11.1 KahaDB disk format. I guess corruption makes it sound more fatal than it is. Corruption would be something I'd want to dig into as a user to understand why, if it's just a downgrade incompatibility, hopefully most users wouldn't bother to investigate or raise a ticket on it - the message is less scary.

Honestly, KahaDB is just terrible, I hate making these work-arounds. I hope they sort this kind of thing out themselves one day. LevelDB is another option I've been keeping an eye on, last time I tried it though it was in a worse state as far as dealing with weird scenarios like this, we should look at it again soon though, it's been about a year - hopefully its more stable.

documentation/trouble_kahadb_corruption.markdown
```diff
@@ -20,8 +20,14 @@ Why does corruption occur?
 In some cases this database may corrupt. Lots of things may cause this:
```
kbarber: ... may corrupt ... "or become unreadable by our software." its not a corruption afaik, lets make it clear.

rbrw:
@wkalt @kbarber perhaps this version will address your concerns.

wkalt:

rbrw:

src/com/puppetlabs/mq.clj
```diff
@@ -132,10 +132,17 @@
...
+       "This is probably due to KahaDB corruption"
+       "(see \"KahaDB Corruption\" in the PuppetDB manual).")
+      (start-broker! (build-embedded-broker brokername dir config)))
+    (catch java.io.IOException e
+      (log/error
+       e "Unable to start broker in" (str (pr-str dir) ".")
+       "This is probably due to KahaDB corruption or version incompatibility"
+       "after a PuppetDB downgrade"
+       "(see \"KahaDB Corruption\" in the PuppetDB manual).")
+      (throw e))))
```
senior: Is there something we're expecting to catch this above? Why not log there?

src/com/puppetlabs/mq.clj
```diff
@@ -132,10 +132,17 @@
...
-             "again to see if that solves it. This is probably due to "
-             "KahaDB corruption."))
-      (start-broker! (build-embedded-broker brokername dir config)))))
+       "Caught EOFException on broker startup, trying again."
+       "This is probably due to KahaDB corruption"
+       "(see \"KahaDB Corruption\" in the PuppetDB manual).")
+      (start-broker! (build-embedded-broker brokername dir config)))
+    (catch java.io.IOException e
+      (log/error
+       e "Unable to start broker in" (str (pr-str dir) ".")
```
senior: Why not just use str here? I'm not sure why you'd use both here.

rbrw: @senior just a simplistic attempt to make paths with whitespace or any other unusual characters a little clearer in some common cases, but it's not rigorous.

senior: I understand why now, makes sense that it's in there.

src/com/puppetlabs/mq.clj
rbrw: @senior ahh, right - hopefully fixed.

rbrw:

PR #1417, opened 2015-05-26
(PDB-1454) Document need to wipe MQ during downgrades

rbrw:
This PR pairs with stable PR: https://github.com/puppetlabs/puppetdb/pull/1416

### DIFF
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -5,6 +5,7 @@ canonical: "/puppetdb/latest/release_notes.html"
 ---
 
 [configure_postgres]: ./configure.html#using-postgresql
+[kahadb_corruption]: ./trouble_kahadb_corruption.html
 [pg_trgm]: http://www.postgresql.org/docs/current/static/pgtrgm.html
 [upgrading]: ./api/query/v4/preparing-for-3.0.html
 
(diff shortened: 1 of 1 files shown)


documentation/release_notes.markdown
```diff
@@ -37,6 +38,14 @@ canonical: "/puppetdb/latest/release_notes.html"
   expected, and is due to internal changes in the way we check if
   catalogs are up-to-date.
 
+### Downgrading
+
+* If you attempt to downgrade to a previous version of PuppetDB, in
+  addition to wiping the database, you will need to clear the queue
+  directory as described [here][kahadb_corruption].  If you see an
```
wkalt: minor note, but I find it more useful to see something like "described in the [documentation on kahadb corruption][kahadb_corruption]" rather than just "here", so that I have an idea of where I'm going. Not a standard we adhere to in general.

mullr: agreed, 'here' links are an html antipattern.

rbrw: @wkalt @mullr fixed.

documentation/release_notes.markdown
```diff
@@ -37,6 +38,15 @@ canonical: "/puppetdb/latest/release_notes.html"
   expected, and is due to internal changes in the way we check if
   catalogs are up-to-date.
 
+### Downgrading
+
+* If you attempt to downgrade to a previous version of PuppetDB, in
+  addition to wiping the database, you will need to clear the queue
+  directory as described in the
+  [documentation on KahaDB corruption][kahadb_corruption].  If you see
```
wkalt: might be good to have a sentence explaining what change occurred to prompt this warning.

PR #1419, opened 2015-05-27
(PDB-1473) Present 404 errors on missing parent data

kbarber:
Before this patch, for paths that contain child data:

* nodes
* environments
* factsets
* catalogs
* reports

Child data for missing parent data (for example factets/<node>/facts, where
<node> is unknown) used to return an empty array.

This patch does a small existance check for its parent, and instead returns
a 404 for missing child data. This means consumers can request child
information directly without having to first check if the parent exists.

### DIFF
--- a/src/puppetlabs/puppetdb/http/catalogs.clj
+++ b/src/puppetlabs/puppetdb/http/catalogs.clj
@@ -10,7 +10,7 @@
             [schema.core :as s]
             [puppetlabs.puppetdb.query.paging :as paging]
             [puppetlabs.puppetdb.middleware :refer [verify-accepts-json validate-query-params
-                                                    wrap-with-paging-options]]
+                                                    wrap-with-paging-options wrap-with-parent-check]]
             [puppetlabs.puppetdb.jdbc :refer [with-transacted-connection]]
             [net.cgrand.moustache :refer [app]]))
 
@@ -53,10 +53,12 @@
                       (str (:url-prefix globals))))
 
     [node "edges" &]
-    (comp (edges/edges-app version false) (partial http-q/restrict-query-to-node node))
+    (-> (comp (edges/edges-app version false) (partial http-q/restrict-query-to-node node))
+        (wrap-with-parent-check version :catalog node))
 
     [node "resources" &]
-    (comp (resources/resources-app version false) (partial http-q/restrict-query-to-node node))))
+    (-> (comp (resources/resources-app version false) (partial http-q/restrict-query-to-node node))
+        (wrap-with-parent-check version :catalog node))))
 
 (defn catalog-app
   [version]
--- a/src/puppetlabs/puppetdb/http/nodes.clj
+++ b/src/puppetlabs/puppetdb/http/nodes.clj
@@ -8,7 +8,7 @@
             [puppetlabs.puppetdb.http :as pl-http]
             [net.cgrand.moustache :refer [app]]
             [puppetlabs.puppetdb.middleware :refer [verify-accepts-json validate-query-params
-                                                    wrap-with-paging-options]]
+                                                    wrap-with-paging-options wrap-with-parent-check]]
             [puppetlabs.puppetdb.jdbc :as jdbc]
             [puppetlabs.puppetdb.http :as http]))
 
@@ -53,10 +53,12 @@
         (validate-query-params {}))}
 
    [node "facts" &]
-   (comp (f/facts-app version) (partial http-q/restrict-query-to-node node))
+   (-> (comp (f/facts-app version) (partial http-q/restrict-query-to-node node))
+       (wrap-with-parent-check version :node node))
 
    [node "resources" &]
-   (comp (r/resources-app version) (partial http-q/restrict-query-to-node node))))
+   (-> (comp (r/resources-app version) (partial http-q/restrict-query-to-node node))
+       (wrap-with-parent-check version :node node))))
 
 (defn node-app
   [version]
--- a/src/puppetlabs/puppetdb/middleware.clj
+++ b/src/puppetlabs/puppetdb/middleware.clj
@@ -1,6 +1,8 @@
 (ns puppetlabs.puppetdb.middleware
   "Ring middleware"
   (:require [puppetlabs.kitchensink.core :as kitchensink]
+            [puppetlabs.puppetdb.jdbc :as jdbc]
+            [puppetlabs.puppetdb.query-eng :as qe]
             [puppetlabs.puppetdb.utils.metrics :refer [multitime!]]
             [puppetlabs.puppetdb.http :as http]
             [ring.util.response :as rr]
@@ -76,13 +78,6 @@
   (fn [req]
     (let [new-req (assoc req :globals globals)]
       (app new-req))))
-
-(defn wrap-read-db-tx
-  "Ring middleware that creates a database transaction on the read-db"
-  [handler]
-  (fn [{:keys [globals] :as req}]
-    (with-transacted-connection (:scf-read-db globals)
-      (handler req))))
 
 (defn wrap-with-paging-options
   "Ring middleware that will add to each request a :paging-options attribute:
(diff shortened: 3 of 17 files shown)


src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -293,3 +290,23 @@
...
       wrap-with-debug-logging))
+
+(defn wrap-with-parent-check
+  "Middleware that checks the parent exists before serving the rest of the
+   application. This ensures we always return 404's on child paths when the
+   parent data is empty."
+  [app version parent id]
+  (fn [{:keys [globals] :as req}]
+    (let [{:keys [scf-read-db url-prefix]} globals
+          status-fn (case parent
```
ajroetker: Could we make this function more generic, having it take an existence check function as an args, and simply wrap the individual handlers in the nodes/env/report endpoints? I'd like to be able to see what function is being used in the endpoints themselves

wkalt:
I'm taking a look at this / testing functionality

src/puppetlabs/puppetdb/http/nodes.clj
```diff
@@ -46,10 +47,12 @@
         (validate-query-params {}))}
```
wkalt: should we get the parent name in the node-status function above now too? These changes make our existing error inconsistent by not specifying "node"
```
curl -X GET http://localhost:8080/pdb/query/v4/nodes/foo
{
  "error" : "No information is known about foo"
}
```
vs
```
curl -X GET http://localhost:880/pdb/query/v4/nodes/foo/facts
{
  "error" : "No information is known about node foo"
}
```

src/puppetlabs/puppetdb/http/catalogs.clj
```diff
@@ -48,10 +48,12 @@
...
+    (-> (comp (edges/edges-app version false) (partial http-q/restrict-query-to-node node))
+        (wrap-with-parent-check version :node node))
 
     [node "resources" &]
-    (comp (resources/resources-app version false) (partial http-q/restrict-query-to-node node))))
+    (-> (comp (resources/resources-app version false) (partial http-q/restrict-query-to-node node))
+        (wrap-with-parent-check version :node node))))
 
 (defn catalog-app
   [version]
```
wkalt: similar comment here as with facts. Now that we 404 in all cases we should change the existing catalogs error so the phrasing matches.

```
curl -X GET http://localhost:8080/pdb/query/v4/catalogs/foo
{
  "error" : "Could not find catalog for foo"
}
```
vs
```
curl -X GET http://localhost:880/pdb/query/v4/catalogs/foo/resources
{
  "error" : "No information is known about node foo"
}
```

src/puppetlabs/puppetdb/http/nodes.clj
kbarber: At a glance, I don't hate this idea. It would involve two queries to just re-use the existing pattern, unless we change the wrap-with-parent-check to deal with using a different testing methodology, one that re-uses the query. Let me ponder.

src/puppetlabs/puppetdb/middleware.clj
kbarber: I just hate passing around functions if you don't have to, it feels like action-at-a-distance unless you absolutely need it. Let me ponder this again.

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -293,3 +290,23 @@
...
+                      :node nodes/node-exists?
+                      :environment envs/environment-exists?
+                      :report reports/report-exists?)]
+      ;; There is a race condition here, in particular we open up 1 transaction
+      ;; for the parent test, but the rest of the query results are done via the
+      ;; streaming query. This can't be solved until we work out a way to
+      ;; pass through an existing db handle through to the streamed query thread.
```
wkalt: what's the user-facing consequence of this race?

kbarber: If a user has a node for example, that gets expired/deactivated during the parent exists? query and the data query, it will return an empty list instead of a 404.

kbarber: Of course, its highly unlikely, and if a user tries again he will get a 404. We should really bring about a middleware that does this for us (Ryan had one, I'm removing it this time because its dead code right now, and minor anyway) and thread that db variable through to the backend thread of the query potential. But @senior and I talked about it, I think the upgrade to clojure.java.jdbc where threading the db pointer becomes mandatory would make this proposed change more clean.

src/puppetlabs/puppetdb/http/catalogs.clj
wkalt: actually, this is a bit more tricky than I had assumed, because the two checks are being done a bit differently. For a node that exists, but has no catalogs, this is our current behavior:

```
curl -X GET http://localhost:8080/pdb/query/v4/catalogs/desktop.localdomain
{
  "error" : "Could not find catalog for desktop.localdomain"
}
```

``
curl -X GET http://localhost:8080/pdb/query/v4/catalogs/desktop.localdomain/resources
[ ]
```

This seems inconsistent to me, but I'm not sure what the right approach is. It feels like something should be thrown in the first case, but in the second case, is the parent the node, or "node with a catalog"?

File: src/puppetlabs/puppetdb/middleware.clj (hunk 4 above)
Comment (wkalt) [inline src/puppetlabs/puppetdb/middleware.clj]: seems okay to me

**Comment by wkalt** [comment]
aside from outstanding comments I think this looks good

File: src/puppetlabs/puppetdb/http/catalogs.clj (hunk 3 above)
Comment (kbarber) [inline src/puppetlabs/puppetdb/http/catalogs.clj]: Well, we should be beholden to the HTTP side of things, if the parent folder returns a 404, so do children so yes this is inconsistent. Whether this is the correct behaviour generally thats a different story, I guess I feel it is correct, in the sense that the key for a catalog is the certname today, and that catalog doesn't exist - regardless of the node state.

File: src/puppetlabs/puppetdb/middleware.clj (hunk 1 above)
Comment (ajroetker) [inline src/puppetlabs/puppetdb/middleware.clj]: I guess what bothers me is that the existence checks should be visible when I'm looking around the code for an endpoint, and right now its kind of buried in that endpoints middleware.

**Comment by senior** [comment]
@kbarber looks like this needs a rebase, and maybe some changes after feedback?

File: src/puppetlabs/puppetdb/query_eng.clj [hunk 5]
```diff
@@ -84,3 +88,39 @@
...
+                                  FROM environments
+                                  WHERE name=?
+                                  LIMIT 1"
+                    :factset "SELECT 1
+                              FROM certnames
+                              INNER JOIN factsets
+                              ON factsets.certname = certnames.certname
+                              WHERE certnames.certname=?
+                                AND deactivated IS NULL
+                               AND expired IS NULL LIMIT 1")]
```
wkalt: I think our uniqueness constraints should make the limits unneccesary in each of these cases

wkalt:
@kbarber I left a couple comments on the test, but they're in the commit unfortunately. This looks good to me otherwise.

kbarber:
@wkalt can you comment in the PR not the commit?

test/puppetlabs/puppetdb/query/reports_test.clj
```diff
@@ -97,7 +98,12 @@
     (testing "should return reports based on hash"
       (let [actual (reports-query-result :v4 ["=" "hash" report-hash])]
         (is (= (munge-expected-reports [basic])
-               (munge-actual-reports actual)))))))
+               (munge-actual-reports actual)))))
+
+    (testing "report-exists? function"
+      (is (= true (qe/object-exists? :report report-hash)))
+      (is (= false (qe/object-exists? :report "chrissyamphlett")))
+      (is (= false (qe/object-exists? :report "terrypratchett"))))))
```
wkalt: @kbarber no, repeating the test is obviously not harmful, but it's also not beneficial or necessary.

kbarber: *sigh* ... not a great use of time, but okay.

wkalt:

PR #1422, opened 2015-05-28
(maint) don't activate the node when it's already active

wkalt:
Before this commit we were updating the deactivated column of certnames to null
even if it was already null. This caused unnecessary contention on the certnames
table with each command submission, causing more frequent retries than needed.

Also corrects the corresponding test logic: if nothing happens, then no records
are updated, and the result of maybe-activate-node should *not* be positive.

### DIFF
--- a/src/com/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/com/puppetlabs/puppetdb/scf/storage.clj
@@ -1159,9 +1159,10 @@
     (add-certname! certname))
   (let [timestamp (to-timestamp time)
         replaced  (sql/update-values :certnames
-                                     ["name=? AND (deactivated<? OR deactivated IS NULL)" certname timestamp]
-                                     {:deactivated nil})]
-    (pos? (first replaced))))
+                                     ["name=? AND deactivated<?" certname timestamp]
+                                     {:deactivated nil})
+        num-updated (first replaced)]
+    (pos? num-updated)))
 
 (pls/defn-validated deactivate-node!
   "Deactivate the given host, recording the current time. If the node is
--- a/test/com/puppetlabs/puppetdb/test/command.clj
+++ b/test/com/puppetlabs/puppetdb/test/command.clj
@@ -3,6 +3,7 @@
             [clojure.java.jdbc :as sql]
             [cheshire.core :as json]
             [com.puppetlabs.puppetdb.scf.storage :as scf-store]
+            [com.puppetlabs.puppetdb.scf.storage-utils :as sutils]
             [com.puppetlabs.puppetdb.catalogs :as catalog]
             [com.puppetlabs.puppetdb.examples.reports :as report-examples]
             [com.puppetlabs.puppetdb.scf.hash :as shash]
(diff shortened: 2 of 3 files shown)


wkalt:
this only affects things when running with command-processing-threads > 1. The easiest way to verify is to make the corresponding change in master and observe your retried commands. Postgres logs will also no longer show this kind of thing (example from master):

```
   9955 2015-05-27 13:05:56.746 PDT [db:puppetdb,sess:5566239e.bab3,pid:47795,vtid:9/3179,tid:7817153] ERROR:  could not serialize access due to concurrent update
   9956 2015-05-27 13:05:56.746 PDT [db:puppetdb,sess:5566239e.bab3,pid:47795,vtid:9/3179,tid:7817153] STATEMENT:  UPDATE certnames SET deactivated=$1, expired=$2 WHERE certname=$3 AND (deactivated<$4 OR deactivated IS NULL) AND (expired<$5 OR expired IS NULL)
```

(i.e that specific statement)

ajroetker:
:+1:

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1159,9 +1159,10 @@
     (add-certname! certname))
   (let [timestamp (to-timestamp time)
         replaced  (sql/update-values :certnames
-                                     ["name=? AND (deactivated<? OR deactivated IS NULL)" certname timestamp]
-                                     {:deactivated nil})]
-    (pos? (first replaced))))
+                                     ["name=? AND deactivated<?" certname timestamp]
```
kbarber: I think, there is a subtlety here we need to be conscious of, while we always updated, we always held some sort of predicate lock during our transaction due to the repeatable read isolation level. Removing this I bet has some locking side-effect in that regard, and while it feels more correct may shake up something hidden we haven't seen before, probably only at scale as well.

Don't get me wrong, I see the benefit of this, its such a subtle find, I like it - but I'm wary. Maybe we can step through the logic at standup or something and see if we consider this a real problem or not. I can't help but think on the surface there isn't really a problem, since deactivated is always going to be null. I definitely think this might speed things up though however ...

test/com/puppetlabs/puppetdb/test/command.clj
```diff
@@ -966,8 +967,11 @@
 
           (test-msg-handler new-facts-cmd publish discard-dir
             (reset! second-message? true)
-            (is (re-matches #".*BatchUpdateException.*(rollback|abort).*"
-                            (extract-error-message publish))))
+            (is (re-matches
+                  (if (sutils/postgres?)
+                    #"(?sm).*ERROR: could not serialize access due to concurrent update.*"
```
kbarber: What changed?

src/com/puppetlabs/puppetdb/scf/storage.clj
wkalt: I've confirmed a speedup. See these screenshots:
selection_003
selection_002

the first is before the change, and the second is after. We should be cautious of unintended consequences like you're describing though.

test/com/puppetlabs/puppetdb/test/command.clj
wkalt: yeah, it's just a different error message on postgres now. Still a rollback.

src/com/puppetlabs/puppetdb/scf/storage.clj
kbarber: I think the predicate locking is actually more around the query side of the update, rather than that update itself btw, so this might be perfectly fine. Still pondering :-).

mullr:
I'm :+1: on this in master, but we decided to skip it in stable.

wkalt:
I'll stick it on master if we still want it there. My assumption based on the conversation this morning is that this is going into stable.

ajroetker:
Yeah sounded like stable is good from this morning

senior:
+1 will merge once test disk issues have been resolved

PR #1423, opened 2015-05-28
(maint) Remove `paged_results.*` wrapper around generated sql

ajroetker:
This commit makes some syntatical changes to the query engine paging
functions as well as removing the `paged_results` wrapper and the sql we
generate for query-engine queries as it is no longer necessary.

### DIFF
--- a/src/puppetlabs/puppetdb/query/paging.clj
+++ b/src/puppetlabs/puppetdb/query/paging.clj
@@ -7,7 +7,7 @@
   (:require [puppetlabs.puppetdb.cheshire :as json]
             [clojure.string :as string]
             [puppetlabs.puppetdb.http :as http]
-            [puppetlabs.kitchensink.core :refer [keyset seq-contains? parse-int order-by-expr?]]
+            [puppetlabs.kitchensink.core :as kitchensink :refer [keyset seq-contains? parse-int order-by-expr?]]
             [puppetlabs.puppetdb.honeysql :as h]))
 
 (def query-params ["limit" "offset" "order_by" "include_total"])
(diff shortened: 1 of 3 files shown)


wkalt:
+1

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -216,17 +216,8 @@
...
 (defn dealias-order-by
   [{:keys [projections] :as query-rec} paging-options]
-  (let [alias-map (reduce-kv (fn [acc k v]
-                               (assoc acc
-                                 (keyword k)
-                                 (-> v :field h/extract-sql)))
-                             {} projections)]
-    (update-in paging-options [:order_by]
-               #(map (partial rename-first alias-map) %))))
+  (let [alias-map (->> (kitchensink/mapvals (comp h/extract-sql :field) projections)
```
mullr: 1. Is this a refactor? if so, it needs to be in its own commit. 
2. Threading with two things is kind of pointless and a bit confusing, would much rather see

    (->> projections 
      (kitchensink/mapvals (comp h/extract-sql :field))
      (kitchensink/mapkeys keyword))

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -216,17 +216,8 @@
...
+                       (kitchensink/mapkeys keyword))]
+    (update paging-options :order_by (partial map #(update % 0 (partial get alias-map))))))
```
mullr: I have no idea what this line does... is it the same as it used to be? if so, I like the old one better.

src/puppetlabs/puppetdb/query/paging.clj
ajroetker: Why 1? Also I see your point with 2 and will change but threading with two things isn't necessarily pointless

mullr: 1 because otherwise figuring out if something is intended to be a refactor or a functional change is like solving a little puzzle every time

ajroetker: But this whole PR is framed as a maint/refactor

ajroetker: I'd agree that if there we're a functional change it would need to be in its own commit, maybe even a note about it in the commit message, but a separate commit I'm not sure about

mullr: for the generated sql. This is incidental to that, yes?

ajroetker: Same codepath, the goal is to eventually have this pipeline refactored so that we don't need to keywordize strings and then turn them back to strings.

mullr: probably shouldn't have picked this time and place to start this discussion. But: especially for someone with imperfect knowledge of the codebase, isolating changes as much as possible to small, interdependent sets that leave the system in a working state is very valuable. Each of them can have its own commit message, in this case 'simplify dealias-order-by' would have been sufficient. That would let any reviewer plus a future history delver know the intent behind the change, which is like 75% of why we do source control in the first place. 

I know this isn't generally what we do, and that I'm asking for a change. I also know that github doesn't support this well. But I feel like we need to figure something out here.

ajroetker:
@mullr updated the commit message at least

ajroetker:

ajroetker:
transient ^

PR #1424, opened 2015-05-28
(PDB-1300) Set conditional beaker options for AIO

johnduarte:
This commit adds a beaker options file for puppetserver_postgres_db
in order to facilitate acceptance testing of PuppetDB with the
server provided by puppetserver and the database provided by
postgres.


acceptance/options/embedded_db.rb
```diff
@@ -3,6 +3,18 @@
 #  a variable 'options_file_path', though.
 require File.expand_path(File.join(File.dirname(options_file_path), 'common.rb'))
 
-common_options_hash.merge({
-  :puppetdb_database => 'embedded',
-})
+defaults_opts = if options[:type] == 'aio' then
```
kbarber: I'm sorry @johnduarte I've given you bad advice here. The options file stuff is parsed by beaker, not by us. In fact its done in an order that means you can prefetch the options hash this way, you have to pluck the value from the environment instead - like ENV["BEAKER_TYPE"]. This sounds scarier than it sounds, the name of the variable is still a static API so its still not going to just underneath us (unless we do a major revision bump to beaker I guess, but it would break other stuff _anyway_ since our CI jobs would need changing).

I'll fire up a job tomorrow with this change before you get online and we can see if it solves it.

johnduarte: @kbarber the eval in the options parser did not seem to like typical conditional statements. I wound up using the tap method to influence the merge on an item-by-item basis. It looks a little weird, but it works. Let me know if you have a better recommendation.

FYI: We will need to update the jenkins job(s) to use the correct options when the `puppetserver_embedded_db.rb` is deleted

PR #1425, opened 2015-05-28
(maint) Remove unneccesary querying macro to improve streaming syntax

ajroetker:
This commit removes the `with-query-results-cursor` macro, as it was
functionally unused. The macro wrapped `with-query-results-cursor*` with a
different interface, but the only thing that used the macro was
`streamed-query-results` which wrapped the macro in the original
interface of `with-query-results-cursor*`. This commit collapses these
functions/macros.
This commit also cleans up some syntax in jdbc, removing unnecessary
parens.

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -185,8 +185,8 @@
   [table]
   (-> "SELECT COUNT(*) as c FROM %s"
       (format table)
-      (query-to-vec)
-      (first)
+      query-to-vec
+      first
       :c))
 
 (def ^{:doc "A more clojurey way to refer to the JDBC transaction isolation levels"}
@@ -202,7 +202,9 @@
    tighter or looser sleep cycles."
   [current-attempt :- s/Int
    base :- (s/either s/Int Double)]
-  (let [sleep-ms (* (- (math/expt base current-attempt) 1) 1000)]
+  (let [sleep-ms (-> (math/expt base current-attempt)
+                     (- 1)
+                     (* 1000))]
     (Thread/sleep sleep-ms)))
 
 (pls/defn-validated retry-sql-or-fail :- Boolean
@@ -288,7 +290,7 @@
      (fn []
        ~@body)))
 
-(defn with-query-results-cursor*
+(defn with-query-results-cursor
   "Executes the given parameterized query within a transaction,
   producing a lazy sequence of rows. The callback `func` is executed
   on the entire sequence.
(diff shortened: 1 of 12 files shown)


src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -298,38 +300,30 @@
 
   The cursor is closed when `func` returns. If an exception is thrown,
   the query is cancelled."
-  [func sql params]
+  [[sql & params] func]
```
mullr: I see how it reduces duplication, but this is a very strange function signature.

ajroetker: It's the same as streamed-query-results was before

mullr: then it was strange before too. but we should leave it.

src/puppetlabs/puppetdb/query/reports.clj
```diff
@@ -84,13 +83,10 @@
...
   [version url-prefix query-sql]
   {:pre [(map? query-sql)]}
-  (let [{[sql & params] :results-query
-         count-query    :count-query} query-sql
-         result {:result (query/streamed-query-result
-                          version sql params
-                          ;; The doall simply forces the seq to be traversed
-                          ;; fully.
-                          (comp doall (munge-result-rows version url-prefix)))}]
+  (let [{:keys [count-query results-query]} query-sql
```
mullr: these functions are all the same, aren't they. we should clean that up at some point.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -288,7 +290,7 @@
      (fn []
        ~@body)))
 
-(defn with-query-results-cursor*
+(defn with-query-results-cursor
```
mullr: I think we should call this streamed-query-result now; it makes more sense at the call site.

src/puppetlabs/puppetdb/jdbc.clj
mullr: ah, there were two overloads before; the one that takes the vector was actually unused.

mullr:
generally looks ok, just some minor comments

src/puppetlabs/puppetdb/jdbc.clj
ajroetker: No it was used by the query-eng, this is the java.jdbc syntax,

src/puppetlabs/puppetdb/query/reports.clj
ajroetker: They are the old query-eng fn's afaik, old code that needs to stick around as long as some of our old endpoints are alive

src/puppetlabs/puppetdb/jdbc.clj
ajroetker: Maybe we could do that in a separate PR, I kind of like it the way it is honestly, streamed-query-results isn't really similar to the java.jdbc functions

src/puppetlabs/puppetdb/query/reports.clj
mullr: I mean that they all look the same structurally, and could use a refactor in the future.

src/puppetlabs/puppetdb/jdbc.clj
mullr: Totally fine with it being in a different PR. My only real objection is that 'with-blah' usually is a macro, so it's a little surprising to see it as a function which takes a callback.

src/puppetlabs/puppetdb/query/reports.clj
ajroetker: Agreed :)

src/puppetlabs/puppetdb/jdbc.clj
ajroetker: Ah wasn't sure about that, yeah how about I make a ticket to rename it and refactor the code from your other comment, that way the refactor will save some renaming hell?

mullr: Sounds good.

ajroetker:
@mullr Also I know this is maint but it relates the hack-day project I wanted to work on... transducers, so I would be really grateful if we could get this merged by tomorrow :D

mullr:
alright, :+1: from me then

wkalt:
looks good to me

PR #1426, opened 2015-05-28
(PDB-1537) dashboard redirect at root

wkalt:
This changes our default bootstrap.cfg to include a new service that redirects
/ to the dashboard at /pdb. The service is injected into the webrouting config
depending on the presence or absence of the service in the bootstrap.cfg,
so to disable the redirect users will need to remove the line from bootstrap.cfg.

### DIFF
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -8,6 +8,7 @@
            [org.joda.time Minutes Days Period])
   (:require [clojure.tools.logging :as log]
             [puppetlabs.kitchensink.core :as kitchensink]
+            [puppetlabs.trapperkeeper.bootstrap :refer [find-bootstrap-config]]
             [clj-time.core :as time]
             [clojure.java.io :as io]
             [fs.core :as fs]
@@ -316,13 +317,20 @@
   (let [default-web-router-service
         {:puppetlabs.puppetdb.cli.services/puppetdb-service "/pdb/query"
          :puppetlabs.puppetdb.dashboard/dashboard-service "/pdb"
-         :puppetlabs.puppetdb.metrics/metrics-service "/metrics"}]
+         :puppetlabs.puppetdb.metrics/metrics-service "/metrics"}
+        bootstrap-cfg (-> (find-bootstrap-config config-data)
+                          slurp
+                          str/split-lines)
+        dashboard-redirect? (contains? (set bootstrap-cfg)
+                                       "puppetlabs.puppetdb.dashboard/dashboard-redirect-service")]
     (doseq [[svc route] default-web-router-service]
       (when (get-in config-data [:web-router-service svc])
         (-> (format "Configuration of the `%s` route is not allowed. This setting defaults to `%s`." svc route)
             utils/println-err)))
     ;; We override the users settings as to make the above routes *not* configurable
-    (update config-data :web-router-service merge default-web-router-service)))
+    (update config-data :web-router-service merge default-web-router-service
+            (when dashboard-redirect?
+              {:puppetlabs.puppetdb.dashboard/dashboard-redirect-service "/"}))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
--- a/src/puppetlabs/puppetdb/dashboard.clj
+++ b/src/puppetlabs/puppetdb/dashboard.clj
@@ -28,3 +28,16 @@
                 (compojure/context (get-route this) [])
                 (add-ring-handler this))
            context)))
+
+(def dashboard-redirect
+  (app ["" &] {:get (fn [req] (rr/redirect "/pdb/dashboard/index.html"))}))
+
+(defservice dashboard-redirect-service
+  [[:PuppetDBServer shared-globals]
+   [:WebroutingService add-ring-handler get-route]]
+
+  (start [this context]
+         (log/info "Redirecting / to the PuppetDB dashboard")
+         (->> dashboard-redirect
+              (add-ring-handler this))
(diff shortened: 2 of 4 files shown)


src/puppetlabs/puppetdb/config.clj
```diff
@@ -316,13 +317,20 @@
   (let [default-web-router-service
         {:puppetlabs.puppetdb.cli.services/puppetdb-service "/pdb/query"
          :puppetlabs.puppetdb.dashboard/dashboard-service "/pdb"
-         :puppetlabs.puppetdb.metrics/metrics-service "/metrics"}]
+         :puppetlabs.puppetdb.metrics/metrics-service "/metrics"}
+        bootstrap-cfg (-> (find-bootstrap-config config-data)
+                          slurp
+                          str/split-lines)
+        dashboard-redirect? (contains? (set bootstrap-cfg)
```
ajroetker: set is unnecessary, vectors work with contains? too

src/puppetlabs/puppetdb/config.clj
```diff
@@ -316,13 +317,20 @@
...
+                                       "puppetlabs.puppetdb.dashboard/dashboard-redirect-service")]
     (doseq [[svc route] default-web-router-service]
       (when (get-in config-data [:web-router-service svc])
         (-> (format "Configuration of the `%s` route is not allowed. This setting defaults to `%s`." svc route)
             utils/println-err)))
     ;; We override the users settings as to make the above routes *not* configurable
-    (update config-data :web-router-service merge default-web-router-service)))
+    (update config-data :web-router-service merge default-web-router-service
+            (when dashboard-redirect?
+              {:puppetlabs.puppetdb.dashboard/dashboard-redirect-service "/"}))))
```
ajroetker: I thought this had to be `""` not `"/"`, does `"/"` not serve the redirect at `//`?

src/puppetlabs/puppetdb/dashboard.clj
```diff
@@ -28,3 +28,16 @@
...
+
+(def dashboard-redirect
+  (app ["" &] {:get (fn [req] (rr/redirect "/pdb"))}))
+
+(defservice dashboard-redirect-service
+  [[:PuppetDBServer shared-globals]
+   [:WebroutingService add-ring-handler get-route]]
+
+  (start [this context]
+         (log/info "Starting dashboard redirect service")
```
ajroetker: Maybe this should just log `Redirecting `/` to the PuppetDB dashboard`?

src/puppetlabs/puppetdb/dashboard.clj
```diff
@@ -28,3 +28,16 @@
                 (compojure/context (get-route this) [])
                 (add-ring-handler this))
            context)))
+
+(def dashboard-redirect
+  (app ["" &] {:get (fn [req] (rr/redirect "/pdb"))}))
```
ajroetker: This should probably be `/pdb/dashboard/index.html` since `/pdb` will yet again redirect you

src/puppetlabs/puppetdb/config.clj
ajroetker: In which case you could probably just do
```
dashboard-redirect? (-> (find-bootstrap-config config-data)
                        slurp
                         str/split-lines
                        (contains? "...."))

Comment (wkalt) [inline src/puppetlabs/puppetdb/config.clj]: vectors don't work with contains?.

```
(contains? ["a"] "a")
=> false
```

Comment (ajroetker) [inline src/puppetlabs/puppetdb/config.clj]: My mistake I think the pattern i am thinking of is
`(some #{"puppetlabs..."} boostrap-cfg)` which is in theory better because you don't always traverse the whole list necessarily

File: src/puppetlabs/puppetdb/config.clj (hunk 2 above)
Comment (wkalt) [inline src/puppetlabs/puppetdb/config.clj]: both "" and "/" serve at localhost:8080, localhost:8080/, and localhost:8080// as far as I can tell

File: src/puppetlabs/puppetdb/dashboard.clj (hunk 3 above)
Comment (wkalt) [inline src/puppetlabs/puppetdb/dashboard.clj]: sure

File: src/puppetlabs/puppetdb/dashboard.clj (hunk 4 above)
Comment (wkalt) [inline src/puppetlabs/puppetdb/dashboard.clj]: sure

File: src/puppetlabs/puppetdb/config.clj (hunk 2 above)
Comment (ajroetker) [inline src/puppetlabs/puppetdb/config.clj]: that's funny, alright nevermind then, what you have is probably best

File: src/puppetlabs/puppetdb/config.clj (hunk 1 above)
Comment (wkalt) [inline src/puppetlabs/puppetdb/config.clj]: it's not better in theory. contains? is a log(n) set lookup and some is a linear vector scan. In practice I don't think it makes a difference here at all.

Comment (wkalt) [inline src/puppetlabs/puppetdb/config.clj]: oh, you're talking about the conversion. fair enough.

File: src/puppetlabs/puppetdb/dashboard.clj [hunk 5]
```diff
@@ -28,3 +28,16 @@
                 (compojure/context (get-route this) [])
                 (add-ring-handler this))
            context)))
+
+(def dashboard-redirect
+  (app ["" &] {:get (fn [req] (rr/redirect "/pdb/dashboard/index.html"))}))
```
mullr: will be very happy to one day slot in reverse routing in places like this

mullr:
:+1:

src/puppetlabs/puppetdb/dashboard.clj
ajroetker: Do you mean like TK's reverse-proxy?

mullr: no, reverse like what comidi does. If routing is url => some structured thing (like a handler fn), then reverse routing is the structured thing => url.

ajroetker:
:+1: if that wasn't clear

PR #1429, opened 2015-06-01
(PDB-1300) Include AIO paths to puppet in ssl-setup

kbarber:
Previously we didn't include the AIO puppet in our PATH setup, we had this
fixed in stable but not in master, since the files changed so dramatically this
was lost in a rollup.


johnduarte:
PR looks good. I will merge when travis run completes.

PR #1431, opened 2015-06-01
(PDB-1508) Deprecate HSQLDB

rbrw:


project.clj
```diff
@@ -46,7 +46,7 @@
                  ;; Job scheduling
                  [overtone/at-at "1.2.0"]
                  ;; Nicer exception handling with try+/throw+
-                 [slingshot "0.10.3"]
+                 [slingshot "0.12.2"]
```
rbrw: We may or may not keep this - I don't think it ended up being important.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -327,20 +338,23 @@
...
-  (adjust-tk-config (f args)))
+  (adjust-and-validate-tk-config (f args)))
 
 (defn process-config!
-  "Accepts a map containing all of the user-provided configuration values
-  and configures the various PuppetDB subsystems."
+  "Accepts a map containing all of the user-provided configuration
+  values and configures the various PuppetDB subsystems.  May throw
+  {:type ::configuration-error :message m} for invalid
+  configurations."
```
rbrw: Vestigial - I'll revert this.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1159,6 +1159,14 @@
...
 ;;; Database support/deprecation
 
+(defn db-deprecated?
+  "Returns a string with a deprecation message if the DB is deprecated,
+   nil otherwise."
+  []
+  (when-not (sutils/postgres?)
+    (str "HSQLDB support has been deprecated and will be removed in a future version. "
+         "Please migrate to PostgreSQL.")))
+
```
rbrw: Resurrected from stable(ish).

config.sample.ini
```diff
@@ -16,20 +16,22 @@ vardir = /var/lib/puppetdb
...
-subprotocol = hsqldb
+# For the embedded DB: hsqldb
+#subprotocol = postgresql
 
-# For the embedded DB: file:/path/to/database;hsqldb.tx=mvcc;sql.syntax_pgs=true
 # For PostgreSQL: //host:port/databaseName
-# Defaults to embedded DB located in <vardir>/db
-subname = file:/var/lib/puppetdb/db/db;hsqldb.tx=mvcc;sql.syntax_pgs=true
+# For the embedded DB: file:/path/to/database;hsqldb.tx=mvcc;sql.syntax_pgs=true
+#subname = //localhost:5234/puppetdb
```
wkalt: 5432 is the PG default. Are we commenting the postgres examples so that the hsqldb defaults still work? If not I think it'd be fine to uncomment them despite them requiring some config.

documentation/configure.markdown
```diff
@@ -200,17 +203,11 @@ PuppetDB can use either **a built-in HSQLDB database** or **a PostgreSQL databas
 >
 > Depending on demand, Oracle support may be forthcoming in a future version of PuppetDB. This hasn't been decided yet.
 
-### Using Built-in HSQLDB
-
-To use an HSQLDB database at the default `<vardir>/db`, you can simply remove all database settings. To configure the DB for a different location, put the following in the `[database]` section:
+### HSQLDB Deprecation {#hsqldb-deprecation}
```
wkalt: I feel like giving this its own section is more attention than it deserves. Would it be better to just stick this blurb in a **Note** under the previous header?

config.sample.ini
rbrw: @wkalt I wasn't sure about this.  I started to uncomment them, and then held off.  Which ones do you think should be uncommented, classname, subprotocol, subname, username, and/or password?

documentation/configure.markdown
rbrw: @wkalt could be.  I initially did that so I could refer to it from two places, but I imagine I could do the same if we moved it (just refer to the section it's in).

test/puppetlabs/puppetdb/config_test.clj
```diff
@@ -53,6 +53,7 @@
       (testing "should default to half the available CPUs, even on single core boxes"
         (is (= (with-ncores 1) 1))))))
 
+;; FIXME: adjust tests for hsqldb deprecation.
 (deftest database-configuration
```
wkalt: status on this fixme?

rbrw: @wkalt in progress.

config.sample.ini
wkalt: I actually changed my mind about this. Commented out is what we currently do -- seems fine to keep it. The port in the example should still change though.

wkalt: actually, that comment is wrong ^. I was looking at the changed file in your patch, not master. I guess putting postgres defaults in there might mess with your misconfiguration detection?

rbrw: @wkalt offhand I think completely commented out means  you'll get hsqldb and the deprecation warning, is that ok for now?

wkalt: I'm on board with that

rbrw: @wkalt wait, no, you'll get the error that you need to edit the config to add values.

wkalt:
this looks fine to me aside from the logging issue

config.sample.ini
rbrw: @wkalt oh, right.  And sounds like that's probably what we really want for now.

documentation/configure.markdown
rbrw: @wkalt fixed, I think.

rbrw:
OK, once this passes tests, likely ready for another look.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -153,12 +154,23 @@
...
+  (when (not-every? database [:classname :subprotocol :subname])
+    (let [default (hsql-default-connection (:vardir global))
+          setting-needed #(when-not (% config)
+                            (format "\n  %s = %s" (name %) (% default)))
+          msg (str "database configuration is now required. "
+                   "Please add this to the [database] section of your config file:"
+                   (setting-needed :classname)
+                   (setting-needed :subprotocol)
+                   (setting-needed :subname))]
+      (throw+ {:type ::configuration-error :message msg})))
```
rbrw: @senior so it looks like this could be (throw+ {:type ::configuration-error} msg), and we could update the receivers, but only if TK core/main is patched to (throw+) instead of (throw+ m).  Or at least changing that throw+ would preserve the throw-context :message, and avoid a stack trace in the log.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -361,11 +363,17 @@
                                     mq-endpoint
                                     (command-names command) version payload)))
 
-(defn -main
+(defn- main
```
mullr: I don't think we want this change. `(defn -main ...)` gives us an external entry point, while `(defn- main ...)` is a ns-private definition.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -361,11 +363,17 @@
...
   [& args]
   (rh/add-hook #'puppetlabs.trapperkeeper.config/parse-config-data #'conf/hook-tk-parse-config-data)
-  (apply main args))
+  (try+
+   (apply tk/main args)
+   (catch [:type ::conf/configuration-error] obj
+     (log/error (:message obj))
+     (throw+ (assoc obj ::utils/exit-status 1)))))
+
+(def -main (utils/wrap-main main))
```
mullr: oh, I see, you wrapped it with this one. That's rather confusing... can the above one be called something different?

src/puppetlabs/puppetdb/config.clj
```diff
@@ -153,12 +154,23 @@
...
 
-(defn defaulted-db-connection
-  "Adds each of `hsql-default-connection` keypairs to `database` if classname, subprotocol and subname
-   are not found in the `database` config"
+(defn validate-db-settings
+  "Throws a {:type ::configuration-error :message m} exception
+  describing the required additions if the [database] configuration
+  doesn't specify classname, subprotocol and subname, all of which are
+  now required."
   [{:keys [global database] :as config :or {database {}}}]
```
mullr: so, 'database' here is a config map? we should rename it to 'database-config' or something if that's the case. As it is, `(when (not-every? database [...]))` makes zero sense.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -153,12 +154,23 @@
...
-  (cond-> config (not-any? database [:classname :subprotocol :subname])
-    (update :database merge (hsql-default-connection (:vardir global)))))
+  (when (not-every? database [:classname :subprotocol :subname])
+    (let [default (hsql-default-connection (:vardir global))
+          setting-needed #(when-not (% config)
+                            (format "\n  %s = %s" (name %) (% default)))
+          msg (str "database configuration is now required. "
+                   "Please add this to the [database] section of your config file:"
```
mullr: We should qualify this with something like 'if you want to use hsqldb, which is deprecated, and you shouldn't use'.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1159,6 +1159,14 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Database support/deprecation
 
+(defn db-deprecated?
```
mullr: This name implies it's a predicate function, which it isn't. Totally fine with the nil punning though.

mullr:
The semantics of this look pretty good. Some of the symbol naming was pretty confusing to me though, and took me awhile to puzzle out.

src/puppetlabs/puppetdb/cli/services.clj
rbrw: @mullr I'll just fold in the wrapper in the next version.

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: @mullr hmm, I don't disagree in general, but I was just mirroring the db-unsupported? function right next to it.

mullr: yeah, we should fix both. I'm fine leaving them for now though.

rbrw:
@mullr may have addressed your concerns with that push

mullr:
looks good now :+1:

wkalt:
@rlb this still seems to log twice to the console and not at all to the log

wkalt:
sorry rlb, meant @rbrw

mullr:
^ transient

src/puppetlabs/puppetdb/config.clj
```diff
@@ -154,12 +155,24 @@
...
-  [{:keys [global database] :as config :or {database {}}}]
-  (cond-> config (not-any? database [:classname :subprotocol :subname])
-    (update :database merge (hsql-default-connection (:vardir global)))))
+(defn validate-db-settings
+  "Throws a {:type ::configuration-error :message m} exception
+  describing the required additions if the [database] configuration
+  doesn't specify classname, subprotocol and subname, all of which are
+  now required."
+  [{global :global db-config :database :or {db-config {}} :as config}]
+  (when (not-every? db-config [:classname :subprotocol :subname])
```
wkalt: should this be not-any?

As written if I omit one of these settings from my postgres config, I get the hsqldb configuration error. On master if I do that it fails with "missing required key: classname".  If I omit all of them, it starts me on hsqldb. Seems like that behavior is preferable.

wkalt: to clarify, seems like the second case is the one we want to catch here. If it's just one item missing of the three we should fail as before.

wkalt:
this looks good to me pending the change to not-any?

rbrw:
@wkalt ok, I changed it to not-any, added a comment about why that's OK, and adjusted the message to presume less about the situation.

wkalt:
@rbrw this needs a rebase

rbrw:

PR #1432, opened 2015-06-02
(maint) Remove dead code and replace testing utility functions

ajroetker:
This commit removes dead code from each of the query/* namespaces whic
was only used in testing code. Previously we were duplicating the
efforts of the produce-streaming-body/streamed-query-results on each of
our endpoints, this commit changes the utility functions in our tests to
use the updated pattern. This commit also removes corresponding unnecessary tests that were testing code we weren't using in production.
This commit also changes a number of update-in calls to update that we're within eyesight while making these changes.

### DIFF
--- a/src/puppetlabs/puppetdb/http.clj
+++ b/src/puppetlabs/puppetdb/http.clj
@@ -249,3 +249,8 @@
    :post [(rr/response? %)]}
   (-> (json-response (:result query-result))
       (add-headers (dissoc query-result :result))))
+
+(defn status-not-found-response
+  "Produces a json response for when an entity (catalog/nodes/environment/...) is not found."
+  [type id]
+  (json-response {:error (format "No information is known about %s %s" type id)} status-not-found))
--- a/src/puppetlabs/puppetdb/http/catalogs.clj
+++ b/src/puppetlabs/puppetdb/http/catalogs.clj
@@ -2,8 +2,7 @@
   (:require [clojure.tools.logging :as log]
             [puppetlabs.puppetdb.catalogs :as catalogs]
             [puppetlabs.puppetdb.http :as http]
-            [puppetlabs.puppetdb.query.catalogs :as c]
-            [puppetlabs.puppetdb.query-eng :refer [produce-streaming-body]]
+            [puppetlabs.puppetdb.query-eng :as eng :refer [produce-streaming-body]]
             [puppetlabs.puppetdb.middleware :as middleware]
             [puppetlabs.puppetdb.http.query :as http-q]
             [puppetlabs.puppetdb.http.edges :as edges]
@@ -18,10 +17,16 @@
 (defn catalog-status
   "Produce a response body for a request to retrieve the catalog for `node`."
   [api-version node db url-prefix]
-  (if-let [catalog (with-transacted-connection db
-                     (c/status api-version node url-prefix))]
-    (http/json-response (s/validate catalogs/catalog-query-schema catalog))
-    (http/json-response {:error (str "Could not find catalog for " node)} http/status-not-found)))
+  (let [catalog (first
+                 (eng/stream-query-result :catalogs
+                                          api-version
+                                          ["=" "certname" node]
+                                          {}
+                                          db
+                                          url-prefix))]
+    (if catalog
+      (http/json-response (s/validate catalogs/catalog-query-schema catalog))
+      (http/status-not-found-response "catalog" node))))
 
 (defn build-catalog-app
   [version entity]
--- a/src/puppetlabs/puppetdb/scf/hash_debug.clj
+++ b/src/puppetlabs/puppetdb/scf/hash_debug.clj
@@ -5,9 +5,10 @@
             [fs.core :as fs]
             [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.catalogs :as catalogs]
+            [puppetlabs.puppetdb.jdbc :as jdbc]
+            [puppetlabs.puppetdb.utils :as utils]
             [puppetlabs.puppetdb.cheshire :as json]
-            [puppetlabs.puppetdb.query.catalogs :as qcat]
-            [puppetlabs.puppetdb.query.resources :as resources]
+            [puppetlabs.puppetdb.query-eng :as eng]
             [puppetlabs.puppetdb.query.edges :as edges]
             [puppetlabs.puppetdb.scf.hash :as shash]
             [puppetlabs.puppetdb.scf.storage-utils :as sutils]))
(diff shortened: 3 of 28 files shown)


src/puppetlabs/puppetdb/http/catalogs.clj
```diff
@@ -18,10 +17,16 @@
...
+  (let [catalog (first
+                 (eng/stream-query-result :catalogs
+                                          api-version
+                                          ["=" "certname" node]
+                                          {}
+                                          db
+                                          url-prefix))]
+    (if catalog
+      (http/json-response (s/validate catalogs/catalog-query-schema catalog))
+      (http/json-response {:error (str "Could not find catalog for " node)} http/status-not-found))))
```
kbarber: While you're digging in this code, @wkalt raised a point about making these messages consistent in https://github.com/puppetlabs/puppetdb/pull/1419. So "No information is known about [type] [id]". For example:

```
# curl 'http://localhost:8080/pdb/query/v4/factsets/node-1/facts'
{
  "error" : "No information is known about factset node-1"
}
```

Its going to be far easier on the conflicts if this change is made here in your patch.

src/puppetlabs/puppetdb/scf/hash_debug.clj
```diff
@@ -35,63 +36,39 @@
...
-  (-> (query-catalog certname :v4)
-      catalogs/catalog-query->wire-v6
-      (update-in [:edges] edge-relationships->kwds)
-      (update-in [:resources] resource-params->kwds)))
+  (let [version :v4
+        query ["=" "certname" certname]]
+    (-> (get-entity :catalogs version query)
+        first
+        (update :resources utils/assoc-when :data (get-entity :resources version query))
+        (update :edges utils/assoc-when :data (get-entity :edges version query))
```
wkalt: I haven't looked to see if this is a regression, but we're making these get-entity calls even on postgres, when the data field exists and nothing will be assoc'd. We should either change assoc-when to be short circuiting, or restructure this to get the same effect.

```
(utils/assoc-when {:data 1} :data (println "foo"))
=> foo                       
{:data 1}
```

src/puppetlabs/puppetdb/http/catalogs.clj
ajroetker: Fixed it up

ajroetker:

test/puppetlabs/puppetdb/testutils/events.clj
```diff
@@ -1,5 +1,6 @@
 (ns puppetlabs.puppetdb.testutils.events
-  (:require [puppetlabs.puppetdb.query.events :as events]
+  (:require [puppetlabs.puppetdb.query-eng :as eng]
+            [puppetlabs.puppetdb.fixtures :as fixt]
             [puppetlabs.puppetdb.jdbc :as jdbc]
             [clojure.walk :as walk]
             [puppetlabs.puppetdb.utils :refer [assoc-when]]
```
wkalt: this isn't used

wkalt: not true ^

src/puppetlabs/puppetdb/scf/hash_debug.clj
```diff
@@ -35,63 +36,45 @@
...
 
 (defn diffable-old-catalog
   "Query for the existing catalog from the DB, prep the results
    for easy diffing."
   [certname]
-  (-> (query-catalog certname :v4)
-      catalogs/catalog-query->wire-v6
-      (update-in [:edges] edge-relationships->kwds)
-      (update-in [:resources] resource-params->kwds)))
+  (let [version :v4
```
wkalt: it'd be nice if we could do this in a more forward compatible way -- prexisting issue though

ajroetker: Yeah I thought about fixing that here, the other issue i'd like to fix is passing down the db-spec so that we don't need to duplicate the stream-query-results logic, but didn't want to make things too big

wkalt:
pending tests this looks great to me

src/puppetlabs/puppetdb/http.clj
```diff
@@ -249,3 +249,8 @@
    :post [(rr/response? %)]}
   (-> (json-response (:result query-result))
       (add-headers (dissoc query-result :result))))
+
+(defn status-not-found-response
+  "Produces a json response for when an entity (catalog/nodes/environment/...) is not found."
+  [type id]
+  (json-response {:error (format "No information is known about %s %s" type id)} status-not-found))
```
mullr: Looks like the acceptance tests need to be updated to account for this message changing slightly

src/puppetlabs/puppetdb/scf/hash_debug.clj
```diff
@@ -35,63 +36,45 @@
...
+  (map #(update % :parameters (partial kitchensink/mapkeys keyword)) resources))
 
-(defn query-resources
-  [certname
-   version]
-  (let [query-sql (resources/query->sql version ["=" "certname" certname])
-        results (resources/query-resources version query-sql "")]
-    (:result results)))
+(defn get-entity
+  "An unfortunate copy of `query_eng.clj`'s `stream-query-result` without the `with-transacted-connection`
```
mullr: Is with-transacted-connection not re-entrant? It seems like it should be fine to just pass *db* down to stream-query result.

ajroetker: nope it is not (from my testing)

test/puppetlabs/puppetdb/query/environments_test.clj
```diff
@@ -18,19 +23,19 @@
     (is (= #{{:name "foo"}
              {:name "bar"}
              {:name "baz"}}
-           (set (:result (query-environments :v4 (query->sql :v4 nil))))))))
+           (set (query-environments :v4))))))
 
 (def jsonify (comp json/parse-strict-string json/generate-string))
```
mullr: what a strange thing to do. in any case, the name makes it sound like it's just generate-string. json-round-trip or something would be better.

ajroetker: yeah this was a fun function to find, made it onto my syntax todo list

ajroetker:
@mullr fixed up the acceptance test issue

src/puppetlabs/puppetdb/scf/hash_debug.clj
mullr: \*sigh\*

(that's a dynamic sigh)

ajroetker: eyo! badumbump

test/puppetlabs/puppetdb/query/event_counts_test.clj
```diff
@@ -57,12 +48,12 @@
...
-                                     "containing_class"
-                                     {}
-                                     {:order_by [[:successes order]]}))]
+              (let [actual (query-event-counts
+                            version
+                            ["=" "certname" "foo.local"]
+                            "containing_class"
+                            {}
+                            {:order_by [[:successes order]]})]
                 (is (= actual expected)))))))
```
mullr: should be `(= expected actual)`. I've already fixed some of these, I imagine there are many 

(don't need to fix this now)

ajroetker: Yeah that's definitely worth it's own ticket, its a problem throughout all our tests

mullr:
Generally fine, I think. Just the acceptance tests issue. There are a few things mixed up in this though; my eyes kind of glazed over towards the end. Would prefer to see things like like the update-in -> update changes in their own commit at least, so we can more easily verify the semantic changes.

ajroetker:
@mullr fair enough, fixed up the acceptance stuff so just waiting on that feedback

PR #1433, opened 2015-06-03
(PDB-1134) turn on strict_validation

wkalt:
now that test_url can be overridden, we should be able to use strict_validation again.

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -350,7 +350,7 @@ module PuppetDBExtensions
       puppetdb_startup_timeout => 120,
       manage_report_processor  => true,
       enable_reports           => true,
-      strict_validation        => false,
+      strict_validation        => true,
       restart_puppet           => false,
       terminus_package         => '#{terminus_package}',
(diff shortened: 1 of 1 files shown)


acceptance/helper.rb
```diff
@@ -350,7 +350,7 @@ class { 'puppetdb::master::config':
       puppetdb_startup_timeout => 120,
       manage_report_processor  => true,
       enable_reports           => true,
-      strict_validation        => false,
+      strict_validation        => true,
```
mullr: where is this actually used? grep is coming up empty, both in pdb and puppet.

wkalt: it's supplied to the connection validator in the module
https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/manifests/master/config.pp#L37-37

mullr:
:shipit:

rbrw:

kbarber:
+1, happy to merge once master pipeline is green.

PR #1435, opened 2015-06-04
(PDB-1536) Support POST queries for nodes

senior:
Webservers have limits on the length of GET request. This commit adds to
the nodes endpoint (root of the endpoint only) the ability to instead
issue the query as a POST, with a JSON document representation of the
query as the body. This new feature needs to be added to all endpoints
eventually, but the one where we see this problem the most is the root
of the nodes endpoint, so this commit starts there. This patch puts new
code in place to centralize handling of the PuppetDB query and related
parameters so that all endpoints can share the same code for handling
GETs and POSTs of queries. Currently only the nodes endpoint calls this
new centralized code. A bigger refactor will need to be made in the
future to switch other endpoints over to the new code.

### DIFF
--- a/src/puppetlabs/puppetdb/http/nodes.clj
+++ b/src/puppetlabs/puppetdb/http/nodes.clj
@@ -30,16 +30,7 @@
   [version]
   (app
    []
-   {:get (comp
-          (fn [{:keys [params globals paging-options]}]
-            (produce-streaming-body
-             :nodes
-             version
-             (params "query")
-             paging-options
-             (:scf-read-db globals)
-             (:url-prefix globals)))
-          http-q/restrict-query-to-active-nodes)}
+   (http-q/query-route :nodes version http-q/restrict-query-to-active-nodes')
 
    [node]
    {:get
--- a/src/puppetlabs/puppetdb/http/query.clj
+++ b/src/puppetlabs/puppetdb/http/query.clj
@@ -4,12 +4,43 @@
    Functions that aid in the parsing, serialization, and manipulation
    of PuppetDB queries embedded in HTTP parameters."
   (:require [puppetlabs.puppetdb.cheshire :as json]
-            [clojure.core.match :as cm]))
+            [clojure.core.match :as cm]
+            [puppetlabs.puppetdb.query-eng :refer [produce-streaming-body produce-streaming-body']]
+            [net.cgrand.moustache :refer [app]]
+            [schema.core :as s]
+            [puppetlabs.puppetdb.schema :as pls]
+            [puppetlabs.puppetdb.query.paging :refer [parse-limit' parse-offset' parse-order-by']]
+            [puppetlabs.puppetdb.utils :as utils]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Schemas
+
+(def puppetdb-query-schema
+  "This schema defines a PuppetDB query and its available
+  parameters. In a GET request this is contained in various query
+  parameters, for POST requests this should be contained in the body
+  of the request"
+  {:query (s/maybe [s/Any])
+   :count? s/Bool
+   (s/optional-key :order_by) (s/maybe [[(s/one s/Keyword "field")
+                                         (s/one (s/enum :ascending :descending) "order")]])
+   (s/optional-key :limit) (s/maybe s/Int)
+   (s/optional-key :offset) (s/maybe s/Int)})
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Query munging functions
 
 (defn- are-queries-different?
+  "DEPRECATED - only works on GET requests, will be replaced by are-queries-different?'"
+  {:deprecated "3.0.0"}
   [req1 req2]
   (not= (get-in req1 [:params "query"])
         (get-in req2 [:params "query"])))
+
+(defn- are-queries-different?'
+  [req1 req2]
+  (not= (:puppetdb-query req1)
+        (:puppetdb-query req2)))
 
 (defn query-criteria
   "Extract the 'criteria' (select) part of the given query"
@@ -50,9 +81,12 @@
             crit)))
 
 (defn restrict-query
-  "Given a criteria that will restrict a query, modify the supplied
+  "DEPRECATED - this function only works on GETs, will be replaced by restrict-query'
+
+  Given a criteria that will restrict a query, modify the supplied
   request so that its query parameter is now restricted according to
   the criteria."
+  {:deprecated "3.0.0"}
   [restriction {:keys [params] :as req}]
   {:pre  [(coll? restriction)]
    :post [(are-queries-different? req %)]}
@@ -61,16 +95,40 @@
                            (add-criteria restriction q))]
     (assoc-in req [:params "query"] (json/generate-string restricted-query))))
 
+(defn restrict-query'
+  "Given a criteria that will restrict a query, modify the supplied
+  request so that its query parameter is now restricted according to
+  `restriction`"
+  [restriction req]
+  {:pre  [(coll? restriction)]
+   :post [(are-queries-different?' req %)]}
+  (update-in req [:puppetdb-query :query] #(add-criteria restriction %)))
+
 (defn restrict-query-to-active-nodes
-  "Restrict the query parameter of the supplied request so that it only returns
+  "DEPRECATED - only works on GET requests, will be replaced by restrict-query-to-active-nodes'
+
+  Restrict the query parameter of the supplied request so that it only returns
   results for the supplied node, unless a node-active criteria is already
   explicitly specified."
+  {:deprecated "3.0.0"}
   [req]
   (if (some-> (get-in req [:params "query"])
               (json/parse-strict-string true)
               find-active-node-restriction-criteria)
     req
     (restrict-query ["=" ["node" "active"] true] req)))
+
+(defn restrict-query-to-active-nodes'
+  "Restrict the query parameter of the supplied request so that it only returns
+  results for the supplied node, unless a node-active criteria is already
+  explicitly specified."
+  [req]
+  (if (some-> req
+              :puppetdb-query
+              :query
+              find-active-node-restriction-criteria)
+    req
+    (restrict-query' ["=" ["node" "active"] true] req)))
 
 
 (defn restrict-query-to-node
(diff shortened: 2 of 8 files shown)


src/puppetlabs/puppetdb/http/nodes.clj
```diff
@@ -30,16 +30,7 @@
...
-          (fn [{:keys [params globals paging-options]}]
-            (produce-streaming-body
-             :nodes
-             version
-             (params "query")
-             paging-options
-             (:scf-read-db globals)
-             (:url-prefix globals)))
-          http-q/restrict-query-to-active-nodes)}
+   (http-q/query-route :nodes version http-q/restrict-query-to-active-nodes')
```
senior: As we flip other endpoints over to support GET/POST, we would switch to query-route providing the handler

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -153,3 +211,88 @@
    :post [(are-queries-different? req %)]}
   (restrict-query ["=" "title" title]
                   req))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Conversion/validation of query parameters
+
+(defn parse-boolean
```
senior: There's a couple of these kinds of functions here. They're to deal with POST data being JSON and already recognized as a boolean/int etc. I'm thinking once we get the endpoints moved over, we move this coercion to being only needed on the GET side and downstream it assumes it's of the right format.

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -80,8 +86,14 @@
           (cond-> (http/json-response* resp)
             count-query (http/add-headers {:count (jdbc/get-result-count count-query)})))))
     (catch com.fasterxml.jackson.core.JsonParseException e
+      (log/errorf e (str "Error executing query '%s' for entity '%s' "
+                         "with paging-options '%s'. Returning a 400 error code.")
+                  (name entity) query paging-options)
       (http/error-response e))
     (catch IllegalArgumentException e
+      (log/errorf e (str "Error executing query '%s' for entity '%s' "
```
senior: This change and the one above I added when having difficulty troubleshooting a bug in my found that a test tripped on. My issue was I had the arguments flipped on a function and so a version (:v4) was being passed in where an entity was expected (:nodes). The error was "IllegalArgumentException No matching clause: :v4", but the above code doesn't log anywhere and only returns an error response. The code above now logs the error + stack before returning an error response

test/puppetlabs/puppetdb/http/server_test.clj
```diff
@@ -17,7 +17,7 @@
   [version versions]
 
   (testing "provides a useful error message"
-    (let [endpoint (str "/" (name version) "/nodes")
```
senior: This test was checking error messages, relying on nodes only accepting GET. I flipped it to reports so that the test still works.

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -219,54 +219,74 @@
...
+   ([paged-test-params]
+    (paged-results* :get paged-test-params))
+   ([method {:keys [app-fn path query params limit total include_total offset] :as paged-test-params}]
+    {:pre [(= #{} (difference
+                   (keyset paged-test-params)
+                   #{:app-fn :path :query :params :limit :total :include_total :offset}))]}
+    (let [params  (merge params
+                         {:limit limit
+                          :offset offset})
+          request (query-request method path query
```
senior: The only real change to this function is swapping the get-response call for the query-response call, and adding a method. I made a multi-arity function here so I wouldn't have to go back and change all of our endpoint tests to use this new form until we're ready to flip it to POST

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -113,8 +115,34 @@
...
+  {:post [(every? order-by-expr? %)]}
+  (when order-by
+    (->> order-by
+         validate-order-by-data-structure
+         validate-no-invalid-order-by-fields
+         parse-required-order-by-fields)))
+
 (defn parse-order-by
-  "Given a map of paging-options, validates that the order_by field conforms to
+  "DEPRECATED - this function will be replaced by parse-order-by'
```
senior: The pattern I've followed here to minimize change is to create a parallel function that works on the query map rather than the request map. When possible I've been making the existing function delegate to the new query map one to minimize the duplication, but the old function will be short lived. Once the endpoints have been flipped over the prime functions will replace the original

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -56,15 +56,21 @@
...
+  the query results.
 
   If the query can't be parsed, a 400 is returned."
+  {:deprecated "3.0.0"}
   [entity version query paging-options db url-prefix]
   (try
     (jdbc/with-transacted-connection db
       (let [[query->sql munge-fn] (entity->sql-fns entity version paging-options url-prefix)
-            {:keys [results-query count-query]} (-> query (json/parse-strict-string true) query->sql)
+            {:keys [results-query count-query]} (query->sql (if (string? query)
```
senior: Another odd area. If we have a JSON doc as the body of a POST, the query has already been parsed. In other areas of the code, we have emitted the json, so that it can be re-parsed here.

senior:
Just found a bug related to order_by and POST, will fix and remove the WIP

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -72,6 +118,18 @@
     req
     (restrict-query ["=" ["node" "active"] true] req)))
 
+(defn restrict-query-to-active-nodes'
```
mullr: Unsure about this convention in the large. We could first change all the existing ones to -old or -dep, then introduce the new one with the regular name?

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -4,13 +4,44 @@
...
+            [schema.core :as s]
+            [puppetlabs.puppetdb.schema :as pls]
+            [puppetlabs.puppetdb.query.paging :refer [parse-limit' parse-offset' parse-order-by']]
+            [puppetlabs.puppetdb.utils :as utils]))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Schemas
+
+(def puppetdb-query-schema
+  "This schema defines a PuppetDB query and it's available
```
mullr: its

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -4,13 +4,44 @@
...
+  parameters. In a GET request this is contained in various query
+  parameters, for POST requests this should be contained in the body
+  of the request"
+  {:query (s/maybe [s/Any])
```
mullr: Very happy to see this schema; would be even happier to see it be a little more concrete down here.

src/puppetlabs/puppetdb/http/query.clj
mullr: I see you've also added :deprecated metadata to the original version; that's convincing me a bit more that this convention is good.

src/puppetlabs/puppetdb/http/query.clj
mullr: That seems reasonable. Can we call them 'coerce-to-boolean', etc? That's how I thought of them at first, and how you described them.

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -153,3 +211,88 @@
...
+  (let [total (get query-map :include_total ::not-found)]
+    (if (= total ::not-found)
+      false
+      (parse-boolean total))))
+
+(pls/defn-validated convert-query-params :- puppetdb-query-schema
+  "This will update a query map to contain the parsed and validated query parameters"
+  [{:keys [order_by limit offset] :as full-query}]
+  (-> full-query
+      (update-in [:order_by] parse-order-by')
```
mullr: we have plain-old `update` now

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -153,3 +211,88 @@
...
+        (let [include-total? (get params "include_total" ::not-found)]
+          (when (not= include-total? ::not-found)
+            [:include_total include-total?]))))
+
+(defn post-req->query
+  "Takes a POST body and parses the JSON to create a pdb query map"
+  [req]
+  (-> req
+      :body
+      clojure.java.io/reader
```
mullr: will parse-stream `.close` this for us?

src/puppetlabs/puppetdb/query/paging.clj
mullr: Yes, this is a sensible strategy. What comes to mind here is - as a medium-term technique for migrating to the prime functions, this is good. If it's solely for the sake of reviewability though, I'd much rather see an initial refactor commit to rename all the old ones and their usages and mark them as deprecated, then the focused commit.

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -113,8 +115,34 @@
...
+  outgoing is of the form:
+
+  [(s/one s/Keyword \"field-name\") (s/one (s/enum :ascending :descending) \"sort-order\")]
+
+  Not applying schema here until the original parse-order-by fn is
+  replaced by this one. There are tests that require bad input
+  parameters to be handled properly inside this function
+
+  If validation fails, this function will throw an exception with
+  an informative error message as to the cause of the failure."
```
mullr: go go gadget docstring!

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -152,43 +179,70 @@
...
-  "Parse the optional `limit` query parameter in the paging options map,
+  "DEPRECATED - this function will be replaced by parse-limit'
+
+  Parse the optional `limit` query parameter in the paging options map,
   and return an updated map with the correct integer value.
   Throws an exception if the provided limit is not a positive non-zero integer."
+  {:deprecated "3.0.0"}
   [paging-options]
-  (update-in paging-options [:limit] #(when-not (nil? %) (validate-limit %))))
+  (update-in paging-options [:limit] parse-limit'))
```
mullr: `update`

mullr: though somewhat more than 50% of me believes that leaving it the way it is is better, in this commit.

src/puppetlabs/puppetdb/query_eng.clj
mullr: Maybe we can pull it out into a `coerce-from-json` fn or similar? Would make the intent a bit clearer.

mullr: Also, the above github comment would be a nice code comment here...

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -80,8 +86,14 @@
           (cond-> (http/json-response* resp)
             count-query (http/add-headers {:count (jdbc/get-result-count count-query)})))))
     (catch com.fasterxml.jackson.core.JsonParseException e
+      (log/errorf e (str "Error executing query '%s' for entity '%s' "
+                         "with paging-options '%s'. Returning a 400 error code.")
+                  (name entity) query paging-options)
```
mullr: THANK YOU

test/puppetlabs/puppetdb/http/nodes_test.clj
```diff
@@ -69,57 +68,61 @@
 ;; TESTS
 
 (deftestseq node-queries
-  [[version endpoint] endpoints]
+  [[version endpoint] endpoints
+   method [:get :post]]
 
-  (let [{:keys [web1 web2 db puppet]} (store-example-nodes)]
+  (let [status-for-node' #(status-for-node method endpoint %)
```
mullr: Where do we stand on point-free-style? I know AJ likes it, you appear to mostly avoid it, and I'm wishy washy as usual.

test/puppetlabs/puppetdb/testutils.clj
mullr: Here, the comment on github is helpful. Adding the arity overload is a good change, but does clearly make the diff messy from github's review UI; the comment helps bme figure out where to look.

wkalt:
might be out of scope for this but do we envision making the query parameter optional and returning all results for the endpoint in the case of an empty post body?

mullr:
Qute happy with this PR; I think it's a good way to handle a necessary midpoint in a larger system change. There are a couple of smallish things mentioned above. 

w.r.t. putting in comments ahead of time: sometimes it's very helpful, namely where we know that the review tool is going to break down. You avoided superfluous refactors, which kept the need for these to a minimum. I like that. 

In other places, I'd almost rather they not be there. Part of the review process is a readability test for the code. If we feel that a chunk of code needs explanation, then it probably deserves a comment, a better docstring, or some more factoring.

src/puppetlabs/puppetdb/http/query.clj
wkalt: given that we support the query=null case, should we go ahead and make the key optional?

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -219,54 +219,74 @@
               :body body
               :params params))))
 
+(defn query-request
+  "Create a ring request map for a PuppetDB query. `http-method` indicates :get or :post.
+  Parameters or headers are optional"
```
wkalt: I don't seem to be able to curl successfully without headers. Is that what you meant here?

src/puppetlabs/puppetdb/http/query.clj
senior: That works, I'll change it

src/puppetlabs/puppetdb/http/query.clj
senior: Yeah, the problem with have a -old or something like that is we then have to rename everywhere that uses it to point to the new name of the old thing. Then visit it again when we can switch it from the new name (old code) to the old name (new code). ' or * is usually how I've seen closely related functions named, but am open to other ideas. I think change existing stuff to a -old name or something just makes work for ourselves though.

src/puppetlabs/puppetdb/http/query.clj
senior: @wkalt I think making that optional would be ok, will verify in the code.

@mullr what should be more concrete? I'm not understanding

src/puppetlabs/puppetdb/http/query.clj
senior: Good question. Looking through the code, I don't think it does. I'll change this

test/puppetlabs/puppetdb/http/nodes_test.clj
senior: I'm guessing you're asking in reference to partial instead of an anonymous function? I don't tend to use partial much, but I do use -> and ->> pretty frequently

src/puppetlabs/puppetdb/http/query.clj
mullr: the s/Any. But that's aspirational.

test/puppetlabs/puppetdb/http/nodes_test.clj
mullr: Yeah. I mention this just because I've seen it in a couple places. I'm fine with doing whatever seems right at the time.

src/puppetlabs/puppetdb/http/query.clj
senior: Right, I'd like to see a schema for our query language. That would be cool, but definitely an undertaking

senior:
@mullr just pushed up the fixes suggested

mullr:
:+1:

PR #1438, opened 2015-06-04
(PDB-1592) retire support for postgres 9.3

wkalt:
Retires support for postgres 9.3.

### DIFF
--- a/acceptance/tests/reports/event_query_with_read_db.rb
+++ b/acceptance/tests/reports/event_query_with_read_db.rb
@@ -11,7 +11,7 @@ test_name "validation of basic PuppetDB resource event queries" do
     second_db_manifest = <<-EOS
     class { 'postgresql::globals':
         manage_package_repo => true,
-        version             => '9.3',
+        version             => '9.4',
     }->
     class { '::postgresql::server':
       ip_mask_allow_all_users => '0.0.0.0/0',
@@ -32,7 +32,10 @@ test_name "validation of basic PuppetDB resource event queries" do
   step "Copy db schema to the new puppetdb2 database" do
 
     duplicate_db_command = <<DUPE
-    su postgres -c "PATH=/usr/pgsql-9.3/bin:/usr/lib/postgresql/9.3/bin:$PATH pg_dump puppetdb | sed -e 's/OWNER TO.*;/OWNER TO puppetdb2;/i' | psql puppetdb2"
+    su postgres -c "PATH=/usr/pgsql-9.4/bin:/usr/lib/postgresql/9.4/bin:$PATH \
+    pg_dump puppetdb \
+    | sed -e 's/OWNER TO.*;/OWNER TO puppetdb2;/i' \
+    | psql puppetdb2"
 DUPE
   
     on database, duplicate_db_command
--- a/src/puppetlabs/puppetdb/honeysql.clj
+++ b/src/puppetlabs/puppetdb/honeysql.clj
@@ -64,16 +64,6 @@
     (keyword (.s raw-sql))
     s))
 
-;; COMPOSITE FUNCTION HELPERS
-
-(pls/defn-validated convert-to-iso8601-utc :- SqlRaw
-  "Wraps to_char() and timezone() to return a UTC fixed ISO8601 format. This is
-  required for PostgreSQL 9.3 especially when relying on the basic row_to_json
-  response which is _not_ by default ISO 8601 when extracting a timezone (fixed
-  in 9.4)."
-  [expr :- key-or-sql]
-  (hcore/raw (str "to_char(timezone('UTC'," (hfmt/to-sql expr) "), 'YYYY-MM-DD\"T\"HH24:MI:SS.US\"Z\"')")))
-
 ;; NEW OPERATORS
 
 ; Custom formatter for PostgreSQL's ~ operator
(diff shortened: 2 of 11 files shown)


src/puppetlabs/puppetdb/honeysql.clj
```diff
@@ -64,16 +64,6 @@
...
 
-;; COMPOSITE FUNCTION HELPERS
-
-(pls/defn-validated convert-to-iso8601-utc :- SqlRaw
-  "Wraps to_char() and timezone() to return a UTC fixed ISO8601 format. This is
-  required for PostgreSQL 9.3 especially when relying on the basic row_to_json
-  response which is _not_ by default ISO 8601 when extracting a timezone (fixed
-  in 9.4)."
-  [expr :- key-or-sql]
-  (hcore/raw (str "to_char(timezone('UTC'," (hfmt/to-sql expr) "), 'YYYY-MM-DD\"T\"HH24:MI:SS.US\"Z\"')")))
```
wkalt: @kbarber this function removal could use a sanity check

acceptance/tests/reports/event_query_with_read_db.rb
```diff
@@ -32,7 +32,7 @@ class { '::postgresql::server':
   step "Copy db schema to the new puppetdb2 database" do
 
     duplicate_db_command = <<DUPE
-    su postgres -c "PATH=/usr/pgsql-9.3/bin:/usr/lib/postgresql/9.3/bin:$PATH pg_dump puppetdb | sed -e 's/OWNER TO.*;/OWNER TO puppetdb2;/i' | psql puppetdb2"
+    su postgres -c "PATH=/usr/pgsql-9.4/bin:/usr/lib/postgresql/9.4/bin:$PATH pg_dump puppetdb | sed -e 's/OWNER TO.*;/OWNER TO puppetdb2;/i' | psql puppetdb2"
```
rbrw: Not critical, but wouldn't mind wrapping this while we're here, i.e.

    su postgres -c "PATH=/usr/pgsql-9.4/bin:/usr/lib/postgresql/9.4/bin:$PATH
      pg_dump puppetdb \
      | sed -e 's/OWNER TO.*;/OWNER TO puppetdb2;/i' \
      | psql puppetdb2" 

or whatever indentation style we prefer...

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -1394,23 +1394,23 @@
...
-      "PostgreSQL" [9 3] nil
+      "PostgreSQL" [8 1] "PostgreSQL DB versions 9.3 and older are no longer supported. Please upgrade Postgres and restart PuppetDB."
+      "PostgreSQL" [8 2] "PostgreSQL DB versions 9.3 and older are no longer supported. Please upgrade Postgres and restart PuppetDB."
+      "PostgreSQL" [8 3] "PostgreSQL DB versions 9.3 and older are no longer supported. Please upgrade Postgres and restart PuppetDB."
+      "PostgreSQL" [8 4] "PostgreSQL DB versions 9.3 and older are no longer supported. Please upgrade Postgres and restart PuppetDB."
+      "PostgreSQL" [9 0] "PostgreSQL DB versions 9.3 and older are no longer supported. Please upgrade Postgres and restart PuppetDB."
+      "PostgreSQL" [9 1] "PostgreSQL DB versions 9.3 and older are no longer supported. Please upgrade Postgres and restart PuppetDB."
+      "PostgreSQL" [9 2] "PostgreSQL DB versions 9.3 and older are no longer supported. Please upgrade Postgres and restart PuppetDB."
+      "PostgreSQL" [9 3] "PostgreSQL DB versions 9.3 and older are no longer supported. Please upgrade Postgres and restart PuppetDB."
```
rbrw: Wonder if "versions older than 9.4 are no longer supported" might be a touch clearer.

rbrw:
Assuming the function removal is OK, then looks good to me.

wkalt:

wkalt:
@rbrw fixed per your comments

wkalt:

src/puppetlabs/puppetdb/honeysql.clj
mullr: This appears to be the relevant thread:

http://www.postgresql.org/message-id/12707.1408322531@sss.pgh.pa.us

mullr:
This looks fine, need feedback from @kbarber re: the time conversion.

src/puppetlabs/puppetdb/honeysql.clj
kbarber: @wkalt yep, the ISO conversion isn't needed, just be wary that I don't believe it will default to UTC.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1470,5 +1469,5 @@
        (str
         "Unable to install optimal indexing\n\n"
         "We are unable to create optimal indexes for your database.\n"
-        "For maximum index performance, we recommend using PostgreSQL 9.3 or\n"
+        "For maximum index performance, we recommend using PostgreSQL 9.4 or\n"
```
kbarber: I'm not sure the version is needed here, since the check for version is no longer valid. In fact this error message now becomes specific to HSQLDB, so we can call that out now "We are unable to create optimal indexes for HSQLDB ..." etc.

kbarber:
To note (but probably out of scope) row_to_json usage could be largely replaced with json_build_object after this patch with 9.4 support.

wkalt:

PR #1439, opened 2015-06-04
(maint) Fix object exists query

mullr:
The object exists query must return data even for nodes which are
expired or deactivated, since the subresources it filters on are
directly addressable from a search which includes deactivated 
and expired nodes.


wkalt:
@mullr might be worth adding a test somewhere for this behavior

wkalt:
otherwise I'm +1 though. extensions is blocking on this.

PR #1441, opened 2015-06-05
(PDB-1125) Detect not found issues with facts find

kbarber:
Once we switched to returning a 404 not found for child endpoints, the no-data
or not-found case wasn't working any more for puppet facts find. This patch just
handles the http 404 case properly.

### DIFF
--- a/puppet/lib/puppet/indirector/facts/puppetdb.rb
+++ b/puppet/lib/puppet/indirector/facts/puppetdb.rb
@@ -64,24 +64,21 @@ class Puppet::Node::Facts::Puppetdb < Puppet::Indirector::REST
           profile("Parse fact query response (size: #{response.body.size})",
                   [:puppetdb, :facts, :find, :parse_response, request.key]) do
             result = JSON.parse(response.body)
-            # Note: the Inventory Service API appears to expect us to return nil here
-            # if the node isn't found.  However, PuppetDB returns an empty array in
-            # this case; for now we will just look for that condition and assume that
-            # it means that the node wasn't found, so we will return nil.  In the
-            # future we may want to improve the logic such that we can distinguish
-            # between the "node not found" and the "no facts for this node" cases.
-            if result.empty?
-              return nil
-            end
+
             facts = result.inject({}) do |a,h|
               a.merge(h['name'] => h['value'])
             end
+
             Puppet::Node::Facts.new(request.key, facts)
           end
         else
           # Newline characters cause an HTTP error, so strip them
           raise "[#{response.code} #{response.message}] #{response.body.gsub(/[\r\n]/, '')}"
         end
+      rescue NotFoundError => e
+        Puppet.warning("Unable to find results for #{request.key} error: #{e.message}")
+        # This is what the inventory service expects when there is no data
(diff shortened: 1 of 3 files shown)


kbarber:
PDB-1125 was already largely done, this just cleans up the 404 not found edge case.

puppet/lib/puppet/util/puppetdb/http.rb
```diff
@@ -55,7 +55,7 @@ def self.action(path_suffix, &http_callback)
             response = nil
           elsif response.is_a? Net::HTTPNotFound
             Puppet.warning("Error connecting to #{url.host} on #{url.port} at route #{route}, error message received was '#{response.message}'. #{SERVER_URL_FAIL_MSG if server_url_config}")
-            response = nil
+            response = :notfound
```
kbarber: Here we mark the results so later on we can return the right error.

puppet/lib/puppet/util/puppetdb/http.rb
```diff
@@ -86,14 +86,21 @@ def self.action(path_suffix, &http_callback)
...
       end
 
-      if response.nil?
-
+      if response.nil? or response == :notfound
         if server_url_config
           server_url_strings = Puppet::Util::Puppetdb.config.server_urls.map {|url| url.to_s}.join(', ')
-          raise Puppet::Error, "Failed to execute '#{path_suffix}' on any of the following 'server_urls': #{server_url_strings}"
+          if response == :notfound
+            raise NotFoundError, "Failed to find '#{path_suffix}' on any of the following 'server_urls': #{server_url_strings}"
```
kbarber: We raise a not found instead of just a puppet::error here, so in the indirector it can be caught, and we can return an empty value - as per the inventory service API.

puppet/spec/unit/indirector/facts/puppetdb_spec.rb
```diff
@@ -158,12 +158,12 @@ def find_facts()
     end
 
     it "should return nil if no facts are found" do
-      body = [].to_json
+      body = {"error" => "No information known about factset some_node"}.to_json
 
-      response = Net::HTTPOK.new('1.1', 200, 'OK')
+      response = Net::HTTPNotFound.new('1.1', 404, 'NotFound')
```
kbarber: Obviously now we need to test against a 404 for 'not found cases'.

puppet/lib/puppet/indirector/facts/puppetdb.rb
```diff
@@ -64,24 +64,21 @@ def find(request)
...
               a.merge(h['name'] => h['value'])
             end
+
             Puppet::Node::Facts.new(request.key, facts)
           end
         else
           # Newline characters cause an HTTP error, so strip them
           raise "[#{response.code} #{response.message}] #{response.body.gsub(/[\r\n]/, '')}"
         end
+      rescue NotFoundError => e
```
kbarber: And now we catch this exception instead of looking for an empty result, as we get a 404 and that is surfaced via exception from the http wrapper.

wkalt:
this looks good to me

PR #1442, opened 2015-06-05
(maint) Don't evaluate unneeded assoc-when values

rbrw:


rbrw:
This could also be broken up into a separate helper function, but doing so may involve either a bit of extra computation at runtime in the helper, or a nonstandard/nonmatching signature for the helper, i.e. taking [[k v] ...] rather than kvs.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -79,17 +79,17 @@
                      (.getPath (apply io/file export-root-dir file-suffix))
                      contents))
 
-(defn assoc-when
-  "Assoc to `m` only when `k` is not already present in `m`"
-  [m & kvs]
+(defmacro assoc-when
+  "Assocs the provided values with the corresponding keys if and only
+  if the key is not already present in map."
+  [map key val & kvs]
```
mullr: why pull out the `key` and `val` params here? it looks like the `kvs` arg can handle them, without needing to add a special case.

rbrw: @mullr just matching assoc which also disallows emtpy arguments like this.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -79,17 +79,17 @@
                      (.getPath (apply io/file export-root-dir file-suffix))
                      contents))
 
-(defn assoc-when
-  "Assoc to `m` only when `k` is not already present in `m`"
-  [m & kvs]
+(defmacro assoc-when
```
mullr: I don't think assoc-when is a very good name for this functionality. I would imagine it only assoc-ing if the map is truthy, or something like that. This feels more like `lazy-assoc-defaults`, maybe?. Speaking of defaults, can we get most of the same behavior by leaning on the `:or` binding?

    (let [{:as m' :or {:a 1, :b 2}} m]
      m')

But this may not get us the shortcutting execution we want.

rbrw:
I'm happy to change the name if there's one everyone prefers, but I might make that a separate PR.

mullr:
@rbrw sounds good. :+1: here

PR #1443, opened 2015-06-05
(maint) Provide /factsets/<node> endpoint

kbarber:
We had missed this as a singleton query endpoint, this patch adds it.

### DIFF
--- a/documentation/api/query/v4/factsets.markdown
+++ b/documentation/api/query/v4/factsets.markdown
@@ -139,6 +139,36 @@ which returns
       "hash" : "d118d161990f202e911b6fda09f79d24f3a5d4f4"
     } ]
 
+## `GET /pdb/query/v4/factsets/<NODE>`
+
+This will return the most recent factset for the given node. Supplying a node
+this way will restrict any given query to only apply to that node, but in
+practice this endpoint is typically used without a query string or URL
+parameters.
+
+The result will be a single map of the factset structure described above, or
+a JSON error message if the factset is not found:
+
+### Examples
+
+    curl 'http://localhost:8080/pdb/query/v4/factsets/kb.local'
+
+    {
+      "certname" : "kb.local",
+      "environment" : "production",
+      "facts" : {...},
+      "hash" : "93253d31af6d718cf81f5bc028be2a671f23ed78",
+      "producer_timestamp" : "2015-06-04T15:27:56.893Z",
+      "timestamp" : "2015-06-04T15:27:56.979Z"
+    }
+
+    curl -X GET http://localhost:8080/pdb/query/v4/factsets/my_fake_hostname
+
+    {
+      "error" : "No information is known about factset my_fake_hostname"
+    }
+
+
 ## `GET /pdb/query/v4/factsets/<NODE>/facts`
 
 This will return all facts for a particular factset, designated by a node certname.
--- a/src/puppetlabs/puppetdb/http/factsets.clj
+++ b/src/puppetlabs/puppetdb/http/factsets.clj
@@ -1,16 +1,31 @@
 (ns puppetlabs.puppetdb.http.factsets
   (:require [net.cgrand.moustache :refer [app]]
+            [puppetlabs.puppetdb.http :as http]
             [puppetlabs.puppetdb.http.facts :as facts]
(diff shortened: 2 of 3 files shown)


documentation/api/query/v4/factsets.markdown
```diff
@@ -139,6 +139,38 @@ which returns
...
+this way will restrict any given query to only apply to that node, but in
+practice this endpoint is typically used without a query string or URL
+parameters.
+
+The result will be a single map of the factset structure described above, or
+a JSON error message if the factset is not found:
+
+### Examples
+
+    curl -X GET http://localhost:8080/pdb/query/v4/catalogs/foo.localdomain
```
wkalt: this example is about catalogs :)

src/puppetlabs/puppetdb/http/factsets.clj
```diff
@@ -25,6 +40,11 @@
    {:get (comp (build-factsets-app version :factsets)
                http-q/restrict-query-to-active-nodes)}
 
+   [node]
+   (fn [{:keys [globals]}]
+     (factset-status version node (:scf-read-db globals)
+                     (str (:url-prefix globals))))
```
wkalt: are we expecting this to be something other than a string?

wkalt:
looks good to me aside from minor comments

src/puppetlabs/puppetdb/http/factsets.clj
```diff
@@ -1,16 +1,31 @@
...
+            [puppetlabs.puppetdb.http :as http]
             [puppetlabs.puppetdb.http.facts :as facts]
             [puppetlabs.puppetdb.http.query :as http-q]
             [puppetlabs.puppetdb.middleware :refer [verify-accepts-json validate-query-params
                                                     wrap-with-paging-options wrap-with-parent-check]]
             [puppetlabs.puppetdb.query.paging :as paging]
-            [puppetlabs.puppetdb.query-eng :refer [produce-streaming-body]]))
+            [puppetlabs.puppetdb.query-eng :as eng]))
+
+(defn factset-status
```
mullr: It seems like this is largely the same as `restrict-query-to-node`, isn't it? Is there some way to lean on the query engine for this kind of thing? i.e. 'start with all factsets, restrict to this node, take the first one'. Maybe that's a larger change.

mullr:
generally lgtm, just the one question

mullr:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

mullr:
Tested locally, works fine.

src/puppetlabs/puppetdb/http/factsets.clj
kbarber: Well, stream-query-results is the query engine really, and this is not about appending a restriction to an existing query per se (query= in a singleton case is ignored) it's an absolute query instead. I can see the overlap you perceive though, but I'm not sure it's exactly the same.

src/puppetlabs/puppetdb/http/factsets.clj
kbarber: yeah its superfluous.

documentation/api/query/v4/factsets.markdown
kbarber: fixing

src/puppetlabs/puppetdb/http/factsets.clj
mullr: But an absolute query is just the universal query (select *) with some restrictions on it.

mullr: (this is, btw, totally out of scope of this change. I'm +1 on the patch)

kbarber: So more or less, I'm sticking with the same pattern for catalogs & reports ... however ... there is a problem.

Actually, queries could be considered valid for singletons, that is - some queries do make sense, like extraction queries that limit what parameters are returned. If we agree this to be true, then all these endpoints need fixing.

kbarber: We'll track this general missing feature here: https://tickets.puppetlabs.com/browse/PDB-1611

PR #1444, opened 2015-06-05
(maint) update catalog timestamp in untimed benchmark run

wkalt:
Without this change, resubmission of a catalog using the benchmark tool will
not reactivate a deactivated node.


mullr:
lgtm

PR #1445, opened 2015-06-09
(maint) Update fs to 1.4.5, matching kitchensink

rbrw:


test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -29,7 +29,7 @@
 
 (deftest whitelisting
   (testing "should log on reject"
-    (let [wl (fs/temp-file)]
+    (let [wl (fs/temp-file "whitelist-log-reject")]
       (.deleteOnExit wl)
```
rbrw: Note that this approach will create temp names like whitelist-log-reject1232135...  I could imagine being fine with that, specifying (temp-dir "foo-") instead, or arguing that a temp-file wrapper should handle it.

test/puppetlabs/puppetdb/command/dlo_test.clj
```diff
@@ -41,12 +41,12 @@
...
-    (let [dir (fs/temp-dir)
+    (let [dir (fs/temp-dir "dlo-compress-intro-old-and-new")
           threshold (-> 7 days)
           stale-timestamp (.getMillis (-> 8 days ago))]
-      (fs/touch (fs/file dir "foo") stale-timestamp)
-      (fs/touch (fs/file dir "bar") stale-timestamp)
-      (fs/touch (fs/file dir "baz"))
+      (doto (fs/file dir "foo") fs/create (#(fs/touch % stale-timestamp)))
+      (doto (fs/file dir "bar") fs/create (#(fs/touch % stale-timestamp)))
+      (fs/create (fs/file dir "baz"))
```
rbrw: Because https://github.com/Raynes/fs/blob/1.4.5/src/me/raynes/fs.clj#L408

test/puppetlabs/puppetdb/command/dlo_test.clj
```diff
@@ -102,8 +102,8 @@
       (is (empty? (fs/list-dir dlo))))
 
     (testing "with subdirectories"
-      (let [subdir (fs/temp-dir dlo)
-            other-subdir (fs/temp-dir dlo)
+      (let [subdir (doto (fs/file dlo "dir-1") fs/mkdir)
+            other-subdir (doto (fs/file dlo "dir-2") fs/mkdir)
```
rbrw: Because: https://github.com/Raynes/fs/blob/1.4.5/src/me/raynes/fs.clj#L254

test/puppetlabs/puppetdb/cli/services_test.clj
kbarber: So you can't leave it blank any longer?

test/puppetlabs/puppetdb/command/dlo_test.clj
kbarber: if you use temp-name with mkdir do we still get the same effect as temp-dir?

kbarber: Oh whats wrong with temp-dir?

test/puppetlabs/puppetdb/cli/services_test.clj
kbarber: I'm guessing this is true: https://github.com/Raynes/fs/blob/1.4.5/src/me/raynes/fs.clj#L323

test/puppetlabs/puppetdb/command/dlo_test.clj
rbrw: @kbarber temp-dir If I recall correctly, temp-dir doesn't allow you to specify an absolute prefix (now?).  So in this case, you can't have a temp-dir in a temp-dir.  It throws an exception, but I'll double-check.

kbarber: Okay.

test/puppetlabs/puppetdb/command/dlo_test.clj
kbarber: Sorry, can you explain to me why touch doesn't do the necessary now?

kbarber: Explained during standup, touch doesn't change the timestamp if its already created.

PR #1446, opened 2015-06-09
(PDB-1601) Only submit catalog and factset once in benchmark

ajroetker:
This commit changes the untimed mode for the benchmarking tool to only
submit the catalog and factset for a node once since we will only ever
use the latest.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/benchmark.clj
+++ b/src/puppetlabs/puppetdb/cli/benchmark.clj
@@ -173,7 +173,7 @@
 (defn update-factset
   "Updates the producer_timestamp to be current, and randomly updates the leaves
    of the factset based on a percentage provided in `rand-percentage`."
-  [factset rand-percentage]
+  [rand-percentage factset]
   (-> factset
       (assoc "producer_timestamp" (time/now))
       (update "values" (partial randomize-map-leaves rand-percentage))))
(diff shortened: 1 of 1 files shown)


ajroetker:
I've run the benchmarking tool a lot in the last few days, the speed-up from this will be very nice

senior:

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -194,9 +194,9 @@
   [{:keys [host lastrun catalog report factset base-url run-interval rand-percentage] :as state} clock]
   (if-not (> (- clock lastrun) run-interval)
     state
-    (let [catalog (some-> catalog update-catalog (maybe-tweak-catalog rand-percentage))
-          report (some-> report update-report-run-fields)
-          factset (some-> factset (update-factset rand-percentage))]
+    (let [catalog (some->> catalog update-catalog (maybe-tweak-catalog rand-percentage))
+          report (some-> report update-report-run-fields json/generate-string)
+          factset (some->> factset (update-factset rand-percentage) json/generate-string)]
```
wkalt: not in a position to test this ATM, but won't this recursively json encode reports and factsets when the -i flag is used?

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -209,34 +209,45 @@
...
             (log/infof "[%s] submitted factset" host)
             (catch Exception e
               (log/errorf "[%s] failed to submit factset: %s" host e)))))
       (assoc state
-        :lastrun clock
-        :catalog catalog))))
-
-(defn update-host
+             :lastrun clock
+             :catalog catalog))))
```
wkalt: answer to the above is no, since only the catalog is being assoc'd here. I'm not sure why that's the case.

ajroetker:
Not what we want after talking to @senior

PR #1447, opened 2015-06-10
(PDB-1605) Don't log 404s in the terminus

mullr:
There are many cases when a 404 from puppetdb is a perfectly
legitimate response, and a scary error message should not be logged.

- In the failover http layer, only treat 'real-looking' 404s as
  not-found responses. Others are treated as errors. 

- In the facts terminus, no longer log a warning on not-found responses.


mullr:

PR #1448, opened 2015-06-11
(PDB-1455) Fix typo for confdir in tests

kbarber:
Packaging was failing due to a typo, fixed.


kbarber:
Cancelled this test ^^ it won't shake up the packaging path anyway.

PR #1450, opened 2015-06-11
(PDB-1590) Support two previous versions of all commands (for upgrades)

senior:
Note these previous versions are not supported at the API level. The
intent here is if the user has upgraded from PuppetDB 2.0.0 to 3.0.0 and
there are still persisted messages in the queue (POSTed, but not
processed) those messages should be consumable by PuppetDB 3.0.0.

This commit takes a previous version of the command and upgrades it,
filling in anything missing, changing the names of keys as necessary to
conform it to the current (supported) command version.

### DIFF
--- a/src/puppetlabs/puppetdb/catalogs.clj
+++ b/src/puppetlabs/puppetdb/catalogs.clj
@@ -350,7 +350,7 @@
 (defmulti parse-catalog
   "Parse a wire-format `catalog` object or string of the specified `version`,
   returning a PuppetDB-suitable representation."
-  (fn [catalog version]
+  (fn [catalog version received-time]
     (match [catalog version]
            [(_ :guard string?) _]
            String
@@ -367,13 +367,32 @@
            (throw (IllegalArgumentException. (format "Catalog must be specified as a string or a map, not '%s'" (class catalog)))))))
 
 (defmethod parse-catalog String
-  [catalog version]
+  [catalog version received-time]
   {:pre   [(string? catalog)]
    :post  [(map? %)]}
-  (parse-catalog (json/parse-string catalog true) version))
+  (parse-catalog (json/parse-string catalog true) version received-time))
+
+(defmethod parse-catalog 4
+  [catalog version received-time]
+  {:pre [(map? catalog)
+         (number? version)]
+   :post [(map? %)]}
+  (-> catalog
+      (assoc :producer_timestamp received-time)
+      (parse-catalog 5 received-time)))
+
+(defmethod parse-catalog 5
+  [catalog version received-time]
+  {:pre [(map? catalog)
+         (number? version)]
+   :post [(map? %)]}
+  (-> catalog
+      (set/rename-keys {:name :certname})
+      utils/dash->underscore-keys
+      (parse-catalog 6 received-time)))
 
 (defmethod parse-catalog 6
-  [catalog version]
+  [catalog version _]
   {:pre [(map? catalog)
          (number? version)]
    :post [(map? %)]}
@@ -383,7 +402,7 @@
        validate))
 
 (defmethod parse-catalog :default
-  [catalog version]
+  [catalog version _]
   (throw (IllegalArgumentException. (format "Unknown catalog version '%s'" version))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -427,4 +446,3 @@
   (map
    catalog-query->wire-v6
    catalogs))
-
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -59,6 +59,7 @@
             [puppetlabs.puppetdb.scf.storage :as scf-storage]
             [puppetlabs.puppetdb.catalogs :as cat]
             [puppetlabs.puppetdb.reports :as report]
+            [puppetlabs.puppetdb.facts :as fact]
             [puppetlabs.puppetdb.mq :as mq]
             [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.cheshire :as json]
@@ -71,7 +72,8 @@
             [puppetlabs.trapperkeeper.services :refer [defservice]]
             [schema.core :as s]
             [puppetlabs.puppetdb.time :refer [to-timestamp]]
-            [clj-time.core :refer [now]]))
+            [clj-time.core :refer [now]]
+            [clojure.set :as set]))
 
 ;; ## Command parsing
 
@@ -175,10 +177,10 @@
 
 (defn replace-catalog*
   [{:keys [payload annotations version]} {:keys [db catalog-hash-debug-dir]}]
-  (let [catalog (upon-error-throw-fatality (cat/parse-catalog payload version))
+  (let [received-timestamp (:received annotations)
+        catalog (upon-error-throw-fatality (cat/parse-catalog payload version received-timestamp))
         certname (:certname catalog)
         id (:id annotations)
-        received-timestamp (:received annotations)
         producer-timestamp (:producer_timestamp catalog)]
     (jdbc/with-transacted-connection' db :repeatable-read
       (scf-storage/maybe-activate-node! certname producer-timestamp)
@@ -193,8 +195,16 @@
   [version command]
   (log/warnf "command '%s' version %s is deprecated, use the latest version" command version))
 
+(defmethod process-command! [(command-names :replace-catalog) 4]
+  [command options]
+  (replace-catalog* command options))
+
+(defmethod process-command! [(command-names :replace-catalog) 5]
+  [command options]
+  (replace-catalog* command options))
+
 (defmethod process-command! [(command-names :replace-catalog) 6]
-  [{:keys [version] :as command} options]
+  [command options]
   (replace-catalog* command options))
 
 ;; Fact replacement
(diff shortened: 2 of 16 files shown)


src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -33,30 +33,6 @@
    #(and (map? %)
          (valid-jdbc-query? (:results-query %)))))
 
-;; ## String operations
-
-(defn dashes->underscores
```
senior: Moved these into utils since they are now useful in commands (and tests)

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -257,3 +258,41 @@
...
+  (let [result (string/replace (name str) \_ \-)]
+    (if (keyword? str)
+      (keyword result)
+      result)))
+
+(defn dash->underscore-keys
+  "Converts all keys (including nested maps) in `m` to use dashes
+  instead of underscores as word separatators"
+  [m]
+  (sp/update [sp/ALL]
```
senior: wow, that was easy...

kbarber:

senior:
transient maven failures ^^^^

src/puppetlabs/puppetdb/catalogs.clj
```diff
@@ -367,13 +367,32 @@
...
+      (parse-catalog 5 received-time)))
+
+(defmethod parse-catalog 5
+  [catalog version received-time]
+  {:pre [(map? catalog)
+         (number? version)]
+   :post [(map? %)]}
+  (-> catalog
+      (set/rename-keys {:name :certname})
+      utils/dash->underscore-keys
```
wkalt: is it possible to have a dashed resource parameter name? seems like that would be underscored here.

mullr: yeah, we don't want to change those.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -215,8 +224,22 @@
...
+      (process-command! config)))
+
+(defmethod process-command! [(command-names :replace-facts) 2]
+  [{:keys [annotations] :as command} config]
+  (-> command
+      (assoc :version 3)
+      (assoc-in [:payload :producer-timestamp] (:received annotations))
+      (process-command! config)))
 
+;; Node deactivation
```
mullr: Nothing for old deactivate-node commands?

src/puppetlabs/puppetdb/command.clj
```diff
@@ -193,6 +194,14 @@
   [version command]
   (log/warnf "command '%s' version %s is deprecated, use the latest version" command version))
 
+(defmethod process-command! [(command-names :replace-catalog) 4]
+  [{:keys [version] :as command} options]
```
mullr: `version` appears to be unused.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -249,6 +272,16 @@
                id (command-names :store-report)
                puppet_version certname)))
 
+(defmethod process-command! [(command-names :store-report) 3]
+  [{:keys [version] :as command} {:keys [db]}]
+  (store-report* 5 db (-> command
+                          (update :payload report/wire-v3->wire-v5 (get-in command [:annotations :received])))))
```
mullr: For reports we have the conversion in the `report` namespace, but for `replace-facts` is inline. Seems like it should be one or the other.

src/puppetlabs/puppetdb/catalogs.clj
```diff
@@ -367,13 +367,32 @@
...
-  (parse-catalog (json/parse-string catalog true) version))
+  (parse-catalog (json/parse-string catalog true) version received-time))
+
+(defmethod parse-catalog 4
+  [catalog version received-time]
+  {:pre [(map? catalog)
+         (number? version)]
+   :post [(map? %)]}
+  (-> catalog
+      (assoc :producer_timestamp received-time)
```
wkalt: I'm a bit concerned that post-upgrade we're substituting (now) for the proper producer_timestamp, and that this will cause the newer catalogs (with timestamps set by the master, before the upgrade occurred) to be rejected in favor of the old ones

wkalt: might have misunderstood here -- are you guarding against the above scenario by using producer_timestamp = (now) on v6 only?

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -348,6 +350,55 @@
...
+    (let [command {:command (command-names :replace-catalog)
+                   :version 5
+                   :payload (get-in wire-catalogs [5 :empty])}
+          certname (get-in command [:payload :name])
+          cmd-producer-timestamp (get-in command [:payload :producer-timestamp])
+          current-time (now)]
+      (with-fixtures
+        (test-msg-handler command publish discard-dir
+
+          ;;names in v5 are hyphenated, this check ensures we're sending a v4 catalog
```
mullr: s/v4/v5/ ?

src/puppetlabs/puppetdb/command.clj
senior: Forgot that we had bumped that, I'll figure out what we need to support here and add it

src/puppetlabs/puppetdb/catalogs.clj
senior: I'll add a test that proves this out, then fix it

src/puppetlabs/puppetdb/catalogs.clj
senior: Not sure exactly what you're asking. I'm using received-time for producer-timestamp only for v4 of the catalog wire format. v5 and v6 should use the producer-timestamp included in the message. We already do this via our SQL migration of old catalogs when users upgrade, this is just handling old stuff that has not yet been persisted. It should have the same end result as the SQL changes.

src/puppetlabs/puppetdb/catalogs.clj
senior: Oddly enough, this isn't a problem as utils/dash->underscore-keys only modifies the top level keys in the map. So although parameters isn't a problem, there's a lot of stuff it should be converting that it is not. I'm going to have to switch to using zippers here.

src/puppetlabs/puppetdb/catalogs.clj
mullr: I think the invocation in question is https://github.com/puppetlabs/puppetdb/pull/1450/files#diff-d47aa6cc1d029cd621e847590c557d6cR28. It appears in this case that, if there's no producer_timestamp, we're using `(now)` to fill it out. We should instead use the `receiver_timestamp`. But that might be a misreading.

wkalt: this was based on a misunderstanding on my part. I don't think this is an issue, and the code with the (now) above is not related.

src/puppetlabs/puppetdb/command.clj
senior: I've moved facts munging to the facts namespace

senior:
@mullr @wkalt Pushed up changes based on feedback. I have more tests now around catalog and report munging. I also added a generative test for the top level dash/underscore swapping.

test/puppetlabs/puppetdb/catalogs_test.clj
```diff
@@ -176,3 +179,46 @@
         (is (not (contains? wire-catalog :metadata)))
         (is (= (dissoc catalog :api_version)
                wire-catalog))))))
+
+(deftest test-v6-conversion
+  (testing "v5->v6"
+    (let [v5-catalog (get-in wire-catalogs [5 :empty])]
```
senior: I wanted to add some schema tests here (like I did for reports) but as I added that, I stubbed my toe on some old v4/v5 code that didn't get cleaned up properly and the change was getting bigger. I wrote this up as [PDB-1630](https://tickets.puppetlabs.com/browse/PDB-1630) to be tackled later.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -215,7 +225,32 @@
...
+    (-> command
+        (fact/v2-wire->v4-wire received-time)
+        (process-command! config))))
+
 ;; Node deactivation
+(defmethod process-command! [(command-names :deactivate-node) 1]
+  [command config]
+  (-> command
+      (assoc :version 3)
+      (update-in [:payload] #(upon-error-throw-fatality (json/parse-string % true)))
```
ajroetker: `update`

wkalt:

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -981,7 +1117,15 @@
               :command {:command (command-names :deactivate-node)
                         :version 3
                         :payload {:certname "bar.example.com"
-                                  :producer_timestamp (now)}}}]]
+                                  :producer_timestamp (now)}}}
+             {:certname "bar.example.com"
+              :command {:command (command-names :deactivate-node)
+                        :version 2
+                        :payload {:certname "bar.example.com"}}}
```
mullr: :payload "bar.example.com"

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -981,7 +1117,15 @@
...
+             {:certname "bar.example.com"
+              :command {:command (command-names :deactivate-node)
+                        :version 1
+                        :payload (json/generate-string {:certname "bar.example.com"})}}]]
```
mullr: :payload (json/generate-string "bar.example.com")

src/puppetlabs/puppetdb/reports.clj
```diff
@@ -153,3 +166,24 @@
   [reports :- [report-query-schema]]
   (map report-query->wire-v5
        reports))
+
+(defn dash->underscore-report-keys [report]
+  (->> report
+       utils/dash->underscore-keys
+       (sp/update [:resource_events sp/ALL sp/ALL]
```
mullr: by my reading of the specter docs, this will narrow down the scope of the transformation as follows:

the report map (post initial dash->underscore-keys)
the :resource_events map
all key-value pairs in the :resource_events map
all individual keys and values in the :resource_events map
Meaning that this will run dashes_underscores on both the keys and values. Is this what you want?

test/puppetlabs/puppetdb/reports_test.clj
```diff
@@ -50,3 +53,33 @@
   (testing "no action on valid reports"
     (let [test-data (:basic reports)]
       (= (sanitize-report test-data) test-data))))
+
+(defn underscore->dash-report-keys [m]
+  (->> m
+       utils/underscore->dash-keys
+       (sp/update [:resource-events sp/ALL sp/ALL]
```
mullr: Same concern as above. Odds are good I simply don't understand the selector mechanism.

mullr:
@senior I think the old deactivate-node command looked different than you remembered. Also have questions about what exactly specter is doing in our usage. Looks good otherwise. 

Also, I *really* like piling up commits the way that you did, it makes further reviews a lot easier. Presumably you'll squash them before merging?

mullr:
looks good, :+1: from me

mullr:

wkalt:
This looks good to me, and works fine with the renamed queue -- just waiting on PDB-1632 so this will work when it goes in.

kbarber:
@senior this will need to be re-opened and targeted at stable now.

senior:
Retargeted at stable here: https://github.com/puppetlabs/puppetdb/pull/1479

PR #1451, opened 2015-06-11
(PDB-1455) Make start_puppetdb/install_puppetdb support legacy installation

kbarber:
After adapting the start_puppetdb/install_puppetdb functions to specify the
newer packages options for the module, we still need a way to flip back to
legacy support for the purposes of upgrades, at least until we ship 3.0.0.

This patch turns all the legacy support flags into 1 flag for those
functions and now also ensures that things like confdir and embedded_path
are set correctly when legacy is true.

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -228,7 +228,14 @@ module PuppetDBExtensions
     pids
   end
 
-  def start_puppetdb(host, test_url="/pdb/meta/v1/version")
+  def start_puppetdb(host, legacy=false)
+    if legacy
+      # PDB 2.x support
+      test_url = "/v3/version"
+    else
+      test_url = "/pdb/meta/v1/version"
+    end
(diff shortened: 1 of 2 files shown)


wkalt:
This looks good to me

acceptance/helper.rb
```diff
@@ -228,7 +228,14 @@ def puppetdb_pids(host)
     pids
   end
 
-  def start_puppetdb(host, test_url="/pdb/meta/v1/version")
+  def start_puppetdb(host, legacy=false)
```
ajroetker: What would you think about casing on the version instead of using `legacy`, i.e. `pdb_ver="3.x"`, or some convention like that?

kbarber: Its interesting, not sure how far we want to go down any road - this code is going soon once we release.

PR #1452, opened 2015-06-11
(docs) Lose the liquid tags, and use triple tildes for best compatibility

nfagerlund:
We recently improved our docs site toolchain to handle syntax highlighting with
fenced code blocks (via prism.js). Yay! But we're still using Kramdown, which
only supports one of the two fenced code block syntaxes specified in CommonMark.
Meh. Anyway, triple tildes good, triple backticks less-good, still best to
use "~~~ ruby" to highlight Puppet snippets for the time being.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -110,7 +110,7 @@ If you _do_ need to change rare settings that the module doesn't manage, you can
 
 Create a new class in a new module (something like `site::puppetdb::server::extra`), declare any number of `ini_setting` resources as shown below, set the class to refresh the `puppetdb::server` class, and assign it to your PuppetDB server.
 
-{% highlight ruby %}
+~~~ ruby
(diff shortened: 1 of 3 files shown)


kbarber:
Those failures are my fault ^^ looks like PR tests are failing on stable. I'll take a look, unrelated to patch obviously.

nfagerlund:
Haha cool, I was like "HOW"

kbarber:

kbarber:
@nfagerlund using your PR as a guinea pig to find out why this is going on, sorry about the noise.

kbarber:
As per discussion with @nfagerlund, here are the list of supported highlights: http://prismjs.com/#languages-list

And examples on our website of what the ruby marking looks like (yes its puppet dsl, but its close enough): http://docs.puppetlabs.com/puppet/latest/reference/lang_classes.html

kbarber:

PR #1453, opened 2015-06-12
(PDB-1587) Optimize aggregate event counts

wkalt:
Author: Wyatt Alt <wyaltster@gmail.com>
Date:   Tue Jun 9 08:33:49 2015 -0700

    (PDB-1587) Optimize aggregate event counts
    
    This commit improves performance on aggregate-event-counts queries.
    Included are the following changes:
    * new (large) index on resource_events(resource_type, resource_title, timestamp)
    * aggregate-event-counts now supports multiple comma-separated summarize_by params
    * union all is used to combine results from multiple AEC summaries
    * AEC output is now a map array, with a new key containing the summary param
    * CTE used to improve performance when multiple summarize_by values are provided
    * fairly expensive migration merged with mega-migration #29
    * AEC no longer compatible with HSQLDB

### DIFF
--- a/documentation/api/query/v4/aggregate-event-counts.markdown
+++ b/documentation/api/query/v4/aggregate-event-counts.markdown
@@ -12,6 +12,9 @@ canonical: "/puppetdb/latest/api/query/v4/aggregate-event-counts.html"
 > **Experimental Endpoint**: The aggregate-event-counts endpoint is designated
 > as experimental. It may be altered or removed in a future release.
 
+> **Not supported on HSQLDB**: The aggregate-event-counts endpoint is not
+> supported on HSQLDB.
+
 Puppet agent nodes submit reports after their runs, and the puppet master forwards these to PuppetDB. Each report includes:
 
 * Some data about the entire run
@@ -37,8 +40,9 @@ This endpoint builds on top of the [`event-counts`][event-counts] endpoint, and 
 * `query`: Required. A JSON array of query predicates in prefix form (`["<OPERATOR>", "<FIELD>", "<VALUE>"]`).
 This query is forwarded to the [`events`][events] endpoint - see there for additional documentation. For general info about queries, see [the page on query structure.][query]
 
-* `summarize_by`: Required. A string specifying which type of object you'd like count. Supported values are
-`resource`, `containing_class`, and `certname`.
+* `summarize_by`: Required. A string specifying which object types you'd like counted. Supported values are
+`resource`, `containing_class`, and `certname`, or any comma-separated
+combination thereof.
 
 * `count_by`: Optional. A string specifying what type of object is counted when building up the counts of
 `successes`, `failures`, `noops`, and `skips`. Supported values are `resource` (default) and `certname`.
@@ -61,16 +65,18 @@ This endpoint builds on top of the [`event-counts`][event-counts] and [`events`]
 
 ### Response Format
 
-The response is a single JSON map containing aggregated event-count information and a `total` for how many
+The response is an array of JSON maps containing the summarize_by parameter,
+aggregated event-count information and a `total` for how many
 event-count results were aggregated.
 
-    {
-      "successes": 2,
-      "failures": 0,
-      "noops": 0,
-      "skips": 1,
-      "total": 3
-    }
+    [ {
+        "summarize_by": "containing_class",
+        "successes": 2,
+        "failures": 0,
+        "noops": 0,
+        "skips": 1,
+        "total": 3
+    } ]
 
 ### Examples
 
@@ -80,7 +86,35 @@ You can use [`curl`][curl] to query information about aggregated resource event 
       --data-urlencode 'query=["=", "certname", "foo.local"]' \
       --data-urlencode 'summarize_by=containing_class'
 
+    [ {
+        "summarize_by" : "containing_class",
+        "successes" : 2,
+        "failures" : 0,
+        "noops" : 0,
+        "skips" : 0,
+        "total" : 2
+    } ]
+
+    curl -G 'http://localhost:8080/pdb/query/v4/aggregate-event-counts'
+    -d 'query=["=","certname","foo.local"]' \
+    -d 'summarize_by=containing_class,certname'
+
+    [ {
+        "summarize_by" : "containing_class",
+        "successes" : 2,
+        "failures" : 0,
+        "noops" : 0,
+        "skips" : 0,
+        "total" : 2
+    }, {
+        "summarize_by" : "certname",
+         "successes" : 1,
+         "failures" : 0,
+         "noops" : 0,
+         "skips" : 0,
+         "total" : 1
+    } ]
+
 ## No Paging
 
-This endpoint always returns a single result, so paging is not necessary.
-
+This endpoint does not support paging options, and results are unordered.
--- a/src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
+++ b/src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
@@ -1,6 +1,8 @@
 (ns puppetlabs.puppetdb.http.aggregate-event-counts
   (:require [puppetlabs.puppetdb.cheshire :as json]
-            [puppetlabs.puppetdb.http.events :as events-http]
+            [puppetlabs.puppetdb.http.events :refer [validate-distinct-options!]]
+            [puppetlabs.puppetdb.http :as http]
+            [puppetlabs.puppetdb.utils :as utils]
             [puppetlabs.puppetdb.query-eng :refer [produce-streaming-body]]
             [puppetlabs.puppetdb.middleware :refer [verify-accepts-json validate-query-params]]
             [clojure.tools.logging :as log]
@@ -9,25 +11,32 @@
 (defn routes
   [version]
   (app
-   [""]
-   {:get (fn [{:keys [params globals]}]
-           (let [{:strs [query summarize_by counts_filter count_by] :as query-params} params
-                 distinct-options (events-http/validate-distinct-options! query-params)
-                 query-options (-> {:counts_filter (when counts_filter (json/parse-string counts_filter true))
-                                    :count_by count_by}
-                                   (merge distinct-options))]
-             (produce-streaming-body
-              :aggregate-event-counts
-              version
-              query
-              [summarize_by query-options]
-              (:scf-read-db globals)
-              (:url-prefix globals))))}))
+    [""]
+    {:get (fn [{:keys [params globals]}]
+            (if (utils/hsql? (:scf-read-db globals))
+              (http/json-response
+                {:error "The aggregate-event-counts endpoint does not support HSQLDB."}
+                http/status-not-implemented)
+              (let [{:strs [query summarize_by counts_filter count_by]} params
+                    distinct-options (validate-distinct-options! params)
+                    counts_filter' (json/parse-string counts_filter true)
+                    query-options (-> {:counts_filter counts_filter'
+                                       :count_by count_by}
+                                      (merge distinct-options))]
+                (log/warn
+                  (str "The aggregate-event-counts endpoint is experimental"
+                       " and may be altered or removed in the future."))
+                (produce-streaming-body
+                  :aggregate-event-counts
+                  version
+                  query
+                  [summarize_by query-options]
+                  (:scf-read-db globals)
+                  (:url-prefix globals)))))}))
 
 (defn aggregate-event-counts-app
   "Ring app for querying for aggregated summary information about resource events."
   [version]
-  (log/warn "The aggregate-event-counts endpoint is experimental and may be altered or removed in the future.")
   (-> (routes version)
       verify-accepts-json
       (validate-query-params {:required ["query" "summarize_by"]
--- a/src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
+++ b/src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
@@ -1,50 +1,81 @@
 (ns puppetlabs.puppetdb.query.aggregate-event-counts
   (:require [puppetlabs.puppetdb.query.event-counts :as event-counts]
+            [clojure.string :as str]
+            [puppetlabs.puppetdb.query.events :as events]
             [puppetlabs.puppetdb.jdbc :as jdbc]
             [puppetlabs.kitchensink.core :as kitchensink]))
 
 (defn- get-aggregate-sql
   "Given the `event-count-sql`, return a SQL string that will aggregate the results."
-  [event-count-sql]
-  {:pre  [(string? event-count-sql)]
+  [event-count-sql summarize_by]
+  {:pre  [(string? event-count-sql)
+          (contains? #{"resource" "containing_class" "certname"} summarize_by)]
    :post [(string? %)]}
-  (format "SELECT SUM(CASE WHEN successes > 0 THEN 1 ELSE 0 END) as successes,
+  (format "(SELECT '%s' as summarize_by,
+           SUM(CASE WHEN successes > 0 THEN 1 ELSE 0 END) as successes,
                   SUM(CASE WHEN failures > 0  THEN 1 ELSE 0 END) as failures,
                   SUM(CASE WHEN noops > 0 THEN 1 ELSE 0 END) as noops,
                   SUM(CASE WHEN skips > 0 THEN 1 ELSE 0 END) as skips,
                   COUNT(*) as total
-           FROM (%s) event_counts" event-count-sql))
+           FROM (%s) event_counts)" summarize_by event-count-sql))
 
-(defn query->sql
-  "Convert an aggregate-event-counts `query` and a value to `summarize_by` into a SQL string.
-  Since all inputs are forwarded to `event-counts/query->sql`, look there for proper documentation."
-  [version query [summarize_by query-options]]
+(defn- assemble-aggregate-sql
+  "Convert an aggregate-event-counts `query` and a value to `summarize_by`
+   into a SQL string."
+  [version query summarize_by query-options]
   {:pre  [(sequential? query)
           (string? summarize_by)
-          ((some-fn map? nil?) query-options)]
-   :post [(jdbc/valid-jdbc-query? (:results-query %))]}
+          ((some-fn map? nil?) query-options)]}
   (let [query-options (if (nil? query-options) {} query-options)
         [count-sql & params] (:results-query
-                              (event-counts/query->sql version query [summarize_by query-options {}]))
-        aggregate-sql        (get-aggregate-sql count-sql)]
-    {:results-query (apply vector aggregate-sql params)}))
+                               (event-counts/query->sql
+                                 true
+                                 version
+                                 query [summarize_by query-options {}]))]
+    (vector (get-aggregate-sql count-sql summarize_by) params)))
+
+(defn query->sql
+  "Convert an aggregate-event-counts `query` and a value(s) to `summarize_by`
+   into a SQL string. Since all inputs are forwarded to
+   `event-counts/query->sql`, look there for proper documentation."
+  [version query [summarize_by {:keys [distinct_resources?] :as query-options}]]
+  {:pre  [(sequential? query)
+          ((some-fn map? nil?) query-options)]
+   :post [(jdbc/valid-jdbc-query? (:results-query %))]}
+  (let [summary-vec (str/split summarize_by #",")
+        nsummarized (count summary-vec)
+        aggregate-fn #(assemble-aggregate-sql version query % query-options)
+        aggregated-sql-and-params (map aggregate-fn summary-vec)
+        common-params (second (first aggregated-sql-and-params))
+        params (if distinct_resources?
+                 ;; when distinct resources is used, the first two parameters
+                 ;; are the distinct start/end times.
+                 ;; extract the distinct start/end times and duplicate the rest
+                 ;; as many times as there are summarize_by's
+                 (concat (take 2 common-params)
+                         (flatten (repeat nsummarized (drop 2 common-params))))
+                 ;; repeat all params as many times as there are summarize_by's
+                 (flatten (repeat nsummarized common-params)))
+        unioned-sql (cond-> (str/join " UNION ALL " (map first aggregated-sql-and-params))
+                      distinct_resources?  events/with-latest-events)]
+    {:results-query (apply vector unioned-sql params)}))
 
 (defn- perform-query
   "Given a SQL query and its parameters, return a vector of matching results."
   [[sql & params]]
   {:pre  [(string? sql)]
-   :post [(vector? %)
-          (= (count %) 1)]}
+   :post [(vector? %)]}
   (jdbc/query-to-vec (apply vector sql params)))
 
 (defn munge-result-rows
   [_ _]
-  (comp (partial kitchensink/mapvals #(if (nil? %) 0 %)) first))
+  (fn [rows]
+    (map (partial kitchensink/mapvals #(if (nil? %) 0 %)) rows)))
 
 (defn query-aggregate-event-counts
   "Given a SQL query and its parameters, return the single matching result map."
   [{:keys [results-query]}]
   {:pre  [(string? (first results-query))]
-   :post [(map? %)]}
+   :post [(seq? %)]}
   (->> (perform-query results-query)
        ((munge-result-rows nil nil))))
--- a/src/puppetlabs/puppetdb/query/event_counts.clj
+++ b/src/puppetlabs/puppetdb/query/event_counts.clj
@@ -116,7 +116,7 @@
   "Helper function to transform the event count subject data from the raw format that we get back from the
   database into the more structured format that the API specifies."
   [summarize_by]
-  (fn [_ _] 
+  (fn [_ _]
    (fn [rows]
      (map (partial munge-subject summarize_by) rows))))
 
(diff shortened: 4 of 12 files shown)


wkalt:
submitting to test acceptance. I want to take one more cleanup pass and compile final performance numbers.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1138,7 +1142,8 @@
                                     :environment_id         (ensure-environment environment)
                                     :status_id              (ensure-status status)}))]
                (->> resource_events
-                    (map (comp convert-containment-path #(assoc % :report_id id)))
+                    (map (comp convert-containment-path #(assoc % :report_id id
+                                                           :certname_id (certname-id certname))))
```
ajroetker: alignment

ajroetker: Also I might just make two separate mappings one for `convert` and  one for `assoc` or just lose the `->>`

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -112,7 +120,11 @@
                                                    (:distinct_start_time query-options)
                                                    (:distinct_end_time query-options))
                                   (default-select select-fields where params))
-        paged-select (jdbc/paged-sql sql paging-options)
+        paged-select (cond-> (jdbc/paged-sql sql paging-options)
```
ajroetker: Looking at this code I saw https://github.com/puppetlabs/puppetdb/pull/1453/files#diff-3c835dc023a53db2db8d483f3e6a7da6R118, that looks wrong, shouldn't it be `[sql & params]`?

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
```diff
@@ -1,50 +1,78 @@
...
+   `event-counts/query->sql`, look there for proper documentation."
+  [version query [summarize_by {:keys [distinct_resources?] :as query-options}]]
+  {:pre  [(sequential? query)
+          ((some-fn map? nil?) query-options)]
+   :post [(jdbc/valid-jdbc-query? (:results-query %))]}
+  (let [summary-vec (str/split summarize_by #",")
+        nsummarized (count summary-vec)
+        aggregate-fn #(assemble-aggregate-sql version query [% query-options])
+        aggregated-sql-and-params (map aggregate-fn summary-vec)
+        common-params (second (first aggregated-sql-and-params))
```
ajroetker: Maybe this could happen via some destructuring above like `[[_ common-params] :as aggregated-sql-and-params]`

src/puppetlabs/puppetdb/query/events.clj
wkalt: no...

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
wkalt: thought about that, but decided against

ajroetker:
@wkalt this is awesome by the way, it actually greatly simplifies some of the transducer code I have

documentation/api/query/v4/aggregate-event-counts.markdown
```diff
@@ -12,6 +12,9 @@ canonical: "/puppetdb/latest/api/query/v4/aggregate-event-counts.html"
 > **Experimental Endpoint**: The aggregate-event-counts endpoint is designated
 > as experimental. It may be altered or removed in a future release.
 
+> **Not supported on HSQLDB**: The aggregate-event-counts endpoint is not
+> supported on HSQLDB.
+
```
mullr: :clap: :clap: :clap:

documentation/api/query/v4/aggregate-event-counts.markdown
```diff
@@ -61,16 +65,18 @@ This endpoint builds on top of the [`event-counts`][event-counts] and [`events`]
...
-      "total": 3
-    }
+    [ {
+        "summarize_by": "containing_class",
+        "successes": 2,
+        "failures": 0,
+        "noops": 0,
+        "skips": 1,
+        "total": 3
+    } ]
```
mullr: Are we going to break any downstream consumers of this api?

wkalt: yup, but these changes are for them. ping @kbrezina heads up on this.

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -9,25 +10,35 @@
...
-             (produce-streaming-body
-              :aggregate-event-counts
-              version
-              query
-              [summarize_by query-options]
-              (:scf-read-db globals)
-              (:url-prefix globals))))}))
+    [""]
+    {:get (fn [{:keys [params globals]}]
+            (if (and (:datasource (:scf-read-db globals))
```
mullr: This is similar in spirit, if not implementation, to `storage-utils/postgres?`. Maybe there should be a similarly-flavored `hsql?` fn elsewhere.

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -9,25 +10,35 @@
...
+                     (re-matches #"jdbc:hsqldb.*"
+                                 (.getJdbcUrl (:datasource
+                                                (:scf-read-db globals)))))
+              (http/json-response
+                {:error "The aggregate-event-counts endpoint does not support HSQLDB."}
+                http/status-not-found)
```
mullr: I think this should be `501 Not Implemented`

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
wkalt: yeah, I could use postgres? but I don't have a database connection at this point and that function requires one. Moving the check inside produce-streaming-body would be out of the AEC jurisdiction, and creating a connection just for the check seems too heavyweight when the info is right there in the config. thoughts?

mullr: Right, I mean in terms of calling it out as a util function somewhere. Giving a name to such a procedure seems valuable.

wkalt: that seems reasonable -- I'll make the change

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -9,25 +10,35 @@
...
+              (let [{:strs [query summarize_by counts_filter count_by]} params
+                    distinct-options (validate-distinct-options! params)
+                    counts_filter' (json/parse-string counts_filter true)
+                    query-options (-> {:counts_filter counts_filter'
+                                       :count_by count_by}
+                                      (merge distinct-options))]
+                (log/warn
+                  (str "The aggregate-event-counts endpoint is experimental"
```
mullr: ^:idle-thought I wonder if we should have a special response for these? I kind of doubt people will see these warnings.

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
mullr: this is just `cadar`

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
```diff
@@ -1,50 +1,80 @@
...
             [puppetlabs.kitchensink.core :as kitchensink]))
 
 (defn- get-aggregate-sql
   "Given the `event-count-sql`, return a SQL string that will aggregate the results."
-  [event-count-sql]
+  [event-count-sql summarize_by]
   {:pre  [(string? event-count-sql)]
    :post [(string? %)]}
-  (format "SELECT SUM(CASE WHEN successes > 0 THEN 1 ELSE 0 END) as successes,
+  (format "(SELECT '%s' as summarize_by,
```
mullr: need to use a parameterized query here, not string format

wkalt: good catch

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
```diff
@@ -1,50 +1,80 @@
...
-           FROM (%s) event_counts" event-count-sql))
+           FROM (%s) event_counts)" summarize_by event-count-sql))
 
-(defn query->sql
-  "Convert an aggregate-event-counts `query` and a value to `summarize_by` into a SQL string.
-  Since all inputs are forwarded to `event-counts/query->sql`, look there for proper documentation."
+(defn assemble-aggregate-sql
+  "Convert an aggregate-event-counts `query` and a value to `summarize_by`
+   into a SQL string."
   [version query [summarize_by query-options]]
```
mullr: we assemble the [summarize_by query-options] vector below seemingly for the sole purpose of unpacking it here.

mullr: Also, It this can be defn-

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
```diff
@@ -1,50 +1,80 @@
...
   (let [query-options (if (nil? query-options) {} query-options)
         [count-sql & params] (:results-query
-                              (event-counts/query->sql version query [summarize_by query-options {}]))
-        aggregate-sql        (get-aggregate-sql count-sql)]
-    {:results-query (apply vector aggregate-sql params)}))
+                              (event-counts/query->sql
+                                :aggregate-event-counts
+                                version
+                                query [summarize_by query-options {}]))]
+    [(get-aggregate-sql count-sql summarize_by) params]))
```
mullr: I think `(cons (get-aggregate-sql ...) params)` might be a somewhat more consistent return value for this fn.

mullr: Or a vector of the same. The docstring of `valid-jdbc-query?` lends credence to this.

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
mullr: On a more serious note: how are these params common? I can see how they're the params of the first query, but is it not possible that the other queries have different params?

wkalt: nope, currently the params are the same for each -- you can see that from the aggregate-fn definition above. They all have the same query-options and differ only in summarize_by.

If I pull summarize_by into a param I'll adjust this. For the record I think the way these five lines work right now is pretty nasty. Definitely open to suggestions -- the issue is that it's doing something with shared code that none of our other endpoints do.

mullr: ok, let's take another look at it after that change. I'm worried about the `(if distinct_resources? ...)` block as well, but can't offer any suggestions. Maybe it'll be clearer next time around.

src/puppetlabs/puppetdb/query/event_counts.clj
```diff
@@ -124,7 +124,10 @@
   "Convert an event-counts `query` and a value to `summarize_by` into a SQL string.
   A second `counts-filter` query may be provided to further reduce the results, and
   the value to `count_by` may also be specified (defaults to `resource`)."
-  ([version query [summarize_by {:keys [counts_filter count_by] :as query-options} paging-options]]
+  ([caller
```
mullr: I'm not sure about this `caller` business. Would much rather call out what it changes explicitly. `include-latest-events`, perhaps, based on my reading.

mullr: If you want to make that a keyword arg so it's explicit at the callsite, I'm cool with that. Just don't like functions knowing about their caller and going into a special mode as a result.

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -47,6 +47,23 @@
...
+  [query]
+  (format
+    "WITH latest_events AS
+           (SELECT certnames.certname,
+                   resource_events.resource_type,
+                   resource_events.resource_title,
+                   resource_events.property,
+                   MAX(resource_events.timestamp) AS timestamp
+                   FROM resource_events
+                   JOIN certnames ON resource_events.certname_id = certnames.id
```
mullr: it's about bleedin' time. At some point we'll want to retrofit the rest of the queries that join on certname as a string.

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
wkalt: if it's any indication we've had complaints about noisy logging in the past. I don't know that I agree we need a separate response -- my main hesitation about what we're currently doing is that it may look weird to PE users.

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -35,7 +36,7 @@
           :catalogs [catalogs/query->sql catalogs/munge-result-rows]
           :nodes [nodes/query->sql (constantly identity)]
           :environments [environments/query->sql (constantly identity)]
-          :events [events/query->sql events/munge-result-rows]
+          :events [(partial events/query->sql :events) events/munge-result-rows]
```
mullr: This is the `caller`, right?

wkalt: yeah

wkalt: I like the idea of changing it to a description of the action though

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1128,8 +1128,25 @@
             (munge-uuid "transaction_uuid") ", environment_id, status_id
             FROM reports"))
 
+    (sql/do-commands
```
mullr: Why is this in the middle of an old migration?

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1128,8 +1128,25 @@
...
+      "CREATE SEQUENCE certname_id_seq CYCLE")
+
+    (sql/create-table :certnames_transform
+                      ;; Rename the 'name' column of certnames to 'certname'.
```
mullr: why?

src/puppetlabs/puppetdb/scf/migrate.clj
wkalt: it's a pretty heavy migration that recreates the same tables. If we can stomach splitting it all out I've got it in that form somewhere -- I think having them separate is far preferable in general.

src/puppetlabs/puppetdb/scf/migrate.clj
wkalt: that's not a new change

mullr: In fact, it's a change that I made.

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: I don't think it's really that heavy, is it? certnames is a small table.

wkalt: actually, now that I consider it you may be right. it's not certnames that's the issue, it's resource_events, which is a massive table in the cases that this patch addresses.
As written this creates a new column on that table at the same time that we are already recreating it, but the really expensive part is creating a new index, and we gain nothing from having that mixed in with this other stuff.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1136,9 +1140,12 @@
...
+                                    :status_id              (ensure-status status)}))
+                   assoc-ids #(assoc %
+                                     :report_id id
+                                     :certname_id (certname-id certname))]
                (->> resource_events
-                    (map (comp convert-containment-path #(assoc % :report_id id)))
+                    (map (comp convert-containment-path assoc-ids))
                     (apply sql/insert-records :resource_events))
                (when update-latest-report?
                  (update-latest-report! certname)))))))
```
mullr: This will turn around and look up the certname by id again, we should pass the id through.

mullr:
Ok, done looking through this. Will wait for the next revision before doing testing on it.

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: there's no certname_id in the reports table as a result of this patch. I'm pretty sure I'd support that change but I think it would hit enough places to warrant its own ticket.

mullr: But it's sitting there in memory, line 1146

mullr: erm.. .no. never mind.

wkalt:
@mullr I'm not sure I've really addressed your concern around the "caller" ugliness, but I got the other stuff. I plopped the hsqldb? function in puppetdb/utils.clj. Not sure that was the best place for it but it didn't feel like a storage-utils function since the functions in that namespace seem to be interacting with the database directly.

wkalt:

wkalt:

senior:

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -257,3 +257,10 @@
       {:href (to-href data)}
       (-> (sutils/parse-db-json data)
           (update :href to-href)))))
+
+(defn hsql?
+  "given the globals configuration bundle, determine if hsqldb is being used."
+  [globals]
```
ajroetker: Usually the pattern I've seen in other projects for this kind of functions is `db-spec` here. I'd prefer we'd do something like that.

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
```diff
@@ -1,50 +1,81 @@
...
-   :post [(vector? %)
-          (= (count %) 1)]}
+   :post [(vector? %)]}
   (jdbc/query-to-vec (apply vector sql params)))
 
 (defn munge-result-rows
   [_ _]
-  (comp (partial kitchensink/mapvals #(if (nil? %) 0 %)) first))
+  (fn [rows]
+    (map (partial kitchensink/mapvals #(if (nil? %) 0 %)) rows)))
```
ajroetker: Unrelated to your PR but this looks like an `(fnil indentity 0)`

src/puppetlabs/puppetdb/utils.clj
wkalt: db-spec means something specific to me in the context of jdbc. It's the classname, subprotocol, subname map formed by the [database] section of our config. globals is our name for a different map that we use in a lot of places, and I think it'd be confusing to call it db-spec here.

ajroetker: I mean we'd just call it `db` but all the other functions for this kind of thing I've seen outside this project use the `db` (`classname`, `subprotocol`, `subname`) as the param

ajroetker: Basically, I could see functions like this being used when you only have access to `db` and not to `globals`.

wkalt: I think I must be misunderstanding what you're proposing here, but just to be clear, this is a different map than the db-spec map I described above -- it has different keys, different values, and db-spec is not contained anywhere inside it. Globals really seems more appropriate than db or db-spec to me. When we use db, or *db* in other functions, we're talking about that db-spec object.

mullr:
I'm not seeing any performance difference. The below test was done with @wkalt 's data, with postgres and pdb running on my mac. Postgres was configured with pgtune to use 8 gigs of memory; it's possible that I've missed something though and haven't set it up right. 

```
1 Aggregate event count patch perf testing
══════════════════════════════════════════

  ┌────
  │ select count(*) from resource_events;
  └────

  ━━━━━━━━━
     count 
  ─────────
   8376036 
  ━━━━━━━━━


1.1 Before patch
────────────────

1.1.1 by certname
╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌

  ┌────
  │ (time curl -s -X GET \
  │       'http://localhost:8080/pdb/query/v4/aggregate-event-counts'  \
  │       -d 'query=[">","report_receive_time",0]' \
  │       -d 'distinct_resources=true' \
  │       -d 'distinct_start_time=2015-05-25T15:39:33.014Z' \
  │       -d 'distinct_end_time=2015-05-26T15:39:33.014Z' \
  │       -d 'summarize_by=certname' \
  │       > /dev/null) \
  │     2>&1
  └────

  ━━━━━━━━━━━━━━━━━
   real  0m52.382s 
   user  0m0.006s  
   sys   0m0.005s  
  ━━━━━━━━━━━━━━━━━


1.1.2 by containing class
╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌

  ┌────
  │ (time curl -s -X GET \
  │       'http://localhost:8080/pdb/query/v4/aggregate-event-counts'  \
  │       -d 'query=[">","report_receive_time",0]' \
  │       -d 'distinct_resources=true' \
  │       -d 'distinct_start_time=2015-05-25T15:39:33.014Z' \
  │       -d 'distinct_end_time=2015-05-26T15:39:33.014Z' \
  │       -d 'summarize_by=containing_class' \
  │       > /dev/null) \
  │     2>&1
  └────

  ━━━━━━━━━━━━━━━━━
   real  0m52.744s 
   user  0m0.006s  
   sys   0m0.005s  
  ━━━━━━━━━━━━━━━━━


1.1.3 by resource
╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌

  ┌────
  │ (time curl -s -X GET \
  │       'http://localhost:8080/pdb/query/v4/aggregate-event-counts'  \
  │       -d 'query=[">","report_receive_time",0]' \
  │       -d 'distinct_resources=true' \
  │       -d 'distinct_start_time=2015-05-25T15:39:33.014Z' \
  │       -d 'distinct_end_time=2015-05-26T15:39:33.014Z' \
  │       -d 'summarize_by=resource' \
  │       > /dev/null) \
  │     2>&1
  └────

  ━━━━━━━━━━━━━━━━━
   real  0m53.693s 
   user  0m0.006s  
   sys   0m0.007s  
  ━━━━━━━━━━━━━━━━━


1.2 After patch
───────────────

1.2.1 by certname
╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌

  ┌────
  │ (time curl -s -X GET \
  │       'http://localhost:8080/pdb/query/v4/aggregate-event-counts'  \
  │       -d 'query=[">","report_receive_time",0]' \
  │       -d 'distinct_resources=true' \
  │       -d 'distinct_start_time=2015-05-25T15:39:33.014Z' \
  │       -d 'distinct_end_time=2015-05-26T15:39:33.014Z' \
  │       -d 'summarize_by=certname' \
  │       > /dev/null) \
  │     2>&1
  └────

  ━━━━━━━━━━━━━━━━━
   real  0m51.215s 
   user  0m0.005s  
   sys   0m0.004s  
  ━━━━━━━━━━━━━━━━━


1.2.2 by containing class
╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌

  ┌────
  │ (time curl -s -X GET \
  │       'http://localhost:8080/pdb/query/v4/aggregate-event-counts'  \
  │       -d 'query=[">","report_receive_time",0]' \
  │       -d 'distinct_resources=true' \
  │       -d 'distinct_start_time=2015-05-25T15:39:33.014Z' \
  │       -d 'distinct_end_time=2015-05-26T15:39:33.014Z' \
  │       -d 'summarize_by=containing_class' \
  │       > /dev/null) \
  │     2>&1
  └────

  ━━━━━━━━━━━━━━━━━
   real  0m51.768s 
   user  0m0.006s  
   sys   0m0.005s  
  ━━━━━━━━━━━━━━━━━


1.2.3 by resource
╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌

  ┌────
  │ (time curl -s -X GET \
  │       'http://localhost:8080/pdb/query/v4/aggregate-event-counts'  \
  │       -d 'query=[">","report_receive_time",0]' \
  │       -d 'distinct_resources=true' \
  │       -d 'distinct_start_time=2015-05-25T15:39:33.014Z' \
  │       -d 'distinct_end_time=2015-05-26T15:39:33.014Z' \
  │       -d 'summarize_by=resource' \
  │       > /dev/null) \
  │     2>&1
  └────

  ━━━━━━━━━━━━━━━━━
   real  0m52.556s 
   user  0m0.006s  
   sys   0m0.005s  
  ━━━━━━━━━━━━━━━━━


1.2.4 all together
╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌

  ┌────
  │ (time curl -s -X GET \
  │       'http://localhost:8080/pdb/query/v4/aggregate-event-counts'  \
  │       -d 'query=[">","report_receive_time",0]' \
  │       -d 'distinct_resources=true' \
  │       -d 'distinct_start_time=2015-05-25T15:39:33.014Z' \
  │       -d 'distinct_end_time=2015-05-26T15:39:33.014Z' \
  │       -d 'summarize_by=resource,certname,containing_class \
  │       > /dev/null) \
  │     2>&1
  └────

  ━━━━━━━━━━━━━━━━━
   real  0m52.556s 
   user  0m0.006s  
   sys   0m0.005s  
  ━━━━━━━━━━━━━━━━━

```

mullr:
Ok, I've been through various permutations of this and learned that a beefier postgres is what does the job. It does even better with all the indices in place. I'm :+1: on this now.

wkalt:
I think there's a little more discussion needed -- don't merge yet

wkalt:

src/puppetlabs/puppetdb/query/events.clj
senior: :+1: :+1: :+1:

test/puppetlabs/puppetdb/fixtures.clj
```diff
@@ -106,6 +106,12 @@
                      (java.util.UUID/randomUUID)
                      ";hsqldb.tx=mvcc;sql.syntax_pgs=true")})
 
+(defn skip-hsqldb
```
senior: I think what you really want here is a test selector. We use the something similar for the opposite case [here] (https://github.com/puppetlabs/puppetdb/blob/master/test/puppetlabs/puppetdb/http/facts_test.clj#L577)

wkalt: thanks

wkalt:

PR #1455, opened 2015-06-12
(PDB-1455) Iterate to install ruby on Debian for AIO

kbarber:
This moves the code to install a real ruby on Debian for AIO so it iterates
on all all database nodes, before it was only installing it on the first one.


senior:
keeping it real

PR #1456, opened 2015-06-12
(PDB-1455) Use AIO/Puppet rake for building packages from source

kbarber:
The packaging tools require facter, but system ruby would not contain this
by default, especially in an AIO world.

So we just use the AIO PATH so that Puppet's rake is used instead, that
already contains facter.


kbarber:
The intention of this is to fix PR testing on the stable branch with AIO.

PR #1457, opened 2015-06-12
(maint) add realistically spaced timestamps for benchmark objects bas…

wkalt:
…ed on -N

This patch computes a starting timestamp based on the -N flag and uses it for
all wire-format timestamps in benchmark command submission,
steping forward in time based on an assumed runinterval of 30 minutes. Does not
alter behavior in -i mode.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/benchmark.clj
+++ b/src/puppetlabs/puppetdb/cli/benchmark.clj
@@ -125,25 +125,30 @@
 
 (defn maybe-tweak-catalog
   "Slightly tweak the given catalog, returning a new catalog, `rand-percentage`
-  percent of the time."
-  [catalog rand-percentage]
-  (if (< (rand 100) rand-percentage)
-    (rand-catalog-mutation catalog)
-    catalog))
+   percent of the time."
+  [catalog rand-percentage stamp]
+  (let [catalog' (assoc catalog "producer_timestamp" stamp)]
+    (if (< (rand 100) rand-percentage)
+      (rand-catalog-mutation catalog')
+      catalog')))
 
-(defn update-catalog [catalog]
-  (assoc catalog "producer_timestamp" (time/now)))
+(defn jitter
+  "jitter a timestamp (rand-int n) seconds in the forward direction"
+  [stamp n]
+  (time/plus stamp (time/seconds (rand-int n))))
 
 (defn update-report-run-fields
   "configuration_version, start_time and end_time should always change
    on subsequent report submittions, this changes those fields to avoid
    computing the same hash again (causing constraint errors in the DB)"
-  [report]
-  (assoc report
-         "configuration_version" (ks/uuid)
-         "start_time" (time/now)
-         "end_time" (time/now)
-         "producer_timestamp" (time/now)))
+  [report stamp]
+  (-> report
+      (update "resource_events" (partial map #(assoc % "timestamp"
+                                                     (jitter stamp 300))))
+      (assoc "configuration_version" (ks/uuid)
+             "start_time" (time/minus stamp (time/seconds 10))
+             "end_time" (time/minus stamp (time/seconds 5))
+             "producer_timestamp" stamp)))
 
 (defn randomize-map-leaf
   "Randomizes a fact leaf based on a percentage provided with `rp`."
@@ -173,9 +178,9 @@
 (defn update-factset
   "Updates the producer_timestamp to be current, and randomly updates the leaves
    of the factset based on a percentage provided in `rand-percentage`."
-  [factset rand-percentage]
+  [factset rand-percentage stamp]
   (-> factset
-      (assoc "producer_timestamp" (time/now))
+      (assoc "producer_timestamp" stamp)
       (update "values" (partial randomize-map-leaves rand-percentage))))
 
 (defn timed-update-host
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -123,27 +123,35 @@
...
    computing the same hash again (causing constraint errors in the DB)"
-  [report]
+  [report stamp]
   (assoc report
-         "configuration_version" (ks/uuid)
-         "start_time" (time/now)
-         "end_time" (time/now)
-         "producer_timestamp" (time/now)))
+    "configuration_version" (ks/uuid)
+    "resource_events" (map #(assoc % "timestamp" (jitter stamp 300))
```
ajroetker: This looks like it could be turned into an `update`.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -123,27 +123,35 @@
...
+                           (get report "resource_events"))
+    "start_time" (time/minus stamp (time/seconds 10))
+    "end_time" (time/minus stamp (time/seconds 5))
+    "producer_timestamp" stamp))
```
ajroetker: Also maybe there should be an `update-producer-timestamp` function since all three entities need it.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -123,27 +123,35 @@
...
-  (assoc catalog "producer_timestamp" (time/now)))
+   percent of the time."
+  [catalog rand-percentage stamp]
+  (let [catalog' (update-catalog catalog stamp)]
+    (if (< (rand 100) rand-percentage)
+      (rand-catalog-mutation catalog')
+      catalog')))
+
+(defn jitter
+  "jitter a timestamp (rand n) milliseconds in the forward direction"
```
senior: I'm a little bit confused here. Comment says milliseconds but the code below says seconds. Also below is using rand, which returns a floating point number, but time/seconds is type hinted as an integer (and the java method it calls also only accepts an integer). There must be some auto casting magic going on. I think using rand-int here and above would be less confusing.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -245,10 +265,15 @@
   [hosts num-msgs]
   (log/infof "Sending %s messages for %s hosts, will exit upon completion"
              num-msgs (count hosts))
-  (loop [mutated-hosts hosts
-         msgs-to-send num-msgs]
-    (when-not (zero? msgs-to-send)
-      (recur (mapv update-host mutated-hosts) (dec msgs-to-send)))))
+  (let [stamp (atom (time/minus (time/now) (time/minutes (* 30 num-msgs))))]
```
senior: Why do we need an atom here? Seems like it could be another loop var pretty easily...

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -123,27 +123,35 @@
                                    (repeatedly children #(random-structured-fact (rand-nth (range depth))
                                                                                  (rand-nth (range children)))))})))
 
+(defn update-catalog [catalog stamp]
```
senior: Seems like this function should have a different name or maybe get slurped into where it's used. Right now it's a pretty generic fn name for something pretty specific.

src/puppetlabs/puppetdb/cli/benchmark.clj
wkalt: I left this as is because I'm not updating the existing value, I'm overriding it, and assoc seems a bit nicer for that to me. As far as I can tell the straightforward way to change this to update would require a nested anonymous function (the function being mapped, plus the function to update with).

src/puppetlabs/puppetdb/cli/benchmark.clj
wkalt: ended up just ditching the update-catalog function to do this inline in all cases

wkalt:
@senior @ajroetker thanks for the comments -- should be addressed

src/puppetlabs/puppetdb/cli/benchmark.clj
ajroetker: You mean `(update report "resource_events" (partial map #(assoc % "timestamp" (jitter stamp 300))))` I don't see why that's a problem, we do that all over the place in Clojure.

ajroetker: I know what you mean about the ugliness of the `update` when it's a list of things, it would be nice if the there was an `update-coll` function that was a wrapper around `update ... (partial map`, but I still think at the top level here you're updating and at the `map` level you're overwriting.

wkalt:
@ajroetker okay, I misunderstood you. I thought you were talking about converting the assoc on the line of your comment. Made the change -- see what you think.

ajroetker:
:+1: @wkalt looks good!

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -125,25 +125,30 @@
...
-  [report]
-  (assoc report
-         "configuration_version" (ks/uuid)
-         "start_time" (time/now)
-         "end_time" (time/now)
-         "producer_timestamp" (time/now)))
+  [report stamp]
+  (-> report
+      (update "resource_events" (partial map #(assoc % "timestamp"
+                                                     (jitter stamp 300))))
```
ajroetker: Don't know how you feel about this but saw this and thought it would be a good kind of thing to use `specter` for https://github.com/nathanmarz/specter. @senior brought it in, in the command versions PR. I think it's basically `(sp/update ["resource_events" sp/ALL] #(assoc % "timestamp" (jitter stamp 300)) report)`

wkalt: I'm not inclined to change it without a performance reason, just because I find specter's update uglier than update in core. Looking at a profiler, this doesn't seem like a particular hotspot in the benchmark code. Any reason to prefer specter here?

ajroetker: No perf reasons, I don't mind either way, I just don't like the `update ... map` syntax, but probably not worth holding up this PR either way

wkalt: I see where you're coming from with respect to map. I'll think about it.

PR #1458, opened 2015-06-15
(maint) Unify cli interfaces

ajroetker:
This PR unifies the CLI handling in PuppetDB.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/anonymize.clj
+++ b/src/puppetlabs/puppetdb/cli/anonymize.clj
@@ -8,9 +8,9 @@
             [puppetlabs.puppetdb.catalogs :as catalogs]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.cli.export :refer [export-metadata-file-name]]
-            [puppetlabs.puppetdb.cli.import :refer [parse-metadata]]
+            [puppetlabs.puppetdb.cli.import :as import]
             [puppetlabs.puppetdb.schema :as pls]
-            [puppetlabs.puppetdb.utils :refer [export-root-dir add-tar-entry]]
+            [puppetlabs.puppetdb.utils :as utils :refer [export-root-dir add-tar-entry]]
             [schema.core :as s]
             [slingshot.slingshot :refer [try+]])
   (:import [org.apache.commons.compress.archivers.tar TarArchiveEntry]
@@ -192,84 +192,84 @@
        archive/read-entry-content
        json/parse-string))
 
-(defn process-tar-entry
+(defn add-anonymized-entity
+  [tar-writer old-path entity file-name data]
+  (let [file-suffix [entity file-name]
+        new-path (.getPath (apply io/file export-root-dir file-suffix))
+        singular-entity (case entity
+                          "facts" "facts"
+                          "catalogs" "catalog"
+                          "reports" "report")]
+    (-> "Anonymizing %s from archive entry '%s' into '%s'"
+        (format singular-entity old-path new-path)
+        println)
+    (add-json-tar-entry tar-writer file-suffix data)))
+
+(pls/defn-validated process-tar-entry
   "Determine the type of an entry from the exported archive, and process it
   accordingly."
-  [^TarGzReader tar-reader ^TarArchiveEntry tar-entry ^TarGzWriter tar-writer config metadata]
-  {:pre  [(instance? TarGzReader tar-reader)
-          (instance? TarArchiveEntry tar-entry)
-          (instance? TarGzWriter tar-writer)]}
-  (let [path    (.getName tar-entry)
-        catalog-pattern (str "^" (.getPath (io/file export-root-dir "catalogs" ".*\\.json")) "$")
-        report-pattern (str "^" (.getPath (io/file export-root-dir "reports" ".*\\.json")) "$")
-        facts-pattern (str "^" (.getPath (io/file export-root-dir "facts" ".*\\.json")) "$")]
+  [tar-reader :- TarGzReader
+   tar-entry :- TarArchiveEntry
+   tar-writer :- TarGzWriter
+   config
+   command-versions]
+  (let [path (.getName tar-entry)]
+    (condp re-find path
+      ;; Process catalogs
+      (import/file-pattern "catalogs")
+      (let [{:strs [certname] :as new-catalog} (->> (next-json-tar-entry tar-reader)
+                                                    (anon/anonymize-catalog config))]
+        (add-anonymized-entity tar-writer path "catalogs" (format "%s.json" certname) new-catalog))
 
-    ;; Process catalogs
-    (when (re-find (re-pattern catalog-pattern) path)
-      (let [new-catalog  (->> (next-json-tar-entry tar-reader)
-                              (anon/anonymize-catalog config))
-            new-hostname (get new-catalog "certname")
-            file-suffix  ["catalogs" (format "%s.json" new-hostname)]
-            new-path     (.getPath (apply io/file export-root-dir file-suffix))]
-        (println (format "Anonymizing catalog from archive entry '%s' into '%s'" path new-path))
-        (add-json-tar-entry tar-writer file-suffix new-catalog)))
+      ;; Process reports
+      (import/file-pattern "reports")
+      (let [cmd-version (:store_report command-versions)
+            {:strs [start_time configuration_version certname] :as new-report}
+            (->> (next-json-tar-entry tar-reader)
+                 (anon/anonymize-report config cmd-version))]
+        (add-anonymized-entity tar-writer path "reports" (->> (str start_time configuration_version)
+                                                              kitchensink/utf8-string->sha1 
+                                                              (format "%s-%s.json" certname)) new-report))
 
-    ;; Process reports
-    (when (re-find (re-pattern report-pattern) path)
-      (let [cmd-version  (get-in metadata [:command_versions :store_report])
-            new-report   (->> (next-json-tar-entry tar-reader)
-                              (anon/anonymize-report config cmd-version))
-            new-hostname (get new-report "certname")
-            new-hash     (kitchensink/utf8-string->sha1
-                          (str
-                           (get new-report "start_time")
-                           (get new-report "configuration_version")))
-            file-suffix  ["reports" (format "%s-%s.json" new-hostname new-hash)]
-            new-path     (.getPath (apply io/file export-root-dir file-suffix))]
-        (println (format "Anonymizing report from archive entry '%s' to '%s'" path new-path))
-        (add-json-tar-entry tar-writer file-suffix new-report)))
-
-    ;; Process facts
-    (when (re-find (re-pattern facts-pattern) path)
-      (let [new-facts    (->> (next-json-tar-entry tar-reader)
-                              (anon/anonymize-facts config))
-            new-hostname (get new-facts "certname")
-            file-suffix  ["facts" (format "%s.json" new-hostname)]
-            new-path     (.getPath (apply io/file export-root-dir file-suffix))]
-        (println (format "Anonymizing facts from archive entry '%s' to '%s'" path new-path))
-        (add-json-tar-entry tar-writer file-suffix new-facts)))))
+      ;; Process facts
+      (import/file-pattern "facts")
+      (let [{:strs [certname] :as new-facts} (->> (next-json-tar-entry tar-reader)
+                                                  (anon/anonymize-facts config))]
+        (add-anonymized-entity tar-writer path "facts" (format "%s.json" certname) new-facts))
+      nil)))
 
 (defn- validate-cli!
   [args]
   (let [profiles (string/join ", " (keys anon-profiles))
-        specs    [["-o" "--outfile OUTFILE" "Path to output file (required)"]
-                  ["-i" "--infile INFILE" "Path to input file (required)"]
-                  ["-p" "--profile PROFILE" (str "Choice of anonymization profile: " profiles) :default "moderate"]
-                  ["-c" "--config CONFIG" "Configuration file path for extra profile definitions (experimental) (optional)"]]
+        specs [["-o" "--outfile OUTFILE" "Path to output file (required)"]
+               ["-i" "--infile INFILE" "Path to input file (required)"]
+               ["-p" "--profile PROFILE" (str "Choice of anonymization profile: " profiles)
+                :default "moderate"]
+               ["-c" "--config CONFIG" "Configuration file path for extra profile definitions (experimental) (optional)"
+                :default {}
+                :parse-fn (comp clojure.edn/read-string slurp)]]
         required [:outfile :infile]]
-    (try+
-     (kitchensink/cli! args specs required)
-     (catch map? m
-       (println (:message m))
-       (case (:type m)
-         :puppetlabs.kitchensink.core/cli-error (System/exit 1)
-         :puppetlabs.kitchensink.core/cli-help  (System/exit 0))))))
+    (utils/try+-process-cli!
+     (fn []
+       (first
+        (kitchensink/cli! args specs required))))))
 
 (defn -main
   [& args]
-  (let [[{:keys [outfile infile profile config]} _] (validate-cli! args)
-        extra-config                                (if (empty? config) {} (clojure.edn/read-string (slurp config)))
-        profile-config                              (get (merge anon-profiles extra-config) profile)
-        metadata                                    (parse-metadata infile)]
+  (let [{:keys [outfile infile profile config]} (validate-cli! args)
+        profile-config (-> anon-profiles
+                           (merge config)
+                           (get profile))
+        metadata (import/parse-metadata infile)]
 
     (println (str "Anonymizing input data file: " infile " with profile type: " profile " to output file: " outfile))
 
-    (with-open [tar-reader (archive/tarball-reader infile)]
-      (with-open [tar-writer (archive/tarball-writer outfile)]
-        ;; Write out the metadata first
-
-        (add-json-tar-entry tar-writer [export-metadata-file-name] metadata)
-        ;; Now process each entry
-        (doseq [tar-entry (archive/all-entries tar-reader)]
-          (process-tar-entry tar-reader tar-entry tar-writer profile-config metadata))))
-    (println (str "Anonymization complete. Check output file contents " outfile " to ensure anonymization was adequate before sharing data"))))
+    (with-open [tar-reader (archive/tarball-reader infile)
+                tar-writer (archive/tarball-writer outfile)]
+      ;; Write out the metadata first
+      (add-json-tar-entry tar-writer [export-metadata-file-name] metadata)
+      ;; Now process each entry
+      (doseq [tar-entry (archive/all-entries tar-reader)]
+        (process-tar-entry tar-reader tar-entry tar-writer profile-config (:command_versions metadata))))
+    (println (str "Anonymization complete. "
+                  "Check output file contents " outfile " to ensure anonymization was adequate before sharing data"))))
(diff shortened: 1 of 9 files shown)


senior:
@ajroetker Broken tests ^^^

rbrw:
Also, I'm happy to defer to the team on this (couldn't recall where we ended up), but I'm still wary of whitespace only changes.  Personally, I tend to just avoid them so I don't even have to consider the possibility of spurious merge conflicts.

ajroetker:
I'm not sure why the `cli-error` and `cli-help` wouldn't be public interfaces, they would seem to be considering `tk/main` strips the namespaces off of the keys so that any namespace could use them. Also I attempted to put all the whitespace changes in the separate commit.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -156,7 +156,7 @@
    :subname     (format "file:%s;hsqldb.tx=mvcc;sql.syntax_pgs=true" (io/file vardir "db"))})
 
 (defn validate-db-settings
-  "Throws a {:type ::configuration-error :message m} exception
+  "Throws a {:type ::cli-error :message m} exception
```
rbrw: I'm not sure it's important, but I'd used a separate configuration-error because I wasn't sure that a config file error should be treated the same as a cli-error, and because Russell asked whether or not the TK cli-error and cli-help exceptions were actually a public interface.

rbrw: Right, I don't know either way wrt cli-error; I just had those two considerations on my mind when I decided to go with configuration-error, and had imagined I might come back and convert a lot of the other IllegalArgumentExceptions to configuration-error, and then handle that higher up.

rbrw: I also suspect we're going to either want changes/enhancements to TK, or to keep our own wrapper function in the medium-term.  Though for that to really work we might at least need to talk to TK about not calling shutdown-agents.  i.e. it should probably either stop doing that, or provide comprehensive cleanup/error-handling facilities.

ajroetker: (relink) I'm not sure why the cli-error and cli-help wouldn't be public interfaces, they would seem to be considering tk/main strips the namespaces off of the keys so that any namespace could use them. Also I would argue that it's a public interface because the `cli!` function uses that interface, exposed from a separate project, kitchensink.

rbrw:
I think it could make sense (if we're going to have (isolated) whitespace only changes) to pull them out into a separate commit if that makes review easier, but that doesn't help at all with respect to the potential for inter-branch merge conflicts.

rbrw:
Personally in most cases, I'd probably just wait to fix the whitespace until I'm changing the relevant code for some other reason.

rbrw:
But that might not be as critical for PuppetDB depending on how often (and what) we merge (where).

ajroetker:
@rbrw removed the whitespace/syntax changes

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -1,6 +1,6 @@
 (ns puppetlabs.puppetdb.anonymizer
   (:require [clojure.string :as str]
-            [puppetlabs.kitchensink.core :refer [regexp? boolean? uuid string-contains?]]
+            [puppetlabs.kitchensink.core :refer [regexp? boolean? uuid string-contains?] :as kitchensink]
```
wkalt: not used?

src/puppetlabs/puppetdb/cli/anonymize.clj
```diff
@@ -8,9 +8,9 @@
             [puppetlabs.puppetdb.catalogs :as catalogs]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.cli.export :refer [export-metadata-file-name]]
-            [puppetlabs.puppetdb.cli.import :refer [parse-metadata]]
+            [puppetlabs.puppetdb.cli.import :as import]
             [puppetlabs.puppetdb.schema :as pls]
-            [puppetlabs.puppetdb.utils :refer [export-root-dir add-tar-entry]]
+            [puppetlabs.puppetdb.utils :as utils :refer [export-root-dir add-tar-entry]]
```
wkalt: can we just use the utils alias for everything? looks like we're already there with add-tar-entry. I try to avoid using :as and :refer in the same line.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -117,66 +113,67 @@
...
-  [rp value]
-  (cond
-   (map? value)
-   (ks/mapvals (partial randomize-map-leaves rp) value)
+  "Runs through a map and randomizes and random percentage of leaves."
+  [value & [rand-percentage]]
+  (let [rand-perc (or rand-percentage (rand-int 100))]
+    (cond
+      (map? value)
+      (kitchensink/mapvals (partial randomize-map-leaves rand-perc) value)
```
wkalt: isn't this going to take rand-perc as the value argument?

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -117,66 +113,67 @@
...
 
-   (coll? value)
-   (map (partial randomize-map-leaves rp) value)
+      (coll? value)
+      (map (partial randomize-map-leaves rand-perc) value)
```
wkalt: same here

src/puppetlabs/puppetdb/cli/benchmark.clj
ajroetker: So I actually did that intentionally, I wasn't really sure what we wanted. If the goal is to have `rp` of factsets altered, then this is what we want. If we pass `rp` through I think we end up in a situate where `rp` of `rp` fact leaves get updated, that's fine just curious what we want.

src/puppetlabs/puppetdb/cli/anonymize.clj
ajroetker: yeah for sure I hate that as well

src/puppetlabs/puppetdb/anonymizer.clj
ajroetker: left over I'll clean that up

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -191,64 +188,86 @@
...
-  (let [catalog (some-> catalog update-catalog (maybe-tweak-catalog rand-percentage))
-        report (some-> report update-report-run-fields)
-        factset (some-> factset (update-factset rand-percentage))]
-    (when catalog (client/submit-catalog base-url 6 (json/generate-string catalog)))
-    (when report (client/submit-report base-url 5 (json/generate-string report)))
-    (when factset (client/submit-facts base-url 4 (json/generate-string factset)))
-    (assoc state :catalog catalog)))
+  [{:keys [catalog report factset] :as state} base-url rand-percentage]
+  (let [catalog (some->> catalog (maybe-tweak-catalog rand-percentage))
+        factset (some->> factset (maybe-tweak-factset rand-percentage))]
```
wkalt: any reason not to combine these with the threading below?

wkalt:
this seems to cause the benchmark tool to fail when I run 
```
lein run benchmark -c postgres.ini -n 10 -N 10
```

error is

```
 Wrong number of args (1) passed to: benchmark/validate-options
```

src/puppetlabs/puppetdb/cli/benchmark.clj
wkalt: rp should be constant for the entire run -- the condition
```
(< (rand-int 100) rp)
```
is meant to determine whether the object's values are altered or not

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -191,64 +188,86 @@
...
    based on the clock)"
-  [{:keys [host lastrun catalog report factset base-url run-interval rand-percentage] :as state}]
-  (let [catalog (some-> catalog update-catalog (maybe-tweak-catalog rand-percentage))
-        report (some-> report update-report-run-fields)
-        factset (some-> factset (update-factset rand-percentage))]
-    (when catalog (client/submit-catalog base-url 6 (json/generate-string catalog)))
-    (when report (client/submit-report base-url 5 (json/generate-string report)))
-    (when factset (client/submit-facts base-url 4 (json/generate-string factset)))
-    (assoc state :catalog catalog)))
+  [{:keys [catalog report factset] :as state} base-url rand-percentage]
```
wkalt: this may be fallout from the discussion about rp above, but are rand-percentage and base-url no longer accessible through state?

src/puppetlabs/puppetdb/cli/benchmark.clj
wkalt: I see why now

src/puppetlabs/puppetdb/cli/benchmark.clj
ajroetker: Correct but does `rp` for factsets refer to the number of factsets that should be tweaked or the number of leaves that should be tweaked over an entire run?

src/puppetlabs/puppetdb/cli/benchmark.clj
ajroetker: correct, I took them out of state and passed them down from the `cli` since they don't change.

src/puppetlabs/puppetdb/cli/benchmark.clj
ajroetker: The idea here was that if I'm running the benchmarking tool with a low or `0` `rand-perc (rp)` that I don't want to do a bunch of `rand 100` (one for each leaf) on every factset check, it's faster to assume `rp` refers to the percentage of `factsets` we want to tweak and for each factset we do tweak, we will tweak and random number of it's leaves.

wkalt: oh I see, misunderstood your last comment.

We've never really nailed down what we mean by rand-perc applied to factsets (or what a "fact duplication rate" is), and we don't have good dashboard metrics to quantify the effect of the parameter yet. The existing doc string does say "updates the leaves" though.  For catalogs the update is all or nothing. I think the way we're doing it now seems more realistic, because having identical catalogs across the infrastructure seems more likely than having identical factsets, or factsets that only differ in certname.

I haven't looked to hard at it, but I don't see your "rp of rp" scenario in the existing code in master. Seems like rand-perc only comes into play on the outermost leaves (in randomize-map-leaf.) Maybe I'm missing something there though.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -271,140 +290,110 @@
...
-      (action-on-error-fn))
-
-    (not-any? #{:nummsgs :runinterval} (keys options))
-    (do
-      (log/error "Error: Either -N/--nummsgs or -i/--runinterval is required.")
-      (action-on-error-fn))
+    (utils/throw+-cli-error! "Error: -N/--nummsgs runs immediately and is not compatable with -i/--runinterval")
+    
+    (and (not (contains? options :runinterval))
+         (not (contains? options :nummsgs)))
```
wkalt: not-any? seemed nicer to me

src/puppetlabs/puppetdb/cli/benchmark.clj
ajroetker: That's a good point, I'll fix that up here too then. Thanks!

ajroetker:
@wkalt Looks like something did indeed break after i rebased, I'll see what's up

src/puppetlabs/puppetdb/cli/benchmark.clj
ajroetker: Oops yeah that wasn't meant to be in this commit anyways, but there's actually a `kitchensink` function called `missing?` that does this a little better

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -271,140 +290,110 @@
...
+  (let [specs [["-c" "--config CONFIG" "Path to config or conf.d directory (required)"
+                :parse-fn config/load-config]
+               ["-F" "--facts FACTS" "Path to a directory containing sample JSON facts (files must end with .json)"]
+               ["-C" "--catalogs CATALOGS" "Path to a directory containing sample JSON catalogs (files must end with .json)"]
+               ["-R" "--reports REPORTS" "Path to a directory containing sample JSON reports (files must end with .json)"]
+               ["-A" "--archive ARCHIVE" "Path to a PuppetDB export tarball. Incompatible with -C, -F, -R, or -D"]
+               ["-i" "--runinterval RUNINTERVAL" "What runinterval (in minutes) to use during simulation"]
+               [nil "--from-classpath FROMCLASSPATH" "Whether to retrieve files from the classpath"
+                :default true
+                :parse-fn kitchensink/parse-bool]
```
wkalt: won't this effectively be always true now? I wonder if that'll mess up imports with -F, -C, and -R. I don't understand making it a CLI option rather than inferring it from the absence of those three params.

wkalt:
done with a first pass. I'll take another look when the outstanding stuff is fixed.

src/puppetlabs/puppetdb/cli/benchmark.clj
ajroetker: Hmm I see your point let me dig

ajroetker:
@wkalt Think I addressed all your comments, thanks for looking!

wkalt:

kbarber:

src/puppetlabs/puppetdb/config.clj
mullr: ... it strips the namespaces off of namespaced keywords? To what end? That's very odd. 

In any case, I've just introduced some code in pdb-ext that `throw+`s a ::configuration error, so we would have to update that as well.

ajroetker: Presumably for exactly the purpose of not needing to know where the `::cli-error` originated, i.e. so that multiple namespaces could use the interface. I'll submit an child PR over there!

mullr: I think a non-namespaced keyword would be better in that case.

ajroetker: Fwiw I do think we should eventually update trapperkeeper to handle configuration errors as well, but I think it's out of the scope of this patch. We could have similar validation for configs as we do for cli's, in fact exactly the same for the most part (required items, parse-functions, validation functions).

ajroetker: @mullr I think "maybe" it could be useful if you had a wrapper that also did something with the error depending on the namespace, but I haven't seen that in any of the projects, so....

src/puppetlabs/puppetdb/cli/anonymize.clj
```diff
@@ -182,7 +181,7 @@
   [tar-writer :- TarGzWriter
    suffix :- [String]
    contents :- {s/Any s/Any}]
-  (add-tar-entry tar-writer {:file-suffix suffix
+  (utils/add-tar-entry tar-writer {:file-suffix suffix
                              :contents (json/generate-pretty-string contents)}))
```
wkalt: indentation

ajroetker:

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -191,64 +185,86 @@
...
             (catch Exception e
               (log/errorf "[%s] failed to submit catalog: %s" host e)))))
       (when report
         (future
           (try
-            (client/submit-report base-url 5 (json/generate-string report))
+            (some->> report
+                     update-report
+                     json/generate-string
+                     (client/submit-report base-url 5))
```
wkalt: is some->> really necessary since this is already wrapped in when? Also, minor point but I don't really understand the rationale or agree with the aesthetics of blowing all these out into threading calls. Seems like the let statement above where they already were was a natural place to have the update calls, and I don't think the submit-X calls really screamed for a change. Also not sure how it relates to the objective of the PR -- seems like it's just unrelated refactoring.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -191,64 +185,86 @@
...
+                 (client/submit-facts base-url 4)) 
             (log/infof "[%s] submitted factset" host)
             (catch Exception e
               (log/errorf "[%s] failed to submit factset: %s" host e)))))
       (assoc state
-        :lastrun clock
-        :catalog catalog))))
+             :lastrun clock
+             :factset factset
+             :catalog catalog))))
```
wkalt: I think this change is appropriate because I think it more closely mirrors real life, but we should verify that it's really what we want because it'll change the way the benchmark tool does factsets. Previously mutations on factsets presented as "aberations" from a baseline, whereas this turns it into more of a continuous mutation where the baseline is never revisited. I think the effect will be less contention on factset submission because previously a mutated factset submission would always be followed by another change relative to the mutated state, whereas now that change will only occur rp% of the time.

Do we have a good reason not to be doing this for reports as well?

wkalt: I see why now -- the report update doesn't do any mutation or take rp into effect at all. That's a bug, pretty sure there's an existing ticket.

wkalt: https://tickets.puppetlabs.com/browse/PDB-1233

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -117,66 +113,64 @@
...
-  "Randomizes a fact leaf based on a percentage provided with `rp`."
-  [rp leaf]
-  (if (< (rand 100) rp)
-    (cond
-     (string? leaf) (random-string (inc (rand-int 100)))
-     (integer? leaf) (rand-int 100000)
-     (float? leaf) (* (rand) (rand-int 100000))
-     (ks/boolean? leaf) (random-bool))
-    leaf))
+  "Randomizes a fact leaf/"
```
wkalt: doc string got mangled

wkalt:
@ajroetker this appears to break benchmark when the -i flag is used

'''
lein run benchmark -c postgres.ini -n 100 -F resources/puppetlabs/puppetdb/benchmark/samples/facts/ -i 1
'''

'''
java.lang.ClassCastException: clojure.lang.MapEntry cannot be cast to clojure.lang.Agent
'''

ajroetker:

ajroetker:
@wkalt fixed for your comments

src/puppetlabs/puppetdb/config.clj
rbrw: Oh, and FWIW, I was imagining, for example, that a configuration-error (assuming that really meant config-file) might include :file and :line number members, and the catch might do something smart with them -- though perhaps you could also handle those as optional members of some shared map.

mullr: That would be downright useful. The right way into this would probably be to have the config parser (in trapperkeeper) annotate the parsed config with file/line metadata, like the clojure reader.

rbrw: Filed this related ticket: https://tickets.puppetlabs.com/browse/PDB-1628

kbarber:
This is a massive change, and very close to release. I'm a little wary of doing this kind of thing right now.

ajroetker:

wkalt:
waiting on the most recent tests, but I've been up and down this a few times and as of last week I was satisfied that the code worked fine. My only comment is that I think review was significantly complicated (and ate up a lot more of my time) in this case by refactors and style changes outside the scope of the objective stated in the commit message and outside of code common to that objective. I think these changes would have been better suited for a separate PR (not even a separate commit.) If it'd be feasible to break them out I'd welcome that, but my guess is that it's all pretty intertwined at this stage.

That aside, I'm satisfied at this point -- probably time for someone else to have a look.

PR #1460, opened 2015-06-15
(maint) Remove wrap-with-globals

ajroetker:
This commit remvoes wrap-with-globals and just passes globals down
through the query-app instead, `assoc`ing the `api-version` at the
top-level so there is no longer a need to have separate files for every
version of the api as most of that logic has been removed from PuppetDB
by now anyways. This change means functions no longer need to rely on
middleware acting on the requests to use data that could be simply
passed through parent functions.


ajroetker:

ajroetker:

wkalt:

src/puppetlabs/puppetdb/http/server.clj
```diff
@@ -30,9 +42,30 @@
     (format "The %s API has been retired; please use v4" version)
     404)))
 
-(def routes
-  (app
-   ["v4" &] {:any v4-app}
+(defn query-app
```
wkalt: how is this going to evolve when new endpoints are added for v5? stick the endpoints in here and put the v4 rejection logic in the relevent http namespace?

src/puppetlabs/puppetdb/http/fact_names.clj
```diff
@@ -3,25 +3,28 @@
             [puppetlabs.puppetdb.query.paging :as paging]
             [puppetlabs.puppetdb.jdbc :refer [with-transacted-connection]]
             [net.cgrand.moustache :refer [app]]
-            [puppetlabs.puppetdb.middleware :refer [verify-accepts-json validate-query-params wrap-with-paging-options validate-no-query-params]]
+            [puppetlabs.puppetdb.middleware
+             :refer [verify-accepts-json validate-query-params wrap-with-paging-options validate-no-query-params]]
```
wkalt: this is getting long -- maybe worth using an alias at this point instead of referring the functions themselves

src/puppetlabs/puppetdb/http/nodes.clj
```diff
@@ -27,34 +27,27 @@
...
-    (-> (fn [{:keys [globals]}]
-          (node-status version
-                       node
-                       (:scf-read-db globals)
-                       (:url-prefix globals)))
-        ;; Being a singular item, querying and pagination don't really make
-        ;; sense here
-        (validate-query-params {}))}
+   {:get (-> (node-status api-version node scf-read-db url-prefix)
+             constantly
```
wkalt: very minor point of preference, but I find the threading through constantly to require a bit of a double take compared to wrapping the node-status call at the top level. There's a couple spots where we're doing it in here -- take it or leave it.

wkalt:
I'm generally +1 on this, aside from a few minor comments. Seems like a good change -- just need to make sure we're leaving space to cleanly integrate the v5 API. I think this already leaves us in a better spot than we were.

src/puppetlabs/puppetdb/http/server.clj
ajroetker: That's what I was imagining, is there something wrong there?

src/puppetlabs/puppetdb/http/fact_names.clj
ajroetker: sounds good, `mid` is what we use elsewhere

src/puppetlabs/puppetdb/http/nodes.clj
ajroetker: yeah I went back and forth on that, I find the `(fn [_]` bloats things for me, but I can change them if you feel strongly

src/puppetlabs/puppetdb/http/server.clj
ajroetker: Also from our conversation this morning it would seem that we're already headed in a direction which is a single endpoint handler for everything and dispatch for a path happens from that endpoint.

wkalt: nothing specifically wrong, but it's nice to be able to see at a glance what endpoints are supported on what AAPI versions. Changing it this way would mean you would have to dig separately into each http namespace to see that information (although if we combine those namespaces the way we discussed today maybe that concern goes away.) Not sure that I have a suggestion that would be less clunky than what we had before. I think what you have is alright.

ajroetker:

src/puppetlabs/puppetdb/http/nodes.clj
wkalt: I actually just meant

```
(-> (constantly (node-status ...
```
although using fn or # you could probably get it all on one line. I also just discovered this, but the semantics aren't the same between (fn []) and constantly. Don't think it'll bite us here but it's a good thing to know

```
puppetlabs.puppetdb.core=> (def x 10)
#'puppetlabs.puppetdb.core/x
puppetlabs.puppetdb.core=> (def foo (constantly x))
#'puppetlabs.puppetdb.core/foo
puppetlabs.puppetdb.core=> (def bar (fn [] x))
#'puppetlabs.puppetdb.core/bar
puppetlabs.puppetdb.core=> (foo)
10
puppetlabs.puppetdb.core=> (bar)
10
puppetlabs.puppetdb.core=> (def x 11)
#'puppetlabs.puppetdb.core/x
puppetlabs.puppetdb.core=> (foo)
10
puppetlabs.puppetdb.core=> (bar)
11
```

ajroetker:
@wkalt changed the constantly

ajroetker:

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -293,10 +285,10 @@
   "Middleware that checks the parent exists before serving the rest of the
    application. This ensures we always return 404's on child paths when the
    parent data is empty."
-  [app version parent id]
-  (fn [{:keys [globals] :as req}]
-    (let [{:keys [scf-read-db url-prefix]} globals]
-      ;; There is a race condition here, in particular we open up 1 transaction
+  [app globals parent id]
+  (let [{:keys [scf-read-db url-prefix api-version]} globals]
+    ;; There is a race condition here, in particular we open up 1 transaction
```
senior: Why is the first line of this comment moved above the fn?

senior:
I'm on the fence with this PR, I'm not seeing a clear benefit of the refactor, seems more like a lateral step. To summarize the change, we've gone from associng globals to the request map via middleware to passing globals to the routes directly. I can see benefits to both approaches. As far as the consolidation of routes into a single ns, I think that's good but isn't necessarily linked to the removal of wrap-with-globals.

Is there something I'm missing with the refactor @ajroetker? I'm not opposed to this PR going in, but would like to hear some other opinions /cc @wkalt @mullr @kbarber @rbrw.

ajroetker:
@senior I'm interested what the benefits are from associng globals to the request map. I know that because we've done that, some functions that could exist in between our routes functions and the endpoint code, has to get pushed done to the endpoint code so that it has access to the request.

mullr:
The middleware / request map pattern is a pretty good design, one that we seem to have converged on for handling web requests in dynamically typed languages (even in the static world, people sometimes use type-y versions of the same). While I have no love for putting most of our data inside the map under a key called 'globals', I *do* think that the request map is the right place for such data. It may, for example, but appropriate to put our database connections under `[:db :read]` and `[:db :write]`. Or perhaps the map should be flat, for ease of destructuring. But I agree with @ajroetker 's implicit assessment that the 'globals' map is a code smell. 

If we're not going to pass down data in the request map, the other apparent design would be to move away from a map parameter and towards explicit function arguments. There are some clear benefits to this; if a function takes only the parameters it needs, then it can be more directly used in more contexts. But we need to balance this with the ease of actually *doing* the composition; having a common request map means that we need only one kind of function, where a fully decomposed design may require a dozen. So I don't think we should do this. 

This patch feels like an incomplete step in the direction of functional decomposition. But I think what we ended up with is just a slightly altered version of what we were already doing with requests maps; that is, it's another instance of the passing-around-a-context-map pattern, except now there are two of them. I would much rather see us stick to the one data structure, with more functions that operate on it, and work on cleaning it up.

ajroetker:
@mullr What are the two context-map's you mention in your comment?

mullr:
The request map, and now globals.

ajroetker:
I think a good example of when this patch is useful is here https://github.com/puppetlabs/puppetdb/pull/1498/files As it stands if we wanted to put a warning associated with anything associated with "facts" say, that also needed access to the product-name, we'd need to duplicate that code across every `get` and `post` clause of the moustache app, instead of wrapping it around the app handler. This doesn't seem right to me. We've seen similar things with the wrap-with-parent check where we have to dig into the endpoint handler instead of the report app for doing the parent-check because we need access to the database. I disagree that we should be passing around a single data-structure which is both the globals and the request map, mostly because nowhere in the code do we use the map as such, look at _every_ endpoint, it immediately destructures globals and passes it on.

I don't see any tangible benefit of associng globals to the request map, if you guys have a concrete use case, I'm all ears.

wkalt:
I'm with @AJ here. I don't understand why passing that common data through explicit function arguments is "the other apparent design" any more than his solution here, and I think introducing it higher than in the http routes will allow us more generality in the future when it comes to attributes shared by subsets of endpoints.

ajroetker:
Also coming at this codebase from a naive perspective, I

ajroetker:
Also I remember tripping over this code a few times when I was first looking at this code, it's not intuitive or functional to have a map with our `db-spec` that "disappears" from the code-path after it gets processed by the config.

senior:
I took a look at what globals looks like today (found [here](https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/cli/services.clj#L321))
and it's a dumping ground! I was surprised that it's grown so large. Definitely agree on that being a code smell.

As far as tangible benefits to associng globals, the example I'd point out is specifically related to databases. Being able to manage resources (i.e. a database connection or a transaction) at the per
request level is very useful. Right now that code is scattered throughout the codebase. Similarly if we have global mutable state (right now I think we only have connection pools) including an
immutable version of that in the request is an easy way to narrow where that mutability spreads.

I understand the concerns you've got on this @ajroetker and @wkalt. I think a lot of
this stems from globals having stuff in there that doesn't belong in there. I think I've found the root of it:

```
{:scf-read-db read-db
 :scf-write-db write-db
 :authorizer authorizer
 :update-server update-server
 :product-name product-name
 :url-prefix url-prefix
 :discard-dir (.getAbsolutePath discard-dir)
 :mq-addr mq-addr
 :mq-dest mq-endpoint
 :mq-threads threads
 :catalog-hash-debug-dir catalog-hash-debug-dir
 :command-mq {:connection mq-connection
                           :endpoint mq-endpoint}}
```

There's 12 things in that map. The following properties are basically just plucked out of config (or a def in services) and slapped directly into globals: update-server, product-name, url-prefix, mq-addr,
mq-endpoint, mq-threads, catalog-hash-debug-dir. So that's over half the items in globals, no changes, just passing through the results of parsing/defaulting the config. And we're attaching this to every web
request! The discard-dir, authorizer and command-mq is very similar. Basically it takes config, munges it slightly and puts the result of that in globals. So looking at globals more closely, in my
opinion, the only thing I think should be in there is the connection pools.

There's a definite smell here. I think the reason for the smell is related to the configuration component of TK. The other services (such as MQ) need to have the defaulted config of PuppetDB. So it can't just reach into the configuration component and get what it needs. It needs our config munging code to run on the config first (defaulting the number of MQ listening threads for example). The hack is to just do that in the PDB service and slap the result of it into the context so others can get at it. Seems totally silly to be attaching the static "product-name" configuration option to every request. Was just a quick way for other components to get access to it.

So I think the instinct here is good. Seems to me like the best refactor here would be to extend the configuration management component to do this defaulting and validation. At that point, all but the connection pools in globals goes away.

ajroetker:
I see what you're saying here @senior, I still don't quite understand what the benefit of assoc'ing the connection-pool to the request vs. passing the connection-pool to the handler is. It seems like if a handler isn't getting the db-pool from it's environment but from each request, that parent-checks and sanity checks for the handler lead to more complicated versions of our middleware, because the middleware needs to also be a handler which does work on the request, instead of being simply middleware that does validation between the request and the handler.

You're right though about a different refactor here, first we should cleanup globals and then we can decide on the more controversial points in a subsequent issue. I'll close this and make a tech-debt ticket related to this.

PR #1463, opened 2015-06-17
(PDB-1629) fix relationships on resource aliases

wkalt:
This corrects a bug where the terminus would scan for resource alias names
using the wrong key, which would cause the list of aliases to fall through to
an empty list.

Correcting this made it clear that we were also incorrectly assuming that
aliases within resources presented as arrays.

### DIFF
--- a/acceptance/tests/catalog/replace_catalog_debugging.rb
+++ b/acceptance/tests/catalog/replace_catalog_debugging.rb
@@ -53,7 +53,7 @@ MANIFEST
   Log.notify "Clearing the catalog hash debugging directory: #{catalog_hash_dir}"
 
   on database, "rm -rf #{catalog_hash_dir}*"
-  
+
   run_agents_with_new_site_pp(master, second_run_manifest)
 
   sleep_until_queue_empty database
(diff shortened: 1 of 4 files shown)


acceptance/tests/catalog/replace_catalog_debugging.rb
```diff
@@ -92,5 +92,33 @@
...
+      alias=>bar,
+    }
+    service{ baz:
+      ensure => running,
+      require => Package[bar]
+    }
+
+MANIFEST
+
+    run_agents_with_new_site_pp(master, manifest, {}, [0,2,4])
```
kbarber: I would have thought the detailed exit codes were enough to analyze if this passed or not, the remaining test to check events would then become unnecessary. Also, optimally if we could run this with noop, an exit code of 0 or 2 would be enough. I bet now a 4 will always occur, as the package doesn't exist.

For the ref for detailed exist codes, look here:

http://docs.puppetlabs.com/references/4.1.0/man/agent.html

Passing --noop into the extra_cli_args is probably the trick.

wkalt: @kbarber it seems like the el5 hosts are throwing a 6 here -- apparently signifies changes and failures. Any idea why there'd be a difference?

kbarber: @wkalt not ... really ... I guess why I was suggesting the noop approach, would make it never fail - since it's a noop. ie. the problem might go away. Also it seems more correct anyway?

kbarber: Point being, I'm more surprised it's not a 6 on everything.

wkalt: I'll give it a shot

ajroetker:
Tests are passing ^

kbarber:
@wkalt we'll need to re-open this on the 2.3.x branch.

wkalt:
reopened on 2.3.x

PR #1464, opened 2015-06-17
(PDB-1632) Transfer old AMQP messages at startup

rbrw:
At startup, transfer any messages found in the old
com.puppetlabs.puppetdb.commands queue to puppetlabs.puppetdb.commands
so that pending commands won't be ignored after an upgrade.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -222,6 +222,21 @@
   [max-frame-size url]
   (format "%s&wireFormat.maxFrameSize=%s&marshal=true" url max-frame-size))
 
+(defn- transfer-old-messages! [connection]
+  (let [[pending exists?]
+        (try+
+         [(mq/queue-size "localhost" "com.puppetlabs.puppetdb.commands") true]
+         (catch [:type ::mq/queue-not-found] ex [0 false]))]
+    (when (pos? pending)
+      (log/infof "Transferring %d commands from legacy queue" pending)
+      (let [n (mq/transfer-messages! "localhost"
+                                     "com.puppetlabs.puppetdb.commands"
+                                     mq-endpoint)]
+        (log/infof "Transferred %d commands from legacy queue" n)))
+    (when exists?
+      (mq/remove-queue! "localhost" "com.puppetlabs.puppetdb.commands")
+      (log/info "Removed legacy queue"))))
+
 (defn start-puppetdb
   [context config service add-ring-handler get-route shutdown-on-error]
   {:pre [(map? context)
@@ -302,6 +317,7 @@
                              (service-id service)
                              #(maybe-check-for-updates product-name update-server read-db)
                              #(stop-puppetdb % true))))]
+      (transfer-old-messages! mq-connection)
       (let [app (->> (server/build-app globals)
                      (compojure/context url-prefix []))]
         (log/info "Starting query server")
--- a/src/puppetlabs/puppetdb/mq.clj
+++ b/src/puppetlabs/puppetdb/mq.clj
@@ -5,9 +5,11 @@
            [javax.jms Connection Message TextMessage BytesMessage Session]
            [org.apache.activemq ActiveMQConnectionFactory]
            [org.apache.activemq.pool PooledConnectionFactory])
-  (:require [puppetlabs.puppetdb.schema :refer [defn-validated]]
+  (:require [clojure.java.jmx :as jmx]
             [clojure.tools.logging :as log]
-            [schema.core :as s]))
+            [puppetlabs.puppetdb.schema :refer [defn-validated]]
+            [schema.core :as s]
+            [slingshot.slingshot :refer [throw+]]))
 
 (defn- set-usage!*
   "Internal helper function for setting `SystemUsage` values on a `BrokerService`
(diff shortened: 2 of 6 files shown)


ajroetker:
Will this require some dashboard changes as well?

wkalt:
I think @ajroetker is right about that. Looks like the test failures may be caused by querying for inaccessible metrics:
https://github.com/puppetlabs/puppetdb/blob/master/acceptance/helper.rb#L562-562

wkalt:
this doesn't seem to solve the root issue yet -- still can't see messages getting transferred on upgrade

ajroetker:
We should also probably have some kind of test for this behavior as part of this PR

rbrw:
Reworked following @senior's suggestion to try JMX commands.

src/puppetlabs/puppetdb/mq.clj
```diff
@@ -264,6 +266,42 @@
...
+   destination))
+
+(defn-validated remove-local-queue!
+  "Deletes the named local queue.  Throws
+  {:type ::queue-not-found :name name} when the queue doesn't exist."
+  [name :- s/Str]
+  ;; When the queue doesn't exist, removeQueue provides no indication.
+  (jmx/invoke-signature local-broker-mb
+                        :removeQueue ["java.lang.String"] name))
+
```
rbrw: Do we want any unit tests for these right now?  Related: https://tickets.puppetlabs.com/browse/PDB-1634

wkalt:
+1, verified it works

rbrw:
Now with (at least some) tests!

test/puppetlabs/puppetdb/mq_test.clj
```diff
@@ -109,8 +110,34 @@
...
                            #"status 404"
                            (svc-utils/current-queue-depth))))))
+
+(deftest queue-operations
+  (with-test-broker "test" conn
+    (remove-queue! "test" "foo")
+    (is (thrown+? [:type ::mq/queue-not-found] (queue-size "test" "foo")))
+    (send-message! conn "foo" "1")
+    (send-message! conn "bar" "1")
+    (bounded-drain-into-vec! conn "foo" 1)
```
wkalt: is the idea here to test that bounded-drain-into-vec successfully empties the queue? If so do we need another check above this to ensure that the message made it in to start with?

rbrw: No, this was just done so that we know that the message reached the queue (i.e. so we don't have to poll or similar).

wkalt: makes sense to me now

wkalt:
looks good to me

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -222,6 +222,21 @@
   [max-frame-size url]
   (format "%s&wireFormat.maxFrameSize=%s&marshal=true" url max-frame-size))
 
+(defn- transfer-old-messages! [connection]
+  (let [[pending exists?]
```
senior: Looks like we might be able to use [this](http://activemq.apache.org/maven/5.10.0/apidocs/org/apache/activemq/broker/jmx/BrokerViewMBean.html#getQueues()) JMX call to get a list of queues that the broker knows about. Then we wouldn't need to infer existence based on queue size (and handle the errors associated with that etc).

test/puppetlabs/puppetdb/mq_test.clj
```diff
@@ -109,8 +110,34 @@
...
 
 (deftest test-jmx-enabled
-  (svc-utils/without-jmx
+  (without-jmx
    (svc-utils/with-puppetdb-instance
      (is (thrown-with-msg? clojure.lang.ExceptionInfo
                            #"status 404"
                            (svc-utils/current-queue-depth))))))
+
+(deftest queue-operations
```
senior: I think this test should be broken up into separate tests for the various operations. With all the operations being tested at once, I find it hard to follow and separate what's being tested in my head.

src/puppetlabs/puppetdb/mq.clj
```diff
@@ -264,6 +266,45 @@
...
+                        :moveMatchingMessagesTo
+                        ["java.lang.String" "java.lang.String"]
+                        "JMSMessageID is NOT NULL"
+                        destination))
+
+(defn-validated remove-queue!
+  "Deletes the named broker queue.  Does nothing if the queue doesn't exist."
+  [broker :- s/Str
+   queue :- s/Str]
+  ;; When the queue doesn't exist, removeQueue provides no indication.
```
senior: Is this comment a restatement of the docstring?

src/puppetlabs/puppetdb/cli/services.clj
senior: Looking at this again, I don't know that I'd change it to the existence call. I can see wanting to log the messages there before the messages have been moved.

src/puppetlabs/puppetdb/mq.clj
rbrw: Practically, I suppose that's correct, though I'd left it because I had to figure it out by testing, and thought it might save someone else time (i.e. so they don't think it's possible, but we just chose to ignore the return value or something).  In any case, I'm happy to remove it.

test/puppetlabs/puppetdb/mq_test.clj
rbrw: Crossed my mind too.  I'll fix it.

wkalt:

PR #1472, opened 2015-06-19
WIP (RE-4779) Migrate config files and associated data on upgrade

melissa:


resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+
+migration_command="puppetdb config-migration"
+
+#############
+# FUNCTIONS #
+#############
+
+# Backs up a file, if it hasn't already been backed up
+#
+# $1 - file to backup
```
mckern: I'd expand this into a clearer usage statement. Here's an example, but feel free to disregard it:

```shell
config-migration.sh takes one argument:
config-migration <path to file to backup>
```

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+    *)
+      usage ;;
+  esac
+done
+
+${force:=false}
+
+set -e
+
+orig_config_dir=/etc/puppetdb/conf.d
```
stahnma: Is this designed only for FOSS?

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+#
+# Example:
+#
+#   backupfile "/etc/myconfig"
+#
+function backupfile() {
+  # Create the global array if it doesn't already exist
+  if [ -z "$backupfile_list" ]; then
+    backupfile_list=()
+  fi
```
mckern: Groooooossssssssssssssssssssssssssssss

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+# $2 - item to search for
+#
+# Example:
+#
+#   myarray=('element1', 'element2')
+#   if contains ${myarray[@]}, "element1'; then
+#     echo "element1 exists in the array"
+#   fi
+#
+function contains() {
```
mckern: ...

melissa:
Okay, @kbarber I'm running into a few issues. I can't run this script in the `preinst` since that is before anything is available. I could squash it down and jam it into the preinst, but that's gross. Running it in the `postinst` I'm also running into problems. That is after the original package has been uninstalled. On rpm, the old config files have been removed if they haven't been modified (but stick around as `/etc/puppetdb/conf.d/config.ini.rpmsave` if they have). I could look for files with that suffix, but that won't work for debian. Debian doesn't even care if you've made any changes to the config files, and just leaves them where they were. 

So, we can look for the old files, look for the .rpmsave files if the old files don't exist. However, there is no guarantee that the .rpmsave files are from this migration. So, another option is archiving the old config files in the preinst, and then spitting out a message saying where the archives can be found, and that the person doing the upgrade really should make sure everything's set correctly. I guess we can also combine those, where we archive them in the preinst, and then run the script in the postinst.

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+  # Create the global array if it doesn't already exist
+  if [ -z "$backupfile_list" ]; then
+    backupfile_list=()
+  fi
+
+  # We check the array to make sure the file isn't already backed up
+  if ! contains ${backupfile_list[@]} "$1"; then
+    local backup_path="$1.bak.$(date +%s)"
+    echo "Backing up $1 to ${backup_path} before making changes"
+    cp -p "$1" "$backup_path"
```
rbrw: I'd recommend checking the exit status here (imagine disk full, etc.), and more generally, I'd suggest checking the exit status of every relevant shell command, whether via explicit "if ! cp -p ..." or "set -e".

Though beware "set -e"'s precise behavior, i.e. this command will actually print 13: `(set -e; false; echo 13) || false` and IIRC the same is true for `foo || false` if foo is a function that tries to use set -e in a subshell.  The details are explained here: https://www.gnu.org/software/bash/manual/bashref.html#The-Set-Builtin

I'd also suggest considering use of "set -o pipefail" (with or without "set -e"), but we'd then need to be careful about situations where a non-zero exit status in a pipeline shouldn't be considered an error (i.e. grep in something like `foo | grep 'might or might not be found' | wc -l`).

Oh, and you may well already know this, but another thing that's bitten me in the past is the behavior of statements like `local foo=$(bar)` because at the time I hadn't realized that the `local` exit status is completely independent of the subcommand's exit status, i.e. after `local foo=$(false)`, the exit status variable $? will be 0.  I believe the same may be true for the other declarations as well.  To avoid that I'll often just handle the declarations separately, i..e. `local foo bar; foo=...`.

Finally, "set -u" can be interesting, but of course you have to be willing to explicitly initialize everything in the affected regions.

Hope this helps.

kbarber:
@melissa in prep for the 3 release, we've moved branches, sorry. This will need to be re-opened and targeted at stable now, as this represents the soon-to-be-released 3.0.0.

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+# Example:
+#
+#    replaceline "^$mysetting.*" "mysetting = myvalue" /etc/myconfig
+#
+function replaceline {
+  backupfile "$3"
+  tmp="$3.tmp.$(date +%s)"
+  sed "s/$1/$(echo "$2" | sed -e 's/[\/&]/\\&/g')/g" "$3" > "$tmp"
+  mv "$tmp" "$3"
+}
```
ajroetker: What's the benefit of doing this over an in-place `sed`?

ajroetker: Sorry I see, I was thinking we we're backing up the files in the preinst.

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+
+# Backs up a file, if it hasn't already been backed up
+#
+# $1 - file to backup
+#
+# Example:
+#
+#   backupfile "/etc/myconfig"
+#
+function backupfile() {
```
rbrw: As an alternative, I wonder if you could set backup_suffix=$(...date...) the the top of the script so that the current invocation would have a consistent suffix, and then just rely on the filesystem itself, instead of the state array  i.e. "if ! test -e "$target"; ...".

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+    echo
+    echo "  Afterwards re-run this tool then restart PuppetDB to complete the SSL"
+    echo "  setup:"
+    echo
+    echo "      ${migration_command} -f"
+    exit 1
+  fi
+  if [ -e "${destination}/${orig_file##*/}" ] ; then
+    if $force ; then
+      echo "Replacing file: ${destination}/${orig_file##*/} with ${orig_file}"
```
rbrw: Wonder if it might make sense to redirect status messages to stderr, i.e. `echo "some message" 1>&2`

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+# This expects various environment variables to have already been set to work.
+#
+# Note: This does not respect directory structures, and will move files blindly
+function migrate_files() {
+  destination=$1
+  orig_file=$2
+  if [ ! -e "$orig_file" ]; then
+    echo "Warning: Unable to find all ini configuration files to copy"
+    echo
+    echo "  This tool requires the following files to exist:"
```
rbrw: Looks like there's only one file possible, so perhaps change this to "file to exist", but I might also consider just changing it to "requires the following path to exist", and then pull out the message into a separate function if that means we could share it with the similar message below.

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+    echo "  certificates may not have been signed yet, probably due to the"
+    echo "  lack of a complete Puppet agent run. Try running puppet first, for"
+    echo "  example:"
+    echo
+    echo "      puppet agent --test"
+    echo
+    echo "  Afterwards re-run this tool then restart PuppetDB to complete the SSL"
+    echo "  setup:"
+    echo
+    echo "      ${migration_command} -f"
```
rbrw: And you might have already rejected this, but (excepting the annoyance of typing tab chars), you might consider a here document (particularly the -FOO variety):
```
    cat 1>&2 <<-EOF
	These files may be missing due to the fact that your host's Puppet
	certificates may not have been signed yet, probably due to the
	...
	EOF
```
Though the tabs are aggravating, so I'm not sure what I'd actually do...

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,243 @@
...
+    h)
+      usage ;;
+    *)
+      usage ;;
+  esac
+done
+
+${force:=false}
+
+set -e
```
rbrw: You might have a good reason not to, but if not, I'd suggest putting this at the top.

melissa:
@kbarber I wanted to ask, we're not shipping PuppetDB 3.x to the PC1 repos. It'll be shipped to nightlies and to PC2 once that's up. Since both of these requires users consciously opt in to upgrading pdb, are we still as worried about automating the upgrade of config files? This is definitely more complicated than I was thinking. There are a lot of hoops to jump through to get this to work with packaging.

melissa:
Closed in favor of https://github.com/puppetlabs/puppetdb/pull/1485 against the stable branch

PR #1473, opened 2015-06-22
(PDB-1646) tighten up regex and subqueries in the legacy query engine

wkalt:
This fixes a couple issues with the legacy query engine, which is used when
the distinct_resources parameter is passed as true:
* regex queries were broken due to use of honeysql instead of strings
* subqueries were broken because of an ambiguous column alias

I've expanded the distinct_resources testing in http/event-counts-test to cover
the fixed behavior.


kbarber:
@wkalt will need to be re-opened and targeted at stable now.

wkalt:
reopened on stable

PR #1474, opened 2015-06-22
(PDB-1629) fix relationships on resource aliases

wkalt:
This corrects a bug where the terminus would scan for resource alias names
using the wrong key, which would cause the list of aliases to fall through to
an empty list.

Correcting this made it clear that we were also incorrectly assuming that
aliases within resources presented as arrays.

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -1049,7 +1049,7 @@ PP
     remote_path
   end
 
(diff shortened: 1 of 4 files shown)


wkalt:

senior:
Tested locally on Puppet 3 and Puppet 4 :+1:

PR #1475, opened 2015-06-22
(PDB-1646) tighten up regex and subqueries in the legacy query engine

wkalt:
This fixes a couple issues with the legacy query engine, which is used when
the distinct_resources parameter is passed as true:
* regex queries were broken due to use of honeysql instead of strings
* subqueries were broken because of an ambiguous column alias

I've expanded the distinct_resources testing in http/event-counts-test to cover
the fixed behavior.

### DIFF
--- a/src/puppetlabs/puppetdb/query.clj
+++ b/src/puppetlabs/puppetdb/query.clj
@@ -62,10 +62,14 @@
    applying ordering constraints."
   (:require [clojure.string :as str]
             [puppetlabs.kitchensink.core :as kitchensink]
+            [puppetlabs.puppetdb.jdbc :as jdbc]
+            [puppetlabs.puppetdb.honeysql :as h]
+            [puppetlabs.puppetdb.utils :as utils]
             [puppetlabs.puppetdb.time :refer [to-timestamp]]
             [puppetlabs.kitchensink.core :refer [parse-number keyset valset order-by-expr?]]
             [puppetlabs.puppetdb.scf.storage-utils :as sutils
-             :refer [db-serialize sql-as-numeric sql-array-query-string sql-regexp-match sql-regexp-array-match]]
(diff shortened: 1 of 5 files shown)


wkalt:
transient failure ^

wkalt:
rebased

src/puppetlabs/puppetdb/query.clj
```diff
@@ -422,21 +434,23 @@
           (:where %)]}
   (match [path]
          ["tag"]
-         {:where (sql-regexp-array-match "catalog_resources" "catalog_resources" "tags")
+         {:where (.s (sql-regexp-array-match "catalog_resources"
```
senior: This causes a reflection warning and I was confused as to what a .s was. There's a similar function that converts a SqlRaw into a keyword, I think a type hinted function there would make this clearer and get rid of the reflection.

src/puppetlabs/puppetdb/query.clj
```diff
@@ -422,21 +434,23 @@
...
+                                             "catalog_resources"
+                                             "tags"))
           :params [value]}
 
          ;; node join.
          ["certname"]
-         {:where  (sql-regexp-match "catalogs.certname")
+         {:where (legacy-sql-regexp-match "catalogs.certname")
```
senior: oh I get it, so we were returning honeysql here, but never converting the honeysql into a string?

senior:
:+1: from me other than the minor comment above

wkalt:
@senior added a type hint. I'm not sure how the reflection warning was being exposed -- PDB wasn't logging it with my config. Can you see if the issue got fixed?

src/puppetlabs/puppetdb/query.clj
wkalt: exactly

senior:
Looks good to me, will merge when tests pass and the pipeline clears.

PR #1477, opened 2015-06-22
(maint) Repair broken mq test

ajroetker:
This commit fixes a broken mq test where there was an `(= ...)`
expression that wasn't wrapped around its second argument. This failed
silently with a normal `lein test` but caused failures with the
humane-test-output plugin for `lein`. This commit also removes some
unnecessary commas and makes sure to cleanup the `somedir` created
during the mq-tests.


rbrw:
Thanks, glad you caught the bug (and the commas -- copy-paste issue).  I might consider breaking this into two patches (in the same PR), but I don't feel strongly about it.  Regardless, looks good to me.

PR #1479, opened 2015-06-22
(PDB-1590) Support two previous versions of all commands (for upgrades)

senior:
Note these previous versions are not supported at the API level. The
intent here is if the user has upgraded from PuppetDB 2.0.0 to 3.0.0 and
there are still persisted messages in the queue (POSTed, but not
processed) those messages should be consumable by PuppetDB 3.0.0.

This commit takes a previous version of the command and upgrades it,
filling in anything missing, changing the names of keys as necessary to
conform it to the current (supported) command version.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -76,7 +76,8 @@
                  [fast-zip-visit "1.0.2"]
                  [robert/hooke "1.3.0"]
                  [honeysql "0.5.2"]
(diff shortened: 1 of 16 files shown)


wkalt:
this is confirmed and was previously reviewed and approved of -- will merge when the pipeline is clear

PR #1480, opened 2015-06-23
(PDB-1652) support include_total with distinct events

wkalt:
This fixes a regression from the aggregate-event-counts optimization patch.
The issue was that the CTE added there for queries with distinct_resources=true
was not being applied to the SQL on which the include_total query was based,
in the case when the include_total param was passed.

include_total has never been supported for event-counts or aggregate-event-counts,
so events is the only thing this will affect and there's no concern about missing
the other endpoints.


wkalt:
@plenkinsro retest this please

wkalt:
@pljenkinsro enkinsro retest this please

wkalt:

senior:
:+1: will merge when the test passes and stable clears

PR #1484, opened 2015-06-23
(packaging) Update to ezbake 0.3.10-SNAPSHOT for service account upgr…

shrug:
…ades


kbarber:
@shrug ezbake snapshots aren't shipped to nexus afaik. We'd need more work to switch to a snapshot I think, last time I did it, it required job changes etc. Why do we want to pin against a snapshot here anyway mate?

kbarber:
Also, if this is a 3.0.0 target, then stable is the correct branch to raise this on.

shrug:
@kbarber I was planning to hold off for a few days to see if any more bugs crop up, but it's easy enough to cut a release. 

I think we probably do want this for 3.0, so I'll retarget

PR #1485, opened 2015-06-23
WIP (RE-4779) Create a script to migrate config files and any needed data

melissa:


melissa:
@kbarber retargeted at stable. The scripts works, but I can't get it to work as a part of packaging. I know you wrote the ssl-setup script, and this work is largely based off that. How did you get that to work in the postinst? It looks like the script is erroring out right away.

rbrw:
@melissa just noticed a comment from the earlier PR.  Might be old news, but if not, note that Debian will leave the config files if dpkg was invoked (directly or indirectly) with remove, but it will unconditionally delete them if it was invoked with "purge". i.e.

dpkg --remove foo
dpkg --purge foo
apt-get dist-upgrade
apt-get dist-upgrade --purge
etc.

The maintainer scripts can tell what mode they're in via "$1" (IIRC) (Perhaps more than you want to know here: https://wiki.debian.org/MaintainerScripts )

kbarber:
@melissa sorry I missed your message here. What does your final install.sh end up looking like after installation? Can you just call install.sh postinst_redhat and have it work? That is, it should work via the same via package install as that methodology.

kbarber:

melissa:
@kbarber okay. The ssl-setup script was failing, and then killing the script so config-migration was never being called. I've updated the script to set +e for only the postinst so any script failures won't cause everything to fail.

kbarber:
@melissa also, maybe better to do the config migration _before_ ssl-setup actually ... if the migration was successful ssl-setup might need to do nothing, but if the migration hasn't happened yet - ssl-setup is just going to mess things up probably, copy files and modify configuration that just gets written over etc.

kbarber:
@melissa let us know when you want us to dive in and do a review of all of this, it will take me a bit to setup the environment for confirming this etc. so I don't want to invest until you are ready :-).

ajroetker:
@melissa this still is targeted at master and not stable fyi

melissa:
Dangit! Okay, try try again

melissa:
https://github.com/puppetlabs/puppetdb/pull/1496 is targeted against stable

PR #1488, opened 2015-06-24
(PDB-1610) deprecate postgres < 9.4

wkalt:
This patch implements the following policy around deprecation messages

FOSS: messages thrown < 9.4
PE: messages thrown < 9.4 *except* 9.2

### DIFF
--- a/src/com/puppetlabs/puppetdb/cli/services.clj
+++ b/src/com/puppetlabs/puppetdb/cli/services.clj
@@ -260,6 +260,7 @@
   (let [{:keys [jetty database read-database global command-processing puppetdb]
          :as config}                            (conf/process-config! config)
         product-name                               (:product-name global)
+        enterprise?                                (= "pe-puppetdb" product-name)
         update-server                              (:update-server global)
         url-prefix                                 (:url-prefix global)
         write-db                                   (pl-jdbc/pooled-datasource database)
@@ -288,9 +289,10 @@
     ;; confused if the database doesn't exist but we open and close a
     ;; connection without creating anything.
     (sql/with-connection write-db
-                         (scf-store/validate-database-version #(System/exit 1))
+                         (scf-store/validate-database-version
+                           enterprise? #(System/exit 1))
                          (migrate!)
-                         (indexes! (:product-name globals)))
+                         (indexes! product-name))
 
     ;; Initialize database-dependent metrics and dlo metrics if existent.
     (pop/initialize-metrics write-db)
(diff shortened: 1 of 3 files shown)


ajroetker:
@wkalt I think I'd rather see an `puppet_enterprise?` boolean being passed around rather than the product name

wkalt:
@ajroetker yeah I like that better too -- I'll push up a change

src/com/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -288,9 +289,10 @@
     ;; confused if the database doesn't exist but we open and close a
     ;; connection without creating anything.
     (sql/with-connection write-db
-                         (scf-store/validate-database-version #(System/exit 1))
+                         (scf-store/validate-database-version
```
ajroetker: Should we even do this validation for `enterprise?` since we can be sure they're using a version of Postgres which is compatible?

ajroetker: On another note, maybe should move this validation to somewhere before we try starting the service? Like right after we process that section of the config? In the spirit of failing fast. That way we can also use the `cli-error` or `config-error` to do the exiting.

test/com/puppetlabs/puppetdb/test/scf/storage.clj
```diff
@@ -1328,18 +1328,33 @@
                    #'sutils/sql-current-connection-database-version (constantly version)}
     f))
 
-(deftest db-deprecation?
+(deftest db-deprecation
   (testing "should return a string if db is deprecated"
-    (are [db version result]
+    (are [db product-name version result]
```
ajroetker: I think this should be `enterprise?` now :)

ajroetker:
lgtm modulo comments! Definitely makes those tests more readable

src/com/puppetlabs/puppetdb/cli/services.clj
wkalt: if you use an external database you can run PE with an incompatible postgres.

Not sure I understand the second comment -- at the moment we don't process the config (except the tk hook) until we attempt to start the service, and the db validation is more or less the first thing we do. Are you talking about moving the validation into the let statement above under process-config, or moving it into config.clj? This comment gives me pause too FWIW: https://github.com/puppetlabs/puppetdb/blob/2.3.x/src/com/puppetlabs/puppetdb/cli/services.clj#L285-289

ajroetker: Actually we already do this validation in the installer for PE whether you're running external or not.

Also I didn't see this was targeted for `2.3.x`, so nevermind about skipping the validation.

ajroetker:
:+1: From me

PR #1489, opened 2015-06-24
(PDB-1613) Add with-logging-to-atom and with-log-level

rbrw:


test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -31,6 +35,56 @@
...
+         logger-id# (if (class? logger-id#) logger-id# (str logger-id#))
+         logger# (.getLogger (LoggerFactory/getILoggerFactory) logger-id#)
+         original-level# (.getLevel logger#)]
+     (.setLevel logger# (case ~level
+                          :trace Level/TRACE
+                          :debug Level/DEBUG
+                          :info Level/INFO
+                          :warn Level/WARN
+                          :error Level/ERROR
+                          :fatal Level/ERROR))
```
wkalt: these are both at Level/ERROR -- should the last be FATAL? Also I think OFF is supported -- might be worth including for completeness.

test/puppetlabs/puppetdb/testutils/services_test.clj
```diff
@@ -0,0 +1,23 @@
...
+      (tgt/with-logging-to-atom Logger/ROOT_LOGGER_NAME log
+        (log/info "wlta-test"))
+      (is (not-any? #(= {:level "INFO" :message "wlta-test"} %)
+                    (map event->map @log)))))
+  (let [log (atom [])]
+    (tgt/with-log-level Logger/ROOT_LOGGER_NAME :info
+      (tgt/with-logging-to-atom Logger/ROOT_LOGGER_NAME log
+        (log/info "wlta-test"))
+      (is (some #(= {:level "INFO" :message "wlta-test"} %)
+                (map event->map @log))))))
```
wkalt: I don't think this construct will ever produce a failure

```
(deftest foo
  (is (some #(= "bar" %) ["foo"])))
```

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -1,5 +1,6 @@
 (ns puppetlabs.puppetdb.testutils.services
-  (:require [puppetlabs.puppetdb.testutils :refer [temp-dir temp-file]]
+  (:require [puppetlabs.kitchensink.core :as sink]
```
ajroetker: this is `kitchensink` in most of our other namespaces

test/puppetlabs/puppetdb/testutils/services_test.clj
wkalt: ^ that's wrong. For some reason it doesn't fail when I run it in the repl, but it fails with lein test

wkalt:
looks good to me aside from the comment about FATAL

test/puppetlabs/puppetdb/testutils/services.clj
rbrw: Assuming I didn't make a mistake, slf4j doesn't support a fatal level (it's redirected to error) https://github.com/clojure/tools.logging/blob/9b33f19dd07fbf3f1c7f7f8f30ff987b6f51f7cc/src/main/clojure/clojure/tools/logging/impl.clj#L76

test/puppetlabs/puppetdb/testutils/services.clj
rbrw: OK, no problem, though that's awfully long...

test/puppetlabs/puppetdb/testutils/services.clj
wkalt: yeah, looks like it. Fine with me then -- just wanted to check.

rbrw: Oh, and I didn't include OFF because clojure.tools.logging doesn't have that.

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -31,6 +35,56 @@
...
+      LifeCycle
+      (start [this] true)
+      (stop [this] true))))
+
+(defmacro with-logging-to-atom
+  "Conjoins all logger-id events produced during the execution of the
+  body to the destination atom, which must contain a collection.  If
+  logger-id is not a class, it is converted via str."
+  [logger-id destination & body]
+  ;; Specify the root logger via org.slf4j.Logger/ROOT_LOGGER_NAME.
```
mullr: I don't understand this comment

mullr: oh, that's what logger-id is, usually. Seems like that's the common case; maybe have a 2-arity overload with that as the default value?

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -31,6 +35,56 @@
...
+  `(let [logger-id# ~logger-id
+         logger-id# (if (class? logger-id#) logger-id# (str logger-id#))
+         logger# (.getLogger (LoggerFactory/getILoggerFactory) logger-id#)
+         appender# (doto (create-log-appender-to-atom ~destination) .start)]
+     (.addAppender logger# appender#)
+     (try
+       (do ~@body)
```
mullr: don't think `do` is needed here is it?

mullr:
minor commonts, lgtm overall

test/puppetlabs/puppetdb/testutils/services.clj
rbrw: Without it, I believe the body can interfere with the try, i.e. (with-foo ... (catch ...)).

test/puppetlabs/puppetdb/testutils/services.clj
rbrw: Perhaps, though I wondered if the caller might normally have a specific logger in mind (like ":sync")?

Though if using the root is common, I wonder if it might be enough to add a convenience var for the Java name (making it a little less slf4j specific too), instead of testing argument types in the macro.

In any case, I'm fine with whatever the team prefers here.

mullr: Let's watch the usages and aggressively refactor if needed. I'm ok keeping it as is.

rbrw: Another thing I've contemplated is (with-logging-to-atom dest [x y ...] ...), i.e. support multiple overrides in one invocation, but thought maybe we should wait to add that if/when it's needed.

mullr: Yeah, agreed. Happy keeping the simple case to start with.

PR #1490, opened 2015-06-25
(PDB-1631) Create the `/pdb/admin` API with import/export endpoints

ajroetker:
This PR adds a `puppetlabs.puppetdb.admin` namespace.
The `admin` namespace contains a trapperkeeper service which serves a handler of the new import/export endpoints.

The new endpoints are served under the `/pdb/admin/v1` api root prefix.

The `archive` endpoint will stream you the tarball via a `GET`.
The `archive` endpoint will stream the tarball to PuppetDB for import via a `POST` of the archive in a multipart request with the archive as the `archive=` param. (`curl ... -F "archive=@foo_backup"`)

The CLI tools (import/export) functionality has been moved to these endpoints and the tools themselves essentially just make calls to these endpoints.

TODO and followups for this PR:
Anonymization could also be an endpoint, to anonymize a tarball in the backups directory.
There could be an anonymization option when downloading `GET /export`, so that the data is streamed to a user anonymized. 
Move the import/export logic out of the `cli` namespaces, maybe their own namespaces.
More validation around the compojure app.
Counts of the imported objects.
Don't reimport things that already exist.
More granular testing around the endpoints would be nice.
Add a benchmarking endpoint for data-loading, there is code duplication with the import tool there.

### DIFF
--- a/documentation/_puppetdb_nav.html
+++ b/documentation/_puppetdb_nav.html
@@ -76,15 +76,20 @@
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/aggregate-event-counts.html">Aggregate Event Counts Endpoint</a></li>
     </ul>
   </li>
-  <li><strong>Metadata API Version 1</strong>
+  <li><strong>Admin API Version 1</strong>
     <ul>
-      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/version.html">Version Endpoint</a></li>
-      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/server-time.html">Server Time Endpoint</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/admin/v1/archive.html">Archive Endpoint</a></li>
     </ul>
   </li>
   <li><strong>Command API Version 1</strong>
     <ul>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/command/v1/commands.html">Commands Endpoint</a></li>
+    </ul>
+  </li>
+  <li><strong>Metadata API Version 1</strong>
+    <ul>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/version.html">Version Endpoint</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/server-time.html">Server Time Endpoint</a></li>
     </ul>
   </li>
   <li><strong>Metrics API Version 1</strong>
--- a/documentation/api/admin/v1/archive.markdown
+++ b/documentation/api/admin/v1/archive.markdown
@@ -1,0 +1,60 @@
+---
+title: "PuppetDB 3.0 » Admin API » v1 » Importing and Exporting PuppetDB Archives"
+layout: default
+canonical: "/puppetdb/latest/api/admin/v1/archive.html"
+---
+
+[curl]: ../curl.html#using-curl-from-localhost-non-sslhttp
+
+The `/archive` endpoint can be used for importing and exporting PuppetDB archives.
+
+## `POST /pdb/admin/v1/archive`
+
+This endpoint can be used for streaming a PuppetDB archive into PuppetDB.
+
+### Request Format
+
+The request should be a multipart POST and have `Content-Type: multipart/mixed`.
+
+### URL Parameters
+
+* `archive`: Required. The archive file to import to the PuppetDB.
+* `command_versions`: Required. A JSON object mapping PuppetDB command-names to their version. The mapping for a given PuppetDB archive can be found in the archive:
+~~~shell
+        tar -xOf <my-pdb-archive>.tar.gz puppetdb-bak/export-metadata.json
+~~~
+
+### Response Format
+
+The response will be in `application/json`, and will return a JSON map
+
+    {"ok": true}
+
+upon successful completion of the importation.
+
+### Examples
+[Using `curl` from localhost][curl]:
+
+        curl -X POST http://localhost:8080/pdb/admin/v1/archive \
+        -F 'command_versions={"replace_facts":4,"store_report":5,"replace_catalog":6}'\
+        -F "archive=@example_backup_archive.tgz"
+
+    {"ok": true}
+
+## `GET /pdb/admin/v1/archive`
+
+This endpoint can be used to stream a tarred, gzipped backup archive of PuppetDB to your local machine.
+
+### URL Parameters
+
+This endpoint takes no URL parameters.
+
+### Response Format
+
+The response will be a `application/octet-stream`, and will return a `tar.gz` archive.
+
+### Examples
+
+[Using `curl` from localhost][curl]:
+
+    curl -X GET http://localhost:8080/pdb/admin/v1/archive -o puppetdb-export.tgz
--- a/src/puppetlabs/puppetdb/admin.clj
+++ b/src/puppetlabs/puppetdb/admin.clj
@@ -1,0 +1,58 @@
+(ns puppetlabs.puppetdb.admin
+  (:require [clojure.java.io :as io]
+            [clojure.tools.logging :as log]
+            [compojure.core :as compojure]
+            [compojure.route :as route]
+            [puppetlabs.kitchensink.core :as kitchensink]
+            [puppetlabs.puppetdb.archive :as archive]
+            [puppetlabs.puppetdb.cheshire :as json]
+            [puppetlabs.puppetdb.cli.export :as export]
+            [puppetlabs.puppetdb.cli.import :as import]
+            [puppetlabs.puppetdb.http :as http]
+            [puppetlabs.puppetdb.middleware :as mid]
+            [puppetlabs.trapperkeeper.core :as trapperkeeper]
+            [ring.middleware.multipart-params :as mp]
+            [clj-time.core :refer [now]]
+            [ring.util.io :as rio]))
+
+(def query-api-version :v4)
+
+(defn munge-nodes-data
+  [nodes]
+  (as-> nodes $
+    (select-keys $ [:certname :facts_timestamp :catalog_timestamp :report_timestamp])
+    (kitchensink/mapvals str $)))
+
+(defn export-app
+  [buffer query-fn]
+  (export/export! buffer
+                  (->> (query-fn :nodes query-api-version nil nil doall)
+                       (map munge-nodes-data)
+                       (mapcat (partial export/get-node-data query-fn)))))
+
+(defn build-app
+  [submit-command-fn query-fn {:keys [authorizer]}]
+  (-> (compojure/routes
+       (mp/wrap-multipart-params
+        (compojure/POST "/v1/archive" request
+                        (let [{{:strs [archive command_versions]} :multipart-params} request]
+                          (import/import! (:tempfile archive)
+                                          (json/parse-string command_versions true)
+                                          submit-command-fn)
+                            (http/json-response {:ok true}))))
+       (compojure/GET "/v1/archive" []
+                      (http/streamed-tar-response #(export-app % query-fn)
+                                                  (format "puppetdb-export-%s.tgz" (now))))
+       (route/not-found "Not Found"))
+      (mid/wrap-with-puppetdb-middleware authorizer)))
+
+(trapperkeeper/defservice admin-service
+  [[:PuppetDBServer shared-globals query]
+   [:PuppetDBCommand submit-command]
+   [:WebroutingService add-ring-handler get-route]]
+
+  (start [this context]
+         (->> (build-app submit-command query (shared-globals))
+              (compojure/context (get-route this) [])
+              (add-ring-handler this))
+         context))
--- a/src/puppetlabs/puppetdb/cli/export.clj
+++ b/src/puppetlabs/puppetdb/cli/export.clj
@@ -7,11 +7,13 @@
    The command will produce a tarball that can then be used with the companion
    `import` PuppetDB command-line tool to import data into another PuppetDB
    database."
-  (:require [clj-http.client :as client]
+  (:require [clj-http.client :as http-client]
             [clj-http.util :refer [url-encode]]
             [clj-time.core :refer [now]]
+            [clojure.java.io :as io]
             [clojure.set :as set]
             [clojure.string :as str]
+            [clojure.tools.logging :as log]
             [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.archive :as archive]
             [puppetlabs.puppetdb.catalogs :as catalogs]
@@ -24,20 +26,45 @@
             [slingshot.slingshot :refer [throw+ try+]])
   (:import [java.net URL]))
 
-(def ^:private api-version :v4)
+(def admin-api-version :v1)
+(def query-api-version :v4)
+(def ^:private command-versions
+  ;; This is not ideal that we are hard-coding the command version here, but
+  ;;  in our current architecture I don't believe there is any way to introspect
+  ;;  on which version of the `replace catalog` matches up with the current
+  ;;  version of the `catalog` endpoint... or even to query what the latest
+  ;;  version of a command is.  We should improve that.
+  {:replace_catalog 6
+   :store_report 5
+   :replace_facts 4})
+
+(defn query-child-href-internally
+  [query-fn child-href]
+  (let [[parent-str identifier child-str] (take-last 3 (clojure.string/split child-href #"/"))
+        entity (case child-str
+                 "metrics" :report-metrics
+                 "logs" :report-logs
+                 (keyword child-str))
+        query (case parent-str
+                "reports" (case child-str
+                            ("metrics" "logs") ["=" "hash" identifier]
+                            ["=" "report" identifier])
+                ("factsets" "catalogs") ["=" "certname" identifier])]
+    (query-fn entity query-api-version query nil doall)))
+
+(defn maybe-expand-href
+  "Expand the child-value's href if the data key isn't present in the child-value"
+  [query-fn
+   {:keys [href] :as child-value}]
+  (utils/assoc-when child-value :data (query-child-href-internally query-fn href)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Internal Schemas
 
-(def node-map {:catalog_timestamp (s/maybe String)
-               :facts_timestamp (s/maybe String)
-               :report_timestamp (s/maybe String)
-               :catalog_environment (s/maybe String)
-               :facts_environment (s/maybe String)
-               :report_environment (s/maybe String)
-               :certname String
-               :deactivated (s/maybe String)
-               :expired (s/maybe String)})
+(def node-map {:catalog_timestamp String
+               :facts_timestamp String
+               :report_timestamp String
+               :certname String})
 
 (def cli-description "Export all PuppetDB catalog data to a backup file")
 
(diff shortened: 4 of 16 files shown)


ajroetker:
The travis failures are a race condition in the roundtrip tests.

ajroetker:

ajroetker:
@senior stuck on two significant issues.

1) Our middleware doesn't play nice with compojure GET/POST/PUT macro handlers, it will end upp apply the handlers to everything under `compojure/routes`.

2) Our middleware for content types doesn't play nice with multi-part headers that we use for uploading the tarball.

Other than that I've actually migrated the old tools to use the new code.

ajroetker:

ajroetker:

ajroetker:

ajroetker:

wkalt:
I'm taking a look at this

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -10,8 +10,10 @@
   (:require [clj-http.client :as client]
             [clj-http.util :refer [url-encode]]
             [clj-time.core :refer [now]]
+            [clojure.java.io :as io]
```
wkalt: we're using the full name below, rather than this alias.

ajroetker: I suspect there might be a few of those lurking, I used the full name while prototyping and moving things around namespaces. I'll fix it up!

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -275,14 +175,31 @@
...
            utils/validate-cli-base-url!)))))
 
+(defn export!
+  [outfile nodes-data]
+  (with-open [tar-writer (archive/tarball-writer outfile)]
+    (let [metadata (export-metadata (now))]
+      (log/info (:msg metadata))
+      (utils/add-tar-entry tar-writer metadata))
+    (doseq [tar-item nodes-data]
+      (log/info (:msg tar-item))
```
ajroetker: Considering that one might want to trigger backups fairly regularly, I'm tempted to remove logging here, and only log the metadata above. Thoughts @wkalt @senior?

src/puppetlabs/puppetdb/cli/import.clj
```diff
@@ -47,32 +48,34 @@
...
           ;; NOTE: these submissions are async and we have no guarantee that they
           ;;   will succeed. We might want to add something at the end of the import
           ;;   that polls puppetdb until the command queue is empty, then does a
           ;;   query to the /nodes endpoint and shows the set difference between
           ;;   the list of nodes that we submitted and the output of that query
-          (client/submit-catalog dest
-                                 (:replace_catalog command-versions)
-                                 (archive/read-entry-content tar-reader)))
+          (command-fn :replace-catalog
+                      (:replace_catalog command-versions)
```
ajroetker: I'm sure there's something nicer we could do here with the command names but I didn't want to dig into that here.

src/puppetlabs/puppetdb/cli/import.clj
```diff
@@ -102,11 +105,20 @@
...
 (defn -main
   [& args]
   (let [{:keys [infile base-url]} (validate-cli! args)
-        command-versions (:command_versions (parse-metadata infile))]
+        base-url-str (utils/base-url->str base-url)]
     ;; TODO: do we need to deal with SSL or can we assume this only works over a plaintext port?
-    (with-open [tar-reader (archive/tarball-reader infile)]
-      (doseq [tar-entry (archive/all-entries tar-reader)]
-        (process-tar-entry tar-reader tar-entry base-url command-versions)))))
+    (clj-http.client/post (str base-url-str "/upload")
```
ajroetker: We use `client` in this namespace to refer to the `pdb-client`

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -42,33 +42,34 @@
...
+            [puppetlabs.puppetdb.query.population :as pop]
+            [puppetlabs.puppetdb.scf.migrate :refer [migrate! indexes!]]
+            [puppetlabs.puppetdb.scf.storage :as scf-store]
+            [puppetlabs.puppetdb.time :refer [to-seconds to-millis parse-period
+                                              format-period period?]]
+            [puppetlabs.puppetdb.utils :as utils]
+            [puppetlabs.trapperkeeper.core :refer [defservice] :as tk]
+            [puppetlabs.trapperkeeper.services :refer [service-id service-context]]
+            [robert.hooke :as rh]
+            [slingshot.slingshot :refer [throw+ try+]])
```
ajroetker: Sorry about the namespace changes above ^ honestly I'm not sure what happened there, I don't remember moving things around. I'm sure I must have had a good reason >.>

ajroetker: The only changes were `clojure.java.io :as io` and requiring the `me.raynes.fs :as fs`

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -259,12 +260,16 @@
...
-        discard-dir (file mq-dir "discarded")
+        backups-dir (str (io/file vardir "backups"))
+        mq-dir (str (io/file vardir "mq"))
+        discard-dir (io/file mq-dir "discarded")
         mq-connection-str (add-max-framesize max-frame-size mq-addr)
         authorizer (when certificate-whitelist
                      (build-whitelist-authorizer certificate-whitelist))]
 
+    (when-not (fs/directory? backups-dir)
+      (fs/mkdir backups-dir))
```
ajroetker: Not sure if this belongs in packaging, @kbarber what do we do for the `mq` dir?

src/puppetlabs/puppetdb/client.clj
```diff
@@ -30,7 +30,7 @@
      (let [message (json/generate-string command-map)
            checksum (kitchensink/utf8-string->sha1 message)
            url (str (utils/base-url->str base-url)
-                    (format "/commands?checksum=%s" checksum))]
+                    (format "?checksum=%s" checksum))]
```
ajroetker: not broken but wrong but I noticed this while testing out my patch

src/puppetlabs/puppetdb/helpers.clj
```diff
@@ -0,0 +1,89 @@
...
+            [ring.util.io :as rio]))
+
+(def api-version :v4)
+
+(defn import-app
+  [infile command-fn]
+  (import/import! infile (partial import/process-tar-entry
+                                  command-fn)))
+
+(defn completion-wrapper
```
ajroetker: This function (as @mullr would say) kind of has a "code-smell", it's basically a translation from an `href` in our entities to an entity/query for our in-process querying.

mullr: I can't really claim credit for the term: http://c2.com/cgi/wiki?CodeSmell

>

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -46,213 +52,107 @@
...
-                           "/reports?query="
-                           (url-encode (format "[\"=\",\"certname\",\"%s\"]"
-                                               node)))
-                      {:accept :json}))]
-      (map #(reports-expansion % base-url) body))))
+(pls/defn-validated report->tar :- utils/tar-item
+  "Create a tar-item map for the `report`"
+  [{:keys [certname configuration_version start_time] :as report}]
+  (let [unique-seed (str (json/parse-string
+                          (json/generate-pretty-string start_time)) configuration_version)
```
wkalt: is this string parsing required now for some reason? seems like it could just be
```
(str start_time configuration_version)
```
if it's just for a hash.

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -46,213 +52,107 @@
...
+        hash (kitchensink/utf8-string->sha1 unique-seed)]
+    {:msg (format "Writing report for node '%s' (start-time: %s version: %s hash: %s)" certname start_time configuration_version hash)
```
wkalt: may as well underscore start-time since start_time is all the user will know about.

src/puppetlabs/puppetdb/helpers.clj
```diff
@@ -0,0 +1,89 @@
...
+        #(select-keys % [:certname :facts_timestamp :catalog_timestamp :report_timestamp])))
+
+(defn export-app
+  [outfile query-fn]
+  (export/export! outfile
+                  (->> (query-fn :nodes api-version nil nil doall)
+                       (map munge-nodes-data)
+                       (mapcat (partial export/get-node-data
+                                        (partial query-wrapper query-fn)
+                                        (partial completion-wrapper query-fn))))))
```
ajroetker: `partial` party, I don't like it and want to remove it, made a note about it in the TODO for the PR description. Basically this code existed side-by-side with the old code so we needed to be able to pass through different kinds of query-fns.

src/puppetlabs/puppetdb/helpers.clj
```diff
@@ -0,0 +1,89 @@
...
+
+(defn build-app
+  [submit-command-fn query-fn {:keys [backups-dir authorizer]}]
+  (-> (compojure/routes
+       ;; TODO Handle the following with the wrap-multipart-params
+       ;; TODO (mid/verify-content-type ["application/octet-stream"])
+       ;; TODO (mid/validate-query-params {:required ["archive"]})
```
ajroetker: I made a ticket about this @wkalt, I know you don't like TODO comments but I figured it was best since this endpoint really doesn't have the right kind of validation we'll want eventually.

src/puppetlabs/puppetdb/http.clj
```diff
@@ -108,6 +108,28 @@
...
+      (rr/header "Content-Type" "application/octet-stream")
+      (rr/header "Content-Disposition" (str "attachment; filename=" filename))
+      (rr/charset "utf-8")
+      (rr/status status-ok)))
+
+(defn streamed-tar-response
+  [producer filename]
+  (tar-response
+   (rio/piped-input-stream producer)
+   filename))
```
ajroetker: @senior Since you wrote the json streaming code, wanna take a look here to make sure I'm not crazy here?

test/puppetlabs/puppetdb/cli/import_export_roundtrip_test.clj
```diff
@@ -14,23 +18,25 @@
...
+  (map (comp tur/munge-report-for-comparison
+             tur/munge-example-report-for-storage)
        (-> report
            utils/vector-maybe)))
 
 (defn munge-catalog
   [catalog]
-  (map (partial tuc/munge-catalog-for-comparison :v6)
+  (map (comp (partial tuc/munge-catalog-for-comparison :v6)
+             #(dissoc % :hash))
```
ajroetker: A subtlety here is that this function now "actually" works on a list of catalogs, which aligns it with the reports function above.

test/puppetlabs/puppetdb/cli/import_export_roundtrip_test.clj
```diff
@@ -82,86 +96,100 @@
...
+         (is (empty? (query-fn :nodes helpers/api-version nil nil doall)))
+
+         (svc-utils/sync-command-post (command-base-url svc-utils/*base-url*) "replace catalog" 6 catalog)
+         (svc-utils/sync-command-post (command-base-url svc-utils/*base-url*) "store report" 5
+                                      (tur/munge-example-report-for-storage report))
+         (svc-utils/sync-command-post (command-base-url svc-utils/*base-url*) "replace facts" 4 facts)
+
+         (is (testutils/=-after? munge-catalog catalog (->> certname
+                                                            (export/catalogs-for-node (partial helpers/query-wrapper query-fn)
+                                                                                      (partial helpers/completion-wrapper query-fn))
```
ajroetker: This is ugly, but like I said before, a refactor PR will follow this one up to clean this up.

src/puppetlabs/puppetdb/cli/export.clj
ajroetker: That was the issue I was having yesterday with the anonymizer hash being different, basically `start_time` is a timestamp and `generate-pretty-string` formats it to a string (like we used to with the old import querying), I decided to do this so that the string basically follows the same codepath it used to.

src/puppetlabs/puppetdb/cli/export.clj
wkalt: I'm in favor of that, now that I watch it work. Maybe that, plus a log line indicating when it finishes?

ajroetker: Yeah that's a good idea, forgot about it. I'll add that it.

src/puppetlabs/puppetdb/cli/export.clj
ajroetker: I think if we're not logging the messages anymore, it doesn't really matter right?

wkalt:
perf improvements confirmed:
https://gist.github.com/wkalt/c57811dd063778819908

~ 22% reduction on 5k nodes with a pretty low number of reports. I'll bet the difference becomes more evident with more reports. I'm guessing the new disparity between real and user time is caused by offloading work from the export process to the main PDB process.

src/puppetlabs/puppetdb/cli/export.clj
wkalt: since we're putting stuff on the wishlist, it would be super cool to have counts of the exported objects in that final line if that would be simple to do.

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -275,14 +175,31 @@
...
-      (utils/add-tar-entry tar-writer (export-metadata))
-      (doseq [node nodes
-              :let [node-data (get-node-data base-url node)]]
-        (doseq [{:keys [msg] :as tar-item} (mapcat node-data [:catalog :reports :facts])]
-          (println msg)
-          (utils/add-tar-entry tar-writer tar-item))))))
+  (let [{:keys [outfile base-url]} (validate-cli! args)]
+    (println (str "Triggering export to " outfile "..."))
+    (trigger-export-via-http! base-url outfile)
+    (println (str "Finished export to " outfile "."))))
```
wkalt: might be redundant with the other logging, but it could be useful to print timestamps in these lines

src/puppetlabs/puppetdb/cli/services.clj
wkalt: looks like you alphabetized them :P

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -24,20 +26,24 @@
             [slingshot.slingshot :refer [throw+ try+]])
   (:import [java.net URL]))
 
-(def ^:private api-version :v4)
+(def ^:private api-version :v1)
+(def ^:private api-command-versions
```
wkalt: it's a bit confusing that we're using "api" to refer to multiple APIs. "cli-api-version" and "command-versions" might be more specific.

src/puppetlabs/puppetdb/client.clj
wkalt: good catch. I guess we never called it with two args.

test/puppetlabs/puppetdb/cli/import_export_roundtrip_test.clj
wkalt: should it be renamed to call that out? also it looks like there's no longer a benefit to the -> below

src/puppetlabs/puppetdb/cli/export.clj
ajroetker: The other logging logs to PuppetDB, this logs to the console when you run the `puppetdb export` command.

wkalt:
Corresponding results for the import:
https://gist.github.com/wkalt/0e1be8318b99e5e44201

This is less meaningful since in both cases the bottleneck was command processing time rather than the import tool. Those timings represent the time taken to dump all the commands onto the queue.

src/puppetlabs/puppetdb/cli/import.clj
wkalt: in the pdb-client namespace we alias http.client as http-client. Might help to use that here so it doesn't get overlooked in the future if someone starts using an alias for it.

src/puppetlabs/puppetdb/helpers.clj
```diff
@@ -0,0 +1,89 @@
+(ns puppetlabs.puppetdb.helpers
```
wkalt: "helpers" seems a little generic to me for what this is, but maybe "cli" doesn't fit right either since we have cli tools that aren't in here. Maybe "cli-services"? "cli-helpers"? A more radical idea would be mounting this on the metadata service, although I could see that making less sense if this is ultimately going to house benchmark as well.

src/puppetlabs/puppetdb/helpers.clj
wkalt: my opinion is that the existence of the ticket makes these comments redundant.

src/puppetlabs/puppetdb/cli/services.clj
ajroetker: hmm i wonder if i hit some sort of clj-refactor function accidentally.

src/puppetlabs/puppetdb/cli/export.clj
ajroetker: I think that would be really easy to do. I'll put it on the list.

wkalt:
@AJ with respect to the blurb in the comment,

```
The export endpoint will either stream you the backup tarball GET, or will trigger an export to the local backups directory PUT.
The streaming GET version takes an archive param which is useful for things like a browser, where the value supplied to the param will be the name of the file downloaded, for curl or Clojure calls to the endpoint, the param isn't necessary.
The local PUT version takes a param (archive) which is the name of the local backup tarball file you wish to make (in the future, this might be a timestamped default).
```

I guess these options aren't available via the cli tool? The thing that's confusing to me is if I'm on the PDB host and do an export to the local filesystem with --outfile, then import with --infile, I end up with two copies of the data -- one in my local directory and one in backups.

Do we need to be persistently archiving data on imports? I may have misunderstood when we talked about this earlier, but I thought that data would only be in /backups for the duration of the import.

test/puppetlabs/puppetdb/cli/import_export_roundtrip_test.clj
ajroetker: Yeah reports is the same way, I can change them here or in my patch to these tests that are lying in another branch.

src/puppetlabs/puppetdb/cli/import.clj
ajroetker: Sounds good.

wkalt:
left a number of minor comments up there. The biggest outstanding one for me is whether to delete the tarball created in /backups once the import is finished -- I'd say yes, not sure what others think. All in all this seems pretty solid.

src/puppetlabs/puppetdb/cli/export.clj
wkalt: yeah, I know, I just mean redundant in the sense that the timestamps are in the log by default. I think it'd be nice to have them in both places so you can see start and end next to each other.

src/puppetlabs/puppetdb/cli/export.clj
wkalt: definitely true

ajroetker:
@wkalt So after thinking about this for a few minutes, deleting the archive would require us to add another endpoint for deleting archive in the backups dir, which is probably a good idea to have but is probably deserving of its own ticket.

ajroetker:

ajroetker:

ajroetker:
transients ^

src/puppetlabs/puppetdb/cli/export.clj
wkalt: I still don't understand -- where's the anonymization code that causes an issue?

ajroetker: The issue is the we're using the in-process querying to grab to report, instead of going over http. The in-process queried report has not had `json/generate-pretty-string` called on it, so it's timestamps haven't been formatted.

The anonymization tool assumes the http/jsonified version of the timestamp when computing the filename-hash for a report, which is correct, but we need to mimic that computation here.

wkalt: okay, that makes sense. Might be good to resolve this as part of a separate refactor, if similar needs will continue to crop up.

ajroetker: I think there might even be a tech-debt ticket for this kind of thing, lemme check.

ajroetker:

documentation/_puppetdb_nav.html
```diff
@@ -78,10 +78,11 @@ <h2 id="puppetdb">PuppetDB NEXT</h2>
...
     </ul>
   </li>
-  <li><strong>Metadata API Version 1</strong>
+  <li><strong>Admin API Version 1</strong>
     <ul>
-      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/version.html">Version Endpoint</a></li>
-      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/server-time.html">Server Time Endpoint</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/admin/v1/upload.html">Version Endpoint</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/admin/v1/import.html">Version Endpoint</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/admin/v1/export.html">Server Time Endpoint</a></li>
```
wkalt: these still say version and server time

documentation/api/admin/v1/export.markdown
```diff
@@ -0,0 +1,51 @@
...
+
+This endpoint can be used to stream a tarred, gzipped backup archive of PuppetDB to your local machine.
+
+### URL Parameters
+
+* `archive`: Required. The name of the archive file PuppetDB will use in the response header.
+
+### Response Format
+
+The response will be a `application/octect-stream`, and will return a `tar.gz` archive.
```
wkalt: octet

documentation/_puppetdb_nav.html
ajroetker: I'm not saying that I copy/pasted, but I'm not saying I didn't either...

ajroetker:

ajroetker:
Failure is because we stopped accepting `./blah.tgz` as valid input to the endpoint, so good news is that the no-relative-paths works, bad news is i have to fix these tests.

ajroetker:

documentation/_puppetdb_nav.html
```diff
@@ -78,10 +78,11 @@ <h2 id="puppetdb">PuppetDB NEXT</h2>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/aggregate-event-counts.html">Aggregate Event Counts Endpoint</a></li>
     </ul>
   </li>
-  <li><strong>Metadata API Version 1</strong>
+  <li><strong>Admin API Version 1</strong>
     <ul>
-      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/version.html">Version Endpoint</a></li>
-      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/server-time.html">Server Time Endpoint</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/admin/v1/upload.html">Upload Endpoint</a></li>
```
kbarber: still pointing to old PR content

ajroetker:

documentation/api/admin/v1/archive.markdown
```diff
@@ -0,0 +1,57 @@
...
+canonical: "/puppetdb/latest/api/admin/v1/archive.html"
+---
+
+[curl]: ../curl.html#using-curl-from-localhost-non-sslhttp
+
+The `/archive` endpoint can be used for importing and exporting PuppetDB archives.
+
+## `POST /pdb/admin/v1/archive`
+
+This endpoint can be used to upload a PuppetDB backup archive to PuppetDB's backup directory.
```
senior: This has changed now, right?

ajroetker: Copy/paste error :dancer:

ajroetker:

documentation/api/admin/v1/archive.markdown
```diff
@@ -0,0 +1,57 @@
...
+This endpoint can be used for streaming a PuppetDB archive into PuppetDB.
+
+### Request Format
+
+The request should be a multipart POST and have `Content-Type: multipart/mixed`.
+
+### URL Parameters
+
+* `archive`: Required. The archive file to import to the PuppetDB.
+* `command_versions`: Required. A JSON object mapping PuppetDB command-names to their version.
```
mullr: We should say something about what to use here, for exports from various versions of PuppetDB.

src/puppetlabs/puppetdb/admin.clj
```diff
@@ -0,0 +1,53 @@
...
+            [puppetlabs.puppetdb.middleware :as mid]
+            [puppetlabs.trapperkeeper.core :as trapperkeeper]
+            [ring.middleware.multipart-params :as mp]
+            [clj-time.core :refer [now]]
+            [ring.util.io :as rio]))
+
+(def query-api-version :v4)
+
+(def munge-nodes-data
+  (comp (partial kitchensink/mapvals str)
```
mullr: It's clearer, faster, and 2 characters shorter to write

(defn munge-nodes-data [nodes]
    (kitchensink/mapvals str
       (select-keys nodes ...)))

Point-free style is great in some places, but I don't think this is it.

src/puppetlabs/puppetdb/admin.clj
```diff
@@ -0,0 +1,53 @@
...
+                  (->> (query-fn :nodes query-api-version nil nil doall)
+                       (map munge-nodes-data)
+                       (mapcat (partial export/get-node-data query-fn)))))
+
+(defn build-app
+  [submit-command-fn query-fn {:keys [authorizer]}]
+  (-> (compojure/routes
+       (mp/wrap-multipart-params
+        (compojure/POST "/v1/archive" [archive command_versions]
+                        (do (import/import! (:tempfile archive)
```
mullr: Can the user squirt something interesting in here by abusing compojure? For example, if they passed `archive='{"tempfile": "/etc/passwd"}'` or something like that. (more likely to work with a form encoding than json)

documentation/api/admin/v1/archive.markdown
ajroetker: What do you mean? Would something like "The command-names to use can be found in the export tarball at `puppetdb-bak/export-metadata.json`." do?

src/puppetlabs/puppetdb/admin.clj
ajroetker: I think they can yes, but even if that file was readable by the PuppetDB user, we don't return anything to the user.

documentation/api/admin/v1/archive.markdown
mullr: If they're actually in the tarball, then we should give them the command line to get the data. This ought to be a cut-and-paste exercise, not requiring the user to go hunt for information.

ajroetker: Alright, but it can't really be copy and paste anymore than I have it already, as I gave them the most recent command version, and I'd have to have something a little more complicated in the example to make it version independent
```
tar -xfvz example.tgz
curl -X POST http://localhost:8080/pdb/admin/v1/upload \
       -F "command_versions='$(cat puppetdb-bak/export-metadata.json)'"\
       -F "archive=@example_backup_archive.tgz"
```

src/puppetlabs/puppetdb/admin.clj
mullr: Event so, the user could throw an arbitrary file on the system at our import function, which isn't really what we want to do. I think this code could be a fair bit tighter on that front. 

One low-impact approach may be to provide our own handler function to wrap-multipart-params which wraps the default temp-file handler, tacking on some metadata, then checking that metadata inside the handler to verify that the data actually came from the place we expect.

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -46,213 +67,103 @@
...
-    value
-    (assoc value :data (retrieve-related-data base-url href))))
+  [query-fn
+   fields :- [s/Keyword]
+   values :- [{s/Any s/Any}]]
+  (->> values
+       (map (partial
+             kitchensink/mapvals
+             #(utils/assoc-when % :data (query-child-href-internally query-fn (:href %)))
+             fields))))
```
mullr: Is there a way to write the above more simply? Maybe as 2 fns or something, or by naming some of the intermediate values. I'm having a hard time working out what it does.

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -46,213 +67,103 @@
...
-              [(catalog->tar certname (->> certname
-                                           (catalogs-for-node base-url)
-                                           first))])})
-
-(pls/defn-validated get-nodes :- (s/maybe (s/pred seq? 'seq?))
-  "Get a list of the names of all active nodes."
-  [base-url :- utils/base-url-schema]
-  (parse-response (client/get (str (utils/base-url->str base-url) "/nodes")
-                              {:accept :json})))
+  (concat (when-not (str/blank? (:facts_timestamp node-data)) (facts-for-node query-fn certname))
```
mullr: destructuring for extra credit

src/puppetlabs/puppetdb/cli/export.clj
```diff
@@ -24,20 +26,39 @@
...
+(def query-api-version :v4)
+(def ^:private command-versions
+  ;; This is not ideal that we are hard-coding the command version here, but
+  ;;  in our current architecture I don't believe there is any way to introspect
+  ;;  on which version of the `replace catalog` matches up with the current
+  ;;  version of the `catalog` endpoint... or even to query what the latest
+  ;;  version of a command is.  We should improve that.
+  {:replace_catalog 6
+   :store_report 5
+   :replace_facts 4})
```
mullr: This needs to always reflect the the latest command versions, right? We *will* forget to update this the next time we bump those. I think it's worth using `(methods)` do gather this up at runtime. @senior, does this sound reasonable?

src/puppetlabs/puppetdb/cli/import.clj
```diff
@@ -102,11 +107,23 @@
...
+
 (defn -main
   [& args]
   (let [{:keys [infile base-url]} (validate-cli! args)
-        command-versions (:command_versions (parse-metadata infile))]
-    ;; TODO: do we need to deal with SSL or can we assume this only works over a plaintext port?
-    (with-open [tar-reader (archive/tarball-reader infile)]
-      (doseq [tar-entry (archive/all-entries tar-reader)]
-        (process-tar-entry tar-reader tar-entry base-url command-versions)))))
+        import-archive (fs/normalized-path infile)
```
mullr: import-archive is a java.io.File, right? Not a string?

src/puppetlabs/puppetdb/cli/export.clj
ajroetker: This is different from what we've done before, I'd probably want to leave that for a separate ticket.

src/puppetlabs/puppetdb/cli/import.clj
ajroetker: fs/normalize-path will turn the string into a file

ajroetker: It's a poorly named function, whose name was actually changed in the most recent version of the library

documentation/api/admin/v1/archive.markdown
mullr: We just need to tell them how to find out what they're going to use in the `command_versions` field. 

    tar -xOf my-backup.tar.gz puppetdb-bak/export-metadata.json

src/puppetlabs/puppetdb/cli/export.clj
mullr: fair 'nuff

mullr: right, since this was moved not added. :+1:

src/puppetlabs/puppetdb/cli/import.clj
mullr: ha, nice. we should really update to that ;)

mullr:
The bulk of this looks good; after we've dealt with the comments I'll be +1

ajroetker:
@mullr i think I've addressed everything if you wanna verify

ajroetker:

ajroetker:

mullr:
:shipit:

senior:

documentation/api/admin/v1/archive.markdown
```diff
@@ -0,0 +1,60 @@
...
+
+### Response Format
+
+The response will be a `application/octet-stream`, and will return a `tar.gz` archive.
+
+### Examples
+
+[Using `curl` from localhost][curl]:
+
+    curl -X GET http://localhost:8080/pdb/admin/v1/export -o puppetdb-export.tgz
```
wkalt: s/export/archive

src/puppetlabs/puppetdb/admin.clj
```diff
@@ -0,0 +1,56 @@
...
+(trapperkeeper/defservice admin-service
+  [[:PuppetDBServer shared-globals query]
+   [:PuppetDBCommand submit-command]
+   [:WebroutingService add-ring-handler get-route]]
+
+  (start [this context]
+         (->> (build-app submit-command query (shared-globals))
+              (compojure/context (get-route this) [])
+              (add-ring-handler this))
+         context))
```
wkalt: we have the same problem with 404s on this admin endpoint as we did with PDB-1509. I don't mind having a separate ticket to fix that at the middleware level, but we should track that work

wkalt:

documentation/_puppetdb_nav.html
```diff
@@ -76,10 +76,9 @@ <h2 id="puppetdb">PuppetDB {{ puppetdb_version }}</h2>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/aggregate-event-counts.html">Aggregate Event Counts Endpoint</a></li>
     </ul>
   </li>
-  <li><strong>Metadata API Version 1</strong>
+  <li><strong>Admin API Version 1</strong>
     <ul>
-      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/version.html">Version Endpoint</a></li>
-      <li><a href="/puppetdb/{{ puppetdb_version }}/api/meta/v1/server-time.html">Server Time Endpoint</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/admin/v1/archive.html">Upload Endpoint</a></li>
```
wkalt: Archive Endpoint

wkalt:
one more residual documentation bug -- looks good otherwise though. I'll merge after that.

ajroetker:
@wkalt fixed the documentation change

ajroetker:
Was only a docs change since the last passing test ^

ajroetker:

wkalt:
+1 will merge when master is green

PR #1491, opened 2015-06-25
(maint) Update tests to use new Facter 3 variables

kbarber:


kbarber:

kbarber:
On second glance most of that older debian 6 & ubuntu 10 JDK stuff can go, I'm going to update this patch.

acceptance/setup/pre_suite/40_install_deps.rb
```diff
@@ -2,35 +2,11 @@
   step "Install other dependencies on database" do
     databases.each do |database|
       os = test_config[:os_families][database.name]
-      db_facts = facts(database.name)
```
kbarber: This is all dead stuff, we don't test on systems that don't have JDK7 any more.

PR #1492, opened 2015-06-25
(maint) skip legacy storeconfigs import export on AIO

wkalt:
This corrects an issue that cropped up due to a facter version change. There's
no longer a puppetversion fact -- dispatch on AIO status instead.

### DIFF
--- a/acceptance/tests/import_export/legacy_storeconfigs_import_export.rb
+++ b/acceptance/tests/import_export/legacy_storeconfigs_import_export.rb
@@ -1,8 +1,6 @@
 test_name "storeconfigs export and import" do
 
-  confine :except, :platform => 'ubuntu-10.04-amd64'
-
-  skip_test "storeconfigs not supported in puppet >= 4.0" unless version_is_less(facts(master.name)['puppetversion'], '4.0.0')
+  skip_test "storeconfigs not supported in puppet >= 4.0" if options[:type] == 'aio'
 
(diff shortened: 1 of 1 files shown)


joshcooper:
There should be `puppetversion` fact. There wasn't initially (and sounds like isn't with the version you're testing against), but it should be there now.

joshcooper:
https://tickets.puppetlabs.com/browse/PUP-4359

wkalt:
@joshcooper it says here that it was dropped in facter 3.0
http://docs.puppetlabs.com/facter/3.0/release_notes.html

and I don't get it on facter 3.0 out the box:
```
[~/work/facter/release] ((HEAD detached at 3.0.0)) $ bin/facter puppetversion

```
ruby facter works fine

```
[~/work/facter/release] ((HEAD detached at 3.0.0)) $ facter puppetversion
3.8.1
```

am I missing something?

joshcooper:
Ah, yeah sorry. Facter used to load puppet code to create the fact, which created a cyclic dependency between facter and puppet code.

So now puppet creates the fact (since it knows what its version is). So yes, `facter puppetversion` is now empty, but running `puppet apply -e 'notice($puppetversion)'` will show the puppetversion fact (or just run `puppet --version`)

wkalt:
@joshcooper gotcha -- yeah we figured using puppet --version would work here as well. Probably the way to go if support for Puppet 3 is in the future.

PR #1494, opened 2015-06-25
(PDB-1665) Guard experimental endpoint warnings

rbrw:
Don't warn about event-counts or aggregate-event-counts endpoints if the
product name isn't puppetdb.

### DIFF
--- a/src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
+++ b/src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
@@ -23,9 +23,10 @@
                     query-options (-> {:counts_filter counts_filter'
                                        :count_by count_by}
                                       (merge distinct-options))]
-                (log/warn
-                  (str "The aggregate-event-counts endpoint is experimental"
-                       " and may be altered or removed in the future."))
+                (when-not (= "puppetdb" (:product-name globals))
+                  (log/warn
+                   (str "The aggregate-event-counts endpoint is experimental"
+                        " and may be altered or removed in the future.")))
                 (produce-streaming-body
                   :aggregate-event-counts
                   version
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/http/event_counts.clj
```diff
@@ -27,8 +27,10 @@
 
 (defn event-counts-app
   "Ring app for querying for summary information about resource events."
-  [version]
-  (log/warn "The event-counts endpoint is experimental and may be altered or removed in the future.")
+  [version globals]
+  (when-not (= "puppetdb" (:product-name globals))
```
wkalt: would moving this check into routes section like in aggregate-event-counts allow you to avoid parameterizing v4-app?

wkalt:
@rbrw this should be on 2.3.x I think

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -23,9 +23,10 @@
                     query-options (-> {:counts_filter counts_filter'
                                        :count_by count_by}
                                       (merge distinct-options))]
-                (log/warn
-                  (str "The aggregate-event-counts endpoint is experimental"
-                       " and may be altered or removed in the future."))
+                (when-not (= "puppetdb" (:product-name globals))
```
senior: I think you're logic is inverted here. The product-name is *not* "puppetdb", meaning it's pe-puppetdb. So when pe-puppetdb, log the warning below, right? We only want this warning on FOSS.

rbrw: Yeah, thanks, I noticed that too when shifting to 2.3.x.

rbrw:
Shifted to 2.3.x: https://github.com/puppetlabs/puppetdb/pull/1498

PR #1496, opened 2015-06-26
(RE-4779) Create a script to migrate config files and any needed data

melissa:


wkalt:
packages here: http://builds.puppetlabs.lan/puppetdb/3.0.0.SNAPSHOT.2015.06.26T1328

wkalt:
should this migrate log files?

wkalt:
I've been testing this on debian wheezy. I still haven't gotten it working due to issues with the migrated ssl files -- more specificity on that when I have it.

Another thing I'm seeing though is that none of the old config is getting removed, although the script appears to try. Here's a gist of what I see when I upgrade:

https://gist.github.com/wkalt/8aa1c150cfa2ba49b2f4

additionally, it seems like /etc/puppet/puppetdb.conf and /etc/puppet/routes.yaml are not being migrated

wkalt:
to add to the above, here's what I'm doing to test:

1. Install Puppet 3 and PuppetDB 2.3.5 according to our usual smoke testing procedure:
https://confluence.puppetlabs.com/display/PP/Smoke+Testing+Guide+for+PDB+Releases

2. do a puppet run and confirm that there's data in hsqldb

3. stop puppetdb and puppet master, install puppet-agent and puppetserver from PC1 repos

4. install puppetdb and terminus from packages above

my understanding is that this should be sufficient (and that there's no required puppet run between steps 3 and 4.) I'm about to try again with an intermediate run, but please correct me if I'm missing a step.

kbarber:
@wkalt General thought: is the right thing to care about here a Puppet 3->4 & PDB 2->3 upgrade? I was just presuming a Puppet 4 only upgrade of PDB 2->3. The problem here is, we can't get involved in a Puppet 3->4 upgrade ourselves, so yeah puppetdb.conf and routes.yaml won't get moved across - but we never manage those files ourselves, they are outside of our influence from a packaging perspective.

Re: log files, the presumption is we won't touch these. I'm not sure how important that is, but log files can be tricky, people might have adjusted their logrotate configuration to do all kinds of weird things I guess. Its entirely possible we could do something - but I don't know if its important enough I guess, or easy if there is "weird stuff". *shrug*.

kbarber:
Results of my test on el7: https://gist.github.com/kbarber/e6c2730d543769807fe7

Things to note:

* https://gist.github.com/kbarber/e6c2730d543769807fe7#file-gistfile1-txt-L45 <- this is related to @shrug's work ... the process is running so usermod will fail, and subsequently /etc/passwd isn't updated. I specifically bumped to 0.3.11 of ezbake to build this package as well, same result. Not related to this patch, but worth noting.
* https://gist.github.com/kbarber/e6c2730d543769807fe7#file-gistfile1-txt-L68 <- this is the wrong path anyway, its a misnomer to use the last /db for a real path ... actually, the last db is a suffix for files (see here: https://gist.github.com/kbarber/35d6c9dd68daf073520f), so the real path to copy should be shortened to just /var/lib/puppetdb/db.

wkalt:
disregard my earlier results -- about to try again on Wheezy just within Puppet 4.

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,244 @@
...
+                echo "Updated setting ${setting} in ${ini_file}."
+                case "$setting" in
+                  vardir|ssl-*)
+                    migrate "$orig_value" "$new_value"
+                    ;;
+                  subname)
+                    orig_value="${orig_value#file:}"
+                    orig_value="${orig_value%/db;hsqldb.tx=mvcc;sql.syntax_pgs=true}"
+                    new_value="${new_value#file:}"
+                    new_value="${new_value%/db;hsqldb.tx=mvcc;sql.syntax_pgs=true}"
```
melissa: @kbarber I've updated this to pull out the last db from the path. It looks like it's working now. I also rebased against stable

melissa: packages with these changes are available at http://builds.puppetlabs.lan/puppetdb/3.0.0.SNAPSHOT.2015.06.29T1119/

kbarber: Testing now.

kbarber: Okay, here is my latest transcript for el7 ... https://gist.github.com/kbarber/07262d963bd42cbf12eb

So there are still some niggles, primarily some restart concerns we need to think through: https://gist.github.com/kbarber/1f0fc222df5042cd8fb9

In short: if you upgrade the terminus, puppetserver needs a restart (no surprises there, not sure how we should handle that) ... and when we migrate SSL we aren't restarting puppetdb either, it all goes well after a restart though.

So these tests I've done are for a happy path (that is, no customisation really - upgrading from package defaults). I'll have to shake up the user customised path as well, in another test most probably.

wkalt:
Just tested the new packages on debian wheezy, upgrading PDB within Puppet 4. Observed the restart issues Ken's talking about but overall it worked great.

One thing to double check though: all of my old config directories are still around after the upgrade in their old spots, including the old mq. It looks like dpkg is trying to remove these on upgrade, but failing: https://gist.github.com/wkalt/6d9f0ed1e34923038cd4

It seems like because of this we could have issues with users updating the wrong config and getting confused. Is this expected, or should these be removed/moved in the script?

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,244 @@
...
+config_file=${config_dir}/config.ini
+ini_files=($config_file $jetty_file $database_file)
+
+# Deal with config files
+echo "Migrating configuration files ${orig_ini_files[*]} to ${config_dir}"
+if [ -e '/tmp/puppetdb-upgrade-config-files.tgz' ] ; then
+  mkdir -p $orig_config_dir
+  tar -xzf '/tmp/puppetdb-upgrade-config-files.tgz' -C $orig_config_dir
+else
+  echo "Config archive not found. Not proceding with migration"
```
wkalt: proceeding

resources/ext/cli/config-migration
wkalt: I've tested this on wheezy with custom vardir and subname -- handles things fine.

wkalt: custom ssl paths are handled nicely too

kbarber:
So the pending concerns (at this point in time) are:

* After upgrading both puppetserver & puppetdb need a second manual restart for this to work. We can probably fix the puppetdb case and we should probably go ahead and have a look at this now @melissa.
* But ... the puppetserver restart case is less obvious. Although we could always restart when the terminus is upgraded? This would be harder if we supported both Puppet 3 & Puppet 4 - which we might in the future, but right now it's not difficult to conceive, but there are no hooks in ezbake to allow the terminus to have postinst 'stuff' I believe today. Not sure how I feel about this, it is a barrier to the upgrade true, fairly easily fixed but the kind of thing I can imagine we'll get hounded on about on IRC & puppet-users (release notes could cover this).
* Old configuration gets left behind, we're concerned people will still try to modify this thinking it will work. We could keep the archive of this after migration perhaps, since the migration script might mess it up, or just cover it in release notes perhaps.
* There is a bug in the usermod code during upgrade @shrug because the process is still running, this was related to your work on ezbake I think, did you want me to raise a separate bug or something?
* @wkalt found a small typo in error reporting.

melissa:
@kbarber @wkalt I've updated the script to clean up the old config files if they are still around. They are archived in the new location, and a backup copy is made if any modifications happen in the script. That being said, this goes against what packaging systems do. As long as your okay with that, this should be fine. Testing now.

melissa:
Okay, here's what I'm thinking with restarting the service on upgrade.

RPM does it already, we just have to reorder what does where. This will require an update to ezbake. Rather than appending out line, we need to be prepending our line.
{code}
postinstall scriptlet (using /bin/sh):
# Reload the systemd units if this is an upgrade
if [ "$1" = "2" ]; then
    systemctl daemon-reload >/dev/null 2>&1 || :
    systemctl try-restart puppetdb.service
fi

if [ $1 -eq 1 ] ; then
        # Initial installation
        /usr/bin/systemctl preset puppetdb.service >/dev/null 2>&1 || :
fi
/opt/puppetlabs/server/apps/puppetdb/scripts/install.sh postinst_redhat
{code}

For debian, we need to do a little more work. We need to actually specify the restart in resources/ext/ezbake.conf. We need to figure out what init system we're running (systemd vs. sysv), so we can make the correct restart call. We should be able to just call `service restart puppetdb
{code}
#!/bin/sh

exec "/opt/puppetlabs/server/apps/puppetdb/scripts/install.sh" postinst_deb
{code}

kbarber:
@melissa did we ever restart for Debian? is that what you mean?

Also, I didn't think we supported Jessie just yet, does this matter yet?

Sounds like the ordering for RHEL is spot on, I'm guessing we need a new hook? I'm only saying this because I presume order assumption is important to current consumers. I really have no idea myself what others might 'expect' :-).

I'm +1 on the direction you have outlined largely @melissa thanks.

We'll test the archiving thing also, I know it feels icky from a packaging perspective, I guess the requirement is ignoring technology to certain extent, but the outcome in the end does seem reasonable right? I'm building out our release notes now and upgrade guide, I'll make sure this behaviour gets a mention.

melissa:
@kbarber both deb and rpm restarts should be happening in ezbake. I think this PR is done in terms of functionality. I'm working on an ezbake pr as we speak.

kbarber:
@melissa thanks very much, looking forward to testing your changes again.

kbarber:
@melissa so this is my latest test transcript:

https://gist.github.com/kbarber/aa2fbd7f2ee07ae7aff3

And then the post-upgrade experience now looks much simpler:

https://gist.github.com/kbarber/e410cd5e0085e486aa24

General Notes:

* No more puppetdb restart needed, yay.
* We're still seeing that usermod problem, /etc/passwd still has the old homedir as a consequence: https://gist.github.com/kbarber/aa2fbd7f2ee07ae7aff3#file-gistfile1-txt-L43. @shrug shall I just raise a ticket on this one perhaps, I don't see any other side-effects for this bug, perhaps we can fix this in a 3.0.1.
* I'm concerned about the warning message here in the puppet agent run for the first run (https://gist.github.com/kbarber/e410cd5e0085e486aa24#file-gistfile1-txt-L4). Its totally ambiguous, not everyone will think to restart puppetserver with an error message like that.

kbarber:
@melissa so @wkalt and I have considered the restart case I noted above, and we feel that perhaps the softest approach (beyond doing nothing that is) is to just drop a message when users upgrade `puppetdb` saying something like "puppetdb-termini/us should also be upgraded, and puppetserver should be restarted". This could be done with simple echo on the post-upgrade block in ezbake perhaps. What do you think?

Another solution is that we could add hooks to the terminus upgrade that displays a similar message, but I don't want to drag this out much longer ... so perhaps we can shelve that idea for later and rely on release notes and this "best effort" for the time we have.

We think this issue is bigger than the current problem, we may want a better handling around this in the future, as its a problem we've always suffered, but its probably not worth it right now. In short, restart puppetserver after a termini upgrade is one idea, but I don't want to rush into this kind of thing.

wkalt:
here's what I see now on debian wheezy, with the packages Ken linked above (http://builds.puppetlabs.lan/puppetdb/3.0.0.SNAPSHOT.2015.07.02T1459):
https://gist.github.com/wkalt/8304b9fe01e91512765e

immediately after installing:
```
root@localhost:/vagrant# service puppetdb status
puppetdb is not running ... failed!
```

getting things working still required a restart of both puppetdb and puppetserver, and it seems like only some of the old config & files have been removed:

```
root@localhost:/vagrant# ls /etc/puppetdb/conf.d/
jetty.ini.bak.1435862023  repl.ini
root@localhost:/vagrant# ls /var/lib/puppetdb/
db  mq
root@localhost:/vagrant# ls /etc/puppetdb/
conf.d  logback.xml  ssl
```

not sure what causes the difference from el7 -- what a headache :(

kbarber:
@wkalt sorry, I had the same effect with left-over configuration on EL7 also.

melissa:
I only remove the three config files we're moving around. I didn't even think about removing the old locations of things we also migrate (/var/lib/puppetdb, etc). I feel uncomfortable moving any of the config files we're not explicitly dealing with since that's default behavior of the packaging system (note, rpm will clean up all old config files unless you've modified them).

I can also add a message in the postinst script that reminds the user to upgrade termini/us and restart puppet server if that is desired. I would prefer to keep that out of ezbake and do that in `resources/ext/ezbake.conf`

melissa:
@wkalt the difference between wheezy and el-7 with files and directories getting left behind is because of the difference between how apt handles uninstalls and how yum handles uninstalls.

kbarber:
> I can also add a message in the postinst script that reminds the user to upgrade termini/us and restart puppet server if that is desired. I would prefer to keep that out of ezbake and do that in resources/ext/ezbake.conf

That would be awesome @melissa, even a basic echo here will do. Thanks a bunch. Its definitely something we can expand on later on, when the dust clears.

wkalt:
@melissa last time I commented I couldn't see the restart happening with wheezy. I think that's the more important difference -- sounds like el7 and wheezy may be acting consistently WRT the untargeted config.

Seems like we should either sort that out or drop a message instructing to restart. Automatic restart would obviously be best, but if the issue is complicated I wouldn't be too bothered by communicating that need with echo as well.

wkalt:
+1 on the terminus upgrade message though

shrug:
@wkalt We merged https://github.com/puppetlabs/ezbake/commit/5a3f21f4606776c1013695477a65c19341610c07 earlier which should handle the service restart on Debian and Ubuntu, I'm about to cut an ezbake release so we can pull that in

wkalt:
@shrug I had understood the packages above to contain that patch -- likely mistaken though. I'll give it another shot.

wkalt:
@shrug I rebuilt the packages with ezbake 0.3.13 and I'm still seeing the restart fail on wheezy. After the upgrade,

```
root@localhost:/vagrant# service puppetdb status
puppetdb is not running ... failed!
```
PuppetDB logging is not yet initialized at this point, so no additional info from there.

wkalt:
these are the packages I used: http://builds.puppetlabs.lan/puppetdb/3.0.0.SNAPSHOT.2015.07.06T1459/repos/deb/wheezy/

shrug:
Yep, we missed something - we are stopping the service in the preun and doing a conditional restart in the post. Because the service is already stopped, it won't restart.

I think we have 2 options:
1. Figure out a reliable and secure way to save state between the scripts - which is made even more fun by the fact that some of the scripts (prerm, postrm) are run from the *old* package.
2. Ask people to restart the service after the upgrade

wkalt:
@shrug @melissa option 2 seems safest and simplest to me. IMO if we do that and drop the terminus upgrade message we'll be adequately covered.

wkalt:
@senior may want to weigh in

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,291 @@
...
+    rm "$old_ini_file"
+  fi
+  if [ -e "${old_ini_file}.rpmsave" ] ; then
+    echo "Removing ${old_ini_file}.rpmsave"
+    rm "${old_ini_file}.rpmsave"
+  fi
+done
+
+echo "Modifications may have been made to PuppetDB config files. To ensure these changes have been picked up, please do the following: "
+echo "  1) upgrade the puppetdb termini "
```
wkalt: maybe "upgrade the puppetdb-termini" package would be clearer?

resources/ext/cli/config-migration
```diff
@@ -0,0 +1,289 @@
...
+#   usage
+#
+function usage() {
+  echo "Usage: ${migration_command}"
+  echo "Configuration helper for migrating config files for PuppetDB upgrade."
+  echo
+  echo "This tool will attempt to copy the necessary PuppetDB config files into "\
+       "place for use by the PuppetDB HTTPS service. It updates the pathing of "\
+       "the files to align with new path expectations in the PuppetDB 3 series. "\
+       "It will ensure the config files  contain either existing edits or up "\
```
ajroetker: extra space in `files  contain`

PR #1498, opened 2015-06-26
(PDB-1665) Guard experimental endpoint warnings

rbrw:
Don't warn about event-counts or aggregate-event-counts endpoints if the
product name isn't puppetdb.


senior:
:+1 will merge when the pipeline is clear

PR #1499, opened 2015-06-29
(PDB-1613) Add with-logging-to-atom and with-log-level

rbrw:
Add with-log-level, allowing us to change the logback log level for a
given "logger-id" (normally a keyword, class, or namespace) during the
execution of a specified body.

Add with-logging-to-atom, which redirects all log events produced by a
specified body to an atom (i.e. redirects the result of all calls to
clojure.tools.logging's log/*), making it easier to test logging
behaviors.

cf. previous incarnation: #1489


wkalt:
looks good to me

wkalt:
travis tests are stalled, but passing locally

PR #1501, opened 2015-06-29
(maint) remove ssl-setup references to AIO pathing

wkalt:
The ssl setup tool uses the existence of the /etc/puppetlabs/puppetdb directory
to assess whether the install target is running PE. This was valid in the olden
days, but with AIO FOSS uses those paths as well. This meant that the ssl
setup tool would set the wrong user in the case of a downgrade from 3.0 to
2.3.x, since at this point the path would exist. This changes the tool to dispatch
on presence of the pe-puppetdb user.

### DIFF
--- a/ext/templates/puppetdb-ssl-setup.erb
+++ b/ext/templates/puppetdb-ssl-setup.erb
@@ -226,7 +226,7 @@ else
   agent_confdir=`puppet agent --configprint confdir`
   agent_vardir=`puppet agent --configprint vardir`
 
-  if [ -d "/etc/puppetlabs/puppetdb" ] ; then
+  if id -u pe-puppetdb > /dev/null 2>&1; then
     puppetdb_confdir="/etc/puppetlabs/puppetdb"
     user=pe-puppetdb
   else


ajroetker:
@shrug @melissa @kbarber ping

ext/templates/puppetdb-ssl-setup.erb
```diff
@@ -226,7 +226,7 @@ else
   agent_confdir=`puppet agent --configprint confdir`
   agent_vardir=`puppet agent --configprint vardir`
 
-  if [ -d "/etc/puppetlabs/puppetdb" ] ; then
+  if id -u pe-puppetdb > /dev/null 2>&1; then
```
shrug: We have traditionally used 'getent' for this, but I have no preference, assuming this works everywhere we need it

shrug:
:+1:

ext/templates/puppetdb-ssl-setup.erb
kbarber: @wkalt we probably want to check it works on all the distros we care about, we have no tests that will shake this path up, pe doesn't use ssl-setup during installation although some users use it standalone for support reasons I believe.

kbarber: Looks like `id` is posix (IEEE 1003.1), `getent` is mainly from glibc, fwiw.

* http://pubs.opengroup.org/onlinepubs/9699919799/utilities/id.html
* http://linux.die.net/man/1/getent

wkalt: tested id -u on debian 7, ubuntu 14.04, centos 7, sles 11, and PE master-supported rhel derivatives. It also works on OS X, which getent does not IIRC.

ajroetker:
:+1: tested and works for me

PR #1504, opened 2015-06-30
(PDB-1674) PuppetDB hangs on shutdown when the database is down 

senior:
This adds connection timeouts to the command processing and query
connection pools. This fixes the problem where the ActiveMQ session
can't be properly shutdown because the command connection pool can't get
a connection to the database (i.e. the database is down). Previously
this would hang, even with a Control-C. This hang was due to the
connection pool software getting caught in a loop trying to get a
connection to the database.

This connection timeout was also applied to the query side of the
application as user provided queries are expected to respond quickly and
a hanging behavior would be bad there as well.

This blocking/retrying behavior is what we want when PuppetDB first
starts up as we need to initialize the schema etc. This commit also
creates a special connection pool (just for startup) that doesn't have
these timeouts. This pool is closed after the startup code that needs a
database connection is finished.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -370,6 +370,10 @@ The maximum time (in minutes) a pooled connection should remain open. Any connec
 
 If not supplied, we won't terminate connections based on their age alone.
 
+### `connection-timeout`
+
+The maximum time to wait (in milliseconds) to acquire a connection from the pool of database connections. If not supplied, defaults to 1000.
+
 ###`statements-cache-size`
 
 This setting defines how many prepared statements are cached automatically. For a large amount of dynamic queries this number could be increased to increase performance, at the cost of memory consumption and database resources.
@@ -446,6 +450,10 @@ If not supplied, we default to 45 minutes.
 The maximum time (in minutes) a pooled connection should remain open. Any connections older than this setting will be closed off. Connections currently in use will not be affected until they are returned to the pool.
 
 If not supplied, we won't terminate connections based on their age alone.
+
+### `connection-timeout`
+
+The maximum time to wait (in milliseconds) to acquire a connection from the pool of database connections. If not supplied, defaults to 500.
 
 
 `[command-processing]` Settings
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -237,6 +237,35 @@
       (mq/remove-queue! "localhost" "com.puppetlabs.puppetdb.commands")
       (log/info "Removed legacy queue"))))
 
+(defn initialize-schema
+  "Ensure the database is migrated to the latest version, and warn if
+  it's deprecated, log and exit if it's unsupported. We do this in a
+  single connection because HSQLDB seems to get confused if the
+  database doesn't exist but we open and close a connection without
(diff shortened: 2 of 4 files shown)


senior:
This just needs docs for the new connection-timeout config

senior:

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -237,6 +237,35 @@
...
+(defn init-with-db
+  "All initialization operations needing a database connection should
+  happen here. This function creates a connection pool using
+  `write-db-config` that will hang until it is able to make a
+  connection to the database. This covers the case of the database not
+  being fully started when PuppetDB starts. This connection pool will
+  be opened and closed within the body of this function."
+  [write-db-config product-name]
+  (with-open [init-db-pool (pl-jdbc/make-connection-pool (assoc write-db-config
+                                                           ;; Block waiting to grab a connection
```
ajroetker: this alignment looks off

senior: Maybe the comment is what throws it off? I reformatted it and nothing changed.

senior:
Added some docs, switched from 2ms connection timeout on the query side to 500ms. Travis is failing with connection timeouts, only on PG. I'm wondering if the issue it's having is lazy init. The first poll for a database connection needs to init the pool, which maybe takes more than 2ms?

senior:
1 transient failing cell ^^^

rbrw:
Not sure which test is failing, but I also wouldn't be surprised if some tests might need to be adjusted to account for these changes, i.e. a query EAGAIN is now more "normal".

documentation/configure.markdown
```diff
@@ -370,6 +370,10 @@ The maximum time (in minutes) a pooled connection should remain open. Any connec
 
 If not supplied, we won't terminate connections based on their age alone.
 
+### `connection-timeout`
+
+The maximum time to wait (in milliseconds) to acquire a connection from the pool of database connections. If not supplied, defaults to 1000.
+
 ###`statements-cache-size`
```
rbrw: Perhaps wrap this (and below)?

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -237,6 +237,35 @@
...
 
+(defn initialize-schema
+  "Ensure the database is migrated to the latest version, and warn if
+  it's deprecated, log and exit if it's unsupported. We do this in a
+  single connection because HSQLDB seems to get confused if the
+  database doesn't exist but we open and close a connection without
+  creating anything."
+  [db-conn-pool product-name]
+  (sql/with-connection db-conn-pool
+    (scf-store/validate-database-version #(System/exit 1))
```
rbrw: Be glad when we finally get a chance to consider our System/exits (but not an issue for this PR).

rbrw:
This all looks good to me, works fine, and I liked the commit breakdown too.

documentation/configure.markdown
senior: Meh, I do like the wrap there, but the stuff around it is not wrapped. The whole file goes in and out of it, but specifically those sections are not. Probably something we should just do on all files at once if we want the change.

PR #1505, opened 2015-06-30
(maint) Update specter dependency to remove `update` conflict

ajroetker:
This commit updates `specter` to the laster version to remove the
conflict with the name `update` in `clojure.core`.


ajroetker:
Tests failed because of the failures on stable, will rebase

PR #1509, opened 2015-06-30
(DOC-2087) Point to correct install docs for PC1...

jtappa:
...and update links to enabling package repositories page.
Also related : DOC-2088


wkalt:
I'm +1 on this. Just for clarity, my understanding is that these PDB changes represent DOCUMENT-2088. The docs this now link to look like they've been updated to reflect DOCUMENT-2087. @senior does this match what you were looking for?

PR #1512, opened 2015-07-02
(PDB-1701) reject old commands over http

wkalt:
This adds some validation on our command handler to reject retired versions of
commands and commands that we don't recognize before enqueuing when submitted
over http.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -299,11 +299,8 @@
   [{:keys [version] :as command} {:keys [db]}]
   (store-report* 5 db command))
 
-(def supported-commands
-  #{"replace facts" "replace catalog" "store report" "deactivate node"})
-
 (def supported-command?
-  (comp supported-commands :command))
+  (comp (kitchensink/valset command-names) :command))
 
 (defservice command-service
   [[:PuppetDBServer shared-globals]
--- a/src/puppetlabs/puppetdb/http/command.clj
+++ b/src/puppetlabs/puppetdb/http/command.clj
@@ -1,12 +1,40 @@
 (ns puppetlabs.puppetdb.http.command
   (:require [puppetlabs.puppetdb.command.constants :refer [command-names]]
             [puppetlabs.trapperkeeper.core :refer [defservice]]
+            [clojure.string :as str]
             [clojure.tools.logging :as log]
             [puppetlabs.puppetdb.command :as command]
             [puppetlabs.puppetdb.http :as http]
+            [puppetlabs.puppetdb.cheshire :as json]
             [net.cgrand.moustache :as moustache]
             [puppetlabs.puppetdb.middleware :as mid]
             [compojure.core :as compojure]))
+
+(def min-supported-commands
+  {"replace catalog" 6
+   "replace facts" 4
+   "store report" 5
+   "deactivate node" 3})
+
+(def valid-commands-str (str/join ", " (sort (vals command-names))))
+
+(defn validate-command-version
+  [app]
+  (fn [{:keys [body-string] :as req}]
+    (let [{:keys [command version]} (json/parse-string body-string true)
+          min-supported (get min-supported-commands command ::invalid)]
+      (cond
+        (= ::invalid min-supported)
+        (http/bad-request-response
+          (format "Supported commands are %s. Received '%s'."
+                  valid-commands-str command))
+
+        (< version min-supported)
+        (http/bad-request-response
+          (format "%s version %s is retired. The minimum supported version is %s."
+                  command version min-supported))
+
+        :else (app req)))))
 
 (defn enqueue-command-handler
   "Enqueues the command in request and returns a UUID"
@@ -27,6 +55,7 @@
         app (moustache/app
              ["v1" &] {:any (enqueue-command-handler connection endpoint)})]
     (-> app
+        validate-command-version
         mid/verify-accepts-json
         mid/verify-checksum
         (mid/validate-query-params {:optional ["checksum"]})
--- a/src/puppetlabs/puppetdb/http.clj
+++ b/src/puppetlabs/puppetdb/http.clj
@@ -254,3 +254,8 @@
   "Produces a json response for when an entity (catalog/nodes/environment/...) is not found."
   [type id]
   (json-response {:error (format "No information is known about %s %s" type id)} status-not-found))
+
+(defn bad-request-response
+  "Produce a json 400 response with an :error key holding message."
+  [message]
+  (json-response {:error message} status-bad-request))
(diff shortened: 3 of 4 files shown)


src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -1,19 +1,45 @@
...
+            [puppetlabs.puppetdb.cheshire :as json]
             [net.cgrand.moustache :as moustache]
             [puppetlabs.puppetdb.middleware :as mid]
             [compojure.core :as compojure]))
 
+(def min-supported-commands
+  {"replace catalog" 6
+   "replace facts" 4
+   "store report" 5
+   "deactivate node" 3})
```
ajroetker: Maybe a map with all supported versions? And we could just reject anything not in the map?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -1,19 +1,45 @@
...
+
+(def valid-commands (str/join ", " (sort (keys min-supported-commands))))
```
ajroetker: There's already something like this in `puppetlabs.puppetdb.command` maybe we could figure out a way to use that?

src/puppetlabs/puppetdb/http/command.clj
wkalt: Not sure I follow -- want to give an example? It sounds like you're suggesting this

```
{"replace catalog" [3 4 5 6]
 "replace facts" [4 5 6]
 ...}
```

I don't like that because it'll require an update every time we create a new command. The only important information here is the minimum supported.

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -1,19 +1,45 @@
...
   "Enqueues the command in request and returns a UUID"
   [connection endpoint]
   (fn [{:keys [body-string] :as request}]
-    (let [uuid (command/enqueue-raw-command! connection endpoint body-string)]
-      (http/json-response {:uuid uuid}))))
+    (let [{:keys [command version]} (json/parse-string body-string true)
+          min-supported (get min-supported-commands command ::invalid)]
+      (cond
+        (= ::invalid min-supported)
+        (http/json-response
```
ajroetker: Is this not just `error-response`?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -1,19 +1,45 @@
...
+
 (defn enqueue-command-handler
   "Enqueues the command in request and returns a UUID"
   [connection endpoint]
   (fn [{:keys [body-string] :as request}]
-    (let [uuid (command/enqueue-raw-command! connection endpoint body-string)]
-      (http/json-response {:uuid uuid}))))
+    (let [{:keys [command version]} (json/parse-string body-string true)
+          min-supported (get min-supported-commands command ::invalid)]
+      (cond
```
ajroetker: What would you think about putting this in a middleware? `validate-command` or some such?

src/puppetlabs/puppetdb/http/command.clj
wkalt: that doesn't return JSON. We use it in places but I think we should swap it out. There's a ticket with links to the company guidelines here: https://tickets.puppetlabs.com/browse/PDB-1511

wkalt:
@ajroetker changed the implementation to match our middleware application

wkalt:

src/puppetlabs/puppetdb/command.clj
```diff
@@ -299,11 +299,8 @@
   [{:keys [version] :as command} {:keys [db]}]
   (store-report* 5 db command))
 
-(def supported-commands
-  #{"replace facts" "replace catalog" "store report" "deactivate node"})
-
 (def supported-command?
-  (comp supported-commands :command))
+  (comp (set (vals command-names)) :command))
```
ajroetker: https://github.com/puppetlabs/clj-kitchensink/blob/master/src/puppetlabs/kitchensink/core.clj#L435

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -1,13 +1,45 @@
 (ns puppetlabs.puppetdb.http.command
   (:require [puppetlabs.puppetdb.command.constants :refer [command-names]]
             [puppetlabs.trapperkeeper.core :refer [defservice]]
+            [clojure.string :as str]
```
ajroetker: Did we decide on what the standard for this namespace should be? AFAIK Clojure core used `string`

ajroetker: Nvm looks like `str` is what core uses.

src/puppetlabs/puppetdb/http/command.clj
ajroetker: Is there anyway to collect the map of supported versions from the `defmulti` @mullr @senior? I've never seen anything like that but seems like you should be able to check everything a defmethod is valid for.

src/puppetlabs/puppetdb/http/command.clj
ajroetker: Maybe there could be a `error-response-json` or `json-response-error` or `json-error-response`, something that wraps what you have here, like `(defn a-good-name [msg] (http/json-response {:error msg} http/status-bad-request)` or something.

src/puppetlabs/puppetdb/http/command.clj
wkalt: even if so, the defmulti contains versions that are not supported -- namely the ones being detected here. Is there a reason you are against using the minimum? Using a list of supported versions will also require a different error if the version is too high, since it wouldn't be retired in that case. As implemented here that's just handled the way it always has been.

ajroetker: My problem with the `min-supported-commands` is just that I could see users wanting something like a `/meta/v1/support_commands` endpoints, or even something in the `/pdb/meta/v1` endpoints like we've talked about, that would give you the the commands we support and the versions we support of those commands. This kind of overlaps with that idea but can't really share code with it.

mullr: @ajroetker `(methods my-method)`

mullr: That said, I think I agree that just our list of supported commands on what's registered is a little janky. I suppose we could add `^:supported` to each method and check that, but that feels a bit magical.

wkalt: I don't know ... ^:supported and the example map I gave above both seem janky to me without actual plans for the endpoint under discussion. If and when we end up making a supported-commands endpoint, I think refactoring this to share code with whatever implementation we choose would be pretty trivial.

ajroetker: Well I don't think this should block the PR being merged either way.

mullr: Agreed that we shouldn't block the PR on this. @wkalt is on the money about the refactor being easy, let's just go with it.

wkalt:
@ajroetker changed it to use the ks function and a wrapper for 400 responses.

wkalt:

ajroetker:
:+1: from me, @mullr?

mullr:
nice clean code, lgtm

PR #1513, opened 2015-07-03
(PDB-1157) Cleanup release notes for 3.0.0

kbarber:
This is a final cleanup for the 3.0.0 release notes, in preparation for
release.

### DIFF
--- a/documentation/release_notes.markdown
+++ b/documentation/release_notes.markdown
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
 [kahadb_corruption]: ./trouble_kahadb_corruption.html
 [pg_trgm]: http://www.postgresql.org/docs/current/static/pgtrgm.html
 [upgrading]: ./api/query/v4/preparing-for-3.0.html
+[puppetdb-module]: https://forge.puppetlabs.com/puppetlabs/puppetdb
 
 3.0.0
 -----
 
+PuppetDB 3.0.0 is a major release which introduces some major new features,
+with some breaking changes.
+
+### Contributors
+
+Andrew Roetker, Andrii Nikitiuk, Chris Price, Eric Sorenson, John Duarte,
+Jarie Tappa, Ken Barber, Brian LaMattery, Lars Windolf, Mathieu Parent,
+Matthaus Owens, Nick Fagerlund, Rob Braden, Rob Browning, Russell Mull,
+Ryan Senior, Scott Garman, Wayne Warren, Wyatt Alt, Jean Bond and
+Darin Perusich.
+
 ### Upgrading
 
-* For the best-possible performance and scaling capacity, we recommend
- PostgreSQL version 9.4 or newer with the [`pg_trgm`][pg_trgm]
- extension enabled, as explained [here][configure_postgres], and we
- have officially retired versions earlier than 9.4. HSQLDB is
- only recommended for local development because it has a number of
- scaling and operational issues.
+* This version of PuppetDB adopts the AIO layout to be consistent with other
+  Puppet Labs layouts, such as Puppet Server.
 
-* Make sure that all of your PuppetDB instances are shut down, and
- only upgrade one at a time.
+    This means the following paths have changed:
 
-* Make sure to upgrade your puppetdb-terminus package (on the host
- where your Puppet Master lives), and restart your master service.
+    * Primary configuration: /etc/puppetdb is now /etc/puppetlabs/puppetdb
+    * Logs: /var/log/puppetdb is now /var/log/puppetlabs/puppetdb
+    * Binaries: /usr/bin is now /opt/puppetlabs/bin
+    * Data: /var/lib/puppetdb is now /opt/puppetlabs/server/data/puppetdb
+    * Rundir: /var/run/puppetdb is now /var/run/puppetlabs/puppetdb
 
-* Ensure during a package upgrade that you analyze any changed
-  configuration files. For Debian you will receive warnings when
-  upgrading interactively about these files, and for RedHat based
-  distributions you will find that the RPM drops .rpmnew files that
-  you should diff and ensure that any new content is merged into your
-  existing configuration.
+    During upgrade we have written a script that will migrate your configuration
+    to the new locations. After upgrade you should double check the migrated
+    configuration to ensure everything is correct. Of course we recommend the
+    use of the [PuppetDB Puppet Module][puppetdb-module] version 5, as it will
+    automate a lot of this kind of management and has been modified specifically
+    for PuppetDB 3 support by default.
 
-* There are some migrations provided in this release, that may mean
-  you have to wait while these run after upgrading. Before upgrading
-  you should also ensure you have as much free disk space as you have
-  data, as migrations often require temporary duplication of table
-  data to succeed.
+* We have retired PostgreSQL earlier than 9.4. We recommended before upgrading
+  Puppet to upgrade PostgreSQL to version 9.4 or newer with the
+  [`pg_trgm`][pg_trgm] extension enabled, as explained
+  [here][configure_postgres], and we have officially retired versions earlier
+  than 9.4. The official PostgreSQL repositories are recommended for packaging, see the
+  ([YUM](https://wiki.postgresql.org/wiki/YUM\_Installation)) instructions or the
+  ([APT](https://wiki.postgresql.org/wiki/Apt)) instructions for more details.
 
-* You may notice some additional system load for the first 30 to 60
-  minutes once the application has started upgrading. This is
-  expected, and is due to internal changes in the way we check if
-  catalogs are up-to-date.
+* PuppetDB now only support Puppet 4 or higher at the moment and we are still looking
+  into what packaging requirements are needed to support older Puppet 3 support.
+  For now you must have installed Puppet 4 from Puppet Collections 1
+  before installing PuppetDB 3.
+
+* Before upgrading, make sure that all of your PuppetDB instances are shut down,
+  and only upgrade one at a time.
+
+* Make sure to upgrade your `puppetdb-terminus` package (on the host where your
+  Puppet Master lives), and restart your master service (this is typically called
+  `puppetserver`). It is worth noting that the `puppetdb-terminus` package has
+  now been renamed to `puppetdb-termini`.
+
+* There are some database schema changes (also known as migrations) provided in
+  this release, that may mean you have to wait while these run after upgrading.
+  Before upgrading you should also ensure you have as much free disk space as
+  you have data, as migrations often require temporary duplication of table
+  data.
+
+* You may notice some additional system load for the first 30 to 60 minutes once
+  the application has started upgrading. This is expected, and is due to
+  internal changes in the way we check if objects (like catalogs) are up-to-date.
 
 ### Downgrading
 
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
 the stable API. The following changes have been made since the last
 stable release of PuppetDB, which included an experimental v4 API.
 
-* Retire all previously deprecated commands ([PDB-840](https://tickets.puppetlabs.com/browse/PDB-840))
+* The version 2 and 3 of the query API are now retired, and v4 experimental is now laid out quite differently.
+
+    Full documentation outlining this change is available [in the API upgrading guide][upgrading].
+
+* We've retired some previously deprecated commands ([PDB-840](https://tickets.puppetlabs.com/browse/PDB-840))
 
     The currently supported commands and versions are now:
 
     * replace catalog, v6
     * replace facts, v4
-    * deactivate node, v2
+    * deactivate node, v3
     * store report, v5
 
-* All API commands and responses now use underscores, instead of dashes.  ([PDB-698](https://tickets.puppetlabs.com/browse/PDB-698))
+    We've insured that internally the previous version of the commands will still be processed, so you
+    don't lose the data being processed in your queue during an upgrade.
+
+* All API commands and responses now use underscores, instead of dashes. ([PDB-698](https://tickets.puppetlabs.com/browse/PDB-698))
 
     This patch changes our API commands and responses to use underscores instead of
     dashes, which will make it easier to interface with PDB using languages like
-    javascript and python.
+    javascript and python. Consult the various API endpoints for the exact new
+    parameters if you are uncertain.
 
-* Split the `/v3/commands` endpoint into its own API served at `/pdb/cmd/v1`.
+* Split the comman endpoint (`/v3/commands` or `/v4/commands`) into their own
+  endpoint namespace served at `/pdb/cmd/v1`. ([PDB-1534](https://tickets.puppetlabs.com/browse/PDB-1534)) ([PDB-1561](https://tickets.puppetlabs.com/browse/PDB-1561))
 
-* Change "name" key to "certname" ([PDB-1099](https://tickets.puppetlabs.com/browse/PDB-1099))
+    The commands endpoint has now been separated from queries so that we can
+    version and manage this endpoint separately.
 
-    For clarity and consistency across endpoints, the "name" key has
-    been changed to "certname" where appropriate. This affects
-    responses from the /pdb/query/v4/catalogs endpoint, queries to it, and the
-    "replace catalog" and "replace facts" commands.
+* The `/metrics` endpoint now has its own top level namespace `/pdb/metrics/v1` ([PDB-800](https://tickets.puppetlabs.com/browse/PDB-800))
+
+    This is something we'd been planning on doing for a while, but separation
+    means we can manage and version this endpoint separately. There are plans
+    to have this endpoint supported in other Puppet Labs products also, and
+    this move gears us up for that.
+
+* Version and server-time now belongs in its own endpoint (meta) ([PDB-1562](https://tickets.puppetlabs.com/browse/PDB-1562))
+
+    These endpoints now live in `/pdb/meta/v1`.
+
+* Change `name` key to `certname` across our API ([PDB-1099](https://tickets.puppetlabs.com/browse/PDB-1099))
+
+    For clarity and consistency across endpoints, the `name` key has
+    been changed to `certname` where appropriate. This affects
+    responses from the `/pdb/query/v4/catalogs` endpoint, queries to it, and the
+    `replace catalog` and `replace facts` commands.
 
 * Expanded data format is now more formalized. ([PDB-1228](https://tickets.puppetlabs.com/browse/PDB-1228))
 
-    In the past in the experimental /v4 endpoint we used to return related data for endpoints directly
+    In the past in the experimental `/v4` endpoint we used to return related data for endpoints directly
     in the JSON response with very little formality. However, due to internal changes that only work for
     PostgreSQL we've decided to formalize the format of returning expanded/related data so that some of
     it may be marked as optional from a contractual perspective.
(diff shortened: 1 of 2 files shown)


documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+
+* We have retired PostgreSQL earlier than 9.4. We recommended before upgrading
+  Puppet to upgrade PostgreSQL to version 9.4 or newer with the
+  [`pg_trgm`][pg_trgm] extension enabled, as explained
+  [here][configure_postgres], and we have officially retired versions earlier
+  than 9.4. The official PostgreSQL repositories are recommended for packaging, see the
+  ([YUM](https://wiki.postgresql.org/wiki/YUM\_Installation)) instructions or the
+  ([APT](https://wiki.postgresql.org/wiki/Apt)) instructions for more details.
+
+* PuppetDB now only support Puppet 4 or higher at the moment and we are still looking
```
ajroetker: Kind of confusing line, could be read as `PuppetDB now only supports Puppet 4` (also a typo here with `support`). Maybe this line should be `At the moment PuppetDB 3 only supports Puppet 4. We are investigating packaging requirements needed to support Puppet 3, and may support Puppet 3 in a future release. For now...`

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+* Before upgrading, make sure that all of your PuppetDB instances are shut down,
+  and only upgrade one at a time.
+
+* Make sure to upgrade your `puppetdb-terminus` package (on the host where your
+  Puppet Master lives), and restart your master service (this is typically called
+  `puppetserver`). It is worth noting that the `puppetdb-terminus` package has
+  now been renamed to `puppetdb-termini`.
+
+* There are some database schema changes (also known as migrations) provided in
+  this release, that may mean you have to wait while these run after upgrading.
```
ajroetker: `wait a while`

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+  Before upgrading you should also ensure you have as much free disk space as
+  you have data, as migrations often require temporary duplication of table
+  data.
```
ajroetker: `Before upgrading you should ensure you have as much free disk space as you have data. Database schema changes (also known as migrations) in this release often require temporary duplication of table data and may mean you have wait during the upgrade as these migrations run.`

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+
```
ajroetker: There should be a `Before upgrading` subsection it feels like, as well as an `On the Master` subsection.

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+* You may notice some additional system load for the first 30 to 60 minutes once
+  the application has started upgrading. This is expected, and is due to
+  internal changes in the way we check if objects (like catalogs) are up-to-date.
```
ajroetker: s/like/such as

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+  now been renamed to `puppetdb-termini`.
+
+* There are some database schema changes (also known as migrations) provided in
+  this release, that may mean you have to wait while these run after upgrading.
+  Before upgrading you should also ensure you have as much free disk space as
+  you have data, as migrations often require temporary duplication of table
+  data.
+
+* You may notice some additional system load for the first 30 to 60 minutes once
+  the application has started upgrading. This is expected, and is due to
```
ajroetker: s/This is expected, and is due/This is due

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+  `puppetserver`). It is worth noting that the `puppetdb-terminus` package has
+  now been renamed to `puppetdb-termini`.
+
+* There are some database schema changes (also known as migrations) provided in
+  this release, that may mean you have to wait while these run after upgrading.
+  Before upgrading you should also ensure you have as much free disk space as
+  you have data, as migrations often require temporary duplication of table
+  data.
+
+* You may notice some additional system load for the first 30 to 60 minutes once
```
ajroetker: `You may notice some additional system load for the up to the 60 minutes after PuppetDB has upgraded...` The 30-60 min implies a lower bound, which negates the may

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
 the stable API. The following changes have been made since the last
 stable release of PuppetDB, which included an experimental v4 API.
 
-* Retire all previously deprecated commands ([PDB-840](https://tickets.puppetlabs.com/browse/PDB-840))
+* The version 2 and 3 of the query API are now retired, and v4 experimental is now laid out quite differently.
```
ajroetker: Maybe `The v4 query API is quite different from the PuppetDB 2 experimental version`

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
...
     The currently supported commands and versions are now:
 
     * replace catalog, v6
     * replace facts, v4
-    * deactivate node, v2
+    * deactivate node, v3
     * store report, v5
 
-* All API commands and responses now use underscores, instead of dashes.  ([PDB-698](https://tickets.puppetlabs.com/browse/PDB-698))
+    We've insured that internally the previous version of the commands will still be processed, so you
```
ajroetker: ensured

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
...
+    don't lose the data being processed in your queue during an upgrade.
```
ajroetker: Older versions of the commands on your queue will still be processed by PuppetDB so you won't lose the data from messages still being processed by PuppetDB before an upgrade.

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
...
+    javascript and python. Consult the various API endpoints for the exact new
+    parameters if you are uncertain.
+
+* Split the comman endpoint (`/v3/commands` or `/v4/commands`) into their own
+  endpoint namespace served at `/pdb/cmd/v1`. ([PDB-1534](https://tickets.puppetlabs.com/browse/PDB-1534)) ([PDB-1561](https://tickets.puppetlabs.com/browse/PDB-1561))
+
+    The commands endpoint has now been separated from queries so that we can
+    version and manage this endpoint separately.
+
+* The `/metrics` endpoint now has its own top level namespace `/pdb/metrics/v1` ([PDB-800](https://tickets.puppetlabs.com/browse/PDB-800))
```
ajroetker: It's actually just `/metrics/v1`, we don't prefix it with `/pdb`

documentation/release_notes.markdown
```diff
@@ -248,6 +320,8 @@ derive data from a PuppetDB export tarball.
 
 * Warn user with unknown config items ([PDB-1067](https://tickets.puppetlabs.com/browse/PDB-1067))
 
+* We now use `honeysql` to generate SQL ([PDB-1228](https://tickets.puppetlabs.com/browse/PDB-1228))
```
ajroetker: Maybe a note about how this affects grabbing the SQL from the debug log?

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
 
-* Remove 'com.' prefix from JMX and ActiveMQ endpoints ([PDB-863](https://tickets.puppetlabs.com/browse/PDB-863))
+* Remove `com.` prefix from JMX and ActiveMQ endpoints ([PDB-863](https://tickets.puppetlabs.com/browse/PDB-863))
 
-    The 'com.' prefix has been removed from the JMX and ActiveMQ endpoints. They are now in the
-    puppetlabs.puppetdb namespace.
+    The `com.` prefix has been removed from the JMX and ActiveMQ endpoints. They are now in the
+    `puppetlabs.puppetdb` namespace. Any monitoring applications you have that
+    expect the old namespace will need to be changed, along with the general
+    URL change from `/v4/metrics` to `/pdb/metrics/v1`.
```
ajroetker: s/`/pdb/metrics/v1`/`/metrics/v1`

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+    configuration to ensure everything is correct. Of course we recommend the
+    use of the [PuppetDB Puppet Module][puppetdb-module] version 5, as it will
+    automate a lot of this kind of management and has been modified specifically
+    for PuppetDB 3 support by default.
+
+* We have retired PostgreSQL earlier than 9.4. We recommended before upgrading
+  Puppet to upgrade PostgreSQL to version 9.4 or newer with the
+  [`pg_trgm`][pg_trgm] extension enabled, as explained
+  [here][configure_postgres], and we have officially retired versions earlier
+  than 9.4. The official PostgreSQL repositories are recommended for packaging, see the
```
wkalt: we say "retired versions older than 9.4" twice in this sentence. "We recommended before upgrading Puppet" is also misworded, and I think Puppet should be PuppetDB. "Users must upgrade PostgreSQL to 9.4 or newer before upgrading PuppetDB" or similar seems a bit cleaner.

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+  ([YUM](https://wiki.postgresql.org/wiki/YUM\_Installation)) instructions or the
+  ([APT](https://wiki.postgresql.org/wiki/Apt)) instructions for more details.
+
+* PuppetDB now only support Puppet 4 or higher at the moment and we are still looking
+  into what packaging requirements are needed to support older Puppet 3 support.
```
wkalt: the last "support" should be dropped at a minimum

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
+    an API response when querying data.
 
-    On factsets the field is queryable and orderable.
+* Support fallback PuppetDB connections ([PDB-100](https://tickets.puppetlabs.com/browse/PDB-100))
 
-* Add `resource_events` key to the `store report` command ([PDB-1072](https://tickets.puppetlabs.com/browse/PDB-1072))
+    When configuring your connection to PuppetDB, you can now specify a fallback.
+    Instead of the old `server` and `port` config, you can now use a
+    `server\_urls` config that contains the full path to one or more PuppetDB
+    URLs. The old server/port format is still supported.
```
ajroetker: The old format is not supported we retired it in 3 and deprecated in 2.

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
+
+* Added `producer\_timestamp` field to reports ([PDB-1487](https://tickets.puppetlabs.com/browse/PDB-1487))
+
+* Added `resource\_events` key to the `store report` command ([PDB-1072](https://tickets.puppetlabs.com/browse/PDB-1072))
+
+    API responses can now be easily used for resubmission as a
+    report. This will be helpful in the HA context, where it will reduce
+    the number of requests needed to synchronize reports.
+
+* The command `deactivate node` now requires a `producer_timestamp` field ([PDB-1310](https://tickets.puppetlabs.com/browse/PDB-1310))
```
ajroetker: Why do we escape the underscore sometimes but not others?

documentation/release_notes.markdown
```diff
@@ -271,17 +345,17 @@ derive data from a PuppetDB export tarball.
...
+* Fix errors on PDB shutdown ([PDB-880](https://tickets.puppetlabs.com/browse/PDB-880)) ([PDB-1102](https://tickets.puppetlabs.com/browse/PDB-1102))
 
-    This patch bumps activemq to 5.7.0.  This seems to prevent two errors that
-    would sometimes occur on shutdown, one where the broker was stopped before its consumers
-    were removed and another where KahaDB batching fails to reset due to the PageFile
-    not being loaded.
+    To fix errors on PDB shutdown, we upgrade ActiveMQ to 5.7.0 (later 5.11.1). This seems to
+    prevent two errors that would sometimes occur on shutdown, one where the broker was stopped
+    before its consumers were removed and another where KahaDB batching fails to reset due to
+    the PageFile not being loaded.
```
ajroetker: This should be reworded, it kind of sounds like we don't really know if it fixed things.
`ActiveMQ was upgraded to ... . This was done to prevent two errors ... .`

documentation/release_notes.markdown
```diff
@@ -306,13 +378,44 @@ derive data from a PuppetDB export tarball.
...
+
+* Improve the error message returned for the `distinct_resources` parameter on the `aggregate-event-counts` endpoint ([PDB-1391](https://tickets.puppetlabs.com/browse/PDB-1391))
+
+* Detangle query-eng database and http streaming to simplify code ([PDB-1465](https://tickets.puppetlabs.com/browse/PDB-1465))
+
+* Handle comparison queries with non-numerics on the `/facts` endpoint ([PDB-1453](https://tickets.puppetlabs.com/browse/PDB-1453))
+
+* Consider reports as part of node expiration, and don't expire nodes that have had recent reports ([PDB-1638](https://tickets.puppetlabs.com/browse/PDB-1638))
+
+* Add a `--url-prefix` option to the import and export commands in the PuppetDB CLI  ([PDB-1068](https://tickets.puppetlabs.com/browse/PDB-1068))
```
ajroetker: I believe this was actually removed.

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+    use of the [PuppetDB Puppet Module][puppetdb-module] version 5, as it will
+    automate a lot of this kind of management and has been modified specifically
+    for PuppetDB 3 support by default.
+
+* We have retired PostgreSQL earlier than 9.4. We recommended before upgrading
+  Puppet to upgrade PostgreSQL to version 9.4 or newer with the
+  [`pg_trgm`][pg_trgm] extension enabled, as explained
+  [here][configure_postgres], and we have officially retired versions earlier
+  than 9.4. The official PostgreSQL repositories are recommended for packaging, see the
+  ([YUM](https://wiki.postgresql.org/wiki/YUM\_Installation)) instructions or the
```
ajroetker: Should this be moved to an alias at the top?

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+  ([APT](https://wiki.postgresql.org/wiki/Apt)) instructions for more details.
```
ajroetker: Should this be moved to an alias at the top?

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
+* Add a `-v` or `--version` flag to the CLI to get the PuppetDB version. ([PDB-992](https://tickets.puppetlabs.com/browse/PDB-922))
+
+* The reports endpoint now implements the `latest_report?` query, which
+  filters the response to only include reports that are from the
+  latest puppet run for each node.
+  ([PDB-1244](https://tickets.puppetlabs.com/browse/PDB-1244))
 
-    This supports a single column or multiple columns. It has been tested for
-    facts, events, nodes, reports and resources
+* We have now added [edges](./api/query/v4/edges.html) querying capability to the top-level namespace. You can also query edges specific to a particular catalog using the new `edges` suffix on the [catalogs](./api/query/v4/catalogs.html) endpoint. ([PDB-1228](https://tickets.puppetlabs.com/browse/PDB-1228)).
```
ajroetker: Should this be moved to an alias at the top?

documentation/release_notes.markdown
```diff
@@ -8,42 +8,72 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+  For now you must have installed Puppet 4 from Puppet Collections 1
+  before installing PuppetDB 3.
+
+* Before upgrading, make sure that all of your PuppetDB instances are shut down,
+  and only upgrade one at a time.
+
+* Make sure to upgrade your `puppetdb-terminus` package (on the host where your
+  Puppet Master lives), and restart your master service (this is typically called
+  `puppetserver`). It is worth noting that the `puppetdb-terminus` package has
+  now been renamed to `puppetdb-termini`.
```
wkalt: seems like the rename should be at the top of this bullet if upgrading puppetdb-terminus will leave users with the wrong package

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
...
+
+* All API commands and responses now use underscores, instead of dashes. ([PDB-698](https://tickets.puppetlabs.com/browse/PDB-698))
 
     This patch changes our API commands and responses to use underscores instead of
     dashes, which will make it easier to interface with PDB using languages like
-    javascript and python.
+    javascript and python. Consult the various API endpoints for the exact new
+    parameters if you are uncertain.
+
+* Split the comman endpoint (`/v3/commands` or `/v4/commands`) into their own
```
wkalt: s/comman/command ; s/their/its

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
...
+  endpoint namespace served at `/pdb/cmd/v1`. ([PDB-1534](https://tickets.puppetlabs.com/browse/PDB-1534)) ([PDB-1561](https://tickets.puppetlabs.com/browse/PDB-1561))
+
+    The commands endpoint has now been separated from queries so that we can
+    version and manage this endpoint separately.
```
wkalt: could be combined with line 118

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
...
+
+* The `/metrics` endpoint now has its own top level namespace `/pdb/metrics/v1` ([PDB-800](https://tickets.puppetlabs.com/browse/PDB-800))
+
+    This is something we'd been planning on doing for a while, but separation
+    means we can manage and version this endpoint separately. There are plans
+    to have this endpoint supported in other Puppet Labs products also, and
+    this move gears us up for that.
+
+* Version and server-time now belongs in its own endpoint (meta) ([PDB-1562](https://tickets.puppetlabs.com/browse/PDB-1562))
```
wkalt: are now mounted under the `metadata` API

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
...
+* Change `name` key to `certname` across our API ([PDB-1099](https://tickets.puppetlabs.com/browse/PDB-1099))
 
-    For clarity and consistency across endpoints, the "name" key has
-    been changed to "certname" where appropriate. This affects
-    responses from the /pdb/query/v4/catalogs endpoint, queries to it, and the
-    "replace catalog" and "replace facts" commands.
+    For clarity and consistency across endpoints, the `name` key has
+    been changed to `certname` where appropriate. This affects
+    responses from the `/pdb/query/v4/catalogs` endpoint, queries to it, and the
+    `replace catalog` and `replace facts` commands.
```
wkalt: also affects the nodes endpoint. If we're listing the specific spots, might as well drop "where appropriate".

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
...
 
 * Expanded data format is now more formalized. ([PDB-1228](https://tickets.puppetlabs.com/browse/PDB-1228))
```
wkalt: we should figure out a way to express this more clearly. "Expanded data format" doesn't mean anything to the consumers of this document, and didn't exist prior to this upgrade.

documentation/release_notes.markdown
```diff
@@ -61,33 +91,56 @@ This release retires all api versions before v4 and promotes v4 to be
...
 
-    In the past in the experimental /v4 endpoint we used to return related data for endpoints directly
+    In the past in the experimental `/v4` endpoint we used to return related data for endpoints directly
```
wkalt: similarly, "related data" doesn't mean anything to users. That could describe all the data in a given blob, whereas we specifically mean nested JSON.

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
     For API consumers this means, the extra `data` key needs to be traversed to access the real related
     object data.
 
-* Add "noop" flag to reports ([PDB-1177](https://tickets.puppetlabs.com/browse/PDB-1177))
+* Add `noop` flag to reports ([PDB-1177](https://tickets.puppetlabs.com/browse/PDB-1177))
```
wkalt: Add a...

ajroetker: Should this be `field`? and not flag

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
 
-    This adds a "noop" field to the reports object, which is a
+    This adds a `noop` field to the reports object, which is a
     required boolean flag indicating whether the run producing the
-    report was a --noop.
+    report was using `--noop` (for example: `puppet agent -t --noop`).
```
wkalt: ... report used `--noop`

documentation/release_notes.markdown
wkalt: to add to that, I'm not sure there's a reason to talk about the endpoint change here. We mention that somewhere else in this doc.

documentation/release_notes.markdown
ajroetker: Or `was run with the `--noop` flag`

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
+    URL change from `/v4/metrics` to `/pdb/metrics/v1`.
+
+* The endpoints `aggregate-event-counts` and `event-counts` have now been marked as experimental.
+
+    These endpoints were introduced to solve specific problems for the PE console,
+    however the design is under question. We're moving these endpoints back to
+    experimental status so that we are able to deliver a better option later on.
+
+* The older puppetdb-\* sub-commands are now retired (`puppetdb-ssl-setup` for example) instead,
+  the new sub-command style should be used `puppetdb ssl-setup`. ([PDB_663](https://tickets.puppetlabs.com/browse/PDB-663))
```
wkalt: what's the difference from 2.x here? This is what we currently document https://docs.puppetlabs.com/puppetdb/latest/install_from_source.html#step-3-option-a-run-the-ssl-configuration-script

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
-    * Many symlinks were removed from /usr/share/puppetdb that were pointing
-      at the /var/lib/puppetdb and /etc/puppetdb/ spaces. This finishes a partially
-      completed migration from the past.
+* PuppetDB now has a new filesystem layout that complies with the overall Puppet AIO layout ([PDB-1455](https://tickets.puppetlabs.com/browse/PDB-1455))
 
-### New Features
+* PuppetDB now uses the Ezbake build system which changes the way we package ([PDB-663](https://tickets.puppetlabs.com/browse/PDB-663))
 
-* Add support for hyphenated classnames, supported by Puppet 2.7.x - 3.7.x. ([PDB-1024](https://tickets.puppetlabs.com/browse/PDB-1024))
+* Due to the Ezbake changes, the `[repl]` block has been renamed to `[nrepl]` and `type` is no longer an accpted parameter ([PDB-663](https://tickets.puppetlabs.com/browse/PDB-663))
```
wkalt: accepted

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
-* Add a '-v' flag to the CLI to get the PuppetDB version.
+* The endpoints contained in `/experimental` have now been removed ([PDB-1127](https://tickets.puppetlabs.com/browse/PDB-1127))
 
-* The `reports` database table was using the hash string as its
-  primary key, we know have switched to using a smaller bigint primary
-  key for that table, which should result in faster joins in most
-  cases, and hopefully smaller foreign key indexes sizes for tables
-  relating to
-  `reports`. ([PDB-1218](https://tickets.puppetlabs.com/browse/PDB-1218))
+* Return proper `404` errors for child endpoints where there parents do not exist, instead of empty lists ([PDB-1473](https://tickets.puppetlabs.com/browse/PDB-1473))
```
wkalt: we should define "child endpoints" somewhere. In previous docs I think we've informally called them "shorthand endpoints".

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
+* We now store report metrics ([PDB-1192](https://tickets.puppetlabs.com/browse/PDB-1192))
 
-* Add pagination to service query method ([PDB-1071](https://tickets.puppetlabs.com/browse/PDB-1071))
+* We now store report logs ([PDB-1192](https://tickets.puppetlabs.com/browse/PDB-1184))
 
-    Add a `paging_options` argument to the service query method.
+* We have added `count` aggregation capabilities to the query API ([PDB-1181](https://tickets.puppetlabs.com/browse/PDB-1181))
 
-* Make `/pdb/query/v4/catalogs` queryable ([PDB-1028](https://tickets.puppetlabs.com/browse/PDB-1028))
+* Support `extract` as a top level query operator ([PDB-207](https://tickets.puppetlabs.com/browse/PDB-207))
```
wkalt: seems like half of this section is written "Added X" and half "We have added support for X". We should standardize.

documentation/release_notes.markdown
wkalt: top-level namespace isn't accurate with the pathing changes. We have added an edges endpoint to the query API.

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
+* We have now added [edges](./api/query/v4/edges.html) querying capability to the top-level namespace. You can also query edges specific to a particular catalog using the new `edges` suffix on the [catalogs](./api/query/v4/catalogs.html) endpoint. ([PDB-1228](https://tickets.puppetlabs.com/browse/PDB-1228)).
 
-#### Operational
+* We've added the -A flag to our benchmark tool to simplify passing sample data ([PDB-1249](https://tickets.puppetlabs.com/browse/PDB-1249))
 
-* Add Apache-style access logging to PuppetDB packages, and enable it by default  ([PDB-477](https://tickets.puppetlabs.com/browse/PDB-477))
+* Now the default behaviour for benchmark is to use some supplied sample data, to lower the barrier for entry for the tooling ([PDB-1249](https://tickets.puppetlabs.com/browse/PDB-1249))
 
-* Add a `--url-prefix` option to the import and export commands in the PuppetDB CLI  ([PDB-1068](https://tickets.puppetlabs.com/browse/PDB-1068))
+* Added new endpoint `/factsets/<node>` for querying a particular node for its factset. ([PDB-1599](https://tickets.puppetlabs.com/browse/PDB-1599))
```
wkalt: child endpoint?

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
-    This commit associates a hash of each factset with the metadata of the factset
-    to allow us to easily tell if fact values have changed.
+    On factsets the field is queryable and orderable.
+
+* Added `producer\_timestamp` field to reports ([PDB-1487](https://tickets.puppetlabs.com/browse/PDB-1487))
+
+* Added `resource\_events` key to the `store report` command ([PDB-1072](https://tickets.puppetlabs.com/browse/PDB-1072))
+
+    API responses can now be easily used for resubmission as a
+    report. This will be helpful in the HA context, where it will reduce
```
wkalt: HA context is a PE concern -- readers may not be familiar with that work

documentation/release_notes.markdown
```diff
@@ -109,131 +162,150 @@ stable release of PuppetDB, which included an experimental v4 API.
...
+* The command `deactivate node` now requires a `producer_timestamp` field ([PDB-1310](https://tickets.puppetlabs.com/browse/PDB-1310))
+
+#### Performance
+
+* Switch from `text` based storage of UUID and hashes, to the PostgreSQL `uuid` and `bytea` types respectively ([PDB-1416](https://tickets.puppetlabs.com/browse/PDB-1416))
+
+* The `reports` database table was using the hash string as its
+  primary key, we know have switched to using a smaller bigint primary
+  key for that table, which should result in faster joins in most
+  cases, and hopefully smaller foreign key indexes sizes for tables
```
wkalt: I think we can drop hopefully

documentation/release_notes.markdown
```diff
@@ -306,13 +378,44 @@ derive data from a PuppetDB export tarball.
...
-* The first request to PuppetDB after DB backend restart fails ([PDB-707](https://tickets.puppetlabs.com/browse/PDB-707))
+* Add better validation to benchmark tool for `--runinterval` and `--numhosts` ([PDB-1268](https://tickets.puppetlabs.com/browse/PDB-1268)) 
+
+* Upgrade `clj-time`, `compojure` and `ring-core` to later revisions ([PDB-1460](https://tickets.puppetlabs.com/browse/PDB-1460))
+
+* Update `me.raynes/fs` to 1.4.5 ([PDB-1606](https://tickets.puppetlabs.com/browse/PDB-1606))
+
+* Improve the error message returned for the `distinct_resources` parameter on the `aggregate-event-counts` endpoint ([PDB-1391](https://tickets.puppetlabs.com/browse/PDB-1391))
+
+* Detangle query-eng database and http streaming to simplify code ([PDB-1465](https://tickets.puppetlabs.com/browse/PDB-1465))
```
wkalt: I can't read detangle without thinking about hair. Disentangle maybe?

documentation/release_notes.markdown
```diff
@@ -306,13 +378,44 @@ derive data from a PuppetDB export tarball.
...
+
+* Handle comparison queries with non-numerics on the `/facts` endpoint ([PDB-1453](https://tickets.puppetlabs.com/browse/PDB-1453))
+
+* Consider reports as part of node expiration, and don't expire nodes that have had recent reports ([PDB-1638](https://tickets.puppetlabs.com/browse/PDB-1638))
```
wkalt: "recent" is configurable via node-ttl

documentation/release_notes.markdown
```diff
@@ -306,13 +378,44 @@ derive data from a PuppetDB export tarball.
...
+
+* Add a `--url-prefix` option to the import and export commands in the PuppetDB CLI  ([PDB-1068](https://tickets.puppetlabs.com/browse/PDB-1068))
+
+    This allows import and export to work against a PuppetDB configured with a url prefix.
+
+* Add factset hash to `replace-facts!` ([PDB-898](https://tickets.puppetlabs.com/browse/PDB-898))
```
wkalt: this almost reads like it's been added to the wireformat, which isn't the case

senior:
I'll work on folding these changes in and resubmitting

documentation/release_notes.markdown
senior: Any opinions on this @nfagerlund ?

nfagerlund: Only bother with that if:

1. It's gonna be used multiple times, or 
2. There are a bunch of really similar links that might need to be updated at some point.

Otherwise there's no particular reason not to do it inline. 

ALSO, for this one, note that you do not want a backslash in front of that underscore. Underscores in link elements don't cause emphasis, so they don't need to be escaped, and escaping them will make the URL incorrect.

documentation/release_notes.markdown
senior: this seems ok to me

documentation/release_notes.markdown
senior: It's a boolean, we also call it flag in our API docs

documentation/release_notes.markdown
senior: We have puppetdb ssl-setup as the "right way" since way back to 1.6 but we hadn't removed/retired those other scripts until 3.0.0.

documentation/release_notes.markdown
senior: It looks like we still support it [here](https://github.com/puppetlabs/puppetdb/blob/stable/puppet/lib/puppet/util/puppetdb/config.rb#L26). Is there something I'm missing?

documentation/release_notes.markdown
senior: I think I'll leave this as is, the SQL is still there, just doesn't use somewhat arbitrary spaces that we embed in a string.

senior:
Closing and reopening as https://github.com/puppetlabs/puppetdb/pull/1518

PR #1515, opened 2015-07-06
(QENG-2545) Remove pe profile.

waynr:
This profile was used to build pe-puppetdb using lein-ezbake. Instead, we will
be building pe-puppetdb from the pe-puppetdb-extensions repo.


waynr:
@shrug @ajroetker @mullr @senior @kbarber ping

senior:
I think this should be targeted at stable, then we'll merge it up to master

PR #1517, opened 2015-07-06
(maint) Update `transaction_uuid` while benchmarking

ajroetker:
Prior to this commit we implicitly assumed that transaction_uuid's are
unique for every report and catalog, but cannot add the constraint to
the schema for a variety of reasons relating to upgrades. This commit
updates the benchmarking tool to change the transaction_uuid for a
catalog and its associated report like we do for timestamps.


senior:
:+1:

PR #1519, opened 2015-07-07
(QENG-2545) Remove pe profile.

waynr:
This profile was used to build pe-puppetdb using lein-ezbake. Instead, we will
be building pe-puppetdb from the pe-puppetdb-extensions repo.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,4 @@
 (def pdb-version "3.0.0-SNAPSHOT")
-(def pe-pdb-version "3.0.0-SNAPSHOT")
 
 (defn deploy-info
   "Generate deployment information from the URL supplied and the username and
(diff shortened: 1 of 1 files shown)


senior:
:+1:

ajroetker:
can https://github.com/puppetlabs/puppetdb/pull/1519/files#diff-0fff143854a4f5c0469a3819b978a483R2 be removed now?

waynr:
@ajroetker As far as I am aware, this profile wasn't being used anywhere in CI so it should be safe to remove whenever.

Someone more familiar with the CI that PuppetDB maintains would need to answer that--maybe @kbarber?

ajroetker:
@waynr Sorry I meant the `pe-puppetdb-version` string, it should be safe to remove right?

ajroetker:
This profile is definitely not used in CI anywhere

waynr:
@ajroetker oh right, I removed `pe-puppetdb-version`.

PR #1520, opened 2015-07-07
(PDB-1731) broken dashboard

wkalt:
We had been initializing dashboard metrics on an ephemeral database connection,
which meant that the connection would be closed when the metrics were accessed.


wkalt:
wasted too much time today writing http metrics tests that wouldn't have caught the original issue here. I'll make a new ticket for the test -- please verify this fix manually.

senior:
I'll test this

senior:
The fix works for me, looks like PR tests are finishing up now

PR #1522, opened 2015-07-09
(maint) fix config migration

wkalt:
Without this we were parsing the subname to get the location of the hsqldb data
directory so it can be migrated. In the default case though, this resides in the vardir, which is
migrated a moment before.

This patch changes things so that we only attempt to migrate that directory if
the original vardir still exists, which is a signal that the vardir is
nonstandard, but the subname still points to the old one (since it's still the
default value.)

### DIFF
--- a/resources/ext/cli/config-migration
+++ b/resources/ext/cli/config-migration
@@ -202,7 +202,9 @@ done
 echo "Checking to see if we need to update default settings"
 for ini_file in "${ini_files[@]}"; do
   if [ -f "${ini_file}" ] ; then
-    # Check settings are correct and fix or warn
+
+    # Check settings are correct and fix or warn. Do not move vardir behind
+    # subname.
     orig_settings=(
       "vardir:/var/lib/puppetdb"
       "logging-config:/etc/puppetdb/logback.xml"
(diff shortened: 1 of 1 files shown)


wkalt:
@senior this is sort of horrible, but I think it might do the trick.

ajroetker:
>Without this we were parsing the subname to get the location of the hsqldb data
>directory so it can be migrated. In the default case though, this resides in the vardir, which is
>migrated a moment before.

>This patch changes things so that we only attempt to migrate that directory if
Which is "that" directory?
>the original vardir still exists, which is a signal that the vardir is
"signal that the user had changed the vardir but not the hsqldb subname"
>nonstandard, but the subname still points to the old one (since it's still the
>default value.)
The commit message and PR message use a lot of pronouns that kind of made this hard to parse without a few rereads.

wkalt:
@ajroetker "that directory" is the hsqldb data directory.

resources/ext/cli/config-migration
```diff
@@ -236,13 +238,18 @@ for ini_file in "${ini_files[@]}"; do
...
-                  new_value="${new_value#file:}"
-                  new_value="${new_value%/db;hsqldb.tx=mvcc;sql.syntax_pgs=true}"
-                  echo "Copying over the contents of ${orig_value} to ${new_value}"
-                  migrate "$orig_value" "$new_value"
-                  rm -rf "$orig_value"
+                  orig_hsql_path="${orig_value#file:}"
+                  orig_hsql_dir="${orig_hsql_path%/db;hsqldb.tx=mvcc;sql.syntax_pgs=true}"
+                  new_hsql_path="${new_value#file:}"
+                  new_hsql_dir="${new_hsql_path%/db;hsqldb.tx=mvcc;sql.syntax_pgs=true}"
+                  ## if the default vardir still exists, then the vardir was
```
ajroetker: s/vardir/vardir config setting

ajroetker: and therefore was not migrated

resources/ext/cli/config-migration
```diff
@@ -236,13 +238,18 @@ for ini_file in "${ini_files[@]}"; do
...
+                  ## changed, but the subname still points to the default
+                  ## vardir.
```
ajroetker: default vardir for PuppetDB 2.x

resources/ext/cli/config-migration
```diff
@@ -236,13 +238,18 @@ for ini_file in "${ini_files[@]}"; do
...
+                  if [ -e "/var/lib/puppetdb" ] ; then
+                    echo "Copying over the contents of ${orig_hsql_dir} to ${new_hsql_dir}"
```
ajroetker: What happens if we find the vardir and subname settings in the opposite order while running the script?

ajroetker: Nevermind we're just looping so that wouldn't happen

ajroetker: Why don't we just run the subname migration first in the list?

resources/ext/cli/config-migration
wkalt: it's not the vardir config setting, it's the actual vardir

resources/ext/cli/config-migration
ajroetker: No we can't do that because the vardir migration will `mv` the old vardir on top of the new vardir, deleting the `mv`d hsqldb.

ajroetker: Can we just check if the old_hsql_dir exists instead of the absolute path?

resources/ext/cli/config-migration
ajroetker: The second `vardir` in this line is not though

PR #1525, opened 2015-07-09
(maint) Pin the module version we use in acceptance tests

ajroetker:
This commit pins the module version we use in acceptance tests as v5 of
the module was released today and contained breaking changes. This
commit is a temporary fix for the issue until we can upgrade to v5 of
the puppetdb module.


ajroetker:
@mullr @senior we should merge this in while my other better fix is in flight

PR #1528, opened 2015-07-10
[PDB 135] improve utf8 warning

kronn:
### Background

I was puzzled by the warning message and a bit worried that something could go wrong with the exported/collected resources.

### Changes

If the strings don't match, I now emit a second warning which includes a hint a file with further information. Since the string sent to PuppetDB is a JSON-string oneliner, I decided to split it up on a few known locations. This enabled the used of standard diff'ing tools for linewise diffing. In turn, I could compile a summary of the mismatching parts of the JSON.

In order to not build big or many files, I decided to only keep the last output. Each warning produces now 2 warnings in the puppet-log and 3 files in /tmp.  

In my personal case, I could determine the problematic areas and change them. 

### Problems / Loose Ends

I have not yet run the tests, so I half-way expect errors.

I also have not written any specs for the ruby-extension. If needed, I can add full spec-coverage to the internal classes.

Currently, I depend on the `diff(1)` command being in `$PATH`.

Since my installation has the puppetmaster and PuppetDB on the same machine, I don't know where the log will be written. As the terminus is writing the file, I expect it be on the puppetmaster's machine.

The current version is intended as a starting point for discussion and further improvement.

### DIFF
--- a/puppet/lib/puppet/util/puppetdb/char_encoding.rb
+++ b/puppet/lib/puppet/util/puppetdb/char_encoding.rb
@@ -234,6 +234,7 @@ module CharEncoding
         @string   = string
         @type     = type
         @filename = "#{base_filename}-#{@type}"
+        @errors   = []
       end
 
       def encoding
@@ -267,12 +268,19 @@ module CharEncoding
       # Splitting the string does the following two things:
       #   - split hashes and array that are values from their keys
       #   - split lists of hashes into new lines
-      def split
+      def split(second_try = false)
         begin
           @string.gsub(/(['"]:)({|\[)/, "\\1\n\\2").gsub(/(}|\]),({|\[)/, "\\1,\n\\2")
         rescue ArgumentError => e
-          @string.encode!(@string.encoding, :invalid => :replace, :undef => :replace, :replace => "?")
-          split
+          if second_try
+            @errors << "The #{@type} string cannot be split into diffable chunks."
+            @string
+          else
+            @string.encode!(@string.encoding, :invalid => :replace, :undef => :replace, :replace => "?")
+            @errors << "Invalid bytes encountered. These have been replaced with '?'."
+            @errors << "This should be enough to produce a helpful diff."
+            split(:second_try)
+          end
         end
       end
     end
@@ -300,6 +308,13 @@ module CharEncoding
         file.puts "diff of #{orig_fn} and #{conv_fn}"
         file.puts ""
         file.puts `diff -u5 #{orig_fn} #{conv_fn}`
+
+        [@original, @converted].each do |str|
+          if str.errors.size > 0
+            file.puts ""
+            file.puts str.errors
+          end
+        end
 
         file.puts diff_footer
 
--- a/puppet/lib/puppet/util/puppetdb/char_encoding.rb
+++ b/puppet/lib/puppet/util/puppetdb/char_encoding.rb
@@ -229,16 +229,36 @@ module CharEncoding
   class EncodingAnalyser
 
     class String
-      def initialize(string)
-        @string = string
-      end
+      attr_reader :filename
 
-      def split!
-        @string.gsub!(/(['"]:)({|\[)/, "\\1\n\\2").gsub!(/(}|\]),({|\[)/, "\\1,\n\\2")
+      def initialize(string, base_filename, type)
+        @string   = string
+        @type     = type
+        @filename = "#{base_filename}-#{@type}"
       end
 
       def encoding
         @string.encoding
+      end
+
+      def size
+        @string.size
+      end
+
+      def write
+        File.open(@filename, 'w') do |file|
+          file.flock(File::LOCK_EX)
+          file.puts split
+          file.chmod(0666)
+          file.flock(File::LOCK_UN)
+        end
+        filename
+      end
+
+      private
+
+      def split
+        @string.gsub(/(['"]:)({|\[)/, "\\1\n\\2").gsub(/(}|\]),({|\[)/, "\\1,\n\\2")
       end
     end
 
@@ -247,11 +267,11 @@ module CharEncoding
     end
 
     def original= string
-      @original = String.new(string)
+      @original = String.new(string, @filename, 'original')
     end
 
     def converted= string
-      @converted = String.new(string)
+      @converted = String.new(string, @filename, 'converted')
     end
 
     def write_summary
@@ -259,8 +279,8 @@ module CharEncoding
         file.flock(File::LOCK_EX)
         file.puts diff_header
 
-        orig_fn = write_string(@original.split!, 'original')
-        conv_fn = write_string(@converted.split!, 'converted')
+        orig_fn = @original.write
+        conv_fn = @converted.write
 
         file.puts "diff of #{orig_fn} and #{conv_fn}"
         file.puts ""
@@ -272,17 +292,6 @@ module CharEncoding
         file.chmod(0666)
         file.flock(File::LOCK_UN)
       end
-    end
-
-    def write_string(string, type)
-      fn = "#{@filename}-#{type}"
-      File.open(fn, 'w') do |file|
-        file.flock(File::LOCK_EX)
-        file.puts string
-        file.chmod(0666)
-        file.flock(File::LOCK_UN)
-      end
-      fn
     end
 
     def diff_header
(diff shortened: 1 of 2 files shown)


ajroetker:
@kronn Looks like there are some broken spec tests, you can run them from the `/puppet` directory with `bundle exec rspec`

kronn:
@ajroetker I adapted the specs in the least way possible to make them green again.

I am looking forward to a review of this (and to the code potentially being torn apart because I missed something).

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -206,6 +220,137 @@ def self.get_byte(str, index)
...
+      File.open(@filename, 'w') do |file|
+        file.flock(File::LOCK_EX)
+        file.puts diff_header
+
+        orig_fn = @original.write
+        conv_fn = @converted.write
+
+        file.puts "diff of #{orig_fn} and #{conv_fn}"
+        file.puts ""
+        file.puts `diff -u5 #{orig_fn} #{conv_fn}`
```
kbarber: External shell dependencies are probably not a great idea here, this terminus primarily now runs in the JVM given that puppet-server is the way forward for puppet masters. I haven't looked into this nor proven it myself, but I presume that a JRuby exec like this uses the clone system call in linux, which requires initially the same amount of RAM as the original JVM before it shrinks down to the necessary size. This can cause linux OOM_Killer to potentially initiate an OOM kill routine at time, unless administrators have tuned for this eventuation. I'm not a massive fan of putting this kind of risk to our users.

I think, if we really wanted to go down the direction this PR states, perhaps an avoidance of external exec's is a good idea. One can always place the two files to be diffed manually instead. Otherwise a diff that doesn't require an exec would be preferred, if such a thing exists without say - C or external dependencies in Ruby.

Not to mention GNU diff and BSD diff can be different, having something hard-coded like this can create compatibility hassles.

Anyway, you get the idea, shell execs are hard for multiple reasons.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -206,6 +220,137 @@ def self.get_byte(str, index)
...
+      @original = String.new(string, @filename, 'original')
+    end
+
+    def converted= string
+      @converted = String.new(string, @filename, 'converted')
+    end
+
+    def write_summary
+      File.open(@filename, 'w') do |file|
+        file.flock(File::LOCK_EX)
```
kbarber: Seems to me the flock here could be avoided completely, if we aren't overwriting the same file each time.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -71,6 +71,20 @@ def self.utf8_string(str)
   def self.warn_if_changed(str, converted_str)
     if converted_str != str
       Puppet.warning "Ignoring invalid UTF-8 byte sequences in data to be sent to PuppetDB"
+
+      begin
+        fn = '/tmp/puppetdb-invalid-utf8-sequences'
```
kbarber: This location seems too open, if we are going to log anything it should be in a log directory of some kind. Not only that, I would think a single file for logging might not be the right direction, users might have multiple problems at different times for example, and having a single file makes that a contention for debugging properly.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -206,6 +220,137 @@ def self.get_byte(str, index)
...
+        [@original, @converted].each do |str|
+          if str.errors.size > 0
+            file.puts ""
+            file.puts str.errors
+          end
+        end
+
+        file.puts diff_footer
+
+        file.chmod(0666)
```
kbarber: This is too permissive, be mindful the contents of these files might be sensitive.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -206,6 +220,137 @@ def self.get_byte(str, index)
...
+
+      def size
+        @string.size
+      end
+
+      def write
+        File.open(@filename, 'w') do |file|
+          file.flock(File::LOCK_EX)
+          file.puts split
+          file.chmod(0666)
```
kbarber: Too permissive, see my other comment.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
kbarber: More to that, even if we wanted to write to a single file, flock's aren't always the best way to do this. Dotfiles patterns are better, and moves are generally atomic which provides a less contentious way of doing this, avoiding the need for a flock again. Currently, this would cause multiple cases where we are receiving these UTF-8 issues to artificially serialize, which is going to slow down the master. Not to mention flock's carry semantic issues and can behave differently depending on the FS, it's just better generally to avoid them if one can for simple cases like this.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
kbarber: Also - any location for this kind of thing should be configurable, presumably in puppetdb.conf. This would also require doc changes, and if we need a new location, packaging concerns fwiw.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -206,6 +220,137 @@ def self.get_byte(str, index)
...
+
+      def errors
+        if @errors.any?
+          ["Errors in #{@type} string", @errors].join("\n")
+        else
+          ''
+        end
+      end
+
+      private
```
kbarber: Meh, don't make methods private, makes them harder to test. We use apidocs with yard conventions to mark methods as private: http://www.rubydoc.info/gems/yard/file/docs/Tags.md#private.

kbarber:
@kronn I don't mind the direction this patch is taking, but I'm wary of rushing this so we'll need to ponder it some more and at the very least get this code into shape a bit, it's a great start though. Some general notes:

* The use of this facility should probably be documented, at least in a FAQ entry perhaps. See documentation/
* Use of diff can be risky, see my comment.
* The pathing locations for the content needs to be considered more, /tmp isn't a great place.
* The use of flock feels like a smell, this needs to be reconsidered.
* Permissions are too permissive.
* The code needs rspec tests, we couldn't take on anything like this without at least rspec tests.
* We document our methods and code with yarddoc, this will need to be done across the board: http://www.rubydoc.info/gems/yard/file/docs/Tags.md
* FWIW our convention for commits is (PDB-135) not [PDB 135], we have bots that rely on this stuff now, so its worth fixing.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -71,6 +71,20 @@ def self.utf8_string(str)
...
+      begin
+        fn = '/tmp/puppetdb-invalid-utf8-sequences'
+
+        analyser = EncodingAnalyser.new(fn)
+        analyser.original = str
+        analyser.converted = converted_str
+        analyser.write_summary
+
+        Puppet.warning "See #{fn} for details of last occurence"
+      rescue => e
```
kbarber: Is it necessary to do a blanket swallow of the exception here? Is it necessary to just swallow and return a warning, what is wrong about the default behaviour with exceptions within Puppet?

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -206,6 +220,137 @@ def self.get_byte(str, index)
...
+      end
+
+      private
+
+      # Splitting the string does the following two things:
+      #   - split hashes and array that are values from their keys
+      #   - split lists of hashes into new lines
+      def split(second_try = false)
+        begin
+          @string.gsub(/(['"]:)({|\[)/, "\\1\n\\2").gsub(/(}|\]),({|\[)/, "\\1,\n\\2")
```
kbarber: Hmm, can you explain more about what the split here does? Do you have sample data that shows an example this need so I can understand it?

kbarber:
Oh yeah, this should be targeted at master, as its a new feature, stable should be mainly for bug fixes and this is probably too intrusive a change to fit that label. You'll need to re-open this PR unfortunately to do that, should have mentioned this before I commented on it, nevermind. When you reopen it just be mindful of the comments I've made, and reference this original PR somewhere in a comment or something.

kronn:
Cool, thanks for the great feedback so far. I will rebase this onto master and make a new PR with the right naming convention (sorry for this mental copy/paste error). 

I understand most of your points and - as stated in the PR description - this was intended as a starting point, not the final solution. Since the general idea seems to be good, I will rework this according to your suggestions.

PR #1536, opened 2015-07-11
(maint) Fix upgrade tests

mullr:
Use 'git describe' to find the most recent released version and use that
to find the base version to install for upgrade tests.

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -246,27 +246,69 @@ module PuppetDBExtensions
                       "https://#{host.node_name}:8081/", [35, 60])
   end
 
-  def get_package_version(host, version = nil)
-    return version unless version.nil?
+  def platform_is_rpm_based(platform)
+    platform.include?('el-') or platform.include?('fedora')
+  end
 
+  def platform_is_deb_based(platform)
+    platform.include?('ubuntu') or platform.include?('debian')
+  end
+
+  def expected_version_for_platform(platform)
+    if platform_is_rpm_based(platform)
+      PuppetDBExtensions.config[:expected_rpm_version]
+    elsif platform_is_deb_based(platform)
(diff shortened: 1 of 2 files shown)


acceptance/helper.rb
```diff
@@ -246,27 +246,68 @@ def sleep_until_started(host)
...
+
+    platform = host['platform']
+
+    if version.nil?
+      expected_version = expected_version_for_platform(platform)
+      return format_package_version_for_platform(platform, expected_version)
+    elsif version =~ /\d+\.\d+\.\d+/
+      actual_version = version
+      if platform_is_deb_based(platform)
+        actual_version += "-1puppetlabs1"
```
ajroetker: ewww ruby

ajroetker: Why not just have the deb clause add this if it doesn't already exist in the version string?

wkalt:
did we come to a decision on this?

acceptance/setup/pre_suite/70_install_released_puppetdb.rb
```diff
@@ -1,8 +1,8 @@
 # We skip this step entirely unless we are running in :upgrade mode.
 if (test_config[:install_mode] == :upgrade)
   step "Install most recent released PuppetDB on the PuppetDB server for upgrade test" do
-    install_puppetdb(database, test_config[:database], 'latest')
+    install_puppetdb(database, test_config[:database], '2.3.5')
```
kbarber: I think if this could be pulling from env variables (like we have for other settings), defaulting back to latest, perhaps this might the patch more flexible for future use.

wkalt:
+1 from me. This won't test anything until we merge it.

ajroetker:
:+1: from me

PR #1537, opened 2015-07-13
(maint) update preinst for redhat to check for 2.x if upgrading

wkalt:
This should fix the issue where we attempted to tar a nonexistent directory in
preinst when upgrading from 3.0.0 -> 3.0.x


wkalt:

mullr:
... this was just merged into master.

PR #1540, opened 2015-07-14
(maint) remove add_el5_postgres?

wkalt:
This removes some old code that seems unneccessary with the recent module
changes. Previously we were doing some special stuff for el5 to install postgres 8.4,
but with the 5.0 module this results in two declarations of the Postgresql class.

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -269,44 +269,20 @@ module PuppetDBExtensions
 
   end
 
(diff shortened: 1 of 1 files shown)


mullr:
This looks fine. To be clear, this will bump us to using pg9.4 on el5, but I'm fine with that too.

ajroetker:
:+1: lgtm

PR #1543, opened 2015-07-15
(PDB-1313) Update deactivated timestamp if newer

rbrw:
If the incoming "deactivate node" timestamp is newer than the one in the
database, replace the database value.  While perhaps reasonable in
general, this is particularly important for pdb sync, i.e. state
convergence.


mullr:
lgtm

mullr:
@rbrw Are we still in "don't merge yet" land or can this go in?

mullr:
(probably needs a rebase too)

rbrw:
@mullr just rebased, and I think this can probably go in now if you're OK with it.

PR #1547, opened 2015-07-16
(maint) Add access to packaging repo

melissa:
This commit allows access to the packaging repo automation. Although
this is no longer needed or used to build packages (in favor of ezbake),
we still have some automation in that packaging repo that we should have
access to. For example, the automation to create release tickets lives
there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ version
 junit
 tmp
 vendor
+ext/packaging
+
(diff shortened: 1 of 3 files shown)


melissa:
cc/ @wkalt this will let you generate the release tickets from the ezbake centric branches of pdb. It's not required by any means so if you'd rather not have this in, that's fine. I figured it might make things a little easier in the future though. I also eventually want to add a shim to the packaging repo so you can run the ticket creation automation independent of a project (so rather than going to a project base and requiring this code in that project base to run the task, eventually you'll be able to clone the packaging repo and simply run the automation from there). In the meantime, though, this might be the best solution.

PR #1549, opened 2015-07-20
(PDB-1812) Analyze all pg tables at least once

rbrw:
Investigation of a performance problem revealed that analyzing the
value_types table improved a PostgreSQL backed http:/.../fact-values
query by a factor of about 4x when run against a ~10m fact benchmark DB
on a tmpfs on Linux.

It appeared that some tables had never been analyzed, and further
testing revealed that the autovacuum process ignored them.  In some
cases, it looks like this is because it ignores small tables that don't
change much, like value_types, which has 6 rows and effectively never
changes.

So make sure that all tables are analyzed at least once (after any
structure changes) by running a "vacuum (analyze, verbose)" whenever we
have to run any migration, and add a dummy migration so that all
existing databases will run the analysis too.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1453,7 +1453,11 @@
    31 coalesce-fact-values
    32 add-producer-timestamp-to-reports
    33 add-certname-id-to-certnames
-   34 add-certname-id-to-resource-events})
+   34 add-certname-id-to-resource-events
+   ;; This dummy migration ensures that even databases that were up to
+   ;; date when the "vacuum analyze" code was added to migrate! will
+   ;; still analyze their existing databases.
+   35 (fn [] true)})
 
 (def desired-schema-version (apply max (keys migrations)))
 
@@ -1491,28 +1495,49 @@
     (into (sorted-map)
           (select-keys migrations pending))))
 
+(defn- sql-or-die [f]
+  "Calls (f) and returns its result.  If f throws an SQLException,
+  logs the exception and its parent and then calls (System/exit 1)"
+  ;; Here we've preserved existing behavior, but this may warrant
+  ;; further consideration later.  If possible, we might want to
+  ;; avoid System/exit in favor of careful exception
+  ;; handling (cf. PDB-1118).
+  (try
+    (f)
+    (catch java.sql.SQLException e
+      (log/error e "Caught SQLException during migration")
+      (let [next (.getNextException e)]
+        (when-not (nil? next)
+          (log/error next "Unravelled exception")))
+      (binding [*out* *err*] (flush)) (flush)
+      (System/exit 1))))
+
 (defn migrate!
-  "Migrates database to the latest schema version. Does nothing if database is
-  already at the latest schema version."
-  []
+  "Migrates database to the latest schema version. Does nothing if
+  database is already at the latest schema version.  Requires a
+  connection pool because some operations may require an indepdendent
+  database connection."
+  [db-connection-pool]
   (if-let [unexpected (first (difference (applied-migrations) (kitchensink/keyset migrations)))]
     (throw (IllegalStateException.
             (format "Your PuppetDB database contains a schema migration numbered %d, but this version of PuppetDB does not recognize that version."
                     unexpected))))
-
   (if-let [pending (seq (pending-migrations))]
-    (sql/transaction
-     (doseq [[version migration] pending]
-       (log/infof "Applying database migration version %d" version)
-       (try
-         (migration)
-         (record-migration! version)
-         (catch java.sql.SQLException e
-           (log/error e "Caught SQLException during migration")
-           (let [next (.getNextException e)]
-             (when-not (nil? next)
-               (log/error next "Unravelled exception")))
-           (System/exit 1)))))
+    (do
+      (sql/transaction
+       (doseq [[version migration] pending]
+         (log/infof "Applying database migration version %d" version)
+         (sql-or-die (fn [] (migration) (record-migration! version)))))
+      (when (sutils/postgres?)
+        ;; Make sure all tables (even small static tables) are
+        ;; analyzed at least once.  Note that vacuum cannot be
+        ;; called from within a transaction block.
+        (sql/with-connection db-connection-pool
+          (log/info "Analyzing database")
+          (sql-or-die (fn []
+                        (-> (doto (sql/find-connection) (.setAutoCommit true))
+                            .createStatement
+                            (.execute "vacuum (analyze, verbose)")))))))
     (log/info "There are no pending migrations")))
 
 ;; SPECIAL INDEX HANDLING
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -246,7 +246,7 @@
   [db-conn-pool product-name]
   (sql/with-connection db-conn-pool
     (scf-store/validate-database-version #(System/exit 1))
-    (migrate!)
+    (migrate! db-conn-pool)
     (indexes! product-name)))
 
 (defn init-with-db
--- a/test/puppetlabs/puppetdb/fixtures.clj
+++ b/test/puppetlabs/puppetdb/fixtures.clj
@@ -26,7 +26,7 @@
       (sql/with-connection *db*
         (sql/transaction
          (clear-db-for-testing!)
-         (migrate!)))
+         (migrate! *db*)))
       (pjdbc/pooled-datasource (assoc db :read-only? read-only?)))))
 
 (defn with-test-db
@@ -35,7 +35,7 @@
   (binding [*db* (test-db)]
     (sql/with-connection *db*
       (clear-db-for-testing!)
-      (migrate!)
+      (migrate! *db*)
       (f))))
 
 (defn without-db-var
--- a/test/puppetlabs/puppetdb/scf/migrate_test.clj
+++ b/test/puppetlabs/puppetdb/scf/migrate_test.clj
@@ -42,7 +42,7 @@
     (testing "should return nothing if the db is completely migrated"
       (sql/with-connection db
         (clear-db-for-testing!)
-        (migrate!)
+        (migrate! db)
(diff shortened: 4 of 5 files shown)


rbrw:
I was able to reproduce this with a database created via something like `benchmark -F resources/puppetlabs/puppetdb/benchmark/samples/facts -n 3000 -N 1 -r 10`.  On a tmpfs a `time curl -o /dev/null -sX GET http://localhost:8080/pdb/query/v4/fact-paths` took about 1m before the analyze, and about 15s after it.

mullr:
On my vm setup:
Before: 6.669s
After: 0.408s

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1492,28 +1497,42 @@
...
+                (System/exit 1))))]
+    (if-let [pending (seq (pending-migrations))]
+      (do
+        (sql/transaction
+         (doseq [[version migration] pending]
+           (log/infof "Applying database migration version %d" version)
+           (sql-or-die #(do (migration) (record-migration! version)))))
+        (when (sutils/postgres?)
+          ;; Make sure all tables (even small static tables) are
+          ;; analyzed at least once.  Note that vaccum cannot be
```
mullr: vacuum

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1492,28 +1497,42 @@
...
-         (migration)
-         (record-migration! version)
-         (catch java.sql.SQLException e
-           (log/error e "Caught SQLException during migration")
-           (let [next (.getNextException e)]
-             (when-not (nil? next)
-               (log/error next "Unravelled exception")))
-           (System/exit 1)))))
-    (log/info "There are no pending migrations")))
+  (letfn [(sql-or-die [f]
```
mullr: why not put this in its own defn?

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1492,28 +1497,42 @@
...
+            (try
+              (f)
+              (catch java.sql.SQLException e
+                (log/error e "Caught SQLException during migration")
+                (let [next (.getNextException e)]
+                  (when-not (nil? next)
+                    (log/error next "Unravelled exception")))
+                ;; See PDB-1118
```
mullr: https://tickets.puppetlabs.com/browse/PDB-1118
Not sure I understand what you mean by referring to this issue. I'd also rather see whatever you want to say written out here; issue trackers come and go, while source code has many lives.

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Knew I'd do that at least once.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1492,28 +1497,42 @@
...
+                (System/exit 1))))]
+    (if-let [pending (seq (pending-migrations))]
+      (do
```
mullr: migrate! is getting a little pudgy; this section may be a good candidate for extracting to a separate function.

mullr:
ok, the fix looks like it works which is the important part. Basically just style notes.

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Mostly just distaste - don't really like it in the first place (system/exit...), and hadn't felt like thinking about its semantics more publicaly (suspect it'll go away, and shouldn't be considered more widely), but I suppose I could just make it private...

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Fair enough - was just an allusion to my concern about System/exit, etc.

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: yeah, `defn-` is a perfectly good tool for that

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: Ah, that's a very good thing to note here.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1492,28 +1497,42 @@
...
+         (doseq [[version migration] pending]
+           (log/infof "Applying database migration version %d" version)
+           (sql-or-die #(do (migration) (record-migration! version)))))
+        (when (sutils/postgres?)
+          ;; Make sure all tables (even small static tables) are
+          ;; analyzed at least once.  Note that vaccum cannot be
+          ;; called from within a transaction block.
+          (sql/with-connection db-connection-pool
+            (log/info "Analyzing database")
+            (sql-or-die #(-> (doto (:connection sqlint/*db*)
```
ajroetker: Is this just an `(fn []`? I usually prefer that just to be explicit that this anon-fn won't take any args

ajroetker: Also you shouldn't need to call `sqlint/*db*` directly, I believe there's a `get-connection` or `find-connection` that does basically that in the normal `jdbc` namespace

ajroetker: Yep, `jdbc/find-connection` is what you want i think, https://github.com/clojure/java.jdbc/blob/java.jdbc-0.1.1/src/main/clojure/clojure/java/jdbc.clj#L57

src/puppetlabs/puppetdb/scf/migrate.clj
ajroetker: Maybe the jira link would be appropriate then?

mullr: explanatory text + the issue number at the end would be ideal

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Fixed - thanks.

rbrw: With respect to the (fn [] ...), since it's short, I'd figured it was pretty easy to see there weren't any %'s in there, but I"m happy to change it if that's preferable.

rbrw:
OK, hopefully addressed the outstanding issues.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1491,28 +1495,49 @@
...
+  ;; handling (cf. PDB-1118).
+  (try
+    (f)
+    (catch java.sql.SQLException e
+      (log/error e "Caught SQLException during migration")
+      (let [next (.getNextException e)]
+        (when-not (nil? next)
+          (log/error next "Unravelled exception")))
+      (binding [*out* *err*] (flush)) (flush)
+      (System/exit 1))))
```
ajroetker: Could we call the trapperkeeper shutdown function? Am I mistaken or do we not do that elsewhere

rbrw: I'm not sure what I'd recommend yet (for our existing exit calls) - there may be some cases where System/exit is appropriate, but even so, I'd argue that perhaps those exit(s) should still be handed at the top, by a throw to an appropriate handler (so that any critical unwinds can fire, we can make sure to always flush, etc.).  

But I'm also assuming that the work should wait for a broader (probably non-stable) overhaul, as proposed in the ticket.

rbrw: (And of course, system/exit should be considered a last resort, since it's very unfriendly, particularly in a multi-tenant scenario.)

mullr: I agree, we should live this alone for now.

rbrw:

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1491,28 +1495,49 @@
...
+         (log/infof "Applying database migration version %d" version)
+         (sql-or-die (fn [] (migration) (record-migration! version)))))
+      (when (sutils/postgres?)
+        ;; Make sure all tables (even small static tables) are
+        ;; analyzed at least once.  Note that vacuum cannot be
+        ;; called from within a transaction block.
+        (sql/with-connection db-connection-pool
+          (log/info "Analyzing database")
+          (sql-or-die (fn []
+                        (-> (doto (sql/find-connection) (.setAutoCommit true))
```
senior: Curious as to why you need autocommit here?

senior:
@rbrw @mullr Do we need to set some expectations for how long with will take on the next startup? Do we have some rough idea of how long this takes on a 5k node install?

rbrw:
For me, it was taking about 10s for the DB mentioned at the top of this thread (10m facts, 3k certnames, generated via benchmark -F (no -C or -R)); I'd be happy to test other arrangements.

senior:
@rbrw I think that's a good idea so that we can put it in the release notes

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Meant to leave a note about that (perhaps a comment?).  Setting it to false was a way I found that would allow us to execute vacuum via the jdbc stack.  There may be other approaches, but if you remove the autocommit setting, you should see the "can't execute vacuum from within a transaction block" exception.

senior:
I'm :+1:. Suggested a comment on the autocommit thing in the channel and getting an idea of duration so that we can let the users know what to expect.

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: I'll add the comment.

rbrw:
Sounds good, I'll test  the full 3k case shortly.

wkalt:
this works for me. with 14129 fact paths inserted with the benchmark tool, the query time went from 36 seconds to 3.5

PR #1550, opened 2015-07-20
(PDB-700) Apply max-frame-size to consumer

mullr:
The command consumer worker threads don't currently apply the
max-frame-size parameter, so the AMQ default of 100MB is used.

### DIFF
--- a/src/com/puppetlabs/puppetdb/cli/services.clj
+++ b/src/com/puppetlabs/puppetdb/cli/services.clj
@@ -246,8 +246,8 @@
 
 (defn add-max-framesize
   "Add a maxFrameSize to broker url for activemq."
-  [config url]
-  (format "%s&wireFormat.maxFrameSize=%s&marshal=true" url (:max-frame-size config)))
+  [max-frame-size url]
+  (format "%s&wireFormat.maxFrameSize=%s&marshal=true" url max-frame-size))
 
 (defn start-puppetdb
   [context config service-id add-ring-handler shutdown-on-error]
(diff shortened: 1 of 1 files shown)


src/com/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -274,7 +274,8 @@
         discard-dir                                (file mq-dir "discarded")
         globals                                    {:scf-read-db          read-db
                                                     :scf-write-db         write-db
-                                                    :command-mq           {:connection-string (add-max-framesize command-processing mq-addr)
+                                                    :command-mq           {:connection-string (add-max-framesize (:max-frame-size command-processing)
```
ajroetker: Should this be named in the `let` so it can be used both here and below?

mullr: yes, it should

mullr:
@ajroetker fixed per your comment

rbrw:
LGTM

src/com/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -272,9 +272,10 @@
         dlo-compression-threshold                  (get command-processing :dlo-compression-threshold)
         mq-dir                                     (str (file (:vardir global) "mq"))
         discard-dir                                (file mq-dir "discarded")
+        command-mq-connection-string               (add-max-framesize (:max-frame-size command-processing) mq-addr)
```
ajroetker: Is the function really necessary anymore?

mullr: not strictly, but I don't mind it being there.

PR #1552, opened 2015-07-20
(maint) rollup 2.3.x to stable

mullr:


mullr:
↑↑↑ not a real failure ↑↑↑
Just a spurious failure on one platform; I don't think we should worry about it.

PR #1553, opened 2015-07-20
(PDB-1809) improve aggregate event counts performance

wkalt:
(PDB-1809) improve aggregate event counts performance
    
    This series of commits attempts to improve the performance of aggregate event
    counts with the following actions:
    
    * explicit C collation in group by used for with-latest-events
    * query parameter for aggregate event counts endpoint is now optional
    * moved some things from distinct-select into with-latest-events so they can be
      reused across multiple summarize_by's

### DIFF
--- a/src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
+++ b/src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
@@ -42,6 +42,7 @@
   [version]
   (-> (routes version)
       verify-accepts-json
-      (validate-query-params {:required ["query" "summarize_by"]
-                              :optional ["counts_filter" "count_by" "distinct_resources"
+      (validate-query-params {:required ["summarize_by"]
+                              :optional ["query"
+                                         "counts_filter" "count_by" "distinct_resources"
                                          "distinct_start_time" "distinct_end_time"]})))
--- a/src/puppetlabs/puppetdb/query.clj
+++ b/src/puppetlabs/puppetdb/query.clj
@@ -140,13 +140,18 @@
   function basically just checks that the operator is known, and then
   dispatches to the function implementing it."
   [ops [op & args :as term]]
-  (when-not (sequential? term)
-    (throw (IllegalArgumentException. (format "%s is not well-formed: queries must be an array" (vec term)))))
-  (when-not op
-    (throw (IllegalArgumentException. (format "%s is not well-formed: queries must contain at least one operator" (vec term)))))
-  (if-let [f (ops op)]
-    (apply f args)
-    (throw (IllegalArgumentException. (format "%s is not well-formed: query operator '%s' is unknown" (vec term) op)))))
+
+  (cond
+    (empty? term)
+    {:where nil :params nil}
+
+    (not op)
+    (throw (IllegalArgumentException. (format "%s is not well-formed: queries must contain at least one operator" (vec term))))
+
+    :else
+    (if-let [f (ops op)]
+      (apply f args)
+      (throw (IllegalArgumentException. (format "%s is not well-formed: query operator '%s' is unknown" (vec term) op))))))
 
 (defn compile-boolean-operator*
   "Compile a term for the boolean operator `op` (AND or OR) applied to
@@ -239,6 +244,27 @@
    "containment_path"       ["resource_events"]
    "containing_class"       ["resource_events"]
    "name"                   ["environments" "environment"]})
+
+(def resource-event-columns
+  {"certname"               ["latest_events"]
+   "configuration_version"  ["latest_events"]
+   "run_start_time"         ["latest_events"]
+   "run_end_time"           ["latest_events"]
+   "report_receive_time"    ["latest_events"]
+   "hash"                   ["latest_events" "report"]
+   "status"                 ["latest_events"]
+   "timestamp"              ["latest_events"]
+   "resource_type"          ["latest_events"]
+   "resource_title"         ["latest_events"]
+   "property"               ["latest_events"]
+   "new_value"              ["latest_events"]
+   "old_value"              ["latest_events"]
+   "message"                ["latest_events"]
+   "file"                   ["latest_events"]
+   "line"                   ["latest_events"]
+   "containment_path"       ["latest_events"]
+   "containing_class"       ["latest_events"]
+   "name"                   ["latest_events" "environment"]})
 
 (def report-columns
   "Return the queryable set of fields and corresponding table names where they reside"
@@ -333,7 +359,7 @@
         columns (case kind
                   :fact fact-columns
                   :resource resource-columns
-                  :event event-columns
+                  :event resource-event-columns
                   :report report-columns
                   :factset factset-columns)
         qualified-field (qualified-column field columns)]
@@ -554,9 +580,9 @@
                                                                                                                  args))))))
 
   (let [timestamp-fields {"timestamp"           "resource_events.timestamp"
-                          "run_start_time"      "reports.start_time"
-                          "run_end_time"        "reports.end_time"
-                          "report_receive_time" "reports.receive_time"}]
+                          "run_start_time"      "latest_events.run_start_time"
+                          "run_end_time"        "latest_events.run_end_time"
+                          "report_receive_time" "latest_events.report_receive_time"}]
     (match [path]
            [(field :guard (kitchensink/keyset timestamp-fields))]
            (if-let [timestamp (to-timestamp value)]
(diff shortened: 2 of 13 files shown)


wkalt:
I need to write a few tests to verify that the query param is no longer needed, but this is ready for some functional review. I'll fix up the commits when we're happy.

src/puppetlabs/puppetdb/query.clj
```diff
@@ -140,13 +140,22 @@
...
+
+  (cond
+
+    (empty? term)
+    {:where nil :params nil}
+    (not (sequential? term))
+    (throw (IllegalArgumentException.
+             (format "%s is not well-formed: queries must be an array" (vec term))))
+
+    (not op)
```
ajroetker: Is this for the `(= nil op)` case? If so I'd prefer `(nil? op)`?

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1453,7 +1453,7 @@
    31 coalesce-fact-values
    32 add-producer-timestamp-to-reports
    33 add-certname-id-to-certnames
-   34 add-certname-id-to-resource-events})
+   34 add-certname-id-to-resource-events })
```
ajroetker: Extra space

src/puppetlabs/puppetdb/query.clj
```diff
@@ -140,13 +140,22 @@
...
   [ops [op & args :as term]]
-  (when-not (sequential? term)
-    (throw (IllegalArgumentException. (format "%s is not well-formed: queries must be an array" (vec term)))))
-  (when-not op
-    (throw (IllegalArgumentException. (format "%s is not well-formed: queries must contain at least one operator" (vec term)))))
+
+  (cond
+
+    (empty? term)
+    {:where nil :params nil}
```
mullr: if we're spacing cond pairs (which I like) there should be a blank line here

wkalt:
this will need a couple docs updates as well

src/puppetlabs/puppetdb/query.clj
```diff
@@ -140,13 +140,22 @@
...
+    (not (sequential? term))
```
mullr: The above call to `empty?` will already throw if term can't be turned into a seq.

wkalt:
I found a bug around using aggregate-event-counts with no query parameter and distinct_resources=false. Working through that now.

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -50,49 +50,74 @@
...
-                   resource_events.property,
-                   MAX(resource_events.timestamp) AS timestamp
-                   FROM resource_events
-                   JOIN certnames ON resource_events.certname_id = certnames.id
-                   WHERE resource_events.timestamp >= ?
-                   AND resource_events.timestamp <= ?
-                   GROUP BY certname, resource_type, resource_title, property)
-           %s" query))
+  (let [collate (if (sutils/postgres?) "COLLATE \"C\"" "")]
+    (format
```
mullr: Whither the whitespace?

wkalt: because I have spent zero time cleaning this up :)

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -50,49 +50,74 @@
...
+       line,
+       containment_path,
+       containing_class,
+       name
+       FROM
+       (SELECT certname_id,
+       resource_type %s AS resource_type,
+       resource_title %s AS resource_title,
+       property %s AS property,
+       MAX(resource_events.timestamp) AS timestamp
```
mullr: calling this latest_event_timestamp or similar would make this easier to understand

src/puppetlabs/puppetdb/query.clj
```diff
@@ -140,13 +140,22 @@
...
+    (throw (IllegalArgumentException.
+             (format "%s is not well-formed: queries must be an array" (vec term))))
+
+    (not op)
+    (throw (IllegalArgumentException. (format "%s is not well-formed: queries must contain at least one operator" (vec term))))
+
+    :else
   (if-let [f (ops op)]
```
ajroetker: Alignment

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -50,49 +50,74 @@
...
-               AND ((resource_events.property = latest_events.property) OR
-                    (resource_events.property IS NULL AND latest_events.property IS NULL))
-               AND resource_events.timestamp = latest_events.timestamp
-         WHERE %s"
-    select-fields
-    where)
-   (concat [distinct-start-time distinct-end-time] params)])
+  (let [where-clause (if where (format "WHERE %s" where) "")]
+    [(format
+       "SELECT %s
```
mullr: smells like sql injection spirit

wkalt: look more carefully

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -42,6 +42,7 @@
   [version]
   (-> (routes version)
       verify-accepts-json
-      (validate-query-params {:required ["query" "summarize_by"]
-                              :optional ["counts_filter" "count_by" "distinct_resources"
+      (validate-query-params {:required ["summarize_by"]
+                              :optional ["query"
```
mullr: docs need to be updated to reflect this

src/puppetlabs/puppetdb/query/events.clj
wkalt: well, truly it should be distinct_events, but then we'd have to realias it as timestamp above to get the return field right, so we'd have
```
distinct_events.distinct_events_timestamp as timestamp
```
up on line 63. I could drop the aliasing there at that point but I don't see the benefit really. Does that seem desirable?

src/puppetlabs/puppetdb/query/events.clj
wkalt: cleared this with @mullr earlier

src/puppetlabs/puppetdb/query/events.clj
wkalt: looking more closely, I'm not sure what whitespace you mean here

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
wkalt: updated

wkalt:

src/puppetlabs/puppetdb/query/events.clj
kbarber: Its all left justified. Some indentation to indicate the SQL blocks would make this much more readable.

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -50,49 +50,74 @@
...
+       (SELECT certname_id,
+       resource_type %s AS resource_type,
+       resource_title %s AS resource_title,
+       property %s AS property,
+       MAX(resource_events.timestamp) AS timestamp
+       FROM resource_events
+       WHERE resource_events.timestamp >= ?
+       AND resource_events.timestamp <= ?
+       GROUP BY certname_id,
+       resource_type %s,
```
senior: I'm not understanding why we would want to collate by ASCII rules here. These come from Puppet as UTF-8 and we store them as UTF-8, what happens when the user has a special character here?

senior: Also why do we need to collate on the projection AND the group by?

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -50,49 +50,74 @@
...
-                   FROM resource_events
-                   JOIN certnames ON resource_events.certname_id = certnames.id
-                   WHERE resource_events.timestamp >= ?
-                   AND resource_events.timestamp <= ?
-                   GROUP BY certname, resource_type, resource_title, property)
-           %s" query))
+  (let [collate (if (sutils/postgres?) "COLLATE \"C\"" "")]
+    (format
+      "WITH latest_events AS
+       (SELECT certname,
```
kbrezina: It appears to be more efficient to use `DISTINCT ON` construct here. See [full](https://gist.github.com/kbrezina/49ea3f93f6348bfb6443#file-resource_events_certname_resource_timestamp-txt) or [reduced](https://gist.github.com/kbrezina/351ae99e67d65dac3dc9) query. Make sure to create [resource_events_certname_resource_timestamp](https://gist.github.com/kbrezina/49ea3f93f6348bfb6443) index first.

src/puppetlabs/puppetdb/query/events.clj
wkalt: fixed

src/puppetlabs/puppetdb/query/events.clj
wkalt: left some discussion of this in PE-11006. Short story is that I like the direction, but the perf benefits don't carry over to queries on shorter time intervals. I think we should explore this as a separate ticket, assuming that the current work gets us in passable shape. @kbrezina sound right to you?

wkalt:

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
```diff
@@ -55,7 +55,8 @@
                  (concat (take 2 common-params)
                          (flatten (repeat nsummarized (drop 2 common-params))))
                  ;; repeat all params as many times as there are summarize_by's
-                 (flatten (repeat nsummarized common-params)))
+                 (if (nil? common-params) []
```
senior: Was the placement of [] intentional here?

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -57,7 +58,7 @@
          (jdbc/with-query-results-cursor results-query (comp row-fn munge-fn)))))))
 
 (defn produce-streaming-body
-  "DEPRECATED - this function will be replaced by produce-streaming-body'
+  " DEPRECATED - this function will be replaced by produce-streaming-body'
```
senior: Extra space

senior:
:+1: two nit picky comments (docs and indentation), looks good otherwise

wkalt:

PR #1556, opened 2015-07-21
(PDB-1739) Upgrade to the latest clojure.java.jdbc `0.3.7`

ajroetker:
This commit upgrades PuppetDB to use the latest clojure.java.jdbc which
consists of using the deprecated namespace in the new version as well as
removing some test fixtures that reached into the old internals
namespace of jdbc to bind the *db* dynamic var.

### DIFF
--- a/test/puppetlabs/puppetdb/fixtures.clj
+++ b/test/puppetlabs/puppetdb/fixtures.clj
@@ -1,7 +1,6 @@
 (ns puppetlabs.puppetdb.fixtures
   (:import [java.io ByteArrayInputStream])
-  (:require [clojure.java.jdbc :as sql]
-            [clojure.java.jdbc.internal :as jint]
+  (:require [clojure.java.jdbc.deprecated :as sql]
             [puppetlabs.puppetdb.http.server :as server]
             [puppetlabs.puppetdb.http.command :as command]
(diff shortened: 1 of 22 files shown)


test/puppetlabs/puppetdb/fixtures.clj
```diff
@@ -38,18 +37,6 @@
...
-(defn without-db-var
-  "Binds the java.jdbc dtabase connection to nil. When running a unit
-   test using `with-test-db`, jint/*db* will be bound. If the routes
-   being tested don't explicitly bind the db connection, it will use
-   one bound in with-test-db. This causes a problem at runtime that
-   won't show up in the unit tests. This fixture can be used around
-   route testing code to ensure that the route has it's own db
-   connection."
-  [f]
-  (binding [jint/*db* nil]
```
senior: Does the deprecated stuff not use this anymore? I'm confused as to why this would need to be removed as part of this patch.

senior:
One question on why we're removing without-db-var, other than that, I'm :+1:

senior:
I think the right ticket number for this patch is PDB-1739?

ajroetker:
@senior i think this is g2g now

ajroetker:
transient failure

PR #1560, opened 2015-07-24
(maint) Add secondsToWaitForCompletion cmd param

mullr:
~~Also includes factoring out a config service and MQService.~~ took this part out

### DIFF
--- a/src/puppetlabs/puppetdb/client.clj
+++ b/src/puppetlabs/puppetdb/client.clj
@@ -22,15 +22,31 @@
     command :- s/Str
     version :- s/Int
     payload]
-     (->> payload
-          (command/assemble-command command version)
-          (submit-command-via-http! base-url)))
+   (submit-command-via-http! base-url
+                             (command/assemble-command command version payload)
+                             nil))
+
+  ([base-url
+    command :- s/Str
+    version :- s/Int
+    payload
+    timeout]
+   (submit-command-via-http! base-url
+                             (command/assemble-command command version payload)
+                             timeout))
+
   ([base-url :- utils/base-url-schema
     command-map :- {s/Any s/Any}]
+   (submit-command-via-http! base-url command-map nil))
+
+  ([base-url :- utils/base-url-schema
+    command-map :- {s/Any s/Any}
+    timeout]
      (let [message (json/generate-string command-map)
            checksum (kitchensink/utf8-string->sha1 message)
            url (str (utils/base-url->str base-url)
-                    (format "?checksum=%s" checksum))]
+                    (format "?checksum=%s" checksum)
+                    (when timeout (format "&secondsToWaitForCompletion=%s" timeout)))]
        (http-client/post url {:body               message
                               :throw-exceptions   false
                               :content-type       :json
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -74,7 +74,8 @@
             [schema.core :as s]
             [puppetlabs.puppetdb.time :refer [to-timestamp]]
             [clj-time.core :refer [now]]
-            [clojure.set :as set]))
+            [clojure.set :as set]
+            [clojure.core.async :as async]))
 
 ;; ## Command parsing
 
@@ -129,12 +130,12 @@
   its id."
   [mq-connection :- mq/connection-schema
    mq-endpoint :- s/Str
-   raw-command :- s/Str]
-  (let [id (kitchensink/uuid)]
-    (mq/send-message! mq-connection mq-endpoint
-                      raw-command
-                      {"received" (kitchensink/timestamp) "id" id})
-    id))
+   raw-command :- s/Str
+   uuid :- s/Str]
+  (mq/send-message! mq-connection mq-endpoint
+                    raw-command
+                    {"received" (kitchensink/timestamp) "id" uuid})
+  uuid)
 
 (defn-validated ^:private do-enqueue-command :- s/Str
   "Submits command to the mq-endpoint of mq-connection and returns
@@ -304,7 +305,7 @@
   (enqueue-command [this connection endpoint command version payload]
     "Annotates the command via annotate-command, submits it to the
     endpoint of the connection, and then returns its unique id.")
-  (enqueue-raw-command [this connection endpoint raw-command]
+  (enqueue-raw-command [this connection endpoint raw-command uuid]
     "Submits the raw-command to the endpoint of the connection and
     returns the command's unique id.")
   (stats [this]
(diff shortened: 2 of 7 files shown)


src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -36,12 +38,34 @@
...
+        topic (str uuid)
+        _ (async/sub response-pub topic response-chan)
+        timeout-chan (async/timeout completion-timeout-ms)
+        _ (do-submit-fn)
+        timed-out (async/alt!! timeout-chan true
+                               response-chan false)]
+    (async/unsub response-pub topic response-chan)
+    (http/json-response {:uuid uuid, :processed (not timed-out)}
+                        (if timed-out 503 200))))
+
```
senior: I think I grok what's going on here but don't know as much about this in core.async. Is there a need to put a close call for the channel and maybe the unsub in a finally clause? The sub call will close the response-chan when the message is published, but what if it doesn't get published? Does response-chan get GC'd? Or if there's a failure and the unsub doesn't get called?

src/puppetlabs/puppetdb/command.clj
```diff
@@ -315,24 +316,35 @@
...
           config {:db scf-write-db
-                  :catalog-hash-debug-dir catalog-hash-debug-dir}]
+                  :catalog-hash-debug-dir catalog-hash-debug-dir}
+          {:keys [response-pub-chan response-pub]} context]
       (register-listener
        supported-command?
        (fn [cmd]
          (let [result (process-command! cmd config)]
            (swap! (:stats context) update :executed-commands inc)
+           (async/>!! response-pub-chan cmd)
```
senior: So commands always publish to this channel?

mullr: Yep. There's a `pub` set up on it, so I *think* that it's a no-op if nobody is listening on its topic.

senior: Yeah, from the docs:
"Items received when there are no matching subs get dropped."

src/puppetlabs/puppetdb/http/command.clj
mullr: There should definitely be a `finally` with a call to `unsub`. I'm less certain about needing to close `response-chan`; http://stackoverflow.com/questions/28888340/should-clojure-core-async-channels-be-closed-when-not-used-anymore indicates that it may not be needed. Looking through https://github.com/clojure/core.async/blob/master/src/main/clojure/clojure/core/async/impl/channels.clj, this *may* be true. Of note are `puts` and `takes` members of `ManyToManyChannel`, which are each a linked list of mutexes. Will continue to investigate.

mullr: conclusion: you don't *have* to close a channel, but if you do you can prevent some otherwise unnecessary work.

mullr:
@senior updated with better cleanup on both the pub and sub sides, and added docs.

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -36,12 +38,57 @@
...
+         (async/unsub p# t# c#)))))
+
+(defn- blocking-submit-command
+  "Submit a command by calling do-submit-fn and block until it completes.
+  Subscribes to response-pub on the topic of the commands uuid, waiting up to
+  completion-timeout-ms."
+  [do-submit-fn response-pub uuid completion-timeout-ms]
+  (with-chan [response-chan (async/chan)]
+    (with-sub {:from response-pub :topic (str uuid) :chan response-chan}
+      (let [timeout-chan (async/timeout completion-timeout-ms)
```
ajroetker: Why not use the `with-chan` here as well?

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -346,6 +346,4 @@
   "Syncronously post a command to PDB by blocking until the message is consumed
    off the queue."
   [base-url cmd version payload]
-  (until-consumed
-   (fn []
-     (pdb-client/submit-command-via-http! base-url cmd version payload))))
+  (pdb-client/submit-command-via-http! base-url cmd version payload 10))
```
ajroetker: Could we pull out the 10 into a let and give it a name?

src/puppetlabs/puppetdb/http/command.clj
mullr: timeout-chan closes itself after the given timeout.

mullr: I guess we could do it early... I'll check and see if it would matter

mullr: It doesn't; in fact it's slightly more work for the computer if we close it early.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -315,27 +316,42 @@
...
    [:MessageListenerService register-listener]]
   (init [this context]
-    (assoc context :stats (atom {:received-commands 0
-                                 :executed-commands 0})))
+    (let [response-pub-chan (async/chan)]
+      (assoc context
+             :stats (atom {:received-commands 0
+                           :executed-commands 0})
+             :response-pub-chan response-pub-chan
+             :response-pub (async/pub response-pub-chan #(-> % :annotations :id str)))))
```
ajroetker: What about `(comp str :id :annotations)`?

src/puppetlabs/puppetdb/client.clj
```diff
@@ -22,15 +22,31 @@
...
+
+  ([base-url :- utils/base-url-schema
+    command-map :- {s/Any s/Any}
+    timeout]
      (let [message (json/generate-string command-map)
            checksum (kitchensink/utf8-string->sha1 message)
            url (str (utils/base-url->str base-url)
-                    (format "?checksum=%s" checksum))]
+                    (format "?checksum=%s" checksum)
+                    (when timeout (format "&secondsToWaitForCompletion=%s" (str timeout))))]
```
ajroetker: Does `format` with `%s` not already call `str` on the arg?

src/puppetlabs/puppetdb/command.clj
mullr: you say point free style, I say shmoint shmree shmile

src/puppetlabs/puppetdb/client.clj
mullr: ah, good point

src/puppetlabs/puppetdb/http/command.clj
ajroetker: Interesting

mullr: see https://github.com/clojure/core.async/blob/master/src/main/clojure/clojure/core/async/impl/timers.clj#L43

mullr:
@ajroetker updated

ajroetker:
Functionally tested and looks good to me!

wkalt:
mentioned to @mullr in person, but we should probably note explicitly that a :processed false response doesn't mean the command won't still get processed. Looks good to me otherwise.

mullr:
@wkalt updated docs and rebased

mullr:

PR #1564, opened 2015-07-28
(PDB-1313) Uniformly format json dates

rbrw:
Previously puppetdb's generate-string just redirected to
cheshire/generate-string, whereas generate-pretty-string would add
puppetdb-specific date formatting (that rendered seconds at "full"
precision) in addition to setting :pretty to true.

Change generate-string to include the puppetdb-specific date formatting,
since not pretty shouldn't also mean not accurate.

This was causing trouble because information (accuracy) was being lost
in transit (say across http).

One of the tests may fail right now, due to an undesirable interaction between pdb-instance and the log tests, but that appears to be a preexisting issue that can be reproduced via `lein test :only puppetlabs.puppetdb.command-test puppetlabs.puppetdb.testutils.services-test`.  One possible fix: https://github.com/puppetlabs/puppetdb/pull/1567

### DIFF
--- a/src/puppetlabs/puppetdb/cheshire.clj
+++ b/src/puppetlabs/puppetdb/cheshire.clj
@@ -38,19 +38,32 @@
 
 (add-common-json-encoders!)
 
-(def default-pretty-opts {:date-format "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'" :pretty true})
+(def default-opts {:date-format "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"})
+(def default-pretty-opts (assoc default-opts :pretty true))
 
-(def generate-string core/generate-string)
-
-(def generate-stream core/generate-stream)
+(defn generate-string
+  "Thinly wraps cheshire.core/generate-string, adding the PuppetDB
+  default date format."
+  ([obj]
+   (core/generate-string obj default-opts))
+  ([obj opts]
+   (core/generate-string obj (merge default-opts opts))))
 
 (defn generate-pretty-string
-  "Thinly wraps cheshire.core/generate-string, adding the PuppetDB default date format
-   and pretty printing from `default-pretty-opts`"
+  "Thinly wraps cheshire.core/generate-string, adding the PuppetDB
+  default date format and pretty printing from `default-pretty-opts`"
   ([obj]
-     (generate-pretty-string obj default-pretty-opts))
+   (core/generate-string obj default-pretty-opts))
   ([obj opts]
-     (generate-string obj (merge default-pretty-opts opts))))
+   (core/generate-string obj (merge default-pretty-opts opts))))
+
+(defn generate-stream
+  "Thinly wraps cheshire.core/generate-stream, adding the PuppetDB
+  default date format."
+  ([obj writer]
+   (core/generate-stream obj writer default-opts))
+  ([obj writer opts]
+   (core/generate-stream obj writer (merge default-opts opts))))
 
 (defn generate-pretty-stream
   "Thinly wraps cheshire.core/generate-stream, adding the PuppetDB default date format
--- a/test/puppetlabs/puppetdb/cheshire_test.clj
+++ b/test/puppetlabs/puppetdb/cheshire_test.clj
@@ -1,9 +1,11 @@
 (ns puppetlabs.puppetdb.cheshire-test
-  (:import (java.io StringWriter StringReader))
   (:require [clj-time.core :as clj-time]
             [puppetlabs.puppetdb.testutils :as tu]
             [clojure.test :refer :all]
-            [puppetlabs.puppetdb.cheshire :refer :all]))
+            [puppetlabs.puppetdb.cheshire :refer :all])
+  (:import [java.io StringWriter StringReader]
+           [java.sql Timestamp]
+           [org.joda.time DateTime]))
 
 (deftest test-generate-string
   (testing "should generate a json string"
@@ -34,6 +36,17 @@
       (generate-pretty-stream (sorted-map :a 1 :b 2) sw)
       (is (= (.toString sw)
              "{\n  \"a\" : 1,\n  \"b\" : 2\n}")))))
+
+(deftest test-date-generation
+  (let [t (java.util.Date. 5928174905781)]
+    (is (= "\"2157-11-09T03:15:05.781Z\""
+           (generate-string t))))
+  (let [t (DateTime. 5928174905781)]
+    (is (= "\"2157-11-09T03:15:05.781Z\""
+           (generate-string t))))
+  (let [t (Timestamp. 5928174905781)]
+    (is (= "\"2157-11-09T03:15:05.781Z\""
+           (generate-string t)))))
 
 (deftest test-parse-string
   (testing "should return a map from parsing a json string"
--- a/test/puppetlabs/puppetdb/command_test.clj
+++ b/test/puppetlabs/puppetdb/command_test.clj
@@ -1,5 +1,6 @@
 (ns puppetlabs.puppetdb.command-test
   (:require [me.raynes.fs :as fs]
+            [clj-http.client :as client]
             [clojure.java.jdbc.deprecated :as sql]
             [cheshire.core :as json]
             [puppetlabs.puppetdb.scf.storage :as scf-store]
@@ -13,7 +14,7 @@
             [puppetlabs.puppetdb.command :refer :all]
             [puppetlabs.puppetdb.testutils :refer :all]
             [puppetlabs.puppetdb.fixtures :refer :all]
-            [puppetlabs.puppetdb.jdbc :refer [query-to-vec]]
+            [puppetlabs.puppetdb.jdbc :refer [query-to-vec] :as jdbc]
             [puppetlabs.puppetdb.examples :refer :all]
             [puppetlabs.puppetdb.testutils.reports :refer [munge-example-report-for-storage]]
             [puppetlabs.puppetdb.testutils.services :as svc-utils]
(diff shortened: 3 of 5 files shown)


mullr:
lgtm

senior:
@rbrw Seems like a test that ensures the timestamps coming out of the top of PuppetDB includes those the fractional seconds would be good here so that we don't break this in the future. Maybe you could use the specific endpoint that was causing you grief in the test?

rbrw:
@senior excellent suggestion.

rbrw:
Added the tests, though they cause the existing testutils/services tests to fail.  I investigated and found that the failure could be induced by adding a dummy test file that just tested (with-pdb-instance ... true) that sorted alphabetically before the commant_test.clj (where I added the new test).  This PR provides one way we could fix the problem: https://github.com/puppetlabs/puppetdb/pull/1567

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1298,6 +1300,45 @@
...
+                              :query-params {"query"
+                                             (json/generate-string
+                                              ["or" ["=" ["node" "active"] true]
+                                               ["=" ["node" "active"] false]])}})
+                 :body
+                 json/parse-string
+                 first
+                 (get "deactivated")
+                 (pt/from-string)))))))
+
```
rbrw: This is fairly heavyweight - In the longer run, I wonder it might make sense to have a more general roundtrip test (if we don't already) that tests more than just a producer_timestamp against the instance.  Hmm, or would an "echo" service be interesting for general testing of the full encoding/decoding path?

senior:
@rbrw from your comment above, it sounds like we shouldn't merge this PR as it causes testutils/services.clj to fail?

rbrw:
@senior actually it turns out that it's is an existing problem on master - both @mullr and @wkalt were able to reproduce it via `lein test :only puppetlabs.puppetdb.command-test puppetlabs.puppetdb.testutils.services-test`, so I think this PR can proceeed.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1298,6 +1300,45 @@
...
+      (enqueue-command connection endpoint
+                       (command-names :deactivate-node) 3
+                       {:certname "foo.local" :producer_timestamp input-stamp})
+      (svc-utils/wait-for-server-processing svc-utils/*server*)
+      ;; While we're here, check the value in the database too...
+      (is (= expected-stamp
+             (jdbc/with-transacted-connection
+               (:scf-read-db globals)
+               :repeatable-read
+               (from-sql-date (scf-store/node-deactivated-time "foo.local")))))
```
senior: Is there a reason to reach inside PDB and invoke this vs POSTing a deactivated node command?

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -349,3 +350,11 @@
   (until-consumed
    (fn []
      (pdb-client/submit-command-via-http! base-url cmd version payload))))
+
+(defn wait-for-server-processing [server]
```
senior: I think this needs to have something that fails after some max time.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1298,6 +1300,45 @@
...
+          shared-globals (partial cli-svc/shared-globals pdb)
+          enqueue-command (partial enqueue-command dispatcher)
+          globals (shared-globals)
+          {{:keys [connection endpoint]} :command-mq} globals
+          deactivate-ms 14250331086887
+          ;; The problem only occurred if you passed a Date to
+          ;; enqueue, a DateTime wasn't a problem.
+          input-stamp (java.util.Date. deactivate-ms)
+          expected-stamp (DateTime. deactivate-ms DateTimeZone/UTC)]
+      (enqueue-command connection endpoint
```
senior: Is there a reason to reach inside PDB and invoke this vs POSTing a deactivated node command?

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1298,6 +1300,45 @@
...
+               (from-sql-date (scf-store/node-deactivated-time "foo.local")))))
+      (is (= expected-stamp
+             (-> (client/get (str (utils/base-url->str svc-utils/*base-url*)
+                                  "/nodes")
+                             {:accept :json
+                              :throw-exceptions true
+                              :throw-entire-message true
+                              :query-params {"query"
+                                             (json/generate-string
+                                              ["or" ["=" ["node" "active"] true]
```
senior: Do you need this query clause because you deactivated the node in the previous command?

test/puppetlabs/puppetdb/http/events_test.clj
```diff
@@ -349,7 +349,7 @@
        :new_value nil
        :containing_class "Foo"
        :report_receive_time "2014-04-16T12:44:40.978Z"
-       :report "99ec099bed6dfb9bff2c7df7828270e95f590147"
+       :report "7d0cfa08901e1e1d80cf2f2f814d356d0e457e09"
```
senior: Did we see the hash change because we were serializing the date to JSON before computing the hash? And thus truncating the milliseconds before computing the hash?

rbrw: I assumed it was just because the rendered formats were slightly different - I think I've seen something like "...8700Z" vs "...087Z"?

test/puppetlabs/puppetdb/command_test.clj
rbrw: Right - I wanted all the nodes, and was just matching what we do in sync/core.

test/puppetlabs/puppetdb/command_test.clj
rbrw: Now that enqueue is a method, I was just testing the relevant subsystem more directly, since we want it to work no matter how you reach it, but if we'd prefer to involve the http endpoint, then perhaps the test should also move to http.comand-test.

test/puppetlabs/puppetdb/command_test.clj
rbrw: If I understand what you're asking correctly, I just added this because it was easy, and tested exactly what was in the DB without the influence of any extra translations on the way out.  It *might* provide a bit of extra safety, but we don't have to have it.

src/puppetlabs/puppetdb/cheshire.clj
```diff
@@ -38,19 +38,32 @@
...
-     (generate-string obj (merge default-pretty-opts opts))))
+   (core/generate-string obj (merge default-pretty-opts opts))))
+
+(defn generate-stream
+  "Thinly wraps cheshire.core/generate-stream, adding the PuppetDB
+  default date format."
+  ([obj writer]
+   (core/generate-stream obj writer default-opts))
+  ([obj writer opts]
+   (core/generate-stream obj writer (merge default-opts opts))))
```
rbrw: I overlooked generate-stream in the previous version.

test/puppetlabs/puppetdb/http/events_test.clj
rbrw: OK, right, so at least as a practical matter, I added some code to print the json in generate string whenever including the pdb custom date format makes a difference, and saw this:

```
2011-01-01T15:00:01Z
2011-01-01T15:00:01.000Z
```
I was assuming that's what changed the hash, though I haven't investigated the report hashing explicitly.  But I can if you like.

test/puppetlabs/puppetdb/cheshire_test.clj
```diff
@@ -35,6 +36,14 @@
       (is (= (.toString sw)
              "{\n  \"a\" : 1,\n  \"b\" : 2\n}")))))
 
+(deftest test-date-generation
+  (let [t (java.util.Date. 5928174905781)]
```
ajroetker: How come your using the Java interop vs clj-time?

rbrw: This is just checks that our generate-string generates the same (expected) output for both a java.util.Date and a joda DateTime.   Before, Date (specifically), didn't work.  And now that you mention it, perhaps we should check Timestamp here too, since that's one of our important types.  I'll add it in a minute.  You can see the failure by changing generate-string to not include default-opts in the core/generate-string calls.

rbrw:
Added a Timestamp test to test-date-generation.

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -352,3 +355,28 @@
   (until-consumed
    (fn []
      (pdb-client/submit-command-via-http! base-url cmd version payload))))
+
+(defn wait-for-server-processing
+  "Returns a truthy value indicating whether the wait was
+  successful (i.e. didn't time out).  Given a false or nil timeout,
+  waits forever."
+  ([server] (wait-for-server-processing server nil))
```
senior: Some of this seems unnecessary. I don't see any usage of the 1 arg version. With that gone I think the code below becomes simpler.

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -352,3 +355,28 @@
...
+         (let [srv (get-service server :PuppetDBCommandDispatcher)
+               stats (dispatch/stats srv)]
+           (= (:executed-commands stats) (:received-commands stats)))
+         true
+
+         :else ;; In theory, polling might be avoided via watcher.
+         (do
+           (Thread/sleep (if-not deadline
+                           100
+                           (max 0 (min 100 (- deadline (now-ms))))))
```
senior: Any reason why we wouldn't always just sleep 100 then recur? Seems like that would make this simpler

test/puppetlabs/puppetdb/testutils/services.clj
rbrw: OK, we were using it before, and I thought I'd preserve it in case someone did want to wait forever, but I suppose they can manage that themselves, or we  can add support later.

test/puppetlabs/puppetdb/testutils/services.clj
rbrw: Just didn't want to miss the deadline.

rbrw: And I forgot, @mullr asked for this function to be busier in its previous incarnation, so I'll just change it to 10ms (as I did there), and omit the extra work.

PR #1565, opened 2015-07-29
(PDB-1721) memory starvation with large structured facts

wkalt:
Our fact storage mechanism previously created some very large prepared
statements when very large structured facts were updated. This changes things
so that our fact-storage related prepared statements use a single array-valued
parameter, so that a single statement should be applicable to all scenarios.

This change appears to have eased the original problem and also improves fact
command processing speed significantly.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -43,7 +43,8 @@
             [metrics.timers :refer [timer time!]]
             [puppetlabs.puppetdb.jdbc :refer [query-to-vec]]
             [puppetlabs.puppetdb.time :refer [to-timestamp]]
-            [honeysql.core :as hcore]))
+            [honeysql.core :as hcore])
+  (:import [org.postgresql.util PGobject]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Schemas
@@ -784,20 +785,32 @@
 
 (defn-validated delete-pending-path-id-orphans!
   "Delete paths in dropped-pids that are no longer mentioned
-  in other factsets."
+   in other factsets."
   [factset-id dropped-pids]
   (when-let [dropped-pids (seq dropped-pids)]
-    (let [in-pids (jdbc/in-clause dropped-pids)]
-      (sql/do-prepared
-       (format
-        "DELETE FROM fact_paths fp
-           WHERE fp.id %s
+    (if (sutils/postgres?)
+      (let [candidate-paths (sutils/array-to-param "bigint" Long dropped-pids)]
+        (sql/do-prepared
+          "DELETE FROM fact_paths fp
+             WHERE fp.id = ANY (?)
              AND NOT EXISTS (SELECT 1 FROM facts f
-                               WHERE f.fact_path_id %s
-                                 AND f.fact_path_id = fp.id
-                                 AND f.factset_id <> ?)"
-        in-pids in-pids)
-       (concat dropped-pids dropped-pids [factset-id])))))
+                             WHERE f.fact_path_id = ANY (?)
+                             AND f.fact_path_id = fp.id
+                             AND f.factset_id <> ?)"
+          (concat [candidate-paths]
+                  [candidate-paths]
+                  [factset-id])))
+      (let [in-pids (jdbc/in-clause dropped-pids)]
+        (sql/do-prepared
+          (format
+            "DELETE FROM fact_paths fp
+               WHERE fp.id %s
+               AND NOT EXISTS (SELECT 1 FROM facts f
+                 WHERE f.fact_path_id %s
+                 AND f.fact_path_id = fp.id
+                 AND f.factset_id <> ?)"
+            in-pids in-pids)
+          (concat dropped-pids dropped-pids [factset-id]))))))
 
 (defn-validated delete-pending-value-id-orphans!
   "Delete values in removed-pid-vid-pairs that are no longer mentioned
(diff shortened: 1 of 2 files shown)


wkalt:
selection_004

the plot above represents two trials of benchmark -N 3 -n 500 -r 20 -F <our facts>, where the first bump is with the patch and the second without. Notice that the swell in queue size is much smaller with the changes.

wkalt:
before:
selection_007

after (after running for significantly longer than above)
selection_010

My experience with the preexisting case is that allowing the benchmark tool to continue to run (with some very large custom facts) results in steady growth of the number of prepared statement objects and retained size of 10+% within an hour or two, whereas in the after case the number wasn't really climbing above 557/2% retained size, even after 45 minutes or so. That's promising, but I'm still not sure what will happen in the case of running under load for weeks (which was the original context). 4% retained size down to 2% doesn't really prove much -- it's a question of whether those objects continue to be opened faster than they are closed, and whether those percentages can continue to grow.

In the case that this doesn't fix the issue, we could a) extend these changes to cases unrelated to facts and see if that helps (although in the reported case, both users were hitting this with huge facts) or b) assume that we're suffering from some naive caching done by BoneCP, and redirect efforts toward the planned move to HikariCP. The root issue can be worked around by supplying statements-cache-size=0 in database.ini. Doing this may hurt storage performance, but eliminates the memory issue.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -784,42 +785,69 @@
...
-       (concat dropped-pids dropped-pids [factset-id])))))
+      (if (sutils/postgres?)
+        (sql/do-prepared
+          "DELETE FROM fact_paths fp
+           WHERE fp.id = ANY (?)
+           AND NOT EXISTS (SELECT 1 FROM facts f
+             WHERE f.fact_path_id = ANY (?)
+             AND f.fact_path_id = fp.id
+             AND f.factset_id <> ?)"
+          (flatten [(sutils/array-to-param "bigint" Long dropped-pids)
```
mullr: Does this have to be two calls to array-to-param, or can we pass the same object in both positions? Also, how does `flatten` work here? That doesn't make sense to me.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -784,42 +785,69 @@
 
 (defn-validated delete-pending-path-id-orphans!
   "Delete paths in dropped-pids that are no longer mentioned
-  in other factsets."
+   in other factsets."
   [factset-id dropped-pids]
   (when-let [dropped-pids (seq dropped-pids)]
     (let [in-pids (jdbc/in-clause dropped-pids)]
```
mullr: looks like this is only used in the postgres case, can we hide it down there?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -784,42 +785,69 @@
...
-                                 AND f.fact_path_id = fp.id
-                                 AND f.factset_id <> ?)"
-        in-pids in-pids)
-       (concat dropped-pids dropped-pids [factset-id])))))
+      (if (sutils/postgres?)
+        (sql/do-prepared
+          "DELETE FROM fact_paths fp
+           WHERE fp.id = ANY (?)
+           AND NOT EXISTS (SELECT 1 FROM facts f
+             WHERE f.fact_path_id = ANY (?)
```
mullr: Can this be skipped? The next line says that fact_path_id = fp.id, which above is already one of dropped-pids.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -784,42 +785,69 @@
...
-        in-vids in-vids in-rm-facts)
-       (flatten [vids vids rm-facts])))))
+               WHERE f.fact_value_id = ANY(?)
+               AND f.fact_value_id = fv.id
+               AND (f.factset_id,
+                 f.fact_path_id,
+                 f.fact_value_id)
+               NOT %s)"
+            in-rm-facts)
+          (flatten [(sutils/array-to-param "bigint" Long vids)
```
mullr: double calls to array-to-param, as above

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -784,42 +785,69 @@
...
              AND NOT EXISTS (SELECT 1 FROM facts f
-                               WHERE f.fact_value_id %s
-                                 AND f.fact_value_id = fv.id
-                                 AND (f.factset_id,
-                                      f.fact_path_id,
-                                      f.fact_value_id) NOT %s)"
-        in-vids in-vids in-rm-facts)
-       (flatten [vids vids rm-facts])))))
+               WHERE f.fact_value_id = ANY(?)
+               AND f.fact_value_id = fv.id
```
mullr: possible redundant conditions, as above

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -784,42 +785,69 @@
...
+               AND (f.factset_id,
+                 f.fact_path_id,
+                 f.fact_value_id)
+               NOT %s)"
```
mullr: As we talked about IRL, it would be great if this was passed as an array parameter too

mullr:
Going to hold off on functional tests until some of the pending changes go through, especially the additional use of array parameters

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: @rbrw you and I talked about this months ago -- do you remember the reason?

rbrw: Took me a bit, but I think I do.  I believe it has to do with the way a correlated subquery like that works.  Without the additional guard, I think it may not correctly match up the subqueries with the candidate delete rows.  Here's (hopefully) an example you can run via psql: https://gist.github.com/rbrw/4c855c8e9c0d42b90e51

If we decide that's correct, I can add some relevant comments to the code this time.

mullr:
tried this out locally, seeing some good improvements in memory usage. Noticed some other things while looking around the heap dump; we really need to evaluate the performance of just shutting off the statement cache. 

:+1: from me

wkalt:
@russ that's next on my agenda

russ:
@wkalt I think I got tagged here wrong.

wkalt:
@russ sorry about that

wkalt:
@mullr @senior I got a bad feeling about the patched delete query on Friday (the followup change with the three arrays that we discussed.) I confirmed that it introduces a pretty serious regression for very large structured facts relative to the original version of this patch by collecting some data from our mbeans during a benchmark run with n=10, N=250, r=20 for stable, 39fcb4, and cfd019, with this massive structured factset added to our sample data https://gist.github.com/wkalt/1b1b74c3c79a254d3a46.

The commands data (combined from the three runs) is here:
 https://gist.github.com/wkalt/a6b238bb1c5dc262b4a3

and the queue data is here: 
https://gist.github.com/wkalt/0ca74979634c06053cd7

memory_usage_in_bytes
one_minute_command_processing_rate
queue_size

For the ordinary benchmark set the differences seem relatively immaterial. cfd019 performs slightly worse than the others, but to me it looks more like some kind of system hiccup that would smooth out with additional trials (I haven't done this though.) That commands data is here: https://gist.github.com/267442c0fa107aaa8ad6 and the queue data is here: https://gist.github.com/1d4929215b3cd9aa37df. The corresponding images for the metrics above are below.

memory_usage_in_bytes
queue_size
one_minute_command_processing_rate

I left the revert as a separate commit so you can check it out yourselves. I'll fix it up when we're satisfied.

mullr:
@wkalt It looks like the top three graphs are all showing the same thing - that commands are processing much more slowly. Do you have a sense of why? Are the explains on that delete query different?

wkalt:
@mullr Correct.

The new query plan is here: https://gist.github.com/wkalt/5edb917b393c68bdd765
http://explain.depesz.com/s/TMpF

Here's a mangled plan for the old one: https://gist.github.com/wkalt/6525e4e7f43bd618e5fd
I pulled out the huge lists of query params.

I think the issue could be deeper than the long query though -- in reality those new queries were basically always getting rolled back due to conflicts. This could just be due to the duration, but I wonder if it's being affected by the concurrent inserts in a different way to the old.

PR #1568, opened 2015-07-31
(PDB-1845) Treat the log config as a global resource 

rbrw:
(PDB-1845) Treat the log config as global resource

Previously pdb-instance invocations would pass a custom config to
trapperkeeper.  Although it probably intended for the changes to be
per-instance, the logging configuration is global, and so whatever
changes were made would apply to everyone, and the original
configuration would never be restored.

As an alternative, add a new with-log-suppressed-unless-notable macro
that can wrap the dynamic lifetime of any number of pdb-instances.  It
will suppress all logging unless an event matches its notable-event?
predicate, and if a notable event is encountered, it will dump the log
to *err* and provide a path to the temporary log file.

Include notable-pdb-event?, which ignores the same spurious errors
previously ignored by modifications introduced via the
removed *extra-appender-config*.

To support of the changes, add with-additional-log-appenders and
with-log-appenders (and because it was trivial and seems useful,
with-log-event-listener).

Finally, rework the existing operations to use the new ones when
appropriate.  Rename puppetdb-instance to call-with-puppetdb-instance to
reflect the way it works, and add two new operations:
call-with-lone-log-suppressed-puppetdb-instance and
with-lone-log-suppressed-puppetdb-instance.  They wrap the previous
operations in with-log-suppressed-unless-notable and allow make it
possible to update the previous call sites without additional
wrapping/indentation.

### DIFF
--- a/test/puppetlabs/puppetdb/testutils/log.clj
+++ b/test/puppetlabs/puppetdb/testutils/log.clj
@@ -1,0 +1,55 @@
+(ns puppetlabs.puppetdb.testutils.log
+  (:require [puppetlabs.kitchensink.core :as kitchensink])
+  (:import [ch.qos.logback.core spi.LifeCycle Appender]
+           [ch.qos.logback.classic Level Logger]
+           [org.slf4j LoggerFactory]))
+
+(defmacro with-log-level
+  "Sets the (logback) log level for the logger specified by logger-id
+  during the execution of body.  If logger-id is not a class, it is
+  converted via str, and the level must be a clojure.tools.logging
+  key, i.e. :info, :error, etc."
+  [logger-id level & body]
+  ;; Assumes use of logback (i.e. logger supports Levels).
+  `(let [logger-id# ~logger-id
+         logger-id# (if (class? logger-id#) logger-id# (str logger-id#))
+         logger# (.getLogger (LoggerFactory/getILoggerFactory) logger-id#)
+         original-level# (.getLevel logger#)]
+     (.setLevel logger# (case ~level
+                          :trace Level/TRACE
+                          :debug Level/DEBUG
+                          :info Level/INFO
+                          :warn Level/WARN
+                          :error Level/ERROR
+                          :fatal Level/ERROR))
+     (try
+       (do ~@body)
+       (finally (.setLevel logger# original-level#)))))
+
+(defn create-log-appender-to-atom
+  [destination-atom]
+  ;; No clue yet if we're supposed to start with a default name.
+  (let [name (atom (str "log-appender-to-atom-" (kitchensink/uuid)))]
+    (reify
+      Appender
+      (doAppend [this event] (swap! destination-atom conj event))
+      (getName [this] @name)
+      (setName [this x] (reset! name x))
+      LifeCycle
+      (start [this] true)
+      (stop [this] true))))
+
+(defmacro with-logging-to-atom
+  "Conjoins all logger-id events produced during the execution of the
+  body to the destination atom, which must contain a collection.  If
+  logger-id is not a class, it is converted via str."
+  [logger-id destination & body]
+  ;; Specify the root logger via org.slf4j.Logger/ROOT_LOGGER_NAME.
+  `(let [logger-id# ~logger-id
+         logger-id# (if (class? logger-id#) logger-id# (str logger-id#))
+         logger# (.getLogger (LoggerFactory/getILoggerFactory) logger-id#)
+         appender# (doto (create-log-appender-to-atom ~destination) .start)]
+     (.addAppender logger# appender#)
+     (try
+       (do ~@body)
+       (finally (.detachAppender logger# appender#)))))
--- a/test/puppetlabs/puppetdb/testutils/log.clj
+++ b/test/puppetlabs/puppetdb/testutils/log.clj
@@ -1,8 +1,35 @@
 (ns puppetlabs.puppetdb.testutils.log
-  (:require [puppetlabs.kitchensink.core :as kitchensink])
-  (:import [ch.qos.logback.core spi.LifeCycle Appender]
-           [ch.qos.logback.classic Level Logger]
-           [org.slf4j LoggerFactory]))
+  (:require
+   [puppetlabs.kitchensink.core :as kitchensink]
+   [puppetlabs.puppetdb.testutils :refer [temp-file]]
+   [me.raynes.fs :as fs])
+  (:import
+   [ch.qos.logback.core Appender FileAppender]
+   [ch.qos.logback.core.spi LifeCycle]
+   [ch.qos.logback.classic Level Logger]
+   [ch.qos.logback.classic.encoder PatternLayoutEncoder]
+   [org.slf4j LoggerFactory]))
+
+(defmacro with-started
+  "Ensures that if a given name's init form executes without throwing
+  an exception, (.stop name) will be called before returning from
+  with-started.  It is the responsibility of the init form to make
+  sure the object has been started.  This macro behaves like
+  with-open, but with respect to .stop instead of .close."
+  [[name init & bindings] & body]
+  (if-let [s (seq bindings)]
+    `(let [~name ~init]
+       (try
+         (with-started ~bindings ~@body)
+         (finally (.stop ~name))))
+    `(let [~name ~init]
+       (try
+         ~@body
+         (finally (.stop ~name))))))
+
+(defn- find-logger [id]
+  (.getLogger (LoggerFactory/getILoggerFactory)
+              (if (class? id) id (str id))))
 
 (defmacro with-log-level
   "Sets the (logback) log level for the logger specified by logger-id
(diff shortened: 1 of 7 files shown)


test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,173 @@
...
+   [ch.qos.logback.core Appender FileAppender]
+   [ch.qos.logback.core.filter EvaluatorFilter]
+   [ch.qos.logback.core.spi FilterReply LifeCycle]
+   [ch.qos.logback.classic Level Logger]
+   [ch.qos.logback.classic.boolex JaninoEventEvaluator]
+   [ch.qos.logback.classic.encoder PatternLayoutEncoder]
+   [org.slf4j LoggerFactory]))
+
+(defn- find-logger [id]
+  (.getLogger (LoggerFactory/getILoggerFactory)
```
mullr: maybe this should delegate to `c.t.l.i.*logger-factory*`

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,173 @@
...
+  "Sets the (logback) log level for the logger specified by logger-id
+  during the execution of body.  If logger-id is not a class, it is
+  converted via str, and the level must be a clojure.tools.logging
+  key, i.e. :info, :error, etc."
+  [logger-id level & body]
+  ;; Specify the root logger via org.slf4j.Logger/ROOT_LOGGER_NAME.
+  ;; Assumes use of logback (i.e. logger supports Levels).
+  `(let [logger# (#'puppetlabs.puppetdb.testutils.log/find-logger ~logger-id)
+         original-level# (.getLevel logger#)]
+     (.setLevel logger# (case ~level
```
mullr: maybe this should be inside the `try`?

mullr:
To answer your question: I think this is a really good approach. It should be easier to deal with than the log file generation.

cprice404:
Would you guys consider targeting some of these changes at trapperkeepers logging testutils rather than just doing them in PuppetDB?

senior:
@cprice404 Sure, @rbrw was pushing this up more as a question, "should we do this?". If you think it's useful to others we can push it forward on the TK side.

cprice404:
@senior awesome, thanks!  I've only skimmed it but it seems pretty generally applicable so that would be cool.

test/puppetlabs/puppetdb/testutils/log.clj
rbrw: Hmm, yeah, not sure - I normally don't initiate the try until successfully establishing the thing I want to undo in the finally, but given the semantics here, it would probably work fine to have the initial setLevel inside the try - as long as setLevel is sane, it shouldn't matter, but in this case it shouldn't hurt and might be safer to have it in the try...

rbrw:
@cprice404 I think we're planning to let this shake out a bit in PDB first, and then we can propose it for TK.  We've already discovered (and fixed) some things that needed reworking.  I've also filed a ticket so we don't forget: https://tickets.puppetlabs.com/browse/PDB-1881

wkalt:
this needs a rebase @rbrw

test/puppetlabs/puppetdb/testutils/log_test.clj
```diff
@@ -0,0 +1,77 @@
...
+        uuid (kitchensink/uuid)]
+    (tgt/with-additional-log-appenders org.slf4j.Logger/ROOT_LOGGER_NAME
+      [(#'puppetlabs.puppetdb.testutils.log/log-event-listener
+        (fn [event] (swap! log conj event)))]
+      (log/error uuid))
+    (is (some #(= {:level "ERROR" :message uuid} %)
+              (map event->map @log)))))
+
+(deftest with-log-appenders
+  ;; Definitely incomplete
```
wkalt: let's address or remove these comments -- not sure what they refer to

rbrw: I believe the main thing I meant there is that both of these tests only test the happy path.  They don't, for example, test that with-log-appenders clobbers all existing appenders, nor that with-additional-log-appenders doesn't.  I just realized that it might be fairly easy to add simple tests for those case.  I'll try.

rbrw: Added additional checks.

rbrw:

rbrw:

rbrw:

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -192,22 +104,30 @@
...
+(defn call-with-lone-log-suppressed-puppetdb-instance
+  "Calls the call-with-puppetdb-instance with args after suppressing
+  all log events.  If there's an error or worse, prints the log to
+  *err*.  Should not be nested, nor nested with calls to
+  with-log-suppressed-unless-notable."
+  [& args]
+  (with-log-suppressed-unless-notable notable-pdb-event?
+    (apply call-with-puppetdb-instance args)))
+
+(defmacro with-lone-log-suppressed-puppetdb-instance
```
mullr: `with-puppetdb-instance-nolog`?

or `(with-puppetdb-instance {:log :suppress} ...)`?

or something like that. 

Also, I don't understand the 'lone' distinction.

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,164 @@
...
+
+(defmacro with-log-level
+  "Sets the (logback) log level for the logger specified by logger-id
+  during the execution of body.  If logger-id is not a class, it is
+  converted via str, and the level must be a clojure.tools.logging
+  key, i.e. :info, :error, etc."
+  [logger-id level & body]
+  ;; Specify the root logger via org.slf4j.Logger/ROOT_LOGGER_NAME.
+  ;; Assumes use of logback (i.e. logger supports Levels).
+  `(let [logger# (#'puppetlabs.puppetdb.testutils.log/find-logger ~logger-id)
```
mullr: I think `#'puppetlabs.puppetdb.testutils.log/` isn't needed

mullr: Oh, it's because find-logger is private isn't it.

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,164 @@
...
+      (stop [this] true))))
+
+(defn- with-additional-log-appenders-fn [logger-id appenders body]
+  (let [logger (find-logger logger-id)]
+    (try
+      (doseq [appender appenders]
+        (.addAppender logger appender))
+      (body)
+      (finally
+        (doseq [appender appenders]
```
mullr: no issue with taking them off in the same order they were added? Might `(doseq [appender (reverse appenders)] ...)` be better? (unlikely to matter, just a thought)

test/puppetlabs/puppetdb/testutils/services.clj
rbrw: This can only be used with one instance since with say two, you need to wrap both with a single with-*.  The "lone" calls just make it possible to replace the existing invocations without any broader reindentation (which we'd need if we added the new wrapper for each).

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,164 @@
...
+
+(defmacro with-log-event-listener
+  "Calls (listen event) for each logger-id event produced during the
+  execution of the body.  If logger-id is not a class, it is converted
+  via str."
+  [logger-id listen & body]
+  ;; Specify the root logger via org.slf4j.Logger/ROOT_LOGGER_NAME.
+  `(with-additional-log-appenders ~logger-id
+     [(doto (#'puppetlabs.puppetdb.testutils.log/log-event-listener ~listen)
+        .start)]
```
mullr: does this need to be stopped?

test/puppetlabs/puppetdb/testutils/log.clj
rbrw: Yes, it's so the expansion will still work.  I'm pretty sure I verified that, and it's also needed iirc in (for example) Guile.

test/puppetlabs/puppetdb/testutils/services.clj
rbrw: Oh, and in this code, I've used the scheme convention that (with-foo ...) is the macro that takes a body, and (call-with-foo ...) is the version that accepts a function.

mullr: Yeah, I noticed that. But with-log-appenders-fn doesn't follow that.

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,164 @@
...
+  body to the destination atom, which must contain a collection.  If
+  logger-id is not a class, it is converted via str."
+  [logger-id destination & body]
+  ;; Specify the root logger via org.slf4j.Logger/ROOT_LOGGER_NAME.
+  `(with-log-event-listener
+     ~logger-id
+     (fn [event#] (swap! ~destination conj event#))
+     ~@body))
+
+(defn- suppressing-appender
```
mullr: suppressing-file-appender, perhaps?

test/puppetlabs/puppetdb/testutils/log.clj
rbrw: No idea if it matters, but I can reverse it "just in case"

mullr: it's just a question of what the implementation of addAppender /
removeAppender actually does. Adding it if it doesn't matter will imply to
future people that it does matter, which I don't want to do.

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,164 @@
...
+        (f))
+      (finally
+        (.stop appender)
+        (when @problem
+          (binding [*out* *err*]
+            (print (slurp log-path))
+            (println "From error log:" log-path)))
+        (when-not @problem (fs/delete log-path))))))
+
+(defmacro with-log-suppressed-unless-notable
```
mullr: Can I do this and also log to an atom? Do I care?

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,164 @@
...
+                 (when-let [cause (.getThrowableProxy event)]
+                   (re-find annoying-peer-error-rx (.getMessage cause)))))))
+
+(defn- with-log-suppressed-unless-notable-fn [notable-event? f]
+  (let [problem (atom false)
+        detector (doto (#'puppetlabs.puppetdb.testutils.log/log-event-listener
+                        (fn [event]
+                          (when (notable-event? event)
+                            (reset! problem true))))
+                   .start)
```
mullr: As above, does this need to be stopped?

test/puppetlabs/puppetdb/testutils/log.clj
rbrw: Right, but I didn't see any promises when I was trying to figure out how it works: http://logback.qos.ch/apidocs/ch/qos/logback/classic/Logger.html#detachAppender%28java.lang.String%29

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,164 @@
...
+    (try
+      (with-log-appenders org.slf4j.Logger/ROOT_LOGGER_NAME
+        [appender detector]
+        (f))
+      (finally
+        (.stop appender)
+        (when @problem
+          (binding [*out* *err*]
+            (print (slurp log-path))
+            (println "From error log:" log-path)))
```
mullr: Ahhh, this is why you're hitting the disk. I like this.

test/puppetlabs/puppetdb/testutils/log.clj
rbrw: I'm fine either way - if I did add the reverse, I had planned to add it with a side comment like "reverse may not be required".

mullr:
Ok, this looks generally good. The only real issue is the lack of a .stop for two .starts, if it's needed. Thumbs up from me after that's been addressed.

test/puppetlabs/puppetdb/testutils/services.clj
rbrw: Oh, right - that predated my decision.  I can fix it.

test/puppetlabs/puppetdb/testutils/log.clj
rbrw: It shouldn't matter in this case since our appender's stop method just returns true, but perhaps still good practice; I'll add that.

test/puppetlabs/puppetdb/testutils/log.clj
rbrw: If I recall correctly, only if you establish the atom logger inside the body (since this macro uses with-log-appenders, which intentionally clobbers everything for the duration).

test/puppetlabs/puppetdb/testutils/log.clj
rbrw: Yeah, in that respect it matches what we had before, just not per-instance.

test/puppetlabs/puppetdb/testutils/log.clj
rbrw: It appears that the current Logger compares the appenders via equal() and adds or removes them accordingly, ignoring duplicates.

rbrw:
OK, hopefully I've accommodated the current comments with the latest push.

rbrw:

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -0,0 +1,183 @@
...
+
+(def ^:private annoying-peer-error-rx #"peer.*vm://localhost.*stopped")
+
+(defn notable-pdb-event? [event]
+  (and (.isGreaterOrEqual (.getLevel event) Level/ERROR)
+       (not (and (-> (.getFormattedMessage event)
+                     (.contains "queue connection error"))
+                 (when-let [cause (.getThrowableProxy event)]
+                   (re-find annoying-peer-error-rx (.getMessage cause)))))))
+
```
rbrw: This of course won't go into any future indepdendent lib...

mullr:
Just reverted this, since we didn't have the corresponding code ready in PE-land

PR #1571, opened 2015-08-03
(maint) default statement cache size to 0

wkalt:
I haven't been able to see a benefit of a non-zero cache size at any scale I've
seen, either for all three commands or just catalogs (measured in terms of
one-minute command processing rate). Setting it to 0 is the easy fix for the
memory leak issues we've been seeing. Seems like defaulting it to that and
leaving it configurable is the way to go.


ajroetker:
Fixes the memory leak from what I saw, 4.2% of the heap to 0.1%, and I didn't observe any hits to command-processing either with a benchmarking tool run

PR #1579, opened 2015-08-05
(PDB-1879) to-timestamp: always return Timestamp

rbrw:
The "timestamp" in clj-time.coerce/to-timestamp refers to
java.sql.Timestamp, which it always returns.  Do the same from
PuppetDB's to-timestamp (which wraps it).

Note that this change, by itself, should not be able to cause any (new)
potential timestamp truncation issues because from-string returns a
DateTime (millisecond resolution) and Timestamp is more
accurate (nanosecond resolution).

### DIFF
--- a/src/puppetlabs/puppetdb/time.clj
+++ b/src/puppetlabs/puppetdb/time.clj
@@ -235,5 +235,6 @@
   [ts]
   (tc/to-timestamp
    (if (string? ts)
-     (from-string ts)
+     (when-let [dt (from-string ts)]
+       (java.sql.Timestamp. (.getMillis dt)))
      ts)))


src/puppetlabs/puppetdb/time.clj
```diff
@@ -235,5 +235,6 @@
   [ts]
   (tc/to-timestamp
    (if (string? ts)
-     (from-string ts)
+     (when-let [dt (from-string ts)]
+       (java.sql.Timestamp. (.getMillis dt)))
```
ajroetker: Should this be happening in `from-string`?

rbrw: My understanding is that from-string is defined to return a DateTime (theirs always does, and ours might, or might return nil).  And to-timestamp is defined to do the same, but for a Timestamp.

rbrw:
I think this was just wrong-headed (I suspect I got confused by the JSON date issue).  In any case, I don't see any way that clj-time could return anything other than a Timestamp, and we have a schema guard on the return type anyway.  Thanks to @ajroetker for asking the relevant questions.

PR #1582, opened 2015-08-06
(packaging) Bump ezbake to 0.3.18

shrug:


wkalt:
@shrug looks like we've got issues with el7 here

PR #1583, opened 2015-08-07
(maint) stop reusing import-export-test functions

wkalt:
Before this patch, we had some functions in our import-export-roundtrip test
that were used in other test namespaces.  This prevented us from using our
fixtures namespace in any of the http tests due to a cyclic dependency, which
I need for PDB-1737. This will break sync testing, but the fix will be
straightforward.


rbrw:
This looks fine to me and passes tests against both DBs.  Though (minor) if we have another version, I might find a commit summary line like "relocate import/export util functions" a little clearer.

PR #1584, opened 2015-08-10
(PDB-1653) Refactor to a one top level web route for /pdb

senior:
This commit moves the configuration of the ring routes to a single namespace (pdb-routing). Previously several components (dashboard, meta, admin) existed solely to attach a new top level pdb
route (i.e. /pdb/meta). This refactor sets us up to better accommodate global (from a PuppetDB perspective) functionality like setting PuppetDB in "maintenance mode" while starting.

### DIFF
--- a/resources/puppetlabs/puppetdb/bootstrap.cfg
+++ b/resources/puppetlabs/puppetdb/bootstrap.cfg
@@ -13,11 +13,8 @@ puppetlabs.trapperkeeper.services.webrouting.webrouting-service/webrouting-servi
 puppetlabs.puppetdb.cli.services/puppetdb-service
 puppetlabs.puppetdb.http.command/puppetdb-command-service
 puppetlabs.puppetdb.metrics/metrics-service
-puppetlabs.puppetdb.dashboard/dashboard-service
-puppetlabs.puppetdb.admin/admin-service
 puppetlabs.puppetdb.mq-listener/message-listener-service
 puppetlabs.puppetdb.command/command-service
-puppetlabs.puppetdb.meta/metadata-service
 
 # NREPL
 puppetlabs.trapperkeeper.services.nrepl.nrepl-service/nrepl-service
--- a/resources/puppetlabs/puppetdb/bootstrap.cfg
+++ b/resources/puppetlabs/puppetdb/bootstrap.cfg
@@ -15,6 +15,8 @@ puppetlabs.puppetdb.http.command/puppetdb-command-service
 puppetlabs.puppetdb.metrics/metrics-service
 puppetlabs.puppetdb.mq-listener/message-listener-service
 puppetlabs.puppetdb.command/command-service
+puppetlabs.puppetdb.pdb-routing/maint-mode-service
+puppetlabs.puppetdb.pdb-routing/pdb-routing-service
 
 # NREPL
 puppetlabs.trapperkeeper.services.nrepl.nrepl-service/nrepl-service
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -274,7 +274,7 @@
         ;;is also sensitive to startup order. get-registered-endpoints
         ;;only has one item in the map at that point (metrics gets
         ;;added later
-        url-prefix (ffirst (get-registered-endpoints))
+        url-prefix (str (ffirst (get-registered-endpoints)) "/query")
 
         write-db (pl-jdbc/pooled-datasource database)
         read-db (pl-jdbc/pooled-datasource (assoc read-database :read-only? true))
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -314,8 +314,6 @@
       (when-not disable-update-checking
         (maybe-check-for-updates product-name update-server read-db))
 
-      (log/info "Starting query server")
-
       ;; Pretty much this helper just knows our job-pool and gc-interval
       (let [job-pool (mk-pool)
             gc-interval-millis (to-millis gc-interval)
@@ -364,7 +362,10 @@
   (set-url-prefix [this url-prefix]
                   (let [old-url-prefix (:url-prefix (service-context this))]
                     (when-not (compare-and-set! old-url-prefix nil url-prefix)
-                      (throw (RuntimeException. (format "Attempt to set url-prefix to %s when it's already been set to %s" url-prefix @old-url-prefix))))))
+                      (throw+ {:url-prefix old-url-prefix
+                               :new-url-prefix url-prefix
+                               :type ::url-prefix-already-set}
+                              (format "Attempt to set url-prefix to %s when it's already been set to %s" url-prefix @old-url-prefix)))))
   (shared-globals [this]
                   (:shared-globals (service-context this)))
   (query [this query-obj version query-expr paging-options row-callback-fn]
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -248,7 +248,7 @@
       (initialize-schema db-pool-map product-name))))
 
 (defn start-puppetdb
-  [context config service get-route]
+  [context config service get-registered-endpoints]
   {:pre [(map? context)
          (map? config)]
    :post [(map? %)
@@ -264,7 +264,18 @@
                  max-frame-size threads]} command-processing
         {:keys [certificate-whitelist
                 disable-update-checking]} puppetdb
-        url-prefix "/pdb";;(get-route service)
+
+        ;;This is a hack to get around startup order, this service no
+        ;;longer manages the root context of PuppetDB, pdb-routing
+        ;;does. Since this feeds off of shared globals (which comes
+        ;;from this service) this URL prefix is needed here. When we
+        ;;refactor away from shared globals, this will go away and
+        ;;will be handed to the query routes by pdb-routing. Note this
+        ;;is also sensitive to startup order. get-registered-endpoints
+        ;;only has one item in the map at that point (metrics gets
+        ;;added later
+        url-prefix (ffirst (get-registered-endpoints))
+
         write-db (pl-jdbc/pooled-datasource database)
         read-db (pl-jdbc/pooled-datasource (assoc read-database :read-only? true))
         mq-dir (str (io/file vardir "mq"))
@@ -353,10 +364,10 @@
   that trapperkeeper will call on exit."
   PuppetDBServer
   [[:ConfigService get-config]
-   [:WebroutingService add-ring-handler get-route]]
+   [:WebroutingService add-ring-handler get-registered-endpoints]]
 
   (start [this context]
-         (start-puppetdb context (get-config) this get-route))
+         (start-puppetdb context (get-config) this get-registered-endpoints))
 
   (stop [this context]
         (stop-puppetdb context))
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -248,10 +248,9 @@
       (initialize-schema db-pool-map product-name))))
 
 (defn start-puppetdb
-  [context config service add-ring-handler get-route]
+  [context config service get-route]
   {:pre [(map? context)
-         (map? config)
-         (ifn? add-ring-handler)]
+         (map? config)]
    :post [(map? %)
           (every? (partial contains? %) [:broker])]}
   (let [{:keys [global   jetty
@@ -265,7 +264,7 @@
                  max-frame-size threads]} command-processing
         {:keys [certificate-whitelist
                 disable-update-checking]} puppetdb
-        url-prefix (get-route service)
+        url-prefix "/pdb";;(get-route service)
         write-db (pl-jdbc/pooled-datasource database)
         read-db (pl-jdbc/pooled-datasource (assoc read-database :read-only? true))
         mq-dir (str (io/file vardir "mq"))
@@ -316,10 +315,7 @@
       (when-not disable-update-checking
         (maybe-check-for-updates product-name update-server read-db))
 
-      (let [app (->> (server/build-app globals)
-                     (compojure/context url-prefix []))]
-        (log/info "Starting query server")
-        (add-ring-handler service app))
+      (log/info "Starting query server")
 
       ;; Pretty much this helper just knows our job-pool and gc-interval
       (let [job-pool (mk-pool)
@@ -360,7 +356,7 @@
    [:WebroutingService add-ring-handler get-route]]
 
   (start [this context]
-         (start-puppetdb context (get-config) this add-ring-handler get-route))
+         (start-puppetdb context (get-config) this get-route))
 
   (stop [this context]
         (stop-puppetdb context))
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -265,17 +265,6 @@
         {:keys [certificate-whitelist
                 disable-update-checking]} puppetdb
 
-        ;;This is a hack to get around startup order, this service no
-        ;;longer manages the root context of PuppetDB, pdb-routing
-        ;;does. Since this feeds off of shared globals (which comes
-        ;;from this service) this URL prefix is needed here. When we
-        ;;refactor away from shared globals, this will go away and
-        ;;will be handed to the query routes by pdb-routing. Note this
-        ;;is also sensitive to startup order. get-registered-endpoints
-        ;;only has one item in the map at that point (metrics gets
-        ;;added later
-        url-prefix (str (ffirst (get-registered-endpoints)) "/query")
-
         write-db (pl-jdbc/pooled-datasource database)
         read-db (pl-jdbc/pooled-datasource (assoc read-database :read-only? true))
         mq-dir (str (io/file vardir "mq"))
@@ -313,7 +302,6 @@
                    :authorizer authorizer
                    :update-server update-server
                    :product-name product-name
-                   :url-prefix url-prefix
                    :discard-dir (.getAbsolutePath discard-dir)
                    :mq-addr mq-addr
                    :mq-dest mq-endpoint
@@ -354,6 +342,7 @@
 
 (defprotocol PuppetDBServer
   (shared-globals [this])
+  (set-url-prefix [this url-prefix])
   (query [this query-obj version query-expr paging-options row-callback-fn]
     "Call `row-callback-fn' for matching rows.  The `paging-options' should
     be a map containing :order_by, :offset, and/or :limit."))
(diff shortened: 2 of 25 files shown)


src/puppetlabs/puppetdb/pdb_routing.clj
```diff
@@ -0,0 +1,47 @@
...
+                                         rr/redirect))
+                     (apply compojure/routes app-routes)))
+
+(tk/defservice pdb-routing-service
+  [[:WebroutingService add-ring-handler get-route]
+   [:PuppetDBServer shared-globals query]
+   [:PuppetDBCommand submit-command]
+   [:PuppetDBCommandDispatcher enqueue-command enqueue-raw-command]]
+  (init [this context]
+        (assoc context ::pdb-routing (atom {})))
```
senior: This needs to be removed, no longer needed

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -263,7 +262,7 @@
                  max-frame-size threads]} command-processing
         {:keys [certificate-whitelist
                 disable-update-checking]} puppetdb
-        url-prefix (get-route service)
+        url-prefix "/pdb";;(get-route service)
```
senior: This is one of the causes of the WIP, I'm spending time now figuring out how to remove this.

senior:
I think I have narrowed the lifecycle issues down to one. This is the cause of the "url-prefix" hard-coding above and what I think will cause us problems with our maintenance mode response. The above routing service has a dependency on PDBServer. The good thing about this is if we're in the start function of routing, we know PDB has started. Since we don't attach the routes until the routing start function, users will get 404s until we have properly started up. Similarly we can have a PE specific service that knows about HA.

What I'm going to work on next is moving the route attaching to the init phase. Some routes require shared-globals to initialize, meaning PDBServer has to have started already (hence why it needs to be in the start phase now). Since I've moved the /pdb handler out of services and into routing, services can no longer interrogate web routing for it's own route (that route is in the pdb-routing ns now). There are web routing calls that get the list of endpoints, but when in PDBServer's start phase, those endpoints haven't been attached, so it just returns an empty map.

src/puppetlabs/puppetdb/pdb_routing.clj
```diff
@@ -0,0 +1,47 @@
...
+            [puppetlabs.puppetdb.http.server :as server]))
+
+
+(defn resource-request-middleware [req]
+  (resource-request req "public"))
+
+(defn pdb-core-routes [shared-globals submit-command-fn query-fn enqueue-raw-command-fn]
+  (map (fn [[uri route]]
+         (compojure/context uri [] route))
+       (partition 2 ["/meta" (meta/build-app shared-globals)
```
mullr: can make this a map to the same effect

senior:
Now that https://github.com/puppetlabs/puppetdb/pull/1589 has been merged, this is ready for review

resources/puppetlabs/puppetdb/bootstrap.cfg
```diff
@@ -13,11 +13,8 @@ puppetlabs.trapperkeeper.services.webrouting.webrouting-service/webrouting-servi
 puppetlabs.puppetdb.cli.services/puppetdb-service
 puppetlabs.puppetdb.http.command/puppetdb-command-service
 puppetlabs.puppetdb.metrics/metrics-service
-puppetlabs.puppetdb.dashboard/dashboard-service
-puppetlabs.puppetdb.admin/admin-service
 puppetlabs.puppetdb.mq-listener/message-listener-service
 puppetlabs.puppetdb.command/command-service
-puppetlabs.puppetdb.meta/metadata-service
```
ajroetker: These probably need a corresponding PR to update them in PE so we don't end up having to do https://github.com/puppetlabs/pe-puppetdb-extensions/pull/78 again

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -355,10 +362,10 @@
   that trapperkeeper will call on exit."
   PuppetDBServer
   [[:ConfigService get-config]
-   [:WebroutingService add-ring-handler get-route]]
+   [:WebroutingService add-ring-handler get-registered-endpoints]]
 
   (start [this context]
-         (start-puppetdb context (get-config) this add-ring-handler get-route))
+         (start-puppetdb context (get-config) this get-registered-endpoints))
```
ajroetker: Any reason to pass the `fn` vs calling the `fn` here and passing the result to `start-puppetdb`?

src/puppetlabs/puppetdb/meta.clj
```diff
@@ -26,58 +26,46 @@
    object containing a `version` key with the version, as well as a `newer` key
    which is a boolean specifying whether the latest version is newer
    than the current version."
-  [{:keys [update-server product-name scf-read-db]}]
-  {:pre [(and update-server product-name scf-read-db)]}
+  [get-shared-globals]
```
ajroetker: similarly here, could we just pass shared globals instead of the callback?

ajroetker:
@senior So is there a reason that we need to pass thunks around instead of passing through their results?

senior:
@ajroetker yes, in the init phase of pdb-routing startup, shared-globals hasn't been created. it's not until the start phase of PDBServer has completed is there a value there. This value isn't needed until the first request goes through (which won't happen until after that service has started anyway). Wrapping this in a thunk gives us the wiggle room we need until we fix that.

test/puppetlabs/puppetdb/pdb_routing_test.clj
```diff
@@ -0,0 +1,100 @@
...
+            [puppetlabs.puppetdb.client :as pdb-client]
+            [clj-time.coerce :refer [to-string]]
+            [clj-time.core :refer [now]]
+            [puppetlabs.puppetdb.testutils.dashboard :as dtu]
+            [puppetlabs.puppetdb.utils :as utils]
+            [puppetlabs.puppetdb.cli.services :as clisvc]
+            [puppetlabs.puppetdb.pdb-routing :refer :all]
+            [puppetlabs.trapperkeeper.app :as tk-app]))
+
+(defn command-base-url
```
ajroetker: I believe there's already a function for this in `utils.clj`

test/puppetlabs/puppetdb/pdb_routing_test.clj
```diff
@@ -0,0 +1,100 @@
...
+                              url-suffix)
+                         {:throw-exceptions false})]
+    (if (tu/json-content-type? resp)
+      (update resp :body #(json/parse-string % true))
+      resp)))
+
+(defn submit-facts [base-url facts]
+  (svc-utils/sync-command-post (assoc base-url
+                                 :prefix "/pdb/cmd"
+                                 :version :v1)
```
ajroetker: should probably be using your `command-base-url` function here

test/puppetlabs/puppetdb/pdb_routing_test.clj
```diff
@@ -0,0 +1,100 @@
...
+                               "replace facts"
+                               4
+                               facts))
+
+(defn query-fact-names [base-url]
+  (pdb-get (assoc svc-utils/*base-url*
+             :prefix "/pdb/query"
+             :version :v4)
```
ajroetker: There's a utils function for this as well

resources/puppetlabs/puppetdb/bootstrap.cfg
senior: https://github.com/puppetlabs/pe-puppetdb-extensions/pull/81

senior:

test/puppetlabs/puppetdb/pdb_routing_test.clj
senior: Changed mine to delegate to the utils one

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -31,23 +31,21 @@
...
+  returned to the client.  If get-authorizer is nil or false, all
+  requests will be accepted.  Otherwise it must accept no arguments
+  and return an authorize function that accepts a request.  The
+  request will be allowed only if authorize returns :authorized.
+  Otherwise, the return value should be a message describing the
+  reason that access was denied."
+  [app get-authorizer]
+  (fn [req]
+    (let [authorize (and get-authorizer (get-authorizer))
+          auth-result (if authorize (authorize req) :authorized)]
```
ajroetker: It really annoys me that we pass the authorizer through here and not just a list of whitelisted certnames

ajroetker: Not a problem with your patch just a general remark :)

src/puppetlabs/puppetdb/http/server.clj
```diff
@@ -38,16 +38,14 @@
...
+  "Generates a Ring application that handles PuppetDB requests.
+  If get-authorizer is nil or false, all requests will be accepted.
+  Otherwise it must accept no arguments and return an authorize
+  function that accepts a request.  The request will be allowed only
+  if authorize returns :authorized.  Otherwise, the return value
+  should be a message describing the reason that access was denied."
+  [get-authorizer get-shared-globals]
   (-> routes
-      (wrap-with-puppetdb-middleware authorizer)
+      (wrap-with-puppetdb-middleware get-authorizer)
```
ajroetker: Am I mistaken or did you say you were going to move this to a higher level? (wrapping `/pdb` as a whole)

src/puppetlabs/puppetdb/middleware.clj
senior: @ajroetker It would annoy me to pass around whitelisted certnames too, what about [this](https://github.com/senior/puppetdb/commit/00ebdc61cd839a57aff420134b3822fce5492af5] instead?

src/puppetlabs/puppetdb/http/server.clj
senior: I was waiting for this to go in first, but I have the change here: https://github.com/senior/puppetdb/commit/00ebdc61cd839a57aff420134b3822fce5492af5

src/puppetlabs/puppetdb/command.clj
```diff
@@ -367,26 +371,34 @@
...
-       #(process-command-and-respond! % config response-pub-chan (:stats context)))
+       #(process-command-and-respond! % config response-chan (:stats context)))
       context))
 
-  (stop [this {:keys [response-pub-chan] :as context}]
-    (async/close! response-pub-chan)
-    (dissoc context :response-pub :response-pub-chan))
+  (stop [this context]
+    (async/unsub-all (:response-pub context))
+    (async/untap-all (:response-mult context))
```
senior: Should we be closing the mult and the response-pub here?

mullr: They aren't closable

src/puppetlabs/puppetdb/http/server.clj
ajroetker: Ah I see I like that much better, sounds good to me!

ajroetker:

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -357,17 +353,23 @@
...
 
   (stop [this context]
         (stop-puppetdb context))
+  (set-url-prefix [this url-prefix]
+                  (let [old-url-prefix (:url-prefix (service-context this))]
+                    (when-not (compare-and-set! old-url-prefix nil url-prefix)
+                      (throw (RuntimeException. (format "Attempt to set url-prefix to %s when it's already been set to %s" url-prefix @old-url-prefix))))))
```
ajroetker: Because this is a "public" exception API, should we be using slingshot here?

src/puppetlabs/puppetdb/config.clj
```diff
@@ -335,13 +335,8 @@
...
   [config-data]
-  (let [default-web-router-service
-        {:puppetlabs.puppetdb.cli.services/puppetdb-service "/pdb/query"
-         :puppetlabs.puppetdb.http.command/puppetdb-command-service "/pdb/cmd"
-         :puppetlabs.puppetdb.meta/metadata-service "/pdb/meta"
-         :puppetlabs.puppetdb.admin/admin-service "/pdb/admin"
-         :puppetlabs.puppetdb.dashboard/dashboard-service "/pdb"
-         :puppetlabs.puppetdb.metrics/metrics-service "/metrics"}
+  (let [default-web-router-service {:puppetlabs.puppetdb.metrics/metrics-service "/metrics"
+                                    :puppetlabs.puppetdb.pdb-routing/pdb-routing-service "/pdb"}
```
ajroetker: I don't think you can do this if you want to swap in a PE version of the routing service, you'll have to do it conditionally like we do for the dashboard-redirect below

test/puppetlabs/puppetdb/pdb_routing_test.clj
```diff
@@ -0,0 +1,89 @@
...
+(defn command-base-url [{:keys [host port]}]
+  (utils/pdb-cmd-base-url host port :v1))
+
+(defn pdb-get [base-url url-suffix]
+  (let [resp (client/get (str (utils/base-url->str base-url)
+                              url-suffix)
+                         {:throw-exceptions false})]
+    (if (tu/json-content-type? resp)
+      (update resp :body #(json/parse-string % true))
+      resp)))
```
ajroetker: seems like this probably belongs in testutils or some such

ajroetker:

src/puppetlabs/puppetdb/pdb_routing.clj
```diff
@@ -0,0 +1,106 @@
...
+            [puppetlabs.trapperkeeper.services :as tksvc]
+            [ring.middleware.resource :refer [wrap-resource resource-request]]
+            [ring.util.request :as rreq]
+            [ring.util.response :as rr]
+            [puppetlabs.puppetdb.meta :as meta]
+            [puppetlabs.puppetdb.admin :as admin]
+            [puppetlabs.puppetdb.http.command :as cmd]
+            [puppetlabs.puppetdb.cli.services :as query]
+            [puppetlabs.puppetdb.http.server :as server]
+            [clojure.tools.logging :as log]))
```
wkalt: a few of these are unused

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -237,6 +237,14 @@
    :prefix "/pdb/cmd"
    :version (or version :v1)})
 
+(defn pdb-meta-base-url
+  [host port & [version]]
+  {:protocol "http"
+   :host host
+   :port port
+   :prefix "/pdb/meta"
+   :version (or version :v1)})
```
wkalt: what's the value of defaulting to v1?

test/puppetlabs/puppetdb/pdb_routing_test.clj
```diff
@@ -0,0 +1,89 @@
...
+            [clj-time.coerce :refer [to-string]]
+            [clj-time.core :refer [now]]
+            [puppetlabs.puppetdb.testutils.dashboard :as dtu]
+            [puppetlabs.puppetdb.utils :as utils]
+            [puppetlabs.puppetdb.cli.services :as clisvc]
+            [puppetlabs.puppetdb.pdb-routing :refer :all]
+            [puppetlabs.trapperkeeper.app :as tk-app]))
+
+(defn command-base-url [{:keys [host port]}]
+  (utils/pdb-cmd-base-url host port :v1))
```
wkalt: we'll have to change this to accept a version arg down the road

test/puppetlabs/puppetdb/pdb_routing_test.clj
```diff
@@ -0,0 +1,89 @@
...
+     (let [pdb-resp (client/get (dtu/dashboard-base-url->str (assoc svc-utils/*base-url*
+                                                               :prefix "/pdb")))]
+       (tu/assert-success! pdb-resp)
+       (is (dtu/dashboard-page? pdb-resp))
+
+       (is (-> (query-server-time svc-utils/*base-url*)
+               (get-in [:body :server_time])
+               time/from-string))
+
+
```
wkalt: extra lines

test/puppetlabs/puppetdb/pdb_routing_test.clj
senior: Yeah, that's where our current base-url stuff falls down I think. The reason I need to provide one here is :v4 isn't want I want (and included in base-url by default). So maybe at this point base-url needs to have a :command-version etc then if we need to override that to test old versions, we would just assoc in a new command-version

src/puppetlabs/puppetdb/utils.clj
senior: I was just following the 4 or 5 other functions in this file that do the same thing with a different :prefix. Are you thinking we shouldn't have that on any of them?

wkalt: yeah, I just noticed that. We probably shouldn't, but if it'd be better as a separate change I wouldn't mind keeping it for now.

src/puppetlabs/puppetdb/cli/services.clj
senior: I prefer the slingshot exceptions, but used runtime here because I thought that's what most of our user facing exceptions used. Have we made the switch to slingshot? Happy to update it if that's the case.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -316,10 +314,7 @@
       (when-not disable-update-checking
         (maybe-check-for-updates product-name update-server read-db))
 
-      (let [app (->> (server/build-app globals)
-                     (compojure/context url-prefix []))]
-        (log/info "Starting query server")
-        (add-ring-handler service app))
+      (log/info "Starting query server")
```
ajroetker: We probably don't need separate log statements for each http app now, just one at the top level

test/puppetlabs/puppetdb/pdb_routing_test.clj
wkalt: makes sense. Probably a separate cleanup.

wkalt:
this looks good to me, aside from AJ's comments. Glad we're making these changes.

ajroetker:
Looks good to me as well, moving on to PE PR

src/puppetlabs/puppetdb/cli/services.clj
ajroetker: Is this user facing? I assumed mostly tests and other trapperkeeper services would need to handle this

src/puppetlabs/puppetdb/cli/services.clj
senior: Good catch, I'm removing this one and putting a starting and finished starting message in pdb_routing

src/puppetlabs/puppetdb/cli/services.clj
senior: good point, I'll slingshot it

test/puppetlabs/puppetdb/pdb_routing_test.clj
senior: created a testutils/http and moved it there

senior:
@wkalt @ajroetker push up fixes based on feedback as a separate commit

senior:

PR #1585, opened 2015-08-10
(maint) Remove unnecessary tools.nrepl dependency

ajroetker:
This commit removes an unnecessary tools.nrepl dependency from the
PuppetDB project.clj which was causing conflicts when using ezbake
commands. The ezbake commands in question were in our Jenkins job
configs and although they were improper commands, they would not have
caused an issue if we didn't explicitly list `tools.nrepl` as a
dependency.


ajroetker:
@waynr this look ok?

waynr:
@ajroetker yep, looks good to me! :+1:

PR #1586, opened 2015-08-10
(PDB-1737) assess postgres? based on driver class, not db connection

wkalt:
This changes our postgres? function to us the list of registered drivers to
determine the current database.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage_utils.clj
+++ b/src/puppetlabs/puppetdb/scf/storage_utils.clj
@@ -35,6 +35,39 @@
 
 ;; FUNCTIONS
 
+(defn get-registered-drivers
+  []
+  (enumeration-seq (java.sql.DriverManager/getDrivers)))
+
+(defn driver-in-use
+  [db-config]
+  (java.sql.DriverManager/getDriver
+    (if-let [datasource (:datasource db-config)]
+      (.getJdbcUrl datasource)
+      (format "jdbc:%s:%s"
+              (:subprotocol db-config)
+              (:subname db-config)))))
+
+(defn deregister-unused-drivers!
+  "Given a database config or connection pool, deregister the driver
+   (hsql or postgres) that isn't being used."
+  [db-config]
+  (let [configured-driver (driver-in-use db-config)
+        initial-drivers (get-registered-drivers)]
+    (doseq [driver initial-drivers
+            :when (not= (class driver) (class configured-driver))]
+      (java.sql.DriverManager/deregisterDriver driver))))
+
+(defn register-drivers!
+  "Given a list of database drivers, register those not already registered."
+  [drivers]
+  (let [registered-classes (->> (get-registered-drivers)
+                                (map class)
+                                set)]
+    (doseq [driver drivers
+            :when (not (contains? registered-classes (class driver)))]
+      (java.sql.DriverManager/registerDriver driver))))
+
 (defn sql-current-connection-database-name
   "Return the database product name currently in use."
   []
@@ -54,7 +87,13 @@
 (pls/defn-validated postgres? :- s/Bool
   "Returns true if currently connected to a Postgres DB instance"
   []
-  (= (sql-current-connection-database-name) "PostgreSQL"))
+  (every? #(instance? org.postgresql.Driver %)
+          (get-registered-drivers)))
+
+(pls/defn-validated hypersql? []
+  "Returns true if currently connected to a HSQLDB instance."
+  (every? #(instance? org.hsqldb.jdbc.JDBCDriver %)
+          (get-registered-drivers)))
 
 (pls/defn-validated db-version? :- s/Bool
   "Returns true if the version list you pass matches the version of the current
--- a/test/puppetlabs/puppetdb/http/facts_test.clj
+++ b/test/puppetlabs/puppetdb/http/facts_test.clj
@@ -19,6 +19,7 @@
                                                    paged-results*
                                                    deftestseq
                                                    parse-result]]
+            [puppetlabs.puppetdb.testutils.services :refer [create-db-map]]
             [puppetlabs.puppetdb.utils :as utils]
             [puppetlabs.kitchensink.core :as ks]))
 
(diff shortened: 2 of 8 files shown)


wkalt:

senior:
Transient failure ^

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -16,6 +16,7 @@
             [clojure.set :refer [difference]]))
 
 (def c-t "application/json")
+(def ^:dynamic *dbase* nil)
```
senior: I don't see this used anywhere

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -19,6 +19,7 @@
                                                    paged-results*
                                                    deftestseq
                                                    parse-result]]
+            [puppetlabs.puppetdb.testutils.services :refer [create-db-map]]
```
senior: Should this be removed?

wkalt: no, I had to move that function out of fixtures to where it is now. Previously it was getting sucked in by the :refer :all

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -35,6 +35,34 @@
...
+  "Given a database config or connection pool, deregister the driver
+   (hsql or postgres) that isn't being used, and return the initial driver
+   list for reregistration in case it's required."
+  [db-config]
+  (let [configured-driver (java.sql.DriverManager/getDriver
+                            (if-let [datasource (:datasource db-config)]
+                              (.getJdbcUrl datasource)
+                              (format "jdbc:%s:%s"
+                                      (:subprotocol db-config)
+                                      (:subname db-config))))
```
ajroetker: I'd be tempted to abstract this to a function on `db-config`, `(defn get-driver [db-config])` or something like that.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -35,6 +35,34 @@
 
 ;; FUNCTIONS
 
+(defn deregister-unused-drivers
```
ajroetker: Should this have a `!` at the end to denote side-effects?

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -35,6 +35,34 @@
...
+    (doseq [driver initial-drivers
+            :when (not= (class driver) (class configured-driver))]
+      (java.sql.DriverManager/deregisterDriver driver))
+    initial-drivers))
+
+(defn register-drivers
+  "Given a list of database drivers, register those not already registered."
+  [drivers]
+  (let [registered-classes (->> (java.sql.DriverManager/getDrivers)
+                                enumeration-seq
```
ajroetker: Looks like `(defn get-drivers [] (enumeration-seq (.../getDrivers)))` would probably be a useful function to be abstracted, here and up above in deregister.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -35,6 +35,34 @@
...
+                              (format "jdbc:%s:%s"
+                                      (:subprotocol db-config)
+                                      (:subname db-config))))
+        initial-drivers (enumeration-seq (java.sql.DriverManager/getDrivers))]
+    (doseq [driver initial-drivers
+            :when (not= (class driver) (class configured-driver))]
+      (java.sql.DriverManager/deregisterDriver driver))
+    initial-drivers))
+
+(defn register-drivers
```
ajroetker: Similarly a `!` here might be useful to denote side-effects. Also should this return the list of all registered drivers?

src/puppetlabs/puppetdb/scf/storage_utils.clj
ajroetker: Also looks like it could be useful below as well.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -35,6 +35,34 @@
...
+                            (if-let [datasource (:datasource db-config)]
+                              (.getJdbcUrl datasource)
+                              (format "jdbc:%s:%s"
+                                      (:subprotocol db-config)
+                                      (:subname db-config))))
+        initial-drivers (enumeration-seq (java.sql.DriverManager/getDrivers))]
+    (doseq [driver initial-drivers
+            :when (not= (class driver) (class configured-driver))]
+      (java.sql.DriverManager/deregisterDriver driver))
+    initial-drivers))
```
ajroetker: I understand why your return `initial-drivers` here after reading through the test fixtures, but it's not really intuitive in terms of what this function does. I would expect this function to return all the drivers that are still registered after deregistering. It seems like anything that needs to keep around `initial-drivers` could do so through a `get-drivers` function.

ajroetker: In regards to the above comment, you could also make `initial-drivers` an argument to your function here.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -35,6 +35,34 @@
...
+
+(defn register-drivers
+  "Given a list of database drivers, register those not already registered."
+  [drivers]
+  (let [registered-classes (->> (java.sql.DriverManager/getDrivers)
+                                enumeration-seq
+                                (map class)
+                                set)]
```
ajroetker: Seems like you could do something similar in the deregister function (using sets that is) which would allow you to use `set/diff` to get the list of things to deregister.

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -163,7 +172,8 @@
                    :host "localhost"
                    :port port
                    :prefix "/pdb/query"
-                   :version :v4}]
+                   :version :v4}
+         initial-drivers (enumeration-seq (java.sql.DriverManager/getDrivers))]
```
ajroetker: Looks like the `get-drivers` abstraction could go here as well.

ajroetker:
Looks like this needs a rebase @wkalt.

wkalt:
@ajroetker thanks, addressed your comments

mullr:
Left comments on the commit. Basically seems fine, but I really don't understand why this is something we want to do. It seems far better to inspect the actually connection to see what you have in hand, rather than relying on the global state inside the jdbc driver registry.

PR #1588, opened 2015-08-12
(PDB-1737) use memoized function get-db-metadata for db metadata functions

wkalt:
Previously we required a database connection to determine the database and
version we were connected to. This patch stashes that data in a dynamic var
on startup.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage_utils.clj
+++ b/src/puppetlabs/puppetdb/scf/storage_utils.clj
@@ -34,45 +34,40 @@
   [s/Int])
 
 ;; FUNCTIONS
-
-(defn sql-current-connection-database-name
-  "Return the database product name currently in use."
-  []
-  (.. (sql/find-connection)
-      (getMetaData)
-      (getDatabaseProductName)))
-
-(pls/defn-validated sql-current-connection-database-version :- db-version
-  "Return the version of the database product currently in use."
+(defn db-metadata-fn
   []
   (let [db-metadata (.. (sql/find-connection)
                         (getMetaData))
         major (.getDatabaseMajorVersion db-metadata)
         minor (.getDatabaseMinorVersion db-metadata)]
-    [major minor]))
+    {:database (.getDatabaseProductName db-metadata)
+     :version [major minor]}))
+
+(def db-metadata
+  (delay (db-metadata-fn)))
 
 (pls/defn-validated postgres? :- s/Bool
   "Returns true if currently connected to a Postgres DB instance"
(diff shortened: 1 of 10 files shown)


src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -51,10 +53,19 @@
         minor (.getDatabaseMinorVersion db-metadata)]
     [major minor]))
 
+(defn get-db-metadata
+  []
+  (let [db-metadata (.. (sql/find-connection)
```
ajroetker: I think you want to us the undeprecated `sql/` namespace here, passing in `db-config` and calling `get-connection` on the `db-config`.

wkalt: based on messing with this for a while today, I think it'll be cleaner to change this function to accept an actual connection object that we'll have once we shift from the undeprecated with-db-connection (which uses a binding) instead of with-connection like we currently do.  We'll need to make a similar change in a lot of places, and I haven't found any mixed approaches that I like better than what's currently here.

wkalt:
this now uses a memoized function, rather than a dynamic var

senior:
:+1: from me, did you want to take a look @mullr ?

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -34,45 +34,40 @@
...
   []
   (let [db-metadata (.. (sql/find-connection)
                         (getMetaData))
         major (.getDatabaseMajorVersion db-metadata)
         minor (.getDatabaseMinorVersion db-metadata)]
-    [major minor]))
+    {:database (.getDatabaseProductName db-metadata)
+     :version [major minor]}))
+
+(def get-db-metadata
```
mullr: I hesitate to mention it, but the more direct way of modeling this may be:

```clojure
(def db-metadata (delay (db-metadata-fn)))
```

Meaning usages turn into `@db-metadata`. This has the nice property of being able to see, at the call site, that db-metadata is kind of a side channel. It should be equally easy to `with-redefs` as well.

mullr:
@senior @wkalt This generally looks fine, the only question is regarding the use of `delay` instead of `memoize`. I feel bad about yanking you around here, but it may be worth considering.

wkalt:
yes, I like that better

mullr:
Who's got two thumbs and approves this change?

...

:+1: THIS GUY! :+1:

PR #1590, opened 2015-08-12
(PDB-1855) Add flag for filtering unchanged resources in the terminus

ajroetker:
Previously we had now way of sending unchanged resources in the
terminus. This commit adds a boolean flag to the terminus config file,
`include_unchanged_resources=<boolean>`, which will _not_ filter
unchanged resources from a report when set to `true`.

As a side note PuppetDB cannot actually handle storing these unchanged resources if you turn them on, that will be handled by a subsequent storage related change.


senior:
:+1:

kbarber:
+1

PR #1591, opened 2015-08-12
(maint) Relay errors for blocking commands

mullr:
When submitting a command in blocking mode, relay any errors that occur
before the timeout in the http response.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -345,6 +345,23 @@
     "Returns a core.async publisher to which commands are written after they
      have been processed. The topic for each message is its uuid." ))
 
+(defn make-cmd-processed-message [cmd ex]
+  (merge {:command cmd
+          :id (-> cmd :annotations :id)}
+         (when ex {:exception ex})))
+
+(defn process-command-and-respond! [cmd config response-pub-chan stats-atom]
+  (try
+    (let [result (process-command! cmd config)]
+      (swap! stats-atom update :executed-commands inc)
+      (async/>!! response-pub-chan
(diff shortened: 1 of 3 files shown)


src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -67,11 +67,24 @@
...
+                           (http/json-response {:uuid uuid
+                                                :processed true
+                                                :timed_out false
+                                                :error (str exception)
+                                                :stack_trace (map str (.getStackTrace exception))}
+                                               503)
+                           (http/json-response {:uuid uuid
+                                                :processed true
+                                                :timed_out false}
+                                               200))))))))
```
ajroetker: Maybe the base response could get named and put in a let, and for the error case the :error and :stack_trace keys could be assoc'd in or merged in

src/puppetlabs/puppetdb/command.clj
```diff
@@ -345,6 +345,25 @@
     "Returns a core.async publisher to which commands are written after they
      have been processed. The topic for each message is its uuid." ))
 
+(defn make-cmd-processed-message [cmd ex]
+  (merge {:command cmd
+          :id (-> cmd :annotations :id)}
+         (if ex
+           {:exception ex}
+           {})))
```
ajroetker: Usually we do this with just `(merge ... (when ex {:exception ex}))` since merge handles nil, same with `conj`.

mullr:
@ajroetker updated

mullr:
⬆️transient

mullr:

mullr:
@ajroetker addressed your comments

PR #1593, opened 2015-08-13
(maint) Make the PuppetDB dashboard javascript more functional

ajroetker:
This commit refactors the dashboard to pass around arguments rather than
relying on global mutable state and simplifies some of logic for
rendering.

This commit should make it easier to make the metrics configurable via Clojure code (good for PE) or moving to clojurescript if wanted to dabble in that sort of thing.

### DIFF
--- a/resources/public/dashboard/charts.js
+++ b/resources/public/dashboard/charts.js
@@ -2,274 +2,180 @@
   Displays a set of counters and a sparkline for a metric (JSON that's
   fetched via AJAX).
 */
-function counterAndSparkline() {
+// Top-line label to use
+
+// Sub-heading describing the metric
+
+// What URL to poll for JSON
+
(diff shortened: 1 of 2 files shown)


resources/public/dashboard/charts.js
```diff
@@ -2,274 +2,180 @@
...
-
-    chart.width = function(_) {
-        if (!arguments.length) return width;
-        width = _;
-        return chart;
+    // Redraw
+    function redraw(now, data) {
+        var datavals = data.map(function(d) { return d.value; });
+
+        // Update our axes (axises?)
```
mullr: axes.

ajroetker: I think that was already there :)

senior:
:+1: tested locally, looks good

PR #1595, opened 2015-08-14
(maint) add jessie to build defaults

wkalt:
This adds jessie to our build defaults.


wkalt:
@melissa any subtleties around this?

PR #1598, opened 2015-08-14
(PDB-1797) testing support for jessie

wkalt:
This sets up the config for acceptance testing on debian 8.


wkalt:

wkalt:
shelving this for now. Everything is in place, but we're blocked on support in Puppet Server.

PR #1600, opened 2015-08-14
(PDB-1853) Query-eng refactor to allow external access

wkalt:


src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -2,59 +2,124 @@
...
-          :edges [edges/query->sql edges/munge-result-rows]
-          :reports [reports/query->sql reports/munge-result-rows]
-          :report-metrics [report-data/metrics-query->sql (report-data/munge-result-rows :metrics)]
-          :report-logs [report-data/logs-query->sql (report-data/munge-result-rows :logs)]
-          :resources [resources/query->sql resources/munge-result-rows])]
-    [#(query->sql version % paging-options) (munge-fn version url-prefix)]))
+  (let [munge-fn (get-munge entity)]
+    (case entity
+      :event-counts
+      ((munge-fn (:summarize_by paging-options)) version url-prefix)
```
ajroetker: since we're special casing it here anyways, maybe the `:event-counts` munge function should be changed to take three arguments here instead of return a function of two args

wkalt: sounds good to me

wkalt:
I think there's more refactoring that can be done due to the removed content from the query namespaces (in theory I think we should be able to delete most of our query tests), but this is already starting to bleed into PDB-1714 so I figured time to cut it.

wkalt:
@pljenkinsro retest that please

wkalt:

src/puppetlabs/puppetdb/query/aggregate_event_counts.clj
```diff
@@ -31,20 +31,22 @@
...
-  [version query [summarize_by {:keys [distinct_resources?] :as query-options}]]
+  [version query {:keys [distinct_resources? summarize_by] :as query-options}]
   {:pre  [((some-fn nil? sequential?) query)
           ((some-fn map? nil?) query-options)]
    :post [(jdbc/valid-jdbc-query? (:results-query %))]}
   (let [summary-vec (str/split summarize_by #",")
         nsummarized (count summary-vec)
-        aggregate-fn #(assemble-aggregate-sql version query % query-options)
+        aggregate-fn #(assemble-aggregate-sql version query (assoc query-options
+                                                                   :summarize_by %))
```
ajroetker: Maybe `#(->> % (assoc query-options :summarize_by) (assemble-aggregate-sql version query))` to make this a little more readable

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -66,23 +131,26 @@
...
 
-  If the query can't be parsed, a 400 is returned."
+   If the query can't be parsed, a 400 is returned."
   {:deprecated "3.0.0"}
   [entity version query paging-options db url-prefix]
   (try
     (jdbc/with-transacted-connection db
-      (let [[query->sql munge-fn] (entity->sql-fns entity version paging-options url-prefix)
-            {:keys [results-query count-query]} (-> query json/coerce-from-json query->sql)
+      (let [query-rec (get-rec entity)
```
ajroetker: Looks like this isn't used anymore

ajroetker:
Might be useful to include a test which replaces one of the query-recs to make sure this gives us the ability to swap queries

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -2,59 +2,124 @@
...
+
+(defn orderable-columns
+  [query-rec]
+  (if-not query-rec
+    []
+    (let [projections (:projections (query-rec))]
+      (->> (keys projections)
+           (remove #{"value"})
+           (filter #(:queryable? (get projections %)))
+           (map keyword)))))
```
ajroetker:
```clojure
(for [[projection-key projection-value] (:projections (query-rec))
  :when (and (not= projection-key "value") (:queryable? projection-value))]
  (keyword projection-key))
```

wkalt:
@AJ added a test and a function for updating the atom

wkalt:

src/puppetlabs/puppetdb/http/fact_names.clj
```diff
@@ -1,27 +1,40 @@
...
+   (app
+     [&]
+     {:get (fn [{:keys [params globals paging-options]}]
+             (let [paging-options (assoc-when paging-options
+                                              :order_by [[:name :ascending]])]
+               (produce-streaming-body
+                 :fact-names
+                 version
+                 (params "query")
+                 paging-options
```
ajroetker: We're calling this `query-options` elsewhere, should we consolidate on one of the two?

test/puppetlabs/puppetdb/query/aggregate_event_counts_test.clj
```diff
@@ -14,7 +14,8 @@
   ([version query summarize_by]
      (aggregate-counts-query-result version query summarize_by {}))
   ([version query summarize_by extra-query-params]
-     (-> (aggregate-event-counts/query->sql version query [summarize_by extra-query-params])
+   (-> (aggregate-event-counts/query->sql version query (assoc extra-query-params
+                                                               :summarize_by summarize_by))
          (aggregate-event-counts/query-aggregate-event-counts))))
```
ajroetker: Alignment

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -2,59 +2,129 @@
...
+(defn assoc-in-idx!
+  "assoc into the entity index as if query recs were maps"
+  [[entity & ks] v]
+  (if (and (= :rec (first ks)) (next ks))
+    ;; if it's a query rec and we're updating a nested component, call the
+    ;; existing rec and wrap val in a function
+    (let [rec' (-> ((get-in @entity-fn-idx [entity :rec]))
+                   (assoc-in (next ks) v))]
+      (swap! entity-fn-idx #(assoc-in % [entity :rec] (fn [] rec'))))
+    (swap! entity-fn-idx #(assoc-in % (cons entity ks) v))))
```
ajroetker: I think the `#(` is redundant you can just use `(swap! entity-fn-idx assoc-in (cons entity ks) v)))`

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -2,59 +2,129 @@
...
+     :resources {:munge resources/munge-result-rows
+                 :rec eng/resources-query}}))
+
+(defn assoc-in-idx!
+  "assoc into the entity index as if query recs were maps"
+  [[entity & ks] v]
+  (if (and (= :rec (first ks)) (next ks))
+    ;; if it's a query rec and we're updating a nested component, call the
+    ;; existing rec and wrap val in a function
+    (let [rec' (-> ((get-in @entity-fn-idx [entity :rec]))
```
ajroetker: What would you think about making the `query-rec`s optionally parametrized with a map that we merge in?

wkalt: conflicted. I feel like it entrenches the nastiness we currently have with function-valued query recs.

wkalt: that's not to say I love the approach above either, it just seems like the stuff I don't like is a bit more consolidated. I'll give it some thought.

src/puppetlabs/puppetdb/http/fact_names.clj
wkalt: I made a ticket for this, PDB-1936. I like the idea of standardizing around query-options, but I think that's best handled in a different patch. As it stands the distinction isn't quite arbitrary -- we're using "query-options" only on the events endpoints, where the map includes non-paging stuff.

ajroetker:
I'm :+1: on this

PR #1601, opened 2015-08-19
(PDB-1921) Swallow exceptions during the `check-for-update!` on startup

ajroetker:
This commit fixes a regression with our check-for-update! function in
which the new shared library we upgraded to does not swallow exceptions
like our previous code. Since the library will log exceptions and
rethrow them, we just need to be sure to swallow the exception in our
wrappers for the library. This commit swallows exceptions for the
version-check library on startup.


wkalt:
:+1:

wkalt:
confirmed the same exceptions are logged, but no longer cause crashes

PR #1603, opened 2015-08-20
(MAINT) Refactor puppetdb-middleware to central location

senior:
Prior to this change, each route (i.e. meta, admin, query etc) included
an explicit wrap of puppetdb-middleware. This pulls that wrapping up to
the top level (where the /pdb route is being defined). This reduces
duplication and centralizes that code to pdb-routing.


ajroetker:
Nice!

ajroetker:
Transient ^

PR #1605, opened 2015-08-21
(PDB-1732) Remove redundant import/export acceptance tests

ajroetker:
This commit remove the import/export acceptance tests as we already have
roundtrip tests in Clojure as well as some acceptance tests to make sure
the bash scripts for the commands themselves work. The facts-only
acceptance tests does not appear to be a use case we care about.


senior:
I think we should add that facts_only test as a clojure test. I think it would be pretty easy and would run quickly. Any reason not to add it?

PR #1606, opened 2015-08-21
(maint) remove the event_query_with_read_db acceptance test

wkalt:
This is duplicated in the http facts clojure tests here: https://github.com/puppetlabs/puppetdb/blob/master/test/puppetlabs/puppetdb/http/facts_test.clj#L579-579


wkalt:
reopening on stable

PR #1607, opened 2015-08-21
(PDB-1934) remove two-database events query acceptance test

wkalt:
This is duplicated in the http facts test.


wkalt:

senior:
I agree that the facts tests [here](https://github.com/puppetlabs/puppetdb/blob/master/test/puppetlabs/puppetdb/http/facts_test.clj#L582) duplicate this test, but one thing the acceptance test covers that the facts one does not is configuration of the two databases. The http/facts test side-steps the configuration side of it.

wkalt:
reopened on stable

PR #1609, opened 2015-08-21
(maint) Move node_ttl acceptance test into a clojure full stack test

senior:
By using a puppetdb-instance call in clojure vs an acceptance test, we
can cut the execution time of the test down from 6 minutes to a few
seconds. The code coverage is still the same, there's just no need to
start/stop the server so many times like the code was doing previously.


wkalt:
+1

senior:
Retargeted at stable https://github.com/puppetlabs/puppetdb/pull/1616

PR #1620, opened 2015-08-25
(maint) install debian-archive-keyring in repo setup

wkalt:
This should correct test failures on debian 7 due to missing keys.


wkalt:
the wheezy pc1 repos have been a bit unstable today. there's a chance this will resolve itself without the fix.

wkalt:

PR #1622, opened 2015-08-26
(PDB-1854) Strip resources blob when converting reports-query->wire-v5

ajroetker:
This commit removes the resources blob when formatting a reports-query
report format to a wire v5 format. This is used only in PE.


ajroetker:
transient failure

PR #1624, opened 2015-08-26
(PDB-1941) Grab db credential variables directly from testutils

ajroetker:
This commit changes the unit tests in PuppetDB to grab their database
credentials from the environment directly rather than through a memoized
function loaded from a Clojure file on the classpath under a `config`
directory (i.e. we're stopping ourselves from doing some silly things).

Relates to https://github.com/puppetlabs/pe-puppetdb-extensions/pull/94.


ajroetker:
transients ^

ajroetker:

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -590,8 +591,8 @@
 (deftestseq ^{:postgres false} two-database-fact-query-config
   [[version endpoint] facts-endpoints]
 
-  (let [read-db-config (create-db-map)
-        write-db-config (create-db-map)
+  (let [read-db-config (create-hsqldb-map)
+        write-db-config (create-hsqldb-map)
```
mullr: What happens to this test when we run with postgres? does it get skipped, or does it run with hsqldb?

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -23,45 +23,36 @@
...
+  {:classname "org.hsqldb.jdbcDriver"
+   :subprotocol "hsqldb"
+   :subname (str "mem:"
+                 (java.util.UUID/randomUUID)
+                 ";hsqldb.tx=mvcc;sql.syntax_pgs=true")})
+
+(defn get-env [var & [default]]
+  (or (System/getenv var) default))
+
+(def create-postgres-map
```
mullr: this is a `def`, but the name indicates it's a function. `create-hsqldb-map` is a function, presumably because it needs to generate a uuid. I think this one should be a function as well for consistency.

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -23,45 +23,36 @@
...
 ;; keep going back to disk for it.
-(def test-config
-  (memoize load-test-config))
-
-(defn test-db
-  "Return a map of connection attrs for the test database"
-  []
-  ((:testdb-config-fn (test-config))))
+(def test-db
+  (memoize test-db*))
```
mullr: I don't think this is what we want for the hsqldb case; we'll get back the same db every time, rather than a fresh `mem://` db.

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -110,11 +110,10 @@
 (defmacro with-puppetdb-instance
   "Convenience macro to launch a puppetdb instance"
   [& body]
-  `(with-redefs [sutils/db-metadata (delay {:database "HSQL Database Engine"
-                                            :version [2 2]})]
+  `(with-redefs [sutils/db-metadata (delay (sutils/db-metadata-fn))]
```
mullr: wtf? I don't get it.

mullr:
@ajroetker some comments, a perhaps real problem, some name pickiness, and some confusion over legacy code. The usual, really.

test/puppetlabs/puppetdb/http/facts_test.clj
ajroetker: It get's skipped, according to the test-selector `^{:postgres false}`, see our project.clj

mullr: You know, I looked right at that and still it didn't enter my brain.

test/puppetlabs/puppetdb/testutils.clj
ajroetker: yeah that's an artifact of something else i was doing, I think I'd rather not have the hsqldb way of doing things as the source of truth though, and just leave this as a map.

test/puppetlabs/puppetdb/testutils.clj
ajroetker: But to your point of consistency, then were testing postgres the same way were testing hsqldb. By reusing one db. We can always create another if that's what we need. And yeah I'll rename it, that was just the legacy name.

test/puppetlabs/puppetdb/testutils.clj
mullr: Can we call it `postgres-map` then?

test/puppetlabs/puppetdb/testutils/services.clj
ajroetker: I just mimicked what we do elsewhere for fetching the db-metadata, since were not always gonna have an HSQLDB connection in this function anymore we have to get the metadata dynamically.

test/puppetlabs/puppetdb/testutils.clj
ajroetker: Yeah that sounds fine to me.

test/puppetlabs/puppetdb/testutils/services.clj
mullr: Sorry, I'm confused by the `delay`. It looks like it was being used as a crappy thunk. Is there any reason to keep it?

test/puppetlabs/puppetdb/http/facts_test.clj
ajroetker: @mullr https://www.youtube.com/watch?v=vJG698U2Mvo

test/puppetlabs/puppetdb/testutils/services.clj
ajroetker: @wkalt ^ do you know more about the `delay`?

test/puppetlabs/puppetdb/testutils.clj
mullr: Well, if we don't care to have a new hsqldb each time then should be able to avoid the memoize here and just have something like `(def test-db (test-db*))` (or a consolidated equivalent)

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -23,45 +23,36 @@
...
-    :testdb-config-fn : a no-arg function that returns a hash of database
-        settings, suitable for passing to the various `clojure.java.jdbc`
-        functions."
+  {:classname "org.hsqldb.jdbcDriver"
+   :subprotocol "hsqldb"
+   :subname (str "mem:"
+                 (java.util.UUID/randomUUID)
+                 ";hsqldb.tx=mvcc;sql.syntax_pgs=true")})
+
+(defn get-env [var & [default]]
```
senior: Environ is really nice for this kind of thing

senior: Sorry, should probably give you a link: https://github.com/weavejester/environ

test/puppetlabs/puppetdb/testutils.clj
senior: Agreed, creating a new DB is easy, easier for sure than cleaning up after yourself. However HSQL won't be around for long, so it sounds like having one is the right thing to do. I suspect we'll be cleaning up after this for a while, i.e. what happens if a failed test happens, leaves some tables around etc.

senior:
:+1: from me after the memoize issue gets addressed

test/puppetlabs/puppetdb/testutils.clj
ajroetker: Yeah I was thinking about using that but didn't see a ton of value considering we have to use `System/getenv` in our project.clj anyways, I can add it in here if we'd like though.

ajroetker:
@mullr think I addressed your memoization issues if you wanna take another look

ajroetker:

senior:
@ajroetker hypersql tests on travis are failing, out of memory, the PR tests are the git issues we've been having

ajroetker:

test/puppetlabs/puppetdb/http/facts_test.clj
wkalt: this was hsqldb only before because our testing didn't support using multiple pg databases. If that's supported now, we should flip it because we actually don't allow separate read/write dbs on hsql.

test/puppetlabs/puppetdb/testutils/services.clj
wkalt: addressed in the channel -- it's a delay because it redefs a delay and is derefed as a delay. It could be an atom, but we'd be going out of our way.

wkalt:
looks good to me

PR #1625, opened 2015-08-27
(PDB-154) allow zero gc interval

mullr:


ajroetker:
lgtm

documentation/configure.markdown
```diff
@@ -274,6 +274,8 @@ Do not use the `username` or `password` settings.
 
 This controls how often, in minutes, to compact the database. The compaction process reclaims space and deletes unnecessary rows. If not supplied, the default is every 60 minutes.
 
+If `gc-interval` is set to 0, all GC processes will be disabled. This may be useful, for example, for a read-only PuppetDB instance using a replicated database. It should not be used for a PuppetDB instance that will receive any commands.
```
wkalt: I don't think the part about receiving commands is accurate. The situation I'm familiar with is a single write database A shared by N PDB's dedicated to command processing, which each write to A and read off replicas of A. Under this scenario you may have a single PDB instance that does nothing but run GC on A, and receives no commands, but it's every instance _except_ that one on which you'd be setting gc-interval=0.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -328,10 +328,11 @@
...
                                      (garbage-collect! write-db)))]
-        ;; Run database maintenance tasks seqentially to avoid
-        ;; competition. Each task must handle its own errors.
-        (gc-task db-maintenance-tasks)
-        (gc-task #(compress-dlo! dlo-compression-threshold discard-dir)))
+        (when (pos? gc-interval-millis)
+          ;; Run database maintenance tasks seqentially to avoid
+          ;; competition. Each task must handle its own errors.
+          (gc-task db-maintenance-tasks)
+          (gc-task #(compress-dlo! dlo-compression-threshold discard-dir))))
```
wkalt: this raises another question -- under the scenario I described above the DLO dirs that matter will be on the command processing slaves, despite the fact that the database we care about GCing is on a different host. Existing issue, but it seems like we need to decouple the DLO compression from the database cleanup.

mullr:
@wkalt ok, added that feature

documentation/configure.markdown
senior: Seems like a more accurate way to put this is, if running more than one PuppetDB instance sharing a single database, only 1 of those instances should have GC enabled. Read vs. write vs. commands doesn't matter.

mullr: I think receiving commands *is* what matters, since setting `gc-interval` to zero will also disable mq compaction.

ajroetker:

wkalt:
I think yesterday we were talking about either making the mq gc non-tunable or making it default to something else in the case of gc-interval=0, to avoid two flags. Is that accurate?

waynr:
test this please

ajroetker:
So, reading @wkalt 's comment, it would seem that this PR is still in progress. Is that the case @mullr?

mullr:
@ajroetker Aye. Working on it now.

mullr:
Ok, I took out `disable-database-gc` and added `dlo-compression-interval`. By default this is equal to `gc-interval`, so it works the same for existing users. Changed the docs to tell people to set it if they want to zero-out `gc-interval`.

wkalt:
+1 from me. test failure looks transient.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -223,6 +225,11 @@
   [config]
   (-> config
       (configure-section :database write-database-config-in write-database-config-out)
+      (update :database (fn [database-section]
+                          ;; Default dlo-compression-interval to gc-interval
+                          (if (:dlo-compression-interval database-section)
+                            database-section
+                            (assoc database-section :dlo-compression-interval (:gc-interval database-section)))))
```
ajroetker: I think this is `assoc-when` or `update-when` from utils

mullr: it's assoc-when-not, but I don't think that's a thing.

wkalt: that's just assoc-when actually https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/utils.clj#L83-94

it's an implicit not I guess

mullr: someone wrote assoc-when-not and called it assoc-when? *sigh*

mullr:

mullr:
test this please

mullr:
@senior ⬆️

PR #1626, opened 2015-08-27
(QENG-2729) Prepare repo for new CI jobs.

waynr:
This commit moves some of the Jenkins shell scripts into this repo in order to
reduce overall churn when converting PuppetDB acceptance testing over to
vmpooler. So the new CI jobs should be usable by both ec2 and vmpooler testing.

Along with these changes some of the variables read in by the Gemfile,
acceptance helper, and jenkins beaker script have been modified to make them
more compatible with variables other CI pipelines might use--bringing us closer
to a world in which our CI definitions can be generically re-usable between
projects.


waynr:
The merge of this PR needs to be coordinated with the merge of puppetlabs/ci-job-configs#469.

ext/jenkins/beaker-tests.sh
```diff
@@ -24,19 +24,12 @@ export BEAKER_PRESERVE_HOSTS=onfail
...
-mkdir vendor
-
-bundle install --path=vendor/bundle --without=test --retry=10
+export BEAKER_CONFIG=$LAYOUT
 
 # Now run our tests
 bundle exec rake beaker:acceptance
+exit_code=$?
+
+$WORKSPACE/ext/jenkins/cleanup-s3.sh
```
kbarber: Does this introduce a potential race? If all the cells are executing this, what if some tests haven't finished yet, and the repos get cleaned early?

waynr: Ah, good point. I was thinking about other reasons to keep this a separate job anyway yesterday (we probably want this step to run regardless of whether some command earlier in the `beaker-tests.sh` script fails). 

I will update puppetlabs/ci-job-configs#469 with a new job to run this command and remove this line.

kbarber: Concur on all points.

This is a "must always run" kind of thing, fwiw this isn't the case today, as the upstream job could fail right now, and the cleanup never occurs. Another idea is to have a real out-of-band sweep to ensure that eventually everything is cleaned up. This concept has been on our radar for years, but we never get around to it.

waynr: An out-of-band sweep sounds good, but I'm not sure it's worthwhile for QENG-2729, especially since we will be working to move away from ec2 for puppetdb testing. I think the problem with an out-of-band sweep that makes it difficult to implement immediately is the potential for race conditions since out-of-band presumably means we will be removing all artifacts rather than specifically-versioned artifacts as seen here: https://github.com/waynr/puppetdb/blob/qeng-2729/ext/jenkins/packaging-cleanup.sh#L14

waynr: Anyway, this has been cleaned up and I am testing the pipeline with the new job now.

Gemfile
```diff
@@ -1,5 +1,5 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
-puppet_branch = ENV['puppet_branch'] || "latest"
+puppet_branch = ENV['PUPPET_VERSION'] || "latest"
```
wkalt: when this goes in we'll need to update our spec tests in jenkins to use this. Easy change though and the name is better.

waynr: puppetlabs/ci-job-configs#469 adds new pipelines that includes spec tests which use this new variable name.

ext/jenkins/lein-test.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+DBNAME=$(echo "puppetdb_${PUPPETDB_BRANCH}_${BUILD_ID}_${BUILD_NUMBER}_${PUPPETDB_DBTYPE}_${JDK}" | tr - _)
+DBUSER="puppetdb"
+DBHOST="fixture-pg94.delivery.puppetlabs.net"
```
mullr: what happens if two tests run at once?

ext/jenkins/packaging.sh
```diff
@@ -0,0 +1,67 @@
...
+pushd "target/staging"
+PACKAGE_BUILD_VERSION=$(rake pl:print_build_param[ref] | tail -n 1)
+
+cat > "${WORKSPACE}/puppetdb.packaging.props" <<PROPS
+PACKAGE_BUILD_VERSION=${PACKAGE_BUILD_VERSION}
+PROPS
+popd
+
+echo "**********************************************"
+echo "NOW DO S3 MAGIC"
```
mullr: ✨

ext/jenkins/lein-test.sh
waynr: They will have different `BUILD_ID`/`BUILD_NUMBER` values and thus have different `DBNAME`

PR #1627, opened 2015-08-28
(PDB-1902) Create a defaulted config service

senior:
Currently configuration is retrieved from the TK config service,
defaulted and consumed by other componentes via shared-globals. This
commit creates a new DefaultedConfig service that has a similar API as
the config service, but applies our validation and defaulting logic
first. This will allow a large amount of shared-globals to be removed
and instead just use the defaulted values from this new service.

This also starts the process of removing things from shared-globals (specifically removing the authorizer). MQ related config probably makes the most sense to work on next (this will be handled in a separate ticket).

### DIFF
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -16,7 +16,9 @@
             [schema.core :as s]
             [slingshot.slingshot :refer [throw+]]
             [puppetlabs.puppetdb.schema :as pls]
-            [puppetlabs.puppetdb.utils :as utils]))
+            [puppetlabs.puppetdb.utils :as utils]
+            [puppetlabs.trapperkeeper.core :refer [defservice] :as tk]
+            [puppetlabs.trapperkeeper.services :refer [service-id service-context]]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Schemas
@@ -375,3 +377,24 @@
       validate-vardir
       convert-config
       configure-catalog-debugging))
+
+
+(defprotocol DefaultedConfig
+  (get-config [this])
+  (get-in-config [this ks] [this ks default]))
+
+(defn create-defaulted-config-service [config-transform-fn]
+  (tk/service
+   DefaultedConfig
+   [[:ConfigService get-config]]
+   (init [this context]
+         (assoc context :config (config-transform-fn (get-config))))
+   (get-config [this]
+               (:config (service-context this)))
+   (get-in-config [this ks]
+                  (get-in (service-context this) ks))
+   (get-in-config [this ks default]
+                  (get-in (service-context this) ks default))))
+
+(def config-service
+  (create-defaulted-config-service process-config!))
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -17,7 +17,7 @@
             [slingshot.slingshot :refer [throw+]]
             [puppetlabs.puppetdb.schema :as pls]
             [puppetlabs.puppetdb.utils :as utils]
-            [puppetlabs.trapperkeeper.core :refer [defservice] :as tk]
+            [puppetlabs.trapperkeeper.core :as tk]
             [puppetlabs.trapperkeeper.services :refer [service-id service-context]]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -378,10 +378,9 @@
       convert-config
       configure-catalog-debugging))
 
-
 (defprotocol DefaultedConfig
   (get-config [this])
-  (get-in-config [this ks] [this ks default]))
+  (get-in-config [this ks]))
 
 (defn create-defaulted-config-service [config-transform-fn]
   (tk/service
@@ -392,9 +391,7 @@
    (get-config [this]
                (:config (service-context this)))
    (get-in-config [this ks]
-                  (get-in (service-context this) ks))
-   (get-in-config [this ks default]
-                  (get-in (service-context this) ks default))))
+                  (get-in (service-context this) ks))))
 
 (def config-service
   (create-defaulted-config-service process-config!))
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -378,6 +378,15 @@
       convert-config
       configure-catalog-debugging))
 
+(defn foss? [config]
+  (= "puppetdb" (get-in config [:global :product-name])))
+
+(defn pe? [config]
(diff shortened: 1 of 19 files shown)


src/puppetlabs/puppetdb/config.clj
```diff
@@ -375,3 +377,23 @@
       validate-vardir
       convert-config
       configure-catalog-debugging))
+
+(defprotocol DefaultedConfig
```
senior: This along with create-defaulted-config function below would be a good candidate for TK or a shared library, see it's usage in the config-service def below. /cc @KevinCorcoran @cprice404

ajroetker:
Looks like you're gonna need to update the bootstrap.cfg in resources for acceptance tests

src/puppetlabs/puppetdb/config.clj
```diff
@@ -16,7 +16,9 @@
             [schema.core :as s]
             [slingshot.slingshot :refer [throw+]]
             [puppetlabs.puppetdb.schema :as pls]
-            [puppetlabs.puppetdb.utils :as utils]))
+            [puppetlabs.puppetdb.utils :as utils]
+            [puppetlabs.trapperkeeper.core :refer [defservice] :as tk]
```
mullr: doesn't look like `defservice` is used.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -375,3 +377,23 @@
       validate-vardir
       convert-config
       configure-catalog-debugging))
+
+(defprotocol DefaultedConfig
+  (get-config [this])
+  (get-in-config [this ks] [this ks default]))
```
mullr: Is the second overload used anywhere? It seems contrary to the idea of a defaulted config service.

test/puppetlabs/puppetdb/middleware_test.clj
```diff
@@ -45,20 +48,26 @@
...
 
+(defn create-authorizing-request [hostname]
+  {:scheme :https
+   :ssl-client-cn hostname})
+
 (deftest wrapping-authorization
   (testing "Should only allow authorized requests"
     ;; Setup an app that only lets through odd numbers
-    (let [handler     (fn [req] (-> (rr/response nil)
+    (let [wl (.getAbsolutePath (temp-file "whitelist-log-reject"))
```
mullr: ick. We should probably plumb this through the regular logger now that we know how to set up separate logs. (I know this is pre-existing)

mullr: wait... I completely misinterpreted this. n/m

mullr:
@senior Mostly minor comments

src/puppetlabs/puppetdb/config.clj
senior: No I was just mimicking the ConfigurationService protocol, I'll drop it

senior:
@mullr fixed up based on your feedback
@ajroetker added the line to bootstrap.cfg

ajroetker:
@senior Is this ready for review? The ticket says it's still in progress.

mullr:
:+1: from me

wkalt:
this appears to break our experimental warnings for event-counts and aggregate-event-counts. They're probably trying to get the product-name from the wrong place now.

wkalt:
looks good to me otherwise

ajroetker:
fwiw I'm :+1: here as well

ajroetker:
@senior will this allow us to get rid of the hook-tk-parse-config (robert hooke) as well?

ajroetker:
needs a rebase

senior:
@ajroetker No, we'll still need the robert hooke hack. I had another attempt at the defaulted config service that would have allowed us to remove that, but it required more state etc.

@wkalt I have updated event-counts and aggregate-event-counts to use product-server from config rather than shared globals. I also moved update-server over as well since it uses product-name similarly, I'll flag the changes with a comment, but it's a separate commit so should hopefully be reviewable by itself

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -6,14 +6,15 @@
...
+            [puppetlabs.puppetdb.config :as conf]))
 
 (defn routes
-  [version]
+  [version config]
   (app
     [""]
     {:get (fn [{:keys [params globals]}]
-            (when (= "puppetdb" (:product-name globals))
+            (when (conf/foss? config)
```
senior: This seems like the simplest way to go about this. We have the static config and a set of functions that can access the pieces we care about. I also thought about an atom that we bash on startup, then this becomes (conf/foss?) and we don't have to thread it through. That might be where we want to go (or something more fancy) but it seemed like this was a better place to start, until we move more things out of shared-globals (i.e. MQ stuff, catalog-hash-debug-dir etc)

rbrw:
The changes look pretty good overall, though the PR fails postgresql tests locally.

src/puppetlabs/puppetdb/pdb_routing.clj
```diff
@@ -26,23 +27,21 @@
 (defn wrap-with-context [uri route]
   (compojure/context uri [] route))
 
-(defn pdb-core-routes [get-shared-globals submit-command-fn query-fn enqueue-raw-command-fn response-pub]
+(defn pdb-core-routes [defaulted-config get-shared-globals submit-command-fn query-fn enqueue-raw-command-fn response-pub]
```
rbrw: Line's a bit long for me.

src/puppetlabs/puppetdb/pdb_routing.clj
```diff
@@ -52,7 +51,7 @@
                                              (format "%s/dashboard/index.html")
                                              rr/redirect))
                          (apply compojure/routes app-routes))
-      (mid/wrap-with-puppetdb-middleware #(:authorizer (get-shared-globals)))))
+      (mid/wrap-with-puppetdb-middleware (get-in defaulted-config [:puppetdb :certificate-whitelist]))))
```
rbrw: This one as well, and I think I noticed a few others, but perhaps not a big deal if they're not *too* long.

wkalt:
tests are failing because config isn't being passed to the routes function in http/aggregate-event-counts. When I fix that, I can see the experimental warning, but I see something else that's weird. On my first query after starting PDB, I get this warning

```
2015-09-08 11:33:41,404 WARN  [c.j.b.BoneCPConfig] JDBC username was not set in config!
2015-09-08 11:33:41,404 WARN  [c.j.b.BoneCPConfig] JDBC password was not set in config!
```

I expect to see that warning on startup, but not after a query.

senior:
@rbrw thanks for catching that, looks like we have some aggregate event counts tests that only run on postgres that I didn't catch, just pushed up a fix

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -6,14 +6,15 @@
...
               (log/warn "The aggregate-event-counts endpoint is experimental"
                         " and may be altered or removed in the future."))
```
wkalt: existing issue, but we're logging this line twice: here and on line 32 below. Also, I think log/warn is handling the spacing for you. I see this rendered as:

```
The aggregate-event-counts endpoint is experimental  and may be altered or removed in the future.
```

and the same on the event-counts meesage

PR #1629, opened 2015-09-01
(PDB-1359) support POST on all endpoints

wkalt:
This refactors our http namespaces and tests to work with POST, and moves most of our query namespace testing into the http test namespace (that's PDB-1714)


wkalt:
more cleanup/docs coming shortly, just want to get tests running on this

wkalt:

wkalt:
just noticed this is using set comparison in a few places where it's testing order by. correcting that now.

ajroetker:
Looks like this needs a rebase

documentation/api/query/curl.markdown
```diff
@@ -81,3 +81,14 @@ If you do this with an endpoint that accepts `GET` requests, **you must also use
 
     curl -G http://localhost:8080/pdb/query/v4/nodes \
       --data-urlencode 'query=["=", ["node", "active"], true]'
+
+## Querying PuppetDB with POST
```
kbarber: Some of tutorial.md mentions GET explicitly in the "How to Query Section", maybe worth going over that, seeing if it needs a mention.

documentation/api/query/curl.markdown
```diff
@@ -81,3 +81,14 @@ If you do this with an endpoint that accepts `GET` requests, **you must also use
...
+
+PuppetDB supports querying by POST, which is useful for particularly large
+queries (exact limits depend on the client and webserver used.)
+
+POST queries use the following syntax:
+
+    curl -X POST http://localhost:8080/pdb/query/v4/nodes \
+    -H 'Content-Type:application/json' \
```
kbarber: Indent the trailing lines 2 space, seems to be a convention we follow in the rest of the examples (also mildly easier to read).

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -1,49 +1,23 @@
...
-                       " and may be altered or removed in the future."))
-                (produce-streaming-body
-                  :aggregate-event-counts
-                  version
-                  query
-                  query-options
-                  (:scf-read-db globals)
-                  (:url-prefix globals)))))}))
+  [version optional-handlers]
+  (let [handlers (or optional-handlers [identity])
```
mullr: Can this be moved into `http-q/query-route`? Looks like it happens in most of the `routes` functions at least. Also, you might want it to be `(or (seq optional-handlers) [identity])` to handle the case where `optional-handlers` is an empty sequence.

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -20,24 +26,26 @@
   parameters. In a GET request this is contained in various query
   parameters, for POST requests this should be contained in the body
   of the request"
-  {:query (s/maybe [s/Any])
-   :count? s/Bool
+  {(s/optional-key :query) (s/maybe [s/Any])
+   (s/optional-key :include_total) (s/maybe s/Bool)
    (s/optional-key :order_by) (s/maybe [[(s/one s/Keyword "field")
                                          (s/one (s/enum :ascending :descending) "order")]])
+   (s/optional-key :distinct_resources) (s/maybe s/Bool)
```
mullr: Does this mean that we're now accepting all of these parameters on every endpoint? What happens in places when we you query /nodes with a counts_filter, for example?

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -20,24 +26,26 @@
...
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Query munging functions
 
-(defn- are-queries-different?
-  "DEPRECATED - only works on GET requests, will be replaced by are-queries-different?'"
-  {:deprecated "3.0.0"}
-  [req1 req2]
-  (not= (get-in req1 [:params "query"])
-        (get-in req2 [:params "query"])))
+(def experimental-entities
```
mullr: I think these are endpoints, not entities.

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -222,36 +182,47 @@
...
-  "If the original query contains the include_total w/o a a value, it
-  will show up here as nil.  We assume that in that case, the caller
-  intended to use it as a flag"
-  [query-map]
-  (let [total (get query-map :include_total ::not-found)]
-    (if (= total ::not-found)
-      false
-      (coerce-to-boolean total))))
+(defn validate-query-params
+  "Given a set of params and a param spec, throw and error if required params
```
mullr: s/and/an/

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -263,34 +234,77 @@
...
      :post (post-req->query req)
-     (throw (IllegalArgumentException. "PuppetDB queries must be made via GET/POST")))))
+     (throw (IllegalArgumentException. "PuppetDB queries must be made via GET/POST")))
+    param-spec))
 
 (defn extract-query
   "Query handler that converts the incoming request (GET or POST)
   parameters/body to a pdb query map"
-  [handler]
+  [handler param-spec]
```
mullr: param-spec shows up often enough that a schema for it would be very nice

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -263,34 +234,77 @@
...
+     #{:distinct_start_time :distinct_end_time}
+     (throw
+       (IllegalArgumentException.
+         "'distinct_resources' query parameter must accompany parameters 'distinct_start_time' and 'distinct_end_time'"))
+     (throw
+       (IllegalArgumentException.
+         "'distinct_resources' query parameter requires accompanying parameters 'distinct_start_time' and 'distinct_end_time'")))))
+
+(defn warn-experimental
+  [entity {:keys [product-name]}]
```
mullr: Wouldn't mind seeing an `:as globals` on this for documentation purposes. Or perhaps better, letting it take product-name directly and pulling it out of globals at the call site.

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -1,49 +1,23 @@
...
-                  (str "The aggregate-event-counts endpoint is experimental"
-                       " and may be altered or removed in the future."))
-                (produce-streaming-body
-                  :aggregate-event-counts
-                  version
-                  query
-                  query-options
-                  (:scf-read-db globals)
-                  (:url-prefix globals)))))}))
+  [version optional-handlers]
```
mullr: It seems odd to me that we flow around `optional-handlers`, when all `query-route` does with it is `comp` it onto its result. Couldn't this be done more cleanly by doing the `comp` on the result of query-route, when it's needed?

(I know we had optional-handlers before; it this is a change for another time, that's fine)

src/puppetlabs/puppetdb/http/fact_names.clj
```diff
@@ -1,40 +1,44 @@
...
 
 (defn routes
-  [query-app]
-  (app
-    []
-    (verify-accepts-json query-app)))
+  [version]
+  (let [param-spec {:optional paging/query-params}]
+    (app
+      []
```
ajroetker: So I've seen this quite a bit in our moustache routes, is there a reason we can't smash routes/query-route together?

ajroetker: My main confusion is why everything needs to be wrapped in `app []`

src/puppetlabs/puppetdb/http/catalogs.clj
```diff
@@ -53,17 +41,15 @@
                       (str (:url-prefix globals))))
 
     [node "edges" &]
-    (-> (comp (edges/edges-app version false) (partial http-q/restrict-query-to-node node))
+    (-> (edges/edges-app version false (partial http-q/restrict-query-to-node node))
```
mullr: Yes, this is an example of where I think the old way is actually better; unless passing the middleware as a parameter really buys us something, I'd rather see one fewer parameter that needs to be passed through in multiple places.

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
mullr: .... or is this because we need the middleware to apply only to requests matched by the route? That could be it.

src/puppetlabs/puppetdb/http/event_counts.clj
```diff
@@ -1,44 +1,27 @@
...
-              :event-counts
-              version
-              query
-              query-options
-              (:scf-read-db globals)
-              (:url-prefix globals))))}))
+  [version optional-handlers]
+  (let [handlers (or optional-handlers [identity])
+        param-spec {:required ["summarize_by"]
+                    :optional (concat ["query"
```
mullr: `paging/query-params` already includes "query"

src/puppetlabs/puppetdb/http/fact_contents.clj
```diff
@@ -2,34 +2,17 @@
...
-  [query-app]
-  (app
-   []
-   (verify-accepts-json query-app)))
+  [version]
+  (let [param-spec {:optional paging/query-params}]
+    (app
+      []
+      (http-q/query-route :fact-contents version param-spec
+                          http-q/restrict-query-to-active-nodes'))))
```
mullr: I think we only have the ' version of this now, maybe we can remove the mark.

src/puppetlabs/puppetdb/http/fact_names.clj
```diff
@@ -1,40 +1,44 @@
...
+  "Returns a route for querying PuppetDB that supports the standard
+   paging parameters. Also accepts GETs and POSTs. Composes
+   `optional-handlers` with the middleware function that executes the
+   query."
+  [entity version param-spec & optional-handlers]
+  (app
+    (http-q/extract-query param-spec)
+    (apply comp
+           (fn [{:keys [params globals puppetdb-query]}]
+             (let [puppetdb-query (if (nil? (:order_by puppetdb-query))
```
mullr: `ks/assoc-when`?

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -222,36 +182,47 @@
...
-  (let [total (get query-map :include_total ::not-found)]
-    (if (= total ::not-found)
-      false
-      (coerce-to-boolean total))))
+(defn validate-query-params
+  "Given a set of params and a param spec, throw and error if required params
+   are missing or unsupported params are present, otherwise return the params."
+  [params param-spec]
+  (let [params (stringify-keys params)]
+    (kitchensink/cond-let
```
mullr: can't say I'm a huge fan of cond-let, but if it simplifies the code I guess it's good.

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -128,57 +129,50 @@
...
-                which accepts a query map
-
-  Given a query, and database connection, return a Ring response with
-  the query results. `query` is either a string (if it's coming from a
-  GET request) or an already parsed clojure data structure (if it's
-  from a POST request).
+  "Given a query, and database connection, return a Ring response with
+   the query results. `query` is either a string (if it's coming from a
+   GET request) or an already parsed clojure data structure (if it's
+   from a POST request).
```
mullr: is this ever a string now? This function doesn't work on strings.

mullr:
Ok, mostly just small comments. :+1: after we deal with them or decide not to. I didn't look at the tests though, since approximately every test changed.

src/puppetlabs/puppetdb/http/query.clj
wkalt: it's existing code, just moved

src/puppetlabs/puppetdb/http/query.clj
wkalt:
```
[~/Documents/puppetdb] (ticket/master/pdb-1359-support-POST) $ curl -X GET http://localhost:8080/pdb/query/v4/nodes -d 'counts_filter=[">","successes",10]'
Unsupported query parameter 'counts_filter'[~/Documents/puppetdb] (ticket/master/pdb-1359-support-POST) $
```

wkalt: ah, but with a POST it succeeds. looking into it.

wkalt: actually that was a get with POST params. with a real POST it fails as expected.

src/puppetlabs/puppetdb/http/query.clj
wkalt: this is consistent with our use of "entity" elsewhere

src/puppetlabs/puppetdb/http/fact_names.clj
wkalt: no, I think we could definitely do that but I judged it to be unrelated to POST support. Worth a ticket IMO.

src/puppetlabs/puppetdb/http/query.clj
wkalt: implemented in validate-query-params

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
wkalt: discussed this in person -- leaving as is to ease consolidation of these functions into query-route down the line

mullr:
ok, I'm :+1: then

PR #1634, opened 2015-09-03
(PDB-1738) (PDB-1739) Drop jdbc.deprecated

rbrw:
```
(PDB-1738) (PDB-1739) Drop jdbc.deprecated

Migrate all code to the newer clojure.jdbc API.

Note that the newer jdbc API handles some database operations like
transactions by storing bookkeeping information in nested database
references.  For example, (with-db-transaction ...) creates a new
database connection map with an incremented level, and the library uses
that level to determine where the "top" is, i.e. to decide when to
commit.

To accommodate that approach (without threading a db argument through
all of the existing code), add a jdbc/*db* variable to refer to the
"current" database, and make sure to bind appropriately it in
puppetdb.jdbc wrappers for the relevant clojure.jdbc calls.

Use (jdbc/db) to refer to *db* when invoking the new jdbc API's
functions.

Use jdbc/ consistently to refer to puppetdb.jdbc in altered files.

While updating migrations, combine adjacent db-do-commands into one
broader do-commands.
```

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -44,7 +44,6 @@
      maintain acceptable performance."
   (:require [clj-time.core :refer [ago]]
             [clojure.java.io :as io]
-            [clojure.java.jdbc.deprecated :as sql]
             [clojure.tools.logging :as log]
             [compojure.core :as compojure]
             [overtone.at-at :refer [mk-pool interspaced]]
@@ -54,7 +53,7 @@
             [puppetlabs.puppetdb.command.dlo :as dlo]
             [puppetlabs.puppetdb.config :as conf]
             [puppetlabs.puppetdb.http.server :as server]
-            [puppetlabs.puppetdb.jdbc :as pl-jdbc :refer [with-transacted-connection]]
+            [puppetlabs.puppetdb.jdbc :as jdbc]
             [puppetlabs.puppetdb.meta.version :as version]
             [puppetlabs.puppetdb.scf.storage-utils :as sutils]
             [puppetlabs.puppetdb.mq :as mq]
@@ -91,7 +90,7 @@
     (kitchensink/demarcate
      (format "sweep of stale nodes (threshold: %s)"
              (format-period node-ttl))
-     (with-transacted-connection db
+     (jdbc/with-transacted-connection db
        (doseq [node (scf-store/stale-nodes (ago node-ttl))]
          (log/infof "Auto-expiring node %s" node)
          (scf-store/expire-node! node))))
@@ -107,7 +106,7 @@
     (kitchensink/demarcate
      (format "purge deactivated and expired nodes (threshold: %s)"
              (format-period node-purge-ttl))
-     (with-transacted-connection db
+     (jdbc/with-transacted-connection db
        (scf-store/purge-deactivated-and-expired-nodes! (ago node-purge-ttl))))
     (catch Exception e
       (log/error e "Error while purging deactivated and expired nodes"))))
@@ -121,7 +120,7 @@
     (kitchensink/demarcate
      (format "sweep of stale reports (threshold: %s)"
              (format-period report-ttl))
-     (with-transacted-connection db
+     (jdbc/with-transacted-connection db
        (scf-store/delete-reports-older-than! (ago report-ttl))))
     (catch Exception e
       (log/error e "Error while sweeping reports"))))
@@ -207,7 +206,7 @@
   database doesn't exist but we open and close a connection without
   creating anything."
   [db-conn-pool config]
-  (sql/with-connection db-conn-pool
+  (jdbc/with-db-connection db-conn-pool
     (scf-store/validate-database-version #(System/exit 1))
     @sutils/db-metadata
     (migrate! db-conn-pool)
@@ -221,11 +220,12 @@
   being fully started when PuppetDB starts. This connection pool will
   be opened and closed within the body of this function."
   [write-db-config config]
-  (with-open [init-db-pool (pl-jdbc/make-connection-pool (assoc write-db-config
-                                                           ;; Block waiting to grab a connection
-                                                           :connection-timeout 0
-                                                           ;; Only allocate connections when needed
-                                                           :pool-availability-threshold 0))]
+  (with-open [init-db-pool (jdbc/make-connection-pool
+                            (assoc write-db-config
+                                   ;; Block waiting to grab a connection
+                                   :connection-timeout 0
+                                   ;; Only allocate connections when needed
+                                   :pool-availability-threshold 0))]
     (let [db-pool-map {:datasource init-db-pool}]
       (initialize-schema db-pool-map config))))
 
@@ -245,8 +245,8 @@
                 max-frame-size threads]} command-processing
         {:keys [disable-update-checking]} puppetdb
 
-        write-db (pl-jdbc/pooled-datasource database)
-        read-db (pl-jdbc/pooled-datasource (assoc read-database :read-only? true))
+        write-db (jdbc/pooled-datasource database)
+        read-db (jdbc/pooled-datasource (assoc read-database :read-only? true))
         mq-dir (str (io/file vardir "mq"))
         discard-dir (io/file mq-dir "discarded")
         mq-connection-str (add-max-framesize max-frame-size mq-addr)]
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -2,7 +2,7 @@
   "Database utilities"
   (:import (com.jolbox.bonecp BoneCPDataSource BoneCPConfig)
            (java.util.concurrent TimeUnit))
-  (:require [clojure.java.jdbc.deprecated :as sql]
+  (:require [clojure.java.jdbc :as sql]
             [clojure.string :as string]
             [clojure.tools.logging :as log]
             [puppetlabs.kitchensink.core :as kitchensink]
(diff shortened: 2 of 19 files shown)


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -54,7 +53,8 @@
             [puppetlabs.puppetdb.command.dlo :as dlo]
             [puppetlabs.puppetdb.config :as conf]
             [puppetlabs.puppetdb.http.server :as server]
-            [puppetlabs.puppetdb.jdbc :as pl-jdbc :refer [with-transacted-connection]]
+            [puppetlabs.puppetdb.jdbc :as jdbc
+             :refer [with-transacted-connection]]
```
ajroetker: might as well move calls to `with-transacted-connection` under the `jdbc` ns as well

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -63,12 +79,12 @@
...
    :post [(or (zero? limit) (<= (count %) limit))]}
   (let [sql-query-and-params (if (string? query) [query] query)]
-    (sql/with-query-results result-set
-      sql-query-and-params
-      (let [limited-result-set (limit-result-set! limit result-set)]
-        (-> limited-result-set
-            convert-result-arrays
-            vec)))))
+    (sql/db-query-with-resultset
+     (jdbc/db)
```
ajroetker: I don't think you need `jdbc/` here

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -2,17 +2,33 @@
...
   (:import (com.jolbox.bonecp BoneCPDataSource BoneCPConfig)
            (java.util.concurrent TimeUnit))
-  (:require [clojure.java.jdbc.deprecated :as sql]
+  (:require [clojure.java.jdbc :as sql]
             [clojure.string :as string]
             [clojure.tools.logging :as log]
             [puppetlabs.kitchensink.core :as kitchensink]
             [clojure.string :as str]
             [puppetlabs.puppetdb.time :as pl-time]
+            [puppetlabs.puppetdb.jdbc :as jdbc]
```
ajroetker: Think this is unnecessary, since this is the ns we're in

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -205,6 +205,6 @@
                               INNER JOIN factsets
                               ON factsets.certname = certnames.certname
                               WHERE certnames.certname=?")]
-    (sql/with-query-results result-set
-      [check-sql id]
-      (pos? (count result-set)))))
+    (sql/db-query-with-resultset (jdbc/db)
+                                 [check-sql id]
+                                 #(boolean (seq (sql/result-set-seq %))))))
```
ajroetker: I think this should probably be a `comp`

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -217,22 +222,21 @@
...
-          :value_type_id 1
-          :value_hash (hash/generic-identity-hash 1)
-          :value_integer 1}))
+      (let [[p1 p2]
+            (map :id (sql/insert! (jdbc/db) :fact_paths
+                                  {:value_type_id 0
+                                   :path "orphan" :name "orphan" :depth 0}
+                                  {:value_type_id 1
+                                   :path "orphan" :name "orphan" :depth 0}))]
+        (sql/insert! (jdbc/db) :fact_values
```
ajroetker: can we just make our own `jdbc/insert!` which does this? Similar for `db-do-commands`? I think that's what @wkalt had in mind for the dynamic var.

wkalt: +1 to that

rbrw: Let's chat about this - I've gotten the impression there are varying opinions (though none categorical), and I started these changes accordingly, but as a practical matter I suspect we might be better off if I rework this (perhaps more in the direction you suggest).

rbrw:

rbrw:

wkalt:
test this please

wkalt:

wkalt:
this needs a rebase

rbrw:
@puppetlabs-jenkins retest this please...

wkalt:

senior:
@rbrw travis is failing right now, looks like migrate-test is hanging?

senior:
I think there's a memory leak here. Travis doesn't say OutOfMemoryError, but I see it locally. For me it ran out of memory at storage-test, but only when I ran the whole test suite. If I run just that test it's fine. Also I only saw this on HyperSQL.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -925,12 +942,13 @@
...
 (defn realize-records!
-  "Inserts the records (maps) into the named database and returns them
+  "Inserts the records (maps) into the named table and returns them
   with their new :id values."
-  [database records]
-  (map #(assoc %2 :id %1)
-       (map :id (apply sql/insert-records database records))
-       records))
+  [table records]
+  (when (seq records)
```
wkalt: did something change to require this?

test/puppetlabs/puppetdb/fixtures.clj
```diff
@@ -17,20 +17,20 @@
             [puppetlabs.puppetdb.scf.migrate :refer [migrate!]]
             [puppetlabs.puppetdb.middleware :as mid]))
 
-(def ^:dynamic *db* nil)
+(def ^:dynamic *db* nil) ;; FIXME: perhaps this should be *db-spec* (cf db.clj)?
```
wkalt: *db* is fine with me

test/puppetlabs/puppetdb/jdbc_test.clj
```diff
@@ -1,36 +1,46 @@
...
+                 (subject/query-to-vec
+                  "SELECT value FROM test WHERE key='foo'"))))))
+    (let [read-pool (-> (assoc *db-spec* :read-only? true)
+                        fixt/defaulted-read-db-config
+                        subject/pooled-datasource)]
+      (with-open [_ (:datasource read-pool)]
+        (subject/with-transacted-connection read-pool
+          (let [msg (try
+                      (insert-map {"bar" 1})
+                      ""
```
wkalt: what does this do?

wkalt: makes failures prettier I guess

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1230,8 +1243,7 @@
                :where [:= :certname certname]
                :order-by [[:producer_timestamp :desc]]
                :limit 1}]
-    (sql/with-query-results result-set (hcore/format query)
-      (:producer_timestamp (first result-set)))))
+    (:producer_timestamp (first (jdbc/query (hcore/format query))))))
```
wkalt: we're using query here and in a couple other spots while it seems like we're using query-with-resultset in most other similar places. I don't mind, but is there any particular reason?

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -191,23 +262,26 @@
      false)))
 
 (pls/defn-validated retry-sql*
-  "Executes f. If an exception is thrown, will retry. At most n retries
-   are done. If still some retryable error state is thrown it is bubbled upwards
-   in the call chain."
-  [remaining :- s/Int
+  "Inovkes (f) up to n times, retrying only if a transient connection
```
wkalt: invokes

wkalt:
managed to get all the way through this. Generally +1, just a few clarifying comments.

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: Yes - I believe insert! chokes when given no arguments, and insert-records didn't.

test/puppetlabs/puppetdb/fixtures.clj
rbrw: OK, the context was that I think the new API may behave differently in some cases when given a an actual DB connection as compared to a config (i.e. spec).  For example, if I recall correctly, if you give with-db-connection a db, it won't open a new connection, it'll just use the one you give it (and close it), whereas "with-connection" (the old one) would create a new connection.  I believe that's what made me wonder.

test/puppetlabs/puppetdb/jdbc_test.clj
rbrw: One change here was to use with-open so that we always close.  Another was to separate\ the read and write pool tests, since they didn't share the connections, and finally, the "read-only" string moved.  It's no longer in the first exception in the "get cause" list, so I added the full-sql-exception-msg function to flatten the text (which might be what you were really asking about...).

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: I tried to make sure I only used query if I knew the query had a (small) result limit, i.e. "limit 1", since I believe query is not lazy; otherwise, I used db-query-with-resultset.

rbrw:
@wkalt thanks for the slog - if my explanations sound plausible, then I'll push the typo fix, and perhaps we'll be set.

test/puppetlabs/puppetdb/jdbc_test.clj
wkalt: heh, I actually just meant the empty string on line 40. My guess is that when the test fails, that allows it to register as a failure rather than an error in re-find.

rbrw: Yes, I think that's right.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -262,24 +261,26 @@
...
   "Returns the id (primary key) from `table` that contain `row-map` values"
   [table :- s/Keyword
    row-map :- {s/Keyword s/Any}]
   (let [cols (keys row-map)
-        where-clause (str "where " (str/join " " (map (fn [col] (str (name col) "=?") ) cols)))]
-    (sql/with-query-results rs (apply vector (format "select id from %s %s" (name table) where-clause) (map row-map cols))
-      (:id (first rs)))))
+        q (format "select id from %s where %s"
+                  (name table)
+                  (str/join " " (map #(str (name %) "=?") cols)))]
```
ajroetker: Not part of this patch, but this seems like a good candidate for honeysql.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -993,133 +989,139 @@
...
+              SELECT id, " (munge-hash "value_hash") ", value_type_id,
+                     value_integer, value_float, value_string, value_boolean,
+                     value_json
+                FROM fact_values")
 
-    (sql/create-table :resource_params_cache_transform
-                      ["resource" hash-type "NOT NULL"]
-                      ["parameters" "TEXT"])
+      (sql/create-table-ddl :resource_params_cache_transform
+                             ["resource" hash-type "NOT NULL"]
```
ajroetker: alignment

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -199,15 +199,14 @@
   "Returns a boolean indicating whether or not the given certname exists in the db"
   [certname]
   {:pre [certname]}
-  (sql/with-query-results result-set
-    ["SELECT 1 FROM certnames WHERE certname=? LIMIT 1" certname]
-    (pos? (count result-set))))
+  (not (empty? (jdbc/query ["SELECT 1 FROM certnames WHERE certname=? LIMIT 1"
```
ajroetker: `not ( empty?` -> `some?`

ajroetker:
Lgtm other than small comments, will there need to be a corresponding PR for PE as well?

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: @ajroetker hmm, I believe query returns a (possibly empty) vector, not nil, so we may need empty? here.

rbrw: @ajroetker I take that back, the default for as-arrays? is false, so it calls doall on the result, not vec.

rbrw: ...in any case, we'd still need (some? (seq (query ...)))

rbrw:
@ajroetker looks like we're fine there.

ajroetker:
Well I g2g on this once the acceptance tests pass!

PR #1640, opened 2015-09-10
(PDB-135) Improve and reduce the number of UTF-8 warnings

senior:
Prior to this commit, any non-ASCII character found in a catalog would
be replaced by /ufffd and a warning would be issued saying it found an
invalid character. The code now takes into account the force_encoding
done by to_pson and attempts to force_encode it back to UTF-8. For most
cases this should be sufficient.

There are times when binary data can appear in a catalog. In this case,
we have characters that we can't represent in UTF-8. For this we
continue to give users a warning. This warning now includes the command
being submitted and the node the command is for. If users have debug
logging enabled, context is given around where this bad character data
occurred. Up to 100 characters before/after this bad data is included in
the Puppet debug log.

### DIFF
--- a/puppet/lib/puppet/util/puppetdb/char_encoding.rb
+++ b/puppet/lib/puppet/util/puppetdb/char_encoding.rb
@@ -31,37 +31,141 @@ module CharEncoding
 
   Utf8ReplacementChar = [ 0xEF, 0xBF, 0xBD ].pack("c*")
 
+  DEFAULT_INVALID_CHAR = "\ufffd"
 
-  def self.utf8_string(str)
+  # @api private
+  def self.all_indexes_of_char(str, char)
+    (0..str.length).find_all{ |i| str[i] == char}
+  end
+
+  # @api private
+  #
+  # Takes an array and returns a sub-array without the last element
+  #
+  # @return [Object]
+  def self.drop_last(array)
+    array[0..-2]
+  end
+
+  # @api private
+  #
+  # Takes an array of increasing integers and collapses the sequential
+  # integers into ranges
+  #
+  # @param index_array an array of sorted integers
+  # @return [Range]
+  def self.collapse_ranges(index_array)
+    ranges = index_array.each.inject([]) do |spans, n|
+      if spans.empty? || spans.last.end != n - 1
+        spans << Range.new(n, n)
+      else
+        drop_last(spans) << Range.new(spans.last.begin,n)
+      end
+    end
+  end
+
+  # @api private
+  #
+  # Scans the string s with bad characters found at bad_char_indexes
+  # and returns an array of messages that give some context around the
+  # bad characters. This will give up to 100 characters prior to the
+  # bad character and 100 after. It will return fewer if it's at the
+  # beginning of a string or if another bad character appears before
+  # reaching the 100 characters
+  #
+  # @param str string coming from to_pson, likely a command to be submitted to PDB
+  # @param bad_char_indexes an array of indexes into the string where invalid characters were found
+  # @return [String]
+  def self.error_char_context(str, bad_char_indexes)
+    bad_char_ranges = collapse_ranges(bad_char_indexes)
+    bad_char_ranges.each_with_index.inject([]) do |state, (r, index)|
+      gap = r.to_a.length
+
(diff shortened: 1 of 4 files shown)


puppet/lib/puppet/util/puppetdb/command.rb
```diff
@@ -42,7 +42,7 @@ def initialize(command, version, certname, payload)
       #
       # This is roughly inline with how Puppet serializes for catalogs as of
       # Puppet 4.1.0. We need a better answer to non-utf8 data end-to-end.
-      }.to_pson)
+      }.to_pson, "Error encoding a '#{command}' command for host '#{certname}'")
```
wkalt: should I see this getting logged with this catalog?
```
exec {"\u006E\u0303":
  command => "/usr/bin/uptime"
}

exec {"\u00F1":
  command => "/usr/bin/date"
}
```

I see this:
```
Warning: Ignoring invalid UTF-8 byte sequences in data to be sent to PuppetDB
```
in the puppet log, but no mention of replace catalog specifically.

wkalt: oh sorry, not on your code

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -31,37 +31,135 @@ module CharEncoding
...
+  # @api private
+  def self.all_indexes_of_char(str, char)
+    str.each_char.with_index.inject([]) do |indexes, (value, index)|
+      if value == char
+        indexes << index
+      else
+        indexes
+      end
+    end
+  end
```
ajroetker:
```
(0 ... str.length).find_all { |i| str[i] == char }
```
A little bit simpler

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -31,37 +31,135 @@ module CharEncoding
...
+  def self.collapse_ranges(index_array)
+    prev = index_array.first
+    index_array.slice_before do |curr_index|
+      old_prev = prev
+      prev = curr_index
+      old_prev + 1 != curr_index
+    end.map do |index_chunk|
+      Range.new(index_chunk.first, index_chunk.last)
+    end
+  end
```
ajroetker: Stollen from stackexchange
```
ranges = a.sort.uniq.inject([]) do |spans, n|
  if spans.empty? || spans.last.last != n - 1
    spans + [n..n]
  else
    spans[0..-2] + [spans.last.first..n]
  end
end
```

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -31,37 +31,135 @@ module CharEncoding
...
+    # warning and give the user more information
+    str_copy.force_encoding("UTF-8")
+    if str_copy.valid_encoding?
+      str_copy.encode!("UTF-8")
+    else
+      # This is force_encoded as US-ASCII to avoid any overlapping
+      # byte related issues that could arise from mis-interpreting a
+      # random extra byte as part of a multi-byte UTF-8 character
+      str_copy.force_encoding("US-ASCII")
+      warn_if_invalid_chars(str_copy.encode!("UTF-8", :invalid => :replace, :undef => :replace, :replace => DEFAULT_INVALID_CHAR),
```
ajroetker: could we break up this long line a little?

wkalt:
+1 from me. functionality looks good.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
ajroetker: Here's the monkey version
```
class Array
  # splits array to sub-arrays wherever two adjacent elements satisfy a condition
  def split_by
    each_cons(2).inject([[first]]){|a, (i, j)|
      a.push([]) if yield(i, j)
      a.last.push j
      a
    }
  end

  # uses split_by to split array to subarrays with consecutive elements, then convert to range
  def to_range
    split_by{|i,j| j-i!=1}.map{|a| a.first..a.last}
  end
end
```

senior:
@ajroetker reworked collapse_ranges based on your feedback, should be good to go now

ajroetker:
Lgtm :+1:

PR #1643, opened 2015-09-15
(PDB-1924) Implicit subqueries

kbarber:
This patch adds the new functionality for implicit subqueries, rather subqueries
that require no explicit column names to be specified.

### DIFF
--- a/documentation/api/query/v4/environments.markdown
+++ b/documentation/api/query/v4/environments.markdown
@@ -6,11 +6,17 @@ canonical: "/puppetdb/latest/api/query/v4/environments.html"
 
 [curl]: ../curl.html#using-curl-from-localhost-non-sslhttp
 [paging]: ./paging.html
-[events]: ./events.html
-[reports]: ./reports.html
-[resources]: ./resources.html
-[facts]: ./facts.html
 [query]: ./query.html
+[subqueries]: ./operators.html#subquery-operators
+[factsets]: ./factsets.html
+[reports]: ./reports.html
+[catalogs]: ./catalogs.html
+[facts]: ./facts.html
+[fact-contents]: ./fact_contents.html
+[events]: ./events.html
+[edges]: ./edges.html
+[resources]: ./resources.html
+
 
 Environments are semi-isolated groups of nodes managed by Puppet. Nodes are assigned to environments by their own configuration, or by the puppet master's external node classifier.
 
@@ -34,6 +40,24 @@ See [the Operators page](./operators.html)
 ### Query Fields
 
 * `"name"` (string): the name of an environment
+
+### Subquery Relationships
+
+Here is a list of relationships that can be queried using implicit subqueries. For
+more information consult the documentation for [subqueries].
+
+Direct relationships:
+
+* [`factsets`][factsets]: Factsets received from this environment.
+* [`reports`][reports]: Reports received from this environment.
+* [`catalogs`][catalogs]: Catalogs received for this this environment.
+
+Transitive relationships:
+
+* [`facts`][facts]: Fact names and values received from this environment.
+* [`fact_contents`][fact-contents]: Factset paths and values received from this environment.
+* [`events`][events]: Report events triggered from this environment.
+* [`resources`][resources]: Catalog resources received for this environment.
 
 ### Response format
 
--- a/documentation/api/query/v4/nodes.markdown
+++ b/documentation/api/query/v4/nodes.markdown
@@ -10,6 +10,15 @@ canonical: "/puppetdb/latest/api/query/v4/nodes.html"
 [paging]: ./paging.html
 [query]: ./query.html
 [8601]: http://en.wikipedia.org/wiki/ISO_8601
+[subqueries]: ./operators.html#subquery-operators
+[factsets]: ./factsets.html
+[reports]: ./reports.html
+[catalogs]: ./catalogs.html
+[facts]: ./facts.html
+[fact-contents]: ./fact_contents.html
+[events]: ./events.html
+[edges]: ./edges.html
+[resources]: ./resources.html
 
 Nodes can be queried by making an HTTP request to the `/nodes` endpoint.
 
@@ -59,6 +68,24 @@ The below fields are allowed as filter criteria and are returned in all response
     Note that nodes which are missing a fact referenced by a `not` query will match
     the query.
 
+### Subquery Relationships
+
+Here is a list of relationships that can be queried using implicit subqueries. For
+more information consult the documentation for [subqueries].
+
+Direct relationships:
+
+* [`factsets`][factsets]: Factsets received from this node.
+* [`reports`][reports]: Reports received from this node.
+* [`catalogs`][catalogs]: Catalogs received for this node.
+
+Transitive relationships:
+
+* [`facts`][facts]: Fact names and values received from this node.
+* [`fact_contents`][fact-contents]: Factset paths and values received from this node.
+* [`events`][events]: Report events triggered from this node.
+* [`edges`][edges]: Catalog edges received for this node.
+* [`resources`][resources]: Catalog resources received for this node.
 
 ### Response format
 
--- a/documentation/api/query/v4/operators.markdown
+++ b/documentation/api/query/v4/operators.markdown
@@ -4,11 +4,17 @@ layout: default
 canonical: "/puppetdb/latest/api/query/v4/operators.html"
 ---
 
-[resources]: ./resources.html
+[catalogs]: ./catalogs.html
+[edges]: ./edges.html
+[environments]: ./environments.html
+[events]: ./events.html
 [facts]: ./facts.html
+[fact-contents]: ./fact-contents.html
+[fact-paths]: ./fact-paths.html
 [nodes]: ./nodes.html
 [query]: ./query.html
-[fact-contents]: ./fact-contents.html
+[reports]: ./reports.html
+[resources]: ./resources.html
 
 PuppetDB's [query strings][query] can use several common operators.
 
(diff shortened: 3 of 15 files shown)


documentation/api/query/v4/environments.markdown
```diff
@@ -35,6 +41,25 @@ See [the Operators page](./operators.html)
...
+### Subquery Relationships
+
+Here is a list of relationships that can be queried using implicit subqueries. For
+more information consult the documentation for [subqueries].
+
+Direct relationships:
+
+* [factsets][`factsets`]: Factsets received from this environment.
+* [reports][`reports`]: Reports received from this environment.
+* [reports][`catalogs`]: Catalogs received for this this environment.
```
kbrezina: [reports] -> [catalogs]

documentation/api/query/v4/nodes.markdown
```diff
@@ -59,6 +68,24 @@ The below fields are allowed as filter criteria and are returned in all response
...
+### Subquery Relationships
+
+Here is a list of relationships that can be queried using implicit subqueries. For
+more information consult the documentation for [subqueries].
+
+Direct relationships:
+
+* [factsets][`factsets`]: Factsets received from this node.
+* [reports][`reports`]: Reports received from this node.
+* [reports][`catalogs`]: Catalogs received for this node.
```
kbrezina: Similar here, "[reports]" should be replaced with "[catalogs]"

documentation/api/query/v4/environments.markdown
kbarber: Let me fix both cases.

kbrezina:
LGTM

documentation/api/query/v4/environments.markdown
```diff
@@ -35,6 +41,25 @@ See [the Operators page](./operators.html)
...
+* [factsets][`factsets`]: Factsets received from this environment.
+* [reports][`reports`]: Reports received from this environment.
+* [catalogs][`catalogs`]: Catalogs received for this this environment.
+
+Transitive relationships:
+
+* [facts][`facts`]: Fact names and values received from this environment.
+* [fact-contents][`fact_contents`]: Factset paths and values received from this environment.
+* [events][`events`]: Report events triggered from this environment.
+* [resources][`resources`]: Catalog resources received for this ebvuribnebt.
```
wkalt: s/ebvuribnebt/environment

kbarber: Hahaha ... I can only presume I lost my home key positioning, and was typing without looking at the screen :-). Thats fabulous.

documentation/api/query/v4/operators.markdown
```diff
@@ -171,10 +177,73 @@ To get the average uptime for your nodes,
...
+        ["=", "title", "Tomcat"]]]
+
+If you wanted to display the entire `networking` fact, if the hosts interfaces uses a certain mac address
+you can do the following on the [facts][`facts`] endpoint:
+
+    ["and",
+      ["=", "name", "networking"]
+      ["subquery", "fact_contents",
+        ["~>", "path", ["networking", ".*", "macaddresses", ".*"]],
+        ["=", "value", "aa:bb:cc:dd:ee:00"]]]
```
wkalt: this needs a comma after the closing bracket on networking, but it also doesn't for me when pasted:


```
[~/work/puppetdb] (ticket/master/pdb-1924-implicit-subqueries) $ curl -X GET http://localhost:8080/pdb/query/v4/facts -d 'query=["and", ["=", "name", "networking"], ["subquery", "fact_contents", ["~>", "path", ["networking", ".*", "macaddresses", ".*"]], ["=", "value", "aa:bb:cc:dd:ee:00"]]]'
Query operator ~> is not allowed on field path
```

not sure what the story is. I haven't read far enough to know yet but it also seems like the subquery clauses of this one and the one above should contain an "and".

wkalt: the issue is that the metadata for the query node being evaluated in validate-binary-operators is nil, when it should be the fact-contents query-rec. This is a separate bug that I actually encountered yesterday, but we shouldn't have nonworking examples whether it's addressed here or not.

documentation/api/query/v4/operators.markdown
```diff
@@ -171,10 +177,73 @@ To get the average uptime for your nodes,
...
+* [facts][`facts`]
+* [fact-contents][`fact_contents`]
+* [fact-paths][`fact_paths`]
+* [nodes][`nodes`]
+* [reports][`reports`]
+* [resources][`resources`]
+
+Internal to PuppetDB, we keep a mapping of how different entities relate to each
+other, and so no other data beyond the entity is needed in this case. This is
+different to explicitly subqueries, for those you must specify yourself how
```
wkalt: explicit

documentation/api/query/v4/operators.markdown
```diff
@@ -171,10 +177,73 @@ To get the average uptime for your nodes,
...
+two entities are related, although functionally they can produce the same results.
 
-Subqueries are unlike the other operators listed above. They always appear together in the following form:
+#### Implicit Subquery Examples
+
+A query string like the following on the [nodes][`nodes`] endpoint will return the list
+of all nodes with the `Package[Tomcat]` resource in its catalog, and a certname starting
```
wkalt: their catalogs

documentation/api/query/v4/operators.markdown
```diff
@@ -229,18 +296,22 @@ Subquery statements are **non-transitive** and take two arguments:
...
+* [catalogs][`select_catalogs`]
+* [edges][`select_edges`]
+* [environments][`select_environments`]
+* [events][`select_events`]
+* [facts][`select_facts`]
+* [fact-contents][`select_fact_contents`]
+* [fact-paths][`select_fact_paths`]
+* [nodes][`select_nodes`]
+* [reports][`select_reports`]
+* [resources][`select_resources`]
```
wkalt: these additions are great

documentation/api/query/v4/nodes.markdown
```diff
@@ -59,6 +68,24 @@ The below fields are allowed as filter criteria and are returned in all response
...
+* [reports][`reports`]: Reports received from this node.
+* [catalogs][`catalogs`]: Catalogs received for this node.
+
+Transitive relationships:
+
+* [facts][`facts`]: Fact names and values received from this node.
+* [fact-contents][`fact_contents`]: Factset paths and values received from this node.
+* [events][`events`]: Report events triggered from this node.
+* [edges][`edges`]: Catalog edges received for this node.
+* [resources][`resources`]: Catalog resources received for this node.
```
wkalt: this bracketed style renders badly in github markdown -- not sure how it'll look on the web, but it's been introduced in a number of places here.

wkalt: maybe they'll be links if the order is swapped?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1284,6 +1344,19 @@
...
+                                     (user-query->logical-obj (str "select_" relationship))
+                                     expr)
+                      nested-qc (:query-context (meta subquery-expr))]
+
+                  {:node (vary-meta ["subquery" relationship
+                                     (vary-meta expr
+                                                assoc :query-context nested-qc)]
+                                    assoc :query-context context)
+                   :state state
+                   :cut true})
```
wkalt: what does this do?

test/puppetlabs/puppetdb/http/catalogs_test.clj
```diff
@@ -141,6 +141,48 @@
...
+   method [:get :post]]
+
+  (testcat/replace-catalog (json/generate-string catalog1))
+  (testcat/replace-catalog (json/generate-string catalog2))
+
+  (are [query expected]
+      (is (= (-> (reader (:body (query-response method endpoint query)))
+                 (json/parse-stream true)
+                 strip-hash
+                 set)
```
wkalt: this probably duplicates testutils/http ordered-query-result

documentation/api/query/v4/environments.markdown
```diff
@@ -35,6 +41,25 @@ See [the Operators page](./operators.html)
...
+Here is a list of relationships that can be queried using implicit subqueries. For
+more information consult the documentation for [subqueries].
+
+Direct relationships:
+
+* [factsets][`factsets`]: Factsets received from this environment.
+* [reports][`reports`]: Reports received from this environment.
+* [catalogs][`catalogs`]: Catalogs received for this this environment.
+
+Transitive relationships:
```
wkalt: I'm not sure the transitive/direct distinction is going to be useful to users.

wkalt:
Code look fine to me generally, but I'm a bit uneasy with the change itself (and I apologize in advance because I know we've already been over it). Interested in what the rest of the team thinks, and if I'm alone I'm happy to defer.

This change gains us an AST-based analog for the implicit subquerying capabilities that are planned for PQL, and also gets us a more concise alternative to the unwieldy in/extract style required for subqueries today. I understand that implicit subqueries are desirable for PQL because we need queries to be as concise as possible, but generally my view is that explicit is always preferable to implicit, and that introducing multiple syntaxes for the same objective is usually bad, because it increases testing burden, makes documentation more confusing, and increases the potential for weird bugs.

There are ways we could improve the existing in/extract stuff similarly to what you have, but leaving things explicit still. If we changed the "name" key of environments to "environment", we might not have to worry about the foreign-key/local-key stuff and could support syntax like this on all endpoints:

```
["subquery",["reports","certname"],[">","received_time","2015-01-01"]]
```

If you issued that against factsets for instance, you'd get all factsets for certnames for which you've received reports since 2015-01-01. The second argument to "subquery" would be a variable length array, with the first element representing the other endpoint and the rest representing the in/extract fields. I think if we reformulated in/extract that way we could deprecate the existing syntax and the result (having to type "certname") would not be especially more burdensome to consumers than making use of the implicit relationships defined in this patch. This would need some extra element for support of cases like

```
/nodes ["in", "report_timestamp",  ["extract", "received_time"], ["select_reports", ["~","certname", ".*"]]]
```

Implicit subqueries in PQL could then be made implicit via some light translation. I know our goal is to support the same functionality on both sides, but I don't really count implicit subquerying as a piece of functionality -- it's just an alternative syntax. If implicit subqueries are desirable in the AST language, the changes to explicit queries above would at least make the two more consistent.

test/puppetlabs/puppetdb/http/catalogs_test.clj
wkalt: sorry, that should just be the regular query-result. strip-hash can be passed as an optional handler.

kbarber:
@wkalt I like your idea, plus I think it will simplify the documentation also if the syntaxes between the two were just slightly different ... that is ... join fields become optional, all we need to do is explain what happens when you don't specify them, to explain implicit subqueries.

src/puppetlabs/puppetdb/query_eng/engine.clj
senior: the :cut true keypair? or something else?

documentation/api/query/v4/environments.markdown
senior: I was a little confused by it. Are you calling it transitive because nodes connect to facts by way of factsets?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -235,6 +249,8 @@
                                        [:= :fs.environment_id :env.id]]
                            :where [:= :fp.depth 0]}
 
+               :relationships {"fact_contents" {:columns ["certname","name"]}}
```
mullr: I fully support this comma

src/puppetlabs/puppetdb/query_eng/engine.clj
wkalt: I meant the :cut key

senior: It's a hint to the zipper functions. This code is a pre-order traversal, meaning the parent in the tree is visited first, then the children. The cut tells the zipper code it should not visit the children of this node. What this code does is attach a query context to each of the expressions in the query. A query is issued for the reports endpoint, but nothing in the JSON query expression says anything about reports. So this code walks the tree and attaches a query context to each piece of that query so it can be processed in isolation. A subquery shouldn't be annotated with reports, but rather events (or whatever was specified in the query). So it doesn't want to walk into that branch of the query and instead, do something different at that part.

documentation/api/query/v4/operators.markdown
```diff
@@ -171,10 +177,73 @@ To get the average uptime for your nodes,
...
+        ["=", "title", "Tomcat"]]]
+
+If you wanted to display the entire `networking` fact, if the hosts interfaces uses a certain mac address
+you can do the following on the [facts][`facts`] endpoint:
+
+    ["and",
+      ["=", "name", "networking"],
+      ["subquery", "fact_contents",
+        ["~>", "path", ["networking", ".*", "macaddresses", ".*"]],
+        ["=", "value", "aa:bb:cc:dd:ee:00"]]]
```
kbarber: Lost comment from @wkalt from a previous rebase:

```
this needs a comma after the closing bracket on networking, but it also doesn't for me when pasted:

[~/work/puppetdb] (ticket/master/pdb-1924-implicit-subqueries) $ curl -X GET http://localhost:8080/pdb/query/v4/facts -d 'query=["and", ["=", "name", "networking"], ["subquery", "fact_contents", ["~>", "path", ["networking", ".*", "macaddresses", ".*"]], ["=", "value", "aa:bb:cc:dd:ee:00"]]]'
Query operator ~> is not allowed on field path
not sure what the story is. I haven't read far enough to know yet but it also seems like the subquery clauses of this one and the one above should contain an "and".
 @wkalt   Collaborator
wkalt added a note 5 days ago
the issue is that the metadata for the query node being evaluated in validate-binary-operators is nil, when it should be the fact-contents query-rec. This is a separate bug that I actually encountered yesterday, but we shouldn't have nonworking examples whether it's addressed here or not.
```

documentation/api/query/v4/environments.markdown
```diff
@@ -35,6 +41,24 @@ See [the Operators page](./operators.html)
...
+Here is a list of relationships that can be queried using implicit subqueries. For
+more information consult the documentation for [subqueries].
+
+Direct relationships:
+
+* [`factsets`][factsets]: Factsets received from this environment.
+* [`reports`][reports]: Reports received from this environment.
+* [`catalogs`][catalogs]: Catalogs received for this this environment.
+
+Transitive relationships:
```
kbarber: Lost comments from rebase from @wkalt:

```
I'm not sure the transitive/direct distinction is going to be useful to users.
```

and from @senior:

```
I was a little confused by it. Are you calling it transitive because nodes connect to facts by way of factsets?
```

kbarber:
This might now be a 'stable' target branch, since master is now 4.0.0. Or not, not sure :-).

kbarber:
I've split out the extra selects, and tests into this PR: https://github.com/puppetlabs/puppetdb/pull/1675 So this will need cleanup once thats merged.

kbarber:
Closing, rebasing against stable.

PR #1644, opened 2015-09-15
(PDB-1210) storage of > bigints should be possible

wkalt:
This changes our value_integer column to be of type numeric(1000,0), which will
allow storage of integers up to 1000 digits.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1495,6 +1495,45 @@
       (sql/do-commands
        "ALTER TABLE certnames ADD CONSTRAINT certnames_reports_id_fkey FOREIGN KEY (latest_report_id) REFERENCES reports(id) ON DELETE SET NULL")))
 
+
+(defn change-value-integer-to-numeric
+  []
+  (let [hash-type (if (sutils/postgres?) "bytea" "varchar(40)")
+        munge-hash (if (sutils/postgres?)
+                     (fn [column] (format "('\\x' || %s)::bytea" column)) identity)]
+    (sql/create-table :fact_values_transform
+                      [ "id" "bigint NOT NULL DEFAULT nextval('fact_values_id_seq')"]
+                      ["value_hash" hash-type "NOT NULL"]
+                      ["value_type_id" "bigint NOT NULL"]
+                      ["value_integer numeric(1000,0)"]
+                      ["value_float" "double precision"]
+                      ["value_string" "text"]
+                      ["value_boolean" "boolean"]
+                      ["value" "text"])
+    (sql/do-commands
+      (str "INSERT INTO fact_values_transform
+            (id, value_hash, value_type_id, value_integer, value_float,
+            value_string, value_boolean, value)
+            SELECT id, " (munge-hash "value_hash") ", value_type_id, value_integer, value_float, value_string, value_boolean, value FROM fact_values"))
+
+    (sql/do-commands
+      "ALTER TABLE facts DROP CONSTRAINT fact_value_id_fk"
+      "ALTER TABLE fact_values DROP CONSTRAINT fact_values_pkey"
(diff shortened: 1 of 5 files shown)


test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -749,6 +750,26 @@
   ([endpoint query paging-options]
    (:results (raw-query-endpoint endpoint query paging-options))))
 
+(deftestseq large-integer-storage
+  [[version endpoint] facts-endpoints
+   method [:get :post]]
+
+  (let [f {:certname "a.local" :name "large-int" :environment "DEV"
+           :value 3546565741691827463891754698375639827814623789421387401348}]
```
kbarber: I wonder if we should test it closer to the maximum, and then test it when it overflows, make sure it returns something meaningful. Really we want to shake up the rest of the stack here also: Java, JSON libraries etc. BigInteger should handle it though.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1495,6 +1495,45 @@
...
+(defn change-value-integer-to-numeric
+  []
+  (let [hash-type (if (sutils/postgres?) "bytea" "varchar(40)")
+        munge-hash (if (sutils/postgres?)
+                     (fn [column] (format "('\\x' || %s)::bytea" column)) identity)]
+    (sql/create-table :fact_values_transform
+                      [ "id" "bigint NOT NULL DEFAULT nextval('fact_values_id_seq')"]
+                      ["value_hash" hash-type "NOT NULL"]
+                      ["value_type_id" "bigint NOT NULL"]
+                      ["value_integer numeric(1000,0)"]
```
kbarber: Looks like the maximum definable "up front" is 1000, otherwise it's 131072 digits. Is that why you chose 1000?

I think that's about 3320 bits of storage, about 415 bytes (without storage overhead). The maximum that decimal can take for a non-float is 435410 bits, about 55 kb per cell.

test/puppetlabs/puppetdb/http/facts_test.clj
kbarber: Ha! I wonder what happens, when you throw a float in JSON, that is larger than what java can handle with the float primitives. Does that turn into a BigInteger, and then attempt to get stored in here?

src/puppetlabs/puppetdb/scf/migrate.clj
kbarber: Nope I'm wrong, from the manual:

```Numeric values are physically stored without any extra leading or trailing zeroes. Thus, the declared precision and scale of a column are maximums, not fixed allocations. (In this sense the numeric type is more akin to varchar(n) than to char(n).) The actual storage requirement is two bytes for each group of four decimal digits, plus three to eight bytes overhead.```

http://www.postgresql.org/docs/current/interactive/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL

Which means 1000 ~ 508 bytes. And the maximum of 131072 ~ 32776 bytes Does that sound about right?

File: src/puppetlabs/puppetdb/scf/migrate.clj [hunk 3]
```diff
@@ -1495,6 +1495,45 @@
...
+                      ["value_float" "double precision"]
```
kbarber: It might sound looney, but this one is possibly going to be a problem too, and it will fall into the same bucket. Thinking ahead, is this theoretical outcome just going to force us into 1 column? I'm just conjecturing, I'm not sure we should change it just yet. One interesting side-effect will be a less complicated cast methodology.

src/puppetlabs/puppetdb/scf/migrate.clj
kbarber: <troll>I wonder what OracleDB limits are</troll>

wkalt:
shelving this until we drop hsqldb support

PR #1645, opened 2015-09-15
(maint) Bump oldes support puppet version in Gemfile

senior:
Packaging currently requires at least Puppet 3.8.1, but the Gemfile is
out of date and lists the olds supported Puppet as 3.5.1. Bumping the
Gemfile to match packaging.


senior:
retargeting at stable https://github.com/puppetlabs/puppetdb/pull/1646

PR #1647, opened 2015-09-15
(maint) Change char encoding context test to > 1.8.7

senior:
The code doesn't work on 1.8.7. The main code paths guard against that,
but the tests did not. This puts that guard around the tests


senior:
this needs to be on master

PR #1648, opened 2015-09-15
(maint) Change char encoding context test to > 1.8.7

senior:
The code doesn't work on 1.8.7. The main code paths guard against that,
but the tests did not. This puts that guard around the tests


ajroetker:
Doesn't appear to fix the issue for me locally

ajroetker:
Looks like `command_spec.rb` needs some guards for Ruby `1.8.7` and those `described_class`es need to be turned into `it "...."; subject.<method>(...); end`s.

PR #1650, opened 2015-09-16
(WIP) use jmsxgroup to reduce certname contention

mullr:
Looking for feedback on this. Clearly bad for perf as it stands, but we can probably do a much lighter weight json parse to only get the info we need.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -124,10 +124,15 @@
    mq-endpoint :- s/Str
    raw-command :- s/Str
    uuid :- (s/maybe s/Str)]
-  (let [uuid (or uuid (kitchensink/uuid))]
+  (let [uuid (or uuid (kitchensink/uuid))
+        parsed (json/parse-string raw-command)
+        certname (get-in parsed ["payload" "certname"])]
     (mq/send-message! mq-connection mq-endpoint
                       raw-command
-                      {"received" (kitchensink/timestamp) "id" uuid})
+                      (merge {"received" (kitchensink/timestamp)
+                              "id" uuid}
+                             (when certname
+                               {"JMSXGroupID" certname})))
     uuid))
 
 (defn-validated ^:private do-enqueue-command :- s/Str


senior:
I think we need some heavy testing on this. One of the things I'm concerned with is making the thundering herd problem worse. What if we get commands dribbled in, all get picked up by the same consumer, then we get slammed with commands. The certnames would all be going to 1 consumer, queue would be really backed up, even if there were other consumers available.

I'm not sure that would be a problem or not, but I didn't see anything in the docs that say there's an initial round-robin or something. It describes a similar problem if one listener starts up quickly and others start slowly (i.e. the first consumes the things, marks them all on it's thread, then all subsequent messages always go to that thread).

mullr:
100% agreed on the testing, of course. 

What I really want is to use hash(certname) % numWorkers to choose a worker predictably, but I'm not sure that's a choice.

wkalt:
I think this may also negatively affect the performance of imports, since currently they go certname by certname.

ajroetker:
@wkalt not anymore!

mullr:
After some discussion with @rbrw and @senior, I think a better approach to this would be manage the scheduling of commands to worker threads ourselves, rather than letting AMQ do it. The AMQ (JMS) semantics are built around a distributed queue, which is not the world we're in here. I think core.async would be a good tool for this job; the interface to AMQ would be two channels, one to put things on the queue and one to take things off. Then we could deserialize the commands as they come off the queue and dispatch them however we like (probably `worker=hashmod(certname, numWorkers)` with work stealing).

wkalt:
@mullr that sounds pretty heavyweight for what I understand to be a cosmetic issue only encountered in the initial loading of a benchmark run -- do we expect other benefits from the change?

mullr:
@wkalt as a solution for just getting rid of those error messages, I agree. But I think this is a good refactor anyway anyway; AMQ's semantics aren't the greatest fit for our system, and it tends to leak its details around the codebase. (callback registration, trying to handle mysterious shutdown errors, and so on) It's cleaner overall to put all our AMQ interaction in one place and communicate with it via core.async.

wkalt:
I guess it comes down to perf testing like ryan's saying then. I understand the cleanliness angle but that seems like it could only increase overhead in the hottest part of the code.

mullr:
I don't think this would increase overhead in any appreciable way. The only real difference is that we'd be deserializing the payload before dispatching it to a worker thread. This wouldn't necessarily happen on the AMQ consumer thread; in fact, it makes more sense to put it in a `go` block so it's on the core.async thread pool. So the fundamental amount of work is the same, it's just moved around a little bit. AMQ doesn't afford us this option; it wants to handle all the dispatch itself.

kbarber:
@mullr this might now be a 'stable' target, just FYI, since master is now 4.0.0.

mullr:
closing this, as we decided not to do it.

PR #1653, opened 2015-09-18
(maint) Store reports without resources again

ajroetker:
This commit fixes a bug with the jdbc upgrade where reports without
resources/resource-events would throw and exception while trying to call
`jdbc/insert!`.


wkalt:
:+1: verified the test and the fix

PR #1654, opened 2015-09-21
(PDB-1848) make max-frame-size tunable

wkalt:
previously we had a schema that was attempting to validate this as a string,
when it should have been an int.


rbrw:
Looks right to me, and passes tests with both databases.

PR #1655, opened 2015-09-22
(PDB-1998) use a coalesce instead of jsonb migration

wkalt:
This changes our jsonb migration to leave the existing json column in place,
and changes our query code to use a coalesce of the two instead of the one-shot
migration to jsonb.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -36,6 +36,7 @@
 
 (def json-agg-row (comp h/json-agg h/row-to-json))
 (def supported-fns #{"sum" "avg" "min" "max" "count"})
+(defn jsonb-type [] (if (su/postgres?) :jsonb :text))
 
 (defn hsql-hash-as-str
   [column-keyword]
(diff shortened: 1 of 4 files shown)


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -359,12 +365,17 @@
       "metrics" {:type :json
                  :queryable? false
                  :field {:select [(h/row-to-json :t)]
-                         :from [[{:select [[:metrics :data] [(hsql-hash-as-str :hash) :href]]} :t]]}
+                         :from [[{:select
+                                  [[(h/coalesce :metrics
+                                                (h/scast :metrics_json (jsonb-type))) :data]
```
wkalt: still testing this with a large dataset. my remaining concern is that this cast could be very expensive for queries returning large numbers of reports.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -297,7 +298,9 @@
   []
   (map->Query {:projections {"logs" {:type :json
                                      :queryable? false
-                                     :field :reports.logs}
+                                     :field (h/coalesce :logs
```
senior: For HyperSQL I guess this hinges on there always being a value in the logs column? What happens in HyperSQL if logs in null?

senior: I think we also need tests around the various scenarios here (data in JSON, data in JSONB etc)

wkalt: @senior added a test to cover these cases

PR #1656, opened 2015-09-22
(PDB-1984) Use wire-v6 of reports in export/anonymization/import

ajroetker:
This PR cleans up some of the anonymization code syntax and moves to use v6 of reports wire-format in our cli related code.

### DIFF
--- a/src/puppetlabs/puppetdb/anonymizer.clj
+++ b/src/puppetlabs/puppetdb/anonymizer.clj
@@ -18,13 +18,15 @@
    (contains? edge "target")
    (contains? edge "relationship")))
 
-(def resource-event-schema-str (utils/str-schema reports/resource-event-v5-wireformat-schema))
+(def event-schema-str (utils/str-schema reports/event-wireformat-schema))
+(def resource-schema-str (utils/str-schema (assoc reports/resource-wireformat-schema
+                                                  :events [event-schema-str])))
 (def metric-schema-str (utils/str-schema reports/metric-wireformat-schema))
 (def log-schema-str (utils/str-schema reports/log-wireformat-schema))
-(def report-schema-str (utils/str-schema (assoc reports/report-v5-wireformat-schema
-                                           :resource_events [resource-event-schema-str]
-                                           :metrics (s/maybe [metric-schema-str])
-                                           :logs (s/maybe [log-schema-str]))))
+(def report-schema-str (utils/str-schema (assoc reports/report-wireformat-schema
+                                                :resources [resource-schema-str]
+                                                :metrics [metric-schema-str]
+                                                :logs [log-schema-str])))
 
 (defn resource?
   "Returns true if it looks like a resource"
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -304,109 +298,89 @@
...
   [resources context config]
   {:pre  [(coll? resources)]
    :post [(coll? %)]}
-  (map #(anonymize-resource % context config) resources))
+  (map #(anonymize-catalog-resource % context config) resources))
 
-(pls/defn-validated anonymize-resource-event :- resource-event-schema-str
-  "Anonymize a resource event from a report"
-  [event :- resource-event-schema-str
+(defn anonymize-event
```
wkalt: did you intend to remove the validation on these functions? seems like we've lost it in a number of places, but may still be covered at the edges

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -304,109 +298,89 @@
...
-
-(defn anonymize-resource-events
-  "Anonymize a collection of resource events from a report"
-  [events context config]
-  {:pre  [(coll? events)]
-   :post [(coll? %)]}
-  (sort-by
-   #(mapv % ["timestamp" "resource_type" "resource_title" "property"])
-   (map #(anonymize-resource-event % context config) events)))
+  (map #(anonymize-log % context config) logs))
```
wkalt: how does this handle existing data with nil-valued logs?

wkalt: oh I see, you've changed it to update-when

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -415,25 +389,25 @@
...
    :post [(catalog? %)]}
-  (let [context {"node" (get catalog ["certname"])}]
+  (let [context {"node" (get catalog "certname")}]
     (-> catalog
-        (update-in ["resources"]        anonymize-resources context config)
-        (update-in ["edges"]            anonymize-edges context config)
-        (update-in ["certname"]         anonymize-leaf :node context config)
-        (update-in ["transaction_uuid"] anonymize-leaf :transaction_uuid context config)
-        (update-in ["environment"] anonymize-leaf :environment context config))))
+        (update "resources" anonymize-catalog-resources context config)
```
wkalt: should this be update-when as well?

wkalt: no, this is catalogs

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -415,25 +389,25 @@
...
   (let [context {"node" (get report "certname")}]
     (-> report
-        (update-in ["certname"]         anonymize-leaf :node context config)
-        (update-in ["resource_events"]  anonymize-resource-events context config)
-        (update-in ["metrics"] anonymize-metrics context config)
-        (update-in ["logs"] anonymize-logs context config)
-        (update-in ["transaction_uuid"] anonymize-leaf :transaction_uuid context config)
-        (update-in ["environment"] anonymize-leaf :environment context config))))
+        (update "certname" anonymize-leaf :node context config)
+        (update "resources" anonymize-report-resources context config)
```
wkalt: should this be update-when?

src/puppetlabs/puppetdb/import.clj
```diff
@@ -40,7 +40,7 @@
       (do (log/infof "Importing report from archive entry '%s'" path)
           (command-fn :store-report
                       (:store_report command-versions)
-                      (reports/sanitize-report (utils/read-json-content tar-reader true))))
```
wkalt: I don't understand why this was needed or why it's being removed

ajroetker: It wasn't used which is ones of the reasons I removed it, another reason it that the only real thing that needs/uses this sanitization appears to be benchmarking, which is still using v5, so I figured I should probably handle that in a different patch, and why upgrade it here when it's not used.

src/puppetlabs/puppetdb/anonymizer.clj
ajroetker: It doesn't actually matter for metrics/logs/resources, because (map foo nil) will return (), which are valid values. But I'll change it for consistency.

src/puppetlabs/puppetdb/reports.clj
```diff
@@ -153,34 +153,40 @@
...
       (update :resource_events resource-events-query->wire-v5)
-      (update :metrics metrics-query->wire-v5)
-      (update :logs logs-query->wire-v5)))
+      (update :metrics :data)
+      (update :logs :data)))
 
 (defn reports-query->wire-v5 [reports]
   (map report-query->wire-v5 reports))
 
+(declare wire-v5->wire-v6)
```
wkalt: seems like this can be avoided by reordering the defns

src/puppetlabs/puppetdb/reports.clj
```diff
@@ -153,34 +153,40 @@
...
-  (:data logs))
-
-(pls/defn-validated metrics-query->wire-v5 :- [metric-wireformat-schema]
-  [metrics :- metrics-expanded-query-schema]
-  (:data metrics))
+  [resource-events :- resource-events-expanded-query-schema]
+  (->> resource-events
+       :data
+       remove-reports-metadata
+       (sort-by #(mapv % [:timestamp :resource_type :resource_title :property]))))
```
wkalt: were you able to nail down why this is needed?

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -287,9 +282,8 @@
 (defn anonymize-containment-path
   "Anonymize a collection of containment path resource references from an event"
   [path context config]
-  {:pre  [(coll? path)]
-   :post [(coll? %)]}
-  (map #(anonymize-containment-path-element % context config) path))
+  (some->> path
+           (map #(anonymize-containment-path-element % context config))))
```
wkalt: I guess we would have barfed on nil containment paths before

wkalt:
left a few minor comments. looks good to me otherwise and seems to work fine. The last little thing would be to rename the branch to conform to our usual scheme.

src/puppetlabs/puppetdb/reports.clj
ajroetker: Oh I validated it wasn't needed in anonymizer, I'll remove it here as well since we don't need it here either

ajroetker:
Opened with a different branch name

PR #1661, opened 2015-09-29
(PDB-2004) Alter tests to fix Travis jdk_switcher

rbrw:
Let Travis handle the jdk selection.  Right now jdk_switcher doesn't
actually exist.  This means Travis generates the test matrix more
implicitly, so use exclusions to make sure we only test ruby on one
db/jdk combination.

Add "set -ueo pipefail" to catch errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,31 @@
 language: clojure
+
 lein: lein2
+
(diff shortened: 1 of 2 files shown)


rbrw:

PR #1662, opened 2015-09-29
(PDB-1948) Use wire-v6 of reports in our testutils

ajroetker:


ajroetker:
Closed in favor of a different PR

PR #1665, opened 2015-09-30
(PDB-2006) Move the facts dual-db test to postgres

rbrw:


rbrw:
This will be WIP until https://github.com/puppetlabs/puppetdb/pull/1663 makes it into master.

ajroetker:
I think we should also move sync to use the second *sync* PuppetDB to use this configuration as well. That way they don't diverge.

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -581,73 +575,74 @@
...
+         (jdbc/with-transacted-connection write-db
+           (scf-store/add-certname! "foo1")
+           (scf-store/add-facts! {:certname "foo1"
+                                  :values facts1
+                                  :timestamp (now)
+                                  :environment "DEV"
+                                  :producer_timestamp (now)}))
+
+         (testing "queries only use the read database"
+           (let [request (get-request endpoint (json/parse-string nil))
```
wkalt: prexisting, but I think these should all work fine with a single endpoint argument

ajroetker:
I'm +1 here, but this is gonna break sync tests so I think we're gonna need to have a patch up there as well before this can get merged.

kbarber:
@rbrw this might now need to be rebased against stable? Since master is now 4.0.0.

rbrw:
@kbarber I think this should be up to date

PR #1666, opened 2015-09-30
(PDB-2003) facts GC overwhelmed with very large updates

wkalt:
This changes our facts GC to do pinpoint deletion in chunks of 3000 elements.
The chunking shouldn't come into play except for in pretty unusual
circumstances, but without it jdbc can truncate the message to the database.

### DIFF
--- a/src/com/puppetlabs/puppetdb/cli/services.clj
+++ b/src/com/puppetlabs/puppetdb/cli/services.clj
@@ -319,7 +319,7 @@
                                              command-mq-connection-string
                                              mq-endpoint
                                              discard-dir
-                                             {:db                     write-db
+                                             {:db write-db
                                               :catalog-hash-debug-dir (:catalog-hash-debug-dir global)})
                                            error-shutdown!)))))
           context (assoc context :command-procs command-procs)
--- a/src/com/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/com/puppetlabs/puppetdb/scf/storage.clj
@@ -55,6 +55,12 @@
 
 (def json-primitive-schema (s/either String Number Boolean))
 
+;; the maximum number of parameters pl-jdbc will admit in a prepared statement
+;; is 32767. delete-pending-value-id-orphans will create a prepared statement
+;; with 5 times the number of invalidated values, so 6000 here keeps us under
+;; that and leaves some room.
+(def gc-chunksize 6000)
+
 (def resource-schema
   (merge resource-ref-schema
          {(s/optional-key :exported) Boolean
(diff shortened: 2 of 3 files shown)


src/com/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -297,6 +298,7 @@
 
     ;; Initialize database-dependent metrics and dlo metrics if existent.
     (pop/initialize-metrics write-db)
+    (alter-var-root (var scf-store/*gc-chunksize*) (constantly gc-chunksize))
```
wkalt: this is stupid; planning to change it. I ran into some really strange issues with call counting in command testing when I tried to thread this through as an argument. I'll pick this up later and figure out what the problem was.

wkalt:
@senior after talking with russ, I removed the config option to keep things simple. The place in jdbc where this is breaking is here:
https://github.com/pgjdbc/pgjdbc/blob/dc0e127c5e986645c3efb879271ccf0d103bc7ad/org/postgresql/core/PGStream.java#L209

the maximum number of prepared statement params is Short.MAX_VALUE, which is 32767. delete-pending-value-id-orphans will generate a statement with 5x the number of invalidated values, so having the chunksize of 3000 should leave us plenty of room.

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -758,42 +764,51 @@
...
-             AND NOT EXISTS (SELECT 1 FROM facts f
-                               WHERE f.fact_path_id %s
-                                 AND f.fact_path_id = fp.id
-                                 AND f.factset_id <> ?)"
-        in-pids in-pids)
-       (concat dropped-pids dropped-pids [factset-id])))))
+    (let [chunked-in-pids (map jdbc/in-clause (partition-all
+                                                gc-chunksize dropped-pids))
+          chunked-dropped-pids (partition-all gc-chunksize dropped-pids)]
+      (doseq [[in-pids-chunk dropped-pids-chunk] (map vector chunked-in-pids
```
rbrw: Might rework these doseqs as parallel map calls, i.e. (map ... in dropped).

src/com/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -758,42 +764,57 @@
...
-                                 AND (f.factset_id,
-                                      f.fact_path_id,
-                                      f.fact_value_id) NOT %s)"
-        in-vids in-vids in-rm-facts)
-       (flatten [vids vids rm-facts])))))
+    (let [vid-chunks (partition-all gc-chunksise
+                                    (map second removed-pid-vid-pairs))
+          removed-fact-chunks (->> removed-pid-vid-pairs
+                                   (map cons (repeat factset-id))
+                                   (partition-all gc-chunksize))
```
rbrw: Not sure, but I wondered about something like this here:
```clojure
(partition-all gc-chunksize
               (for [[p v] removed-pid-vid-pairs]
                 [factset-id p v]))
```

PR #1668, opened 2015-10-05
(PDB-1614) anonymize path and value independently

wkalt:
This changes our anonymization code so that structured facts like
{:foo {:bar "baz"}} and {:foo {:bar "biz"}} will be identical when anonymized
except for the element that differs. This also changes our handling of arrays so
[1 2 3] and [1 2 4] will only differ in the last element when anonymized.
Finally, this bumps our use of (rand-int 300) to (rand-int (min value 20)),
which should ensure that the integers we get are sufficiently diverse.

### DIFF
--- a/src/puppetlabs/puppetdb/anonymizer.clj
+++ b/src/puppetlabs/puppetdb/anonymizer.clj
@@ -115,11 +115,9 @@
   [value]
   (cond
    (string? value) (random-string 30)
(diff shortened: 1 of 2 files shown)


ajroetker:
Looks good code-wise, I'll test this out.

ajroetker:
This worked for me as advertised.

PR #1672, opened 2015-10-06
(PDB-841) Retire the legacy storeconfigs tool

ajroetker:
This commit removes support for the legacy storeconfigs export tool as
we no longer support Puppet <4 in PuppetDB.


acceptance/tests/storeconfigs/basic_collection.rb
```diff
@@ -1,47 +0,0 @@
-test_name "general collection should get all exported resources except the host's" do
```
kbarber: Why is this test being removed?

ajroetker: I mistakenly removed the whole directory `storeconfigs`

acceptance/tests/storeconfigs/collections_with_queries.rb
```diff
@@ -1,120 +0,0 @@
-test_name "collections with queries" do
```
kbarber: Same again, not sure why this is being removed? So the ticket was to remove the _legacy_ storeconfigs export tool, these tests are still relevant without the tool right? That is, we still have storeconfigs without the legacy export mechanism don't we?

acceptance/tests/storeconfigs/legacy_storeconfigs_import_export.rb
```diff
@@ -1,75 +0,0 @@
-puppetdb_query_url = "http://localhost:8080/pdb/query"
```
kbarber: Now this test makes sense to remove, I'm not sure anything else in this directory does though ...

acceptance/tests/storeconfigs/collections_with_queries.rb
ajroetker: Yep, responded on the other comment, removed the whole directory by mistake.

kbarber:
@ajroetker looking good, I think we just have removed too many tests :-).

ajroetker:
transient travis failure ^

PR #1674, opened 2015-10-07
(PDB-1931) Retire catalog-hash-conflict-debugging

ajroetker:
This commit retires catalog-hash-conflict-debugging as a
service/configuration item.


documentation/trouble_low_catalog_duplication.markdown
```diff
@@ -1,172 +0,0 @@
...
-resources or properties. One example of that scenario is in a two
-puppet master setup, one puppet master would send its resources in a
-different order than the other puppet master. This would cause the
-catalogs to hash differently, even though their content was the same.
-This issue has been fixed in PuppetDB 1.6.
-
-How Can I Detect the Cause of Catalogs Hashing Differently?
------
-
-Unfortunately, diagnosing why a catalog is hashing differently can be
```
wkalt: I'd be nice if we could give some kind of SQL query that would be helpful in this circumstance. Down to this point the doc is still relevant I'd say. I'll give it some thought.

senior: The user would need more than a SQL query, there's really three pieces to this. The incoming catalog, the stored catalog and the difference between the two. Getting the hash from the stored catalog is easy. Computing the hash for the new catalog is not (i.e. the user would need to invoke one of our functions). The comparison is probably the most difficult aspect of this as we have to hydrate the catalog into the right structures and format the new catalog in the same way. We did this by munging/sorting both sides and spitting to a file. From there the user needed to use diff. Even then it would probably be some help, but could still be difficult to trace down.

I don't think this kind of debugging info is needed anymore. I have not heard of a user having an issue related to de-duplication rate since before 1.6.0. Back then we'd always flush the whole catalog/factset and reinsert everything fresh. Now we're much more targeted, which I think is why the user probably wouldn't even notice a lower duplication rate. I think the users that would, the change would be pretty obvious. You'd have to have a large number of resources changing each time. Compare that with what we had before, which any change (or even accidental change in order of resources) would trigger big I/O on the DB. So the differences now would be much more obvious, before were possibly not even related to a user action.

senior:
:+1:

PR #1678, opened 2015-10-07
(PDB-1459) Support extract with no subexpression

ajroetker:
This commit adds support for extracts with no subexpression.

### DIFF
--- a/src/puppetlabs/puppetdb/http/query.clj
+++ b/src/puppetlabs/puppetdb/http/query.clj
@@ -82,6 +82,9 @@
     [["extract" columns nil]]
     ["extract" columns crit]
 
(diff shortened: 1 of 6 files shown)


kbarber:
@ajroetker do we have any tests for this? I see we cleaned up a test, but there are no new tests, or am I missing something?

ajroetker:
@kbarber added some tests for ya ;)

wkalt:
this needs docs updates

wkalt:
also shouldn't this work?
```
curl -X GET http://localhost:8080/pdb/query/v4/facts -d 'query=["extract","certname"]'
```

ajroetker:
@wkalt should be all good now!

PR #1679, opened 2015-10-07
(PDB-2039) WIP - Proof of concept code on securing at the node level

senior:
This commit hacks in the ssl client cert to the JMS message header to be
validated when the command is pulled off the queue.

There's still plenty of work left. Currently this is just a hack. There
needs to be a config option to turn this on. It needs to be added to
each command type (currently it's only on catalogs) and we also need to
have a cert whitelist check. There's probably a bunch of broken tests too.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -97,11 +97,10 @@
           (number? (:version %))
           (map? (:annotations %))]}
   (let [message     (json/parse-string body true)
-        received    (get headers :received (kitchensink/timestamp))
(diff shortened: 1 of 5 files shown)


ajroetker:
I started working on this but it become clear that if we are going to do this for every command that we should probably get some consistency around the code in `command.clj`, I submitted https://github.com/puppetlabs/puppetdb/pull/1681 for this reason.

kbarber:
Closing, looks like AJ is evolving this now: https://tickets.puppetlabs.com/browse/PDB-2039

PR #1682, opened 2015-10-12
(pdb-2034) schema error on replace-facts with escape-quoted strings

wkalt:
Without this commit we're broken on escape-quoted fact names and fact names
containing #~. Some escape-quoting we do before storage was causing different
names to come out of realize-paths! than went in, which caused an empty key
lookup that led to the error.

### DIFF
--- a/src/puppetlabs/puppetdb/facts.clj
+++ b/src/puppetlabs/puppetdb/facts.clj
@@ -69,7 +69,7 @@
 (defn unescape-string
   "Strip escaped quotes from a string."
   [^String s]
-  (if (= \" (.charAt s 0))
+  (if (= \" (first s) (last s))
     (subs s 1 (dec (.length s)))
     s))
 
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -958,6 +958,7 @@
   (if-let [pathstrs (seq pathstrs)]
     (let [existing-path-ids (existing-row-ids :fact_paths "path" pathstrs identity
                                               String "text")
+
           missing-db-paths (set/difference (set pathstrs)
                                            (set (keys existing-path-ids)))]
       (merge existing-path-ids
@@ -1022,7 +1023,8 @@
       (insert-facts-pv-pairs! (certname-to-factset-id certname)
                               (map #(vector (get paths-to-ids %1)
                                             (get vhashes-to-ids %2))
-                                   pathstrs vhashes))))))
+                                   (map facts/unescape-string pathstrs)
+                                   vhashes))))))
 
 (defn-validated update-facts!
   "Given a certname, querys the DB for existing facts for that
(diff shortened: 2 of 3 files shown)


src/puppetlabs/puppetdb/facts.clj
```diff
@@ -69,7 +69,7 @@
 (defn unescape-string
   "Strip escaped quotes from a string."
   [^String s]
-  (if (= \" (.charAt s 0))
+  (if (= \" (first s) (last s))
```
ajroetker: So this will only escape strings like `"\"foo\""` now instead of `"\"foo"` as well?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -958,6 +958,7 @@
   (if-let [pathstrs (seq pathstrs)]
     (let [existing-path-ids (existing-row-ids :fact_paths "path" pathstrs identity
                                               String "text")
+
```
ajroetker: extra line here

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1022,7 +1023,8 @@
       (insert-facts-pv-pairs! (certname-to-factset-id certname)
                               (map #(vector (get paths-to-ids %1)
                                             (get vhashes-to-ids %2))
-                                   pathstrs vhashes))))))
+                                   (map facts/unescape-string pathstrs)
```
ajroetker: I think this could probably just go into the let

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1051,7 +1053,8 @@
          ;; Add new facts and remove obsolete facts.
          replacement-pv-pairs (set (map #(vector (paths-to-ids %1)
                                                  (vhashes-to-ids %2))
-                                        pathstrs vhashes))
+                                        (map facts/unescape-string pathstrs)
```
ajroetker: same here

src/puppetlabs/puppetdb/scf/storage.clj
ajroetker: So we don't want `realize-paths` to also have the escaped quotes stripped?

src/puppetlabs/puppetdb/facts.clj
wkalt: yes -- before what we were actually escaping \"foo to fo

wkalt: actually I've screwed something up here

ajroetker: Do we not want to strip escaped quotes on a single side as well? "\"foo" -> "foo"

wkalt: no, I think that should behave the same as foo\"bar.

I'm not honestly sure why we're escaping quotes to begin with, and I wonder if it could be hsql related, but we can't change that behavior now.

wkalt:
I'll reopen this on stable

PR #1684, opened 2015-10-13
(PDB-1924) Implicit subqueries

kbarber:


wkalt:
reviewing

documentation/api/query/v4/nodes.markdown
```diff
@@ -59,6 +68,19 @@ The below fields are allowed as filter criteria and are returned in all response
...
+more information consult the documentation for [subqueries].
+
+* [`factsets`][factsets]: Factsets received from this node.
+* [`reports`][reports]: Reports received from this node.
+* [`catalogs`][catalogs]: Catalogs received for this node.
+* [`facts`][facts]: Fact names and values received from this node.
+* [`fact_contents`][fact-contents]: Factset paths and values received from this node.
+* [`events`][events]: Report events triggered from this node.
+* [`edges`][edges]: Catalog edges received for this node.
+* [`resources`][resources]: Catalog resources received for this node.
```
wkalt: I find the "this node" phrasing a bit confusing, since the subquery is actually constraining the list of nodes returned via some arbitrary field on the target table. I'm not sure what the right thing to say is

```
[resources][resources]: Constrain result to nodes matching a resource query
```
kind of clunky -- any thoughts?

wkalt: same applies to some of the other doc updates btw

documentation/api/query/v4/operators.markdown
```diff
@@ -180,8 +180,72 @@ Subqueries allow you to correlate data from multiple sources or multiple
...
+* [`facts`][facts]
+* [`fact_contents`][fact-contents]
+* [`fact_paths`][fact-paths]
+* [`nodes`][nodes]
+* [`reports`][reports]
+* [`resources`][resources]
+
+Internal to PuppetDB, we keep a mapping of how different entities relate to each
+other, and so no other data beyond the entity is needed in this case. This is
+different to explicit subqueries, for those you must specify yourself how
```
wkalt: s/to/than

wkalt:
a couple minor notes, but I'm +1.

documentation/api/query/v4/operators.markdown
kbarber: 'to', 'than' and 'from' are all valid forms of english, even in the United States. Seems like 'from' is a more acceptable international choice, so I'll switch it to that.

http://www.oxforddictionaries.com/words/different-from-than-or-to

documentation/api/query/v4/nodes.markdown
kbarber: Okay, I've got an idea which is different again, let me tap it up.

PR #1687, opened 2015-10-14
(pdb-2034) schema error on replace-facts with escape-quoted strings

wkalt:
Without this commit we're broken on escape-quoted fact names and fact names
containing #~. Some escape-quoting we do before storage was causing different
names to come out of realize-paths! than went in, which caused an empty key
lookup that led to the error.


ajroetker:
Solves the original issue for me, +1 codewise

PR #1688, opened 2015-10-15
(PDB-1014) Make factsets hash field not nullable

senior:


senior:
@kbarber is this what you had in mind for [PDB-1014](https://tickets.puppetlabs.com/browse/PDB-1014)? It seems like this could have been done when it was first introduced, so I'm not entirely sure what you had in mind.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1465,6 +1465,12 @@
       (format "ALTER TABLE reports ADD COLUMN logs %s DEFAULT NULL" jsonb-type)
       (format "ALTER TABLE reports ADD COLUMN resources %s DEFAULT NULL" jsonb-type))))
 
+(defn factset-hash-field-not-nullable
+  []
+  (jdbc/do-commands
+   "UPDATE factsets SET hash='' WHERE hash is NULL"
```
wkalt: is this going to do anything weird with the existing unique constraint on hash?

wkalt: yeah, if you have null-valued hashes this migration causes a constraint violation.

wkalt: just to echo what I said in the channel the other day, I think a simple way to make this work would be to set the hash to the stringified factset id.

kbarber: So I'm trying to figure out the idea here, behind why the ticket was raised.

I can't help but think that setting this to anything other than a real hash will cause problems with HA for older data that doesn't have one already at migration time. I mean, if you send this content over via HA, won't it generate a different hash for comparison? /cc @mullr

wkalt: I don't think we're using hashes for that, otherwise we'd have problems when we change hash fields too.

senior: I think we'd be fine with a pk as a string for that value. It'll be updated on next facset anyway. Why I am asking is we could have always done that, there's nothing 4.0.0 specific about that.

kbarber: I'm pretty sure our thinking was to generate a hash later on, due to HA concerns. If this is not a thing, then yeah it didn't/doesn't matter.

senior:
@wkalt fixed

wkalt:
looks good to me, just waiting for tests

PR #1689, opened 2015-10-15
(PDB-1996) Make query-recs defs rather than defns

ajroetker:
This commit makes our query-recs defs rather than defns, now that hsqldb
has been removed.

### DIFF
--- a/test/puppetlabs/puppetdb/query_eng_test.clj
+++ b/test/puppetlabs/puppetdb/query_eng_test.clj
@@ -2,7 +2,7 @@
   (:require [clojure.test :refer :all]
             [puppetlabs.puppetdb.scf.storage :as scf-store]
             [puppetlabs.puppetdb.query-eng.engine :refer :all]
-            [puppetlabs.puppetdb.query-eng :refer [entity-fn-idx assoc-in-idx!]]
+            [puppetlabs.puppetdb.query-eng :refer [entity-fn-idx]]
             [clj-time.core :refer [now]]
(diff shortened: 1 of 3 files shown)


test/puppetlabs/puppetdb/query_eng_test.clj
```diff
@@ -95,82 +94,3 @@
...
-        (assoc-in-idx! [:fact-names :munge] (fn [_ _] identity))
-
-        (assoc-in-idx! [:fact-names :rec :projections "depth"] {:type :integer
-                                                                :queryable? true
-                                                                :field :depth})
-        (let [request (get-request endpoint)
-              {:keys [status body]} (fixt/*app* request)
-              result (vec (parse-result body))]
-          (is (= status http/status-ok))
-          (is (= result (map #(hash-map :name % :depth 0) expected-result)))))
```
wkalt: I assume you're moving to inline swaps by deleting assoc-in-idx!, but this still seems like something we should be testing

ajroetker: So I went back and forth on this, we would really just be testing that the data structure is an atom.

wkalt: we'd be testing the whole pipeline like we do here -- at the moment that seems like a test that it's an atom, but testing it insulates us against future changes

ajroetker: We also have PE tests for this, so were not completely uninsulated. I went ahead and added the text back either way.

PR #1690, opened 2015-10-16
(PDB-2056) Publicize testutils.logging call-with-*

rbrw:
Add a call-with-log-level to parallel with-log-level, and make the
call-with-* functions that support the public with-* macros public so
that people can use them when they're more suitable.

Fix up the docstring and change "body" to "f" in the call-with-*
functions.


kbarber:
@rbrw was this intended for 4.0.0? The ticket has no fix version so I can't tell.

rbrw:
Closing because the changes we're making for TK have already superseded these.

PR #1691, opened 2015-10-16
(PDB-2007) array params with undef don't sort

wkalt:
This fixes a discrepancy between JVM puppetserver and Puppet with respect to our
terminus, where manifests like

@@notify { "test":
    tag => [undef, "a", "b"],
}

would fail due to inability to compare symbols and strings. With this patch, we
sort by string representation rather than sorting the array directly.

### DIFF
--- a/puppet/lib/puppet/indirector/catalog/puppetdb.rb
+++ b/puppet/lib/puppet/indirector/catalog/puppetdb.rb
@@ -214,7 +214,7 @@ class Puppet::Resource::Catalog::Puppetdb < Puppet::Indirector::REST
         params = resource['parameters']
         UnorderedMetaparams.each do |metaparam|
           if params[metaparam].kind_of? Array then
-            values = params[metaparam].sort
+            values = params[metaparam].sort_by {|x| x.to_s}
             params[metaparam] = values unless values.empty?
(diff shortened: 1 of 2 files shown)


puppet/lib/puppet/indirector/catalog/puppetdb.rb
```diff
@@ -214,7 +214,7 @@ def sort_unordered_metaparams(hash)
         params = resource['parameters']
         UnorderedMetaparams.each do |metaparam|
           if params[metaparam].kind_of? Array then
-            values = params[metaparam].sort
+            values = params[metaparam].sort_by {|x| x.to_s}
```
ajroetker: why do we sort at all?

wkalt: probably for hashing reasons

ajroetker:
lgtm +1

puppet/lib/puppet/indirector/catalog/puppetdb.rb
wkalt: discussed in channel, but it's needed for the checksum computation we do before command submission

PR #1692, opened 2015-10-19
(PDB-2049) Allow aggregates to omit subexpression

ajroetker:
This commit fixes a bug in `engine.clj` where the clauses for dealing
with `extract` were misordered so we had incorrect behavior when
handeling aggregates with a group-by clause without a subexpression.

This commit also removes some documentation claiming that a user must
supply a subexpression.


wkalt:
+1

PR #1695, opened 2015-10-20
(PDB-2064) stop converting value equality queries to hash comparisons

wkalt:
This changes our fact queries to use the proper string or boolean column
instead of converting equality queries to a comparison on value_hash.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -211,6 +211,14 @@
                                             :query-only? true
                                             :queryable? false
                                             :field :fv.value_float}
+                             "value_string" {:type :string
+                                            :query-only? true
+                                            :queryable? false
+                                            :field :fv.value_string}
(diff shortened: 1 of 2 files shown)


wkalt:
there should be an evident difference with and without this commit if merged on top of 2a18d97553a5f6b4426955b860b1a3228d6b3536 for comparison with @kbrezina's dataset.

For this query: https://gist.github.com/9106de649fb52b4d4957

I see 100 seconds elapsed before and 39 seconds after, and when "metrics" is removed from the extract, 63 seconds before and 15 seconds after. I'm optimistic that we can cut it down a bit more, but I'll have to keep looking at it.

wkalt:
note that that applies to the full unpaginated result set -- paginating by 50 seems pretty quick at this point. I'm not sure if there's still an issue here or not.

wkalt:
this has a little bug with nodes queries of form ["=", ["fact",<name>], <boolean>]. Don't merge without a fix.

wkalt:
note above has been fixed, WIP removed

PR #1696, opened 2015-10-20
(PDB-1797) testing support for jessie

wkalt:
This sets up the config for acceptance testing on debian eight.


wkalt:
reopening on stable

PR #1702, opened 2015-10-21
(PDB-2096) Add not-found route to top-level `/pdb` route

ajroetker:
This commit adds the `compojure.route/not-found` route at the top-level
of the PuppetDB routes so that new PuppetDB routes won't need to add the
not-found route. This also fixes an error where if a user hit any
non-existent route under `/pdb`, PDB would throw an exception related to
our middleware not playing nicely with not-founds which produced a
confusing `nil...status` error.

### DIFF
--- a/src/puppetlabs/puppetdb/admin.clj
+++ b/src/puppetlabs/puppetdb/admin.clj
@@ -1,6 +1,5 @@
 (ns puppetlabs.puppetdb.admin
   (:require [compojure.core :as compojure]
-            [compojure.route :as route]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.export :as export]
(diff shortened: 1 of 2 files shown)


ajroetker:
Needs https://github.com/puppetlabs/pe-puppetdb-extensions/pull/125 in extensions to be merged as well.

ajroetker:
@wkalt @senior Think this should fix the problems john and jon saw as well as fixing the preexisting not-found issue.

kbarber:
Should we be targeting this for the 3.2.0 release? ie. stable?

kbarber:
@ajroetker oh yeah, I'm pretty sure in your comments on PDB-2096 you wanted this in stable not master ... rebase/re-open when you can mate.

ajroetker:
@kbarber oops yep thanks dude!

PR #1704, opened 2015-10-22
(PDB-2096) Add not-found route to top-level `/pdb` route

ajroetker:
This commit adds the `compojure.route/not-found` route at the top-level
of the PuppetDB routes so that new PuppetDB routes won't need to add the
not-found route. This also fixes an error where if a user hit any
non-existent route under `/pdb`, PDB would throw an exception related to
our middleware not playing nicely with not-founds which produced a
confusing `nil...status` error.

### DIFF
--- a/src/puppetlabs/puppetdb/pdb_routing.clj
+++ b/src/puppetlabs/puppetdb/pdb_routing.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.puppetdb.pdb-routing
   (:require [puppetlabs.trapperkeeper.core :as tk]
             [compojure.core :as compojure]
+            [compojure.route :as route]
             [puppetlabs.trapperkeeper.services :as tksvc]
             [ring.middleware.resource :refer [resource-request]]
             [ring.util.request :as rreq]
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/pdb_routing.clj
```diff
@@ -40,7 +41,8 @@
            "/cmd" (cmd/command-app get-shared-globals
                                    enqueue-raw-command-fn get-response-pub)
            "/query" (server/build-app get-shared-globals)
-           "/admin" (admin/build-app enqueue-command-fn query-fn)]))))
+           "/admin" (admin/build-app enqueue-command-fn query-fn)
+           (route/not-found "Not Found")]))))
```
mullr: seems like we have a separate top-level route thing for PE. Does this need to be done over there as well?

wkalt:
How much work would it be to get consistent behavior when people try to use the old urls too?

```
curl -X GET http://localhost:8080/v4/facts
```

ajroetker:
That would be a pretty big chunk of work, right now we only have control over "/pdb" and sometimes "/" now @wkalt, but the error messages you'd get with a non-existent top level url i.e. /v4 /rbac or whatever is consistent across all our PL products @wkalt, it's whatever the jetty default is.

mullr:
Tested this locally, 404s seem to work fine.

src/puppetlabs/puppetdb/pdb_routing.clj
ajroetker: Already done

PR #1705, opened 2015-10-22
(PDB-2070) Add cmd broadcast, min_successful_writes to terminus

mullr:
This includes a sizable refactor of the http module, may be best to review the commits.

### DIFF
--- a/documentation/puppetdb_connection.markdown
+++ b/documentation/puppetdb_connection.markdown
@@ -48,9 +48,27 @@ You can use a comma separated list of URLs if there are multiple PuppetDB instan
 
     server_urls = https://puppetdb1.example.com:8081,https://puppetdb2.example.com:8081
 
-The PuppetDB terminus will always attempt to connect to the first PuppetDB instance specified (listed above as puppetdb1). If a server-side exception occurs, or the request takes too long (see [`server_url_timeout`](#server_url_timeout)), the PuppetDB terminus will attempt the same operation on the next instance in the list.
+The default value is `https://puppetdb:8081`.
 
-The default value is https://puppetdb:8081
+For read operations, the PuppetDB terminus will always attempt to connect to the first PuppetDB instance specified (listed above as puppetdb1). If a server-side exception occurs, or the request takes too long (see [`server_url_timeout`](#server_url_timeout)), the PuppetDB terminus will attempt the same operation on the next instance in the list.
+
+For write operations, all commands are sent to all PuppetDB instances in the list. 
+
+Configurations with multiple `server_urls`, along with the `min_successful_submissions` and `submit_only_server_urls` settings, can be used with the Puppet Enterprise version of PuppetDB to set up a high-availability PuppetDB deployment. Please note that, absent the synchronization support in Puppet Enterprise, multiple PuppetDB instances can easily diverge from each other due to normal transient network issues. We do not recommend using these settings outside of a supported high-availability deployment. `
+
+### `min_successful_submissions`
+
+When writing data (submitting commands) to PuppetDB, this is the minimum number of machines to which the command must be successfully sent to consider the write successful. The default value is 1, which should be appropriate for most single- or dual-PuppetDB deployments. You may wish to change this value if you have particularly high data durability requirements. For example, if you want to ensure that you can withstand the loss of a node even when running PuppetDB in a degraded state, you could deploy a cluster of a 3 PuppetDB instances and set `min_successful_submissions` to 2. 
+
+### `submit_only_server_urls`
+
+This setting allows you specify PuppetDB instances to which commands should be sent, but which shouldn't ever be queried for data needed during a Puppet run. It uses the same format as `server_urls`. For example:
+
+    submit_only_server_urls = https://puppetdb-submit-only.example.com:8081
+
+If a server is listed in `submit_only_server_urls`, it shouldn't be listed in `server_urls`; the two lists should be disjoint.
+
+Successful command submission to the PuppetDB instances in this list *do* count towards the `min_successful_submissions` setting, so you should consider incrementing that if you use this setting.
 
 ### `server_url_timeout`
 
--- a/documentation/puppetdb_connection.markdown
+++ b/documentation/puppetdb_connection.markdown
@@ -8,6 +8,7 @@ canonical: "/puppetdb/latest/puppetdb_connection.html"
 [connect_to_puppetdb]: ./connect_puppet_master.html
 [confdir]: /puppet/latest/reference/dirs_confdir.html
 [puppetdb_conf]: ./connect_puppet_master.html#edit-puppetdb\.conf
+[ha_guide]: ./ha_guide.html
 
 The `puppetdb.conf` file contains the hostname and port of the [PuppetDB][puppetdb_root] server. It is only used if you are using PuppetDB and have [connected your Puppet master to it][connect_to_puppetdb].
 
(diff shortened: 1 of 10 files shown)


ajroetker:
There should also be a ticket for adding these settings to the PDB module as well, I think we're already missing the ability to configure multiple server_urls there.

ajroetker:
I'm not really a fan of the `min_successful_writes` setting, it implies that the writes will indeed be successful in the command submits successfully.

mullr:
@ajroetker This is a write in the 'data is sitting on a disk' sense. It makes sense when you're thinking about it from a data durability perspective. At worst, even if the command invalid, they're going to end up in a DLO somewhere. That still counts to me. 

Which is not to say that I don't want a better name, since I always want better names for everything. Suggestions are welcome.

mullr:
Channel traffic says:
- min_successful_writes => min_successful_submissions
- write_only_server_urls => submit_only_server_urls
- Fix the puppetdb "failed to submit command" error to say "failed to process command", which is what it really means

puppet/lib/puppet/util/puppetdb/config.rb
```diff
@@ -14,6 +14,8 @@ def self.load(config_file = nil)
         :soft_write_failure        => false,
```
wkalt: seems like min_successful_writes and soft_write_failure should be mutually exclusive with some sort of config error

documentation/puppetdb_connection.markdown
```diff
@@ -48,9 +48,25 @@ You can use a comma separated list of URLs if there are multiple PuppetDB instan
...
+
+Configurations with multiple `server_urls`, along with the `min_successful_writes` and `write_only_server_urls` settings, can be used with the Puppet Enterprise version of PuppetDB to set up a high-availability PuppetDB deployment. Please note that, absent the synchronization support in Puppet Enterprise, multiple PuppetDB instances can easily diverge from each other due to normal transient network issues. We do not recommend using these settings outside of a supported high-availability deployment. `
+
+### `min_successful_writes`
+
+When writing data (submitting commands) to PuppetDB, this is the minimum number of machines to which the command must be successfully sent to consider the write successful. The default value is 1, which should be appropriate for most single- or dual-PuppetDB deployments. You may wish to change this value if you have particularly high data durability requirements. For example, if you want to ensure that you can withstand the loss of a node even when running PuppetDB in a degraded state, you could deploy a cluster of a 3 PuppetDB instances and set `min_successful_writes` to 2. 
+
+### `write_only_server_urls`
```
wkalt: does this do anything on foss?

kbarber:
@dalen this might be interesting to you.

documentation/puppetdb_connection.markdown
mullr: Yes, it does exactly what it says in both foss and PE. Even in foss, you can use it to set up a second server and send all your data to it for testing.

puppet/lib/puppet/util/puppetdb/config.rb
mullr: I think soft_write_failure is the same as min_successful_writes=0, isn't it? I'll just make them use the same mechanism and forbid configuring them together.

mullr: Ah, it's actually subtly different. I'll leave the code in and just change the config code.

mullr:
@ajroetker @wkalt renamed things and made min_successful_submissions exclusive with soft_write_failure

mullr:
@ajroetker @wkalt I added a commit for query-side sticky failover as well, since it wasn't that much more. I'll stop now!

puppet/lib/puppet/util/puppetdb/http.rb
```diff
@@ -26,94 +29,200 @@ def self.concat_url_snippets(snippet1, snippet2)
...
-      server_url_config = config.server_url_config?
 
-      for url in Puppet::Util::Puppetdb.config.server_urls
-        begin
-          route = concat_url_snippets(url.request_uri, path_suffix)
-          http = Puppet::Network::HttpPool.http_instance(url.host, url.port)
-          request_timeout = config.server_url_timeout
+      last_good_index = @@last_good_query_server_url_index.deref()
+      server_count = config.server_urls.length
+      server_try_order = (0...server_count).map { |i| (i + last_good_index) % server_count }
```
wkalt: should this be server_count - 1?

mullr: `...` is `[start, end)`

wkalt: ah, missed the extra dot

mullr:
:arrow_up: looks like a transient...

```
	Failed to issue method call: Activation of org.freedesktop.hostname1 timed out>
```

wkalt:
This generally works well. The only issues I can see are
* we should validate that min_successful_submissions is less than or equal to the number of server urls
* the master log will report failure to submit to any hosts when a single host is down
* when there are fewer hosts up than min_successful_submissions, catalogs will fail to process but reports and factsets will succeed

mullr:
This generally works well. The only issues I can see are
* we should validate that min_successful_submissions is less than or equal to the number of server urls
 /fixed/

* the master log will report failure to submit to any hosts when a single host is down
/fixed/

* when there are fewer hosts up than min_successful_submissions, catalogs will fail to process but reports and factsets will succeed
/This is actually a puppet issue; going to file it over there/

mullr:
*when there are fewer hosts up than min_successful_submissions, catalogs will fail to process but reports and factsets will succeed*

Wyatt and I have talked about this at length; basically, it seems like puppet is doing the right thing, and that we should probably document our way out of these. We should talk about it as a team though, since the whole thing is fairly subtle.

puppet/lib/puppet/util/puppetdb/command.rb
```diff
@@ -58,7 +58,7 @@ def submit
 
     begin
       response = profile("Submit command HTTP post", [:puppetdb, :command, :submit]) do
-        Http.action("#{CommandsUrl}?checksum=#{checksum}") do |http_instance, path|
+        Http.action("#{CommandsUrl}?checksum=#{checksum}", :command) do |http_instance, path|
```
ajroetker: seems kind of odd that our url tells us whether the request is a `:command` or a `:query` and you have to specify it as well

puppet/lib/puppet/util/puppetdb/config.rb
```diff
@@ -51,6 +53,11 @@ def self.load(config_file = nil)
       main_section = result['main'] || {}
       # symbolize the keys
       main_section = main_section.inject({}) {|h, (k,v)| h[k.to_sym] = v ; h}
+
+      if main_section[:soft_write_failure] and main_section[:min_successful_submissions]
+        raise "soft_write_failure and min_successful_submissions cannot be configured together"
```
ajroetker: cannot both be `true` together. `false` is still configured right?

mullr: fixed

puppet/lib/puppet/util/puppetdb/command.rb
mullr: I'd really rather not parse the URL to figure out which kind of request it is.

mullr:
@wkalt @ajroetker Updated, have flags now for `sticky_read_failover` and `command_broadcast`

puppet/lib/puppet/util/puppetdb/command.rb
ajroetker: Why not construct the url-prefix based on what type of request it is?

mullr: I thought that was different for various queries...

documentation/puppetdb_connection.markdown
```diff
@@ -70,3 +81,32 @@ The default value is false.
 
 Hostname and port of the PuppetDB instance. `server_urls` takes precedence if both are defined.
 
+### High Availability configuration
+
+These settings are designed for use in a high-availability deployment of PuppetDB, available in Puppet Enterprise. While they may appear to work in other configurations, we discourage their use. In particular, absent the synchronization support in Puppet Enterprise, multiple PuppetDB instances can easily diverge from each other due to normal transient network issues. See the [PuppetDB HA Configuration Guide][ha_guide] for detailed configuration information.
```
senior: https://www.youtube.com/watch?v=E09LU6XVyxs maybe just s/we discourage their use/their use is discouraged/

documentation/puppetdb_connection.markdown
```diff
@@ -70,3 +81,32 @@ The default value is false.
...
+
+#### `sticky_read_failover`
+
+*PE-only*
+
+When using multiple `server_urls`, this flag can be set to `true` to cause queries to be made to the last PuppetDB instance that was successfully contacted. 
+
+The default value is `false`.
```
senior: Seems strange that this is defaulted to false, is that the case in PE?

mullr: We'll tell people to change it to true for HA... I don't think the defaults should be different for foss vs. PE.

puppet/spec/unit/util/puppetdb/http_spec.rb
```diff
@@ -19,120 +19,371 @@
...
+          response.message.should == 'OK'
+        end
       end
 
-      response.code.should == 200
-      response.message.should == 'OK'
     end
+
+
+
```
senior: Some extra newlines here

senior:
Thought through some issues with multiple JRuby instances and these new settings with @cprice404 and @mullr. There needs to be a a PR on the puppet-server side, but other than that I'm :+1:

mullr:
@cprice404 @senior addressed the shared state issues as discussed; see also https://github.com/puppetlabs/puppet-server/pull/755

mullr:
@senior all rebased up and comments addressed, I think. It'll still work in puppetserver w/o the new code there, it'll just print a warning and fall back to local storage.

puppet/lib/puppet/util/puppetdb/http.rb
```diff
@@ -3,14 +3,28 @@
...
+require 'puppet/util/puppetdb/atom'
 
 module Puppet::Util::Puppetdb
   class Http
+    SERVER_URL_FAIL_MSG = "Failing over to the next PuppetDB server_url in the 'server_urls' list"
 
-    SERVER_URL_FAIL_MSG = "Failing over to the next PuppetDB url in the 'server_urls' list"
+    if RUBY_PLATFORM == 'java'
+      begin
+        @@last_good_query_server_url_index = com::puppetlabs::puppetserver::PuppetDBTerminusState.getSharedStateAtom()
```
senior: so if you're running a PuppetDB with a Java puppet-server, but doesn't happen to include ^^^ it's going to throw a class not found sort of exception and assign a "local" Atom for that instance, is that how I'm reading this?

mullr: yes, that's what `NameError` is about.

PR #1706, opened 2015-10-22
(PDB-1964) don't swallow errors in applied-migrations

wkalt:
We need to swallow the error that occurs when the schema_migrations table
doesn't exist, because that means it's a fresh DB. Everything else should be thrown.


wkalt:
You can test this manually by observing the difference in the errors on this patch and the errors you see on master when you issue
```
alter table schema_migrations owner to foo;
```
and attempt to start puppetdb. I had trouble thinking of a way to test this without making some assumptions about the permissions of the puppetdb testing user that might not hold on our test boxes/personal machines. Suggestions welcome.

kbarber:
@wkalt Any reason this can't be in 3.2.0?

wkalt:
reopened on stable

PR #1707, opened 2015-10-23
(PDB-1964) don't swallow errors in applied-migrations

wkalt:
We need to swallow the error that occurs when the schema_migrations table
doesn't exist. Everything else should be thrown.


mullr:
:+1:

PR #1708, opened 2015-10-27
(maint) Retire the old server and port terminus settings

ajroetker:
This commit retires the deprecated server and port terminus settings
which were replaced by the server_urls setting.


kbarber:
@ajroetker couple more tests that might need changing: spec/unit/indirectory/catalog/puppetdb_spec.rb & spec/unit/indirectory/resource/puppetdb_spec.rb. These both stub :server and :port still.

kbarber:
Otherwise, I think I'm +1.

ajroetker:
Fixed those up @kbarber

PR #1711, opened 2015-10-28
(PDB-2078) Rethink migrate.clj

senior:
There are three commits in this PR. The first puts in place some test code that allows some basic comparison of the database schema. 

The second commit adds a new "rolled up" migration that includes migrations 1-28 (all existed prior to 3.0.0). This was obtained by running the first 28 migrations, using `pg_dump` to get the schema, then putting that in the migrate.clj (after cleaning it up a bit). There is also a test in that commit that ensures the new rolled up migration is identical (schema-wise) to the old set of migrations.

The final commit removes the old migrations (and the rolled up test) along with the associated migrate legacy code.


kbarber:
So I think we should be wary about what happens when you try to run this code from say, migration 24, when there is no path between 24 and 28. I would argue this should fail badly, with some meaningful thing about "upgrade to here, before using this code blah blah".

senior:
@kbarber I added some validation on upgrade path. It fails now if it finds an older migration than is supported.

kbarber:
Here is my naive diff, with & without this patch: https://gist.github.com/kbarber/2bff01c3888294b364c5

senior:
Is that good?

kbarber:
@senior yes, its good :-). All the differences are expected.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -79,868 +78,290 @@
...
-      REFERENCES resource_params_cache(resource) ON DELETE CASCADE"
-   "ALTER TABLE resource_params ADD FOREIGN KEY (resource)
-      REFERENCES resource_params_cache(resource) ON DELETE CASCADE"))
-
-(defn add-event-status-index
-  "Add an index to the `status` column of the event table."
-  []
-  (jdbc/do-commands
-   "CREATE INDEX idx_resource_events_status ON resource_events(status)"))
+(defn init-through-2-y-z []
```
mullr: Why not 3.0.0? We only support migrating from one version back, and master is going to be 4.x, right?

mullr:
lgtm, just the one question.

src/puppetlabs/puppetdb/scf/migrate.clj
senior: Are you just talking about the name of the function should be init-through-3-y-z? I collapsed all migrations that appear in a 2.y.z release and before. I did not collapse the migrations that appear in 3.0.0. Or are you saying that I should collapse the migrations in 3.0.0 and include them in this?

mullr: I think the migrations that go from 2.last to 3.0.0 should be included in this.

senior: Alright, I'll rework this

senior:
@mullr @kbarber whew, that was rough. We were pretty sloppy on our 3.0.0 migrations. They're rolled up in the latest push. I needed to add some better output to the testutils diffing code to help me keep tabs on what things I missed in the rollup. It's clean and should be the same (just like before) up to migration 34, which was the last migration in 3.0.0

mullr:
```ALTER TABLE ONLY edges
    ADD CONSTRAINT edges_certname_fkey FOREIGN KEY (certname) REFERENCES certnames(certname) ON DELETE CASCADE``` lost ``ON UPDATE NO ACTION```, but that seems like it's probably the default anyway. otherwise lgtm.

**Comment by senior** [comment]
@mullr right
```
NO ACTION
Produce an error indicating that the deletion or update would create a foreign key constraint violation. If the constraint is deferred, this error will be produced at constraint check time if there still exist any referencing rows. This is the default action
```

PR #1713, opened 2015-10-28
(docs) Change version numbers in docs pages/nav to 3.2

nfagerlund:


kbarber:
This failure ^^ is legit, but nothing to do with this patch. It was introduced by another patch, @wkalt is looking into the fix.

PR #1715, opened 2015-10-29
(PDB-2075) Speed unit tests via template copies

rbrw:


ext/travisci/test.sh
```diff
@@ -5,18 +5,61 @@ set -x
...
+  
+  export PDB_TEST_DB_HOST=127.0.0.1
+  export PDB_TEST_DB_PORT=5432
+  export PDB_TEST_DB_USER=puppetdb
+  export PDB_TEST_DB_ADMIN=pdb_test_admin
+  set +x
+  export PDB_TEST_DB_USER_PASSWORD="$passwd"
+  export PDB_TEST_DB_ADMIN_PASSWORD="$admin_passwd"
+  set -x
+  lein2 test :only puppetlabs.puppetdb.jdbc-test/query-to-vec
```
ajroetker: Fyi our jenkins unit tests will also need this change

rbrw:
For some of the changes it may be easier to consult the individual commits, and you'll almost certainly want to ignore whitespace: https://github.com/puppetlabs/puppetdb/pull/1715/files?w=1

mullr:
@rbrw Ok, I've been through the bulk of it, left notes on the commits. One thing I'm missing though is the actual use of the template... where is that? I'd expect to see a `create database with template` or something.

mullr:
I'm :+1: on this. Running locally I see:

```
Ran 382 tests containing 5480 assertions.
0 failures, 0 errors.
```

Does that seem right?

wkalt:
taking a look at this now

test-resources/bin/pdb-test-env
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
```
wkalt: It'd be nice to have a short comment at the top describing what this is for. Same with setup-pdb-pg.

test/puppetlabs/puppetdb/config_test.clj
```diff
@@ -92,8 +93,7 @@
 (deftest garbage-collection
   (let [config-with (fn [base-config]
                       (-> base-config
-                          (update :database merge
-                                  (first tu/available-postgres-configs))
+                          (update :database merge sample-db-config)
```
wkalt: swapping sample-db-config for available-postgres-configs works here, but available-postgres-configs is used in the sync tests so I don't think we'll be able to drop it as we have

wkalt:
This looks pretty good to me generally, and speeds the tests up by a factor of > 2. It breaks the extensions tests with the removal of available-postgres-configs; there might be more issues with those once we get past that one.

Also left a minor note about explanatory comments in the new shell scripts.

test/puppetlabs/puppetdb/config_test.clj
rbrw: Right, I'm working on a patch for that too.

test-resources/bin/pdb-test-env
rbrw: @wkalt ok, right - for now I just expanded the usage information.  Just shout if that doesn't address your concerns.

wkalt: :+1:

ext/travisci/test.sh
```diff
@@ -1,22 +1,25 @@
...
 ulimit -u 4096
 
+run-unit-tests()
+(
+  pgdir="$(pwd)/test-resources/var/pg"
+  readonly pgdir
+
+  export PGHOST=127.0.0.1
+  export PGPORT=5432
+  test-resources/bin/setup-pdb-pg "$pgdir"
```
kbarber: Minor comment: So the test-resources directory is really for clojure related stuff, any reason why these scripts don't just live somewhere in ext/travisci or something like that? At least then all the related items are together also.

rbrw:
I'm working on the related PR for extensions.

rbrw:
I just realized that I need to fix the doc strings for some of the new functions.  i.e. I wrote (defn foo [...] "bar" ...)

rbrw:
There were some rebase conflicts and other changes, so probably warrants another look, but if everything still seems OK (and the tests pass), I think this may be ready now.

wkalt:
@rbrw sorry for getting to this late -- needs a rebase though

PR #1718, opened 2015-10-29
add code coverage tool

cwcowellshah:
Puppet QA would like to measure what percent of lines and forms are covered by PuppetDB's unit tests. We will generate code coverage numbers using the third-party package "cloverage", which lives at https://github.com/lshift/cloverage. Cloverage is run via a leinegen plugin called "lein-cloverage".

This PR does two things:
* add lein-cloverage plugin to `project.clj`
* add 2 dependencies to dev profile in `project.clj` to avoid dependency clashes with cloverage

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -88,7 +88,9 @@
   :repositories [["releases" "http://nexus.delivery.puppetlabs.net/content/repositories/releases/"]
                  ["snapshots" "http://nexus.delivery.puppetlabs.net/content/repositories/snapshots/"]]
 
-  :plugins [[lein-release "1.0.5"]]
+  :plugins [[lein-release "1.0.5"]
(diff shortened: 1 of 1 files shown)


wkalt:
maybe this should go on stable?

cwcowellshah:
@wkalt Ideally both, I guess. Is it easier to first merge into stable and then merge up to master? Or should I issue a separate PR for stable? That would probably be just as easy.

wkalt:
@cwcowellshah if you reopen the PR on stable we'll merge it up to master after it's merged.

cwcowellshah:
Cloverage doesn't seem to work with the stable branch (I suspect there are dependency conflicts causing trouble), so let's just put it in the master branch and not worry about the stable branch for now. Maybe we can tackle it again in stable when we have more time. Ha.

PR #1722, opened 2015-10-30
(maint) Stop encoding json child data twice

ajroetker:
This commit removes the munge functions for catalogs, reports and
factsets that were parsing the json child data for those entities. This
commit adds a json encoder that can handle encoding the raw database
json objects so we don't encode the child data twice.


ajroetker:
Note that this removes our JSON pretty printing for catalogs/factsets/reports

ajroetker:
@senior @wkalt wanna take a quick glance and see if this is what y'all had in mind. I'll add documentation if it is.

src/puppetlabs/puppetdb/cheshire.clj
```diff
@@ -19,6 +19,12 @@
   "Non-memoize version of add-common-json-encoders!"
   []
   (generate/add-encoder
+   org.postgresql.util.PGobject
+   (fn [data jsonGenerator]
+     (if (.getPrettyPrinter jsonGenerator)
```
mullr: This needs a comment about why getPrettyPrinter may return null or not, and what that means. and a type annotation on jsonGenerator would be nice too.

src/puppetlabs/puppetdb/export.clj
```diff
@@ -2,6 +2,7 @@
   (:require [clj-time.core :refer [now]]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
+            [puppetlabs.puppetdb.scf.storage-utils :as su]
```
mullr: This is usually aliased as sutils

src/puppetlabs/puppetdb/http.clj
```diff
@@ -190,9 +190,11 @@
 (defn stream-json
   "Serializes the supplied sequence to `buffer`, which is a `Writer`
   object."
-  [coll buffer]
+  [jsonable buffer pretty-print]
```
mullr: jsonable sounds like a boolean param to me... I only know it's the thing to be converted because I see it used to be called `coll`. If your intent is that it's no longer just a collection, the docstring needs to be updated as well.

src/puppetlabs/puppetdb/http/catalogs.clj
```diff
@@ -5,6 +5,8 @@
             [puppetlabs.puppetdb.http.query :as http-q]
             [puppetlabs.puppetdb.http.edges :as edges]
             [puppetlabs.puppetdb.http.resources :as resources]
+            [puppetlabs.kitchensink.core :as kitchensink]
+            [puppetlabs.puppetdb.scf.storage-utils :as su]
```
mullr: `sutils`

src/puppetlabs/puppetdb/export.clj
ajroetker: I copied and pasted this :)

src/puppetlabs/puppetdb/cheshire.clj
ajroetker: Yeah I'll add the comment, we're not using a type annotation below though?

src/puppetlabs/puppetdb/export.clj
ajroetker: I'll change it though, to be clear

src/puppetlabs/puppetdb/http.clj
ajroetker: I'll change it back, I meant to add an instance check for JSONable as well

ajroetker:

src/puppetlabs/puppetdb/cheshire.clj
mullr: Yeah, I know... the reason I'm looking for it in particular here is because this is doing something interesting (.getPrettyPrinter) to the parameter, and I want to know where to go look for documentation.

ajroetker:
@wkalt Where do you think the documentation for the new `pretty_print` query param belongs? Similarly where do you think we should document the change of the default (non-pretty printing)?

ajroetker:
@wkalt Added docs and changed the parameter name.

src/puppetlabs/puppetdb/http/fact_names.clj
```diff
@@ -25,7 +25,8 @@
                  version
                  (http-q/validate-distinct-options! (merge (keywordize-keys params) puppetdb-query))
                  (:scf-read-db globals)
-                 (:url-prefix globals))))
+                 (:url-prefix globals)
+                 (:pretty-print globals))))
```
wkalt: minor, but maybe worth pointing out that this param doesn't do anything for fact-names

ajroetker: Url-prefix doesn't do anything for fact-names really either though :)

src/puppetlabs/puppetdb/query/resources.clj
```diff
@@ -20,25 +21,7 @@
...
-(def resource-schema
-  "Schema for validating a single resource."
-  (query/wrap-with-supported-fns
-    {(s/optional-key :certname) s/Str
-     (s/optional-key :environment) (s/maybe s/Str)
-     (s/optional-key :exported) s/Bool
-     (s/optional-key :file) (s/maybe s/Str)
-     (s/optional-key :line) (s/maybe s/Int)
-     (s/optional-key :parameters) resource-parameters-schema
+     (s/optional-key :parameters) (s/maybe PGobject)
```
wkalt: I think this is a change in semantics -- are we ever expecting nil here?

wkalt: oh I see it was getting intercepted in parse-params

wkalt:
explicitly supplying pretty_print=false seems not to work

src/puppetlabs/puppetdb/http/fact_names.clj
wkalt: sure, but that's not user-facing. I mean it may be worth stating something explicit with respect to fact-names and the pretty_print param

documentation/api/query/curl.markdown
```diff
@@ -82,6 +82,17 @@ If you do this with an endpoint that accepts `GET` requests, **you must also use
     curl -G http://localhost:8080/pdb/query/v4/nodes \
       --data-urlencode 'query=["=", ["node", "active"], true]'
 
+## Pretty querying of PuppetDB
+
+PuppetDB returns unprettified JSON by default. PuppetDB provides the option of
+prettifying your JSON responses with the `pretty` parameter, setting it to any
+non-`null` value will work. Be aware pretty printing comes at the cost of
```
wkalt: I don't like accepting any non-null value for this (especially the value false) -- seems like we should restrict the options to true/false.

wkalt:
This looks good to me aside from outstanding comments. The one additional thing I'd add is that I think we should make pretty printing configurable at the config level as well, so you don't need to specify it with each request if you'd rather not. Anyone whose primary interface with PDB is curl, or who mainly uses PDB for querying smaller amounts of data benefits less from this change in defaults compared to someone who is pulling over large amounts per request. My guess is that most queries are targeted enough that the perf improvement from this is minor, and the UX change becomes more significant in curl/manual inspection usage -- where it's really helpful is transfers of large amounts of data.

It's also the case for me that curling to the console is about 20% slower without pretty printing than with, but writing to a file is substantially faster (> 2x). I think that's another reason to allow a global switch on the default, although my guess is that it's pretty rare that people would be curling for that much data without writing it.

documentation/api/query/curl.markdown
```diff
@@ -82,6 +82,17 @@ If you do this with an endpoint that accepts `GET` requests, **you must also use
...
+performance on some of our endpoints, such as `/v4/catalogs`, `/v4/reports` and
```
wkalt: also, it may be worth saying something a little more detailed here that ties this to json/jsonb usage specifically. People might otherwise interpret us as saying "pretty printing has performance costs", when that's not appreciably true in isolation and the real costs stem from our implementation.

src/puppetlabs/puppetdb/http/fact_names.clj
ajroetker: I don't know I think it's pretty clear that trying to pretty-print a json string will have no effect

documentation/api/query/curl.markdown
ajroetker: The only kind of precedence I've seen for a param like this is GitHub and the `?w=<anything>` param.

ajroetker:
@wkalt I think having the config item is a little contentious to me and I think if we did do it, that I would want it in a different patch.

documentation/api/query/curl.markdown
ajroetker: I don't think this makes any generalizations about pretty-printing, especially because it states that only some of our endpoints experience this performance cost.

documentation/api/query/curl.markdown
wkalt: have to say, I don't find the github argument compelling. Accepting anything here seems sloppier than requiring a bool and validating like we do for our other params. In the case of false, I think the behavior is confusing as written.

src/puppetlabs/puppetdb/http/fact_names.clj
wkalt: it's not a string, it's a json array

ajroetker: Looks like it does do *something*, i.e. it adds some whitespace in between the list entries.

wkalt:
+1

documentation/api/query/curl.markdown
```diff
@@ -82,6 +82,18 @@ If you do this with an endpoint that accepts `GET` requests, **you must also use
     curl -G http://localhost:8080/pdb/query/v4/nodes \
       --data-urlencode 'query=["=", ["node", "active"], true]'
 
+## Pretty querying of PuppetDB
+
+PuppetDB returns unprettified JSON by default. PuppetDB provides the option of
```
senior: Does this need to be accompanied with a "deprecation" of pretty by default on the 3.x side?

src/puppetlabs/puppetdb/cheshire.clj
```diff
@@ -19,6 +19,14 @@
   "Non-memoize version of add-common-json-encoders!"
   []
   (generate/add-encoder
+   org.postgresql.util.PGobject
+   (fn [data jsonGenerator]
+     ;; The .getPrettyPrinter method on the Jackson jsonGenerator will return
+     ;; nil if `:pretty` is not set as a cheshire option
+     (if (.getPrettyPrinter jsonGenerator)
```
senior: I bet this throws reflection warnings, along with down below, I'll check

documentation/api/query/curl.markdown
ajroetker: Yep already working on it.

src/puppetlabs/puppetdb/cheshire.clj
senior: It does, should get that cleaned up as we'll be hitting this code a lot

src/puppetlabs/puppetdb/query/reports.clj
```diff
@@ -1,35 +0,0 @@
...
-  (-> event
-      (update :old_value json/parse-string)
-      (update :new_value json/parse-string)))
-
-(pls/defn-validated row->report
-  "Convert a report query row into a final report format."
-  [base-url :- s/Str]
-  (fn [row]
-    (-> row
-        (utils/update-when [:resource_events] utils/child->expansion :reports :events base-url)
```
senior: this is GREAT!

PR #1725, opened 2015-11-03
(PDB-2057) Teach tree-zipper to traverse maps

mullr:


mullr:
also, @senior basically wrote this

PR #1727, opened 2015-11-04
(PDB-2118) "Deprecate" default prettified JSON responses

ajroetker:
This commit deprecates the default prettified JSON responses from
PuppetDB because in 4.0.0 we have removed the re-encoding of child data
from the database and can't return prettified responses performantly.


kbarber:
Stable?

PR #1730, opened 2015-11-05
(MAINT) Don't fail when old migration versions are present

senior:
This fixes a bug created by the collapsed 1.x and 2.x migrations. The
bug would occur in upgrade scenarios. If a pre-3.0 migration version was
found in the database, it would fail, even if the database was a 3.0.0
database (and upgrades are supported). This commit fixes that and
separates out the supported upgrade versions and the known previous
version that are no longer referenced in PuppetDB.

### DIFF
--- a/test/puppetlabs/puppetdb/scf/migrate_test.clj
+++ b/test/puppetlabs/puppetdb/scf/migrate_test.clj
@@ -14,7 +14,8 @@
             [clojure.set :refer :all]
             [puppetlabs.puppetdb.jdbc :as jdbc :refer [query-to-vec]]
             [puppetlabs.puppetdb.testutils :refer [clear-db-for-testing! test-db]]
-            [puppetlabs.puppetdb.testutils.db :refer [schema-info-map diff-schema-maps]])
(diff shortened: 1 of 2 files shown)


test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -242,3 +243,15 @@
     (is (thrown-with-msg? IllegalStateException
                           #"Found an old and unuspported database migration.*"
                           (migrate! *db*)))))
+
+(deftest test-upgrade-migration
```
ajroetker: Should we have one for a 3.x pdb as well?

wkalt: seems to me like the code comment is wrong, and that this is actually testing a 3.x upgrade

senior: I could just remove the comment entirely if it's confusing. The issue is when there a row in the migration table from 1.x or 2.x, the upgrade will fail. This test covers that by adding a row for each prior migration (all of 1.x and 2.x)

PR #1731, opened 2015-11-05
(MAINT) Add an INSTALL_MODE var to beaker-tests.sh

senior:
We are currently not testing upgrades. The JJB jobs use an INSTALL_TYPE
environment variable to indicate whether we are testing an install or an
upgrade. PuppetDB does not recognize that variable and always assumes
were testing an install. This commit just sets the value of INSTALL_MODE
to what JJB has set for INSTALL_TYPE. This should fix things so our
upgrade tests go through the upgrade process.


waynr:
:+1:

PR #1733, opened 2015-11-05
(PDB-2100) ssl setup fails on nonproduction environments

wkalt:
This changes our ssl-setup script to use "puppet agent" rather than "puppet
master" when gathering its hostname and config file locations.


wkalt:
reopening on stable

PR #1732, opened 2015-11-05
(PDB-2097) Disable Puppet Server startup checkins to Dujour

ajroetker:
Prior to this commit Puppet Server would checkin to Dujour when
puppetserver was started up because we munged `/etc/hosts` after
puppetserver had already been installed. This commit separates the
installation of puppet and puppetserver so that we can put the munging
of `/etc/hosts` in the middle.

This commit also removes a lot of dead code that was getting in the way
of the refactor.

PR #1740, opened 2015-11-13
(PDB-2097) Disable Puppet Server startup checkins to Dujour

ajroetker:
Prior to this commit Puppet Server would checkin to Dujour when
puppetserver was started up because we munged `/etc/hosts` after
puppetserver had already been installed.
This commit uses a `puppet resource` call in between `puppet-server` and
`puppet-agent` package installs so our ec2 nodes won't check in with
dujour.


ajroetker:
This is the `stable` patch corresponding to https://github.com/puppetlabs/puppetdb/pull/1732

senior:
:+1: will merge once the pipeline is green again

PR #1749, opened 2015-11-14
(PDB-2075) Clean up template DB via shutdown hook

rbrw:


senior:
Is this cleanup only supposed to kick in when :pdb-test-keep-db-on-fail is set to true?

PR #1752, opened 2015-11-16
(PDB-2098) tk status integration

wkalt:
This cherry picks the status-service commits from master, changes our status implementation's maintenance mode response from "running" to "starting", and upgrades our tk-status version to enable the change.


kbarber:
@wkalt this is a feature patch in a Z release, since stable is destined only for 3.2.x changes. Can you explain what we're trying to do here?

PR #1753, opened 2015-11-17
(maint) change maint-mode state of status implementation to :starting

wkalt:
This changes our status implementation to return :starting while in maint mode,
and bumps to the status-service version that enables the change.

### DIFF
--- a/documentation/api/status/v1/status.markdown
+++ b/documentation/api/status/v1/status.markdown
@@ -36,7 +36,10 @@ following:
 * `detail_level`: info is currently the only level
 * `service_status_version`: version of the status API
 * `service_version`: version of PuppetDB
-* `state`: "running" if read and write databases are up, "error" if down
+* `state`: short description of PuppetDB's current state:
+    * "starting" if PuppetDB is in maintenance mode
+    * "running" if not in maintenance mode and read and write databases are up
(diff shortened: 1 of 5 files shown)


kbarber:
How do I artificially create this maintenance mode so I can manually test? :-). Sorry, I presume this is HA stuff I'm not across :-).

Incidentally, while I was starting and just seeing what it did while I kept curling, I get an 'unknown' state: https://gist.github.com/kbarber/e36fcd54373cb8c75247

wkalt:
Nope, not HA stuff. The easiest thing to do is probably query
```
curl -X GET http://localhost:8080/status/v1/services
```
in a bash loop during PDB startup, and observe that you get a few "starting" statuses.

wkalt:
To your second point that's a bug with the mq-depth component. I'll add some handling there.

kbarber:
@wkalt okay, well in that case here are my results: https://gist.github.com/kbarber/490684628d9acb69521c

I don't see the transition, I'm guessing its hidden behind these exceptions. Notice however, there is another problem with db-spec? I think we have at least 2 cases of failure in this mechanism today it seems.

kbarber:
@wkalt looks good doing manual testing now: https://gist.github.com/kbarber/9ca6f28ff1745b866a27

PR #1754, opened 2015-11-17
(QENG-3111) Make use of the PUPPETDB_DBTYPE CI variable.

waynr:
Without this patch the acceptance tests just always use `postgres` even if PUPPETDB_DBTYPE is set to something else.

Previously the name `embedded_db` was the alternative to `postgres` despite the fact that the unit test jobs actually used `hsqldb` vs `postgres` as the distinction. In https://github.com/puppetlabs/ci-job-configs/pull/715 I removed the use of `embedded_db` entirely to be replaced with `hsqldb` where appropriate.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,7 +4,7 @@ RAKE_ROOT = File.dirname(__FILE__)
 
 def run_beaker(test_files)
   config = ENV["BEAKER_CONFIG"] || "ec2-west-dev"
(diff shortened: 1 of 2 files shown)


kbarber:
@waynr @wkalt Umm, master doesn't care about hsqldb, we should probably fix this on stable shouldn't we?

PR #1760, opened 2015-11-21
(PDB-2104) Update to Clojure 1.8.0-RC2

rbrw:


test/puppetlabs/puppetdb/zip_test.clj
```diff
@@ -1,6 +1,7 @@
 (ns puppetlabs.puppetdb.zip-test
   (:require [puppetlabs.puppetdb.zip :refer :all]
             [clojure.test :refer :all]
+            [clojure.pprint :refer [pprint]]
```
wkalt: unused

kbarber:
@rbrw looks like there is a problem with this bump + testing with pe-puppetdb-extensions: https://gist.github.com/kbarber/599b2ee0c624c89861e9 ...

rbrw:
@kbarber any chance you had "lein install"ed this branch, but weren't on the RC2 branch in extensions when you ran the tests?  I think that may be what caused an error like that for me before.

PR #1761, opened 2015-11-23
(PDB-2206) indexes not getting used for hash/uuid queries

wkalt:
Without this patch, equality queries on our bytea fields trigger full table
scans because we're doing type coercion on the left hand side of the
comparison. This adds expression indexes for those coercions to enable fast
queries.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1465,6 +1465,29 @@
       (format "ALTER TABLE reports ADD COLUMN logs %s DEFAULT NULL" jsonb-type)
       (format "ALTER TABLE reports ADD COLUMN resources %s DEFAULT NULL" jsonb-type))))
 
+(defn add-expression-indexes-for-bytea-queries
+  []
+  (if (sutils/postgres?)
+    (jdbc/do-commands
+      "CREATE UNIQUE INDEX reports_hash_expr_idx ON reports(trim(leading '\\x' from hash::text))"
+      "ALTER TABLE REPORTS DROP CONSTRAINT reports_hash_key"
+      "DROP INDEX reports_transaction_uuid_idx"
+      "CREATE INDEX reports_tx_uuid_expr_idx ON reports(CAST(transaction_uuid AS text))"
+
+      ;; leave the existing resources/rpc indices for resources-query join
+      "CREATE INDEX resources_hash_expr_idx ON catalog_resources(trim(leading '\\x' from resource::text))"
+      "CREATE INDEX rpc_hash_expr_idx ON resource_params_cache(trim(leading '\\x' from resource::text))"
+      "CREATE INDEX resource_params_hash_expr_idx ON resource_params(trim(leading '\\x' from resource::text))"
+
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1465,6 +1465,31 @@
...
       (format "ALTER TABLE reports ADD COLUMN resources %s DEFAULT NULL" jsonb-type))))
 
+(defn add-expression-indexes-for-bytea-queries
+  []
+  (if (sutils/postgres?)
+    (jdbc/do-commands
+      "CREATE UNIQUE INDEX reports_hash_expr_idx ON reports(trim(leading '\\x' from hash::text))"
+      "ALTER TABLE REPORTS DROP CONSTRAINT reports_hash_key"
+      "DROP INDEX reports_transaction_uuid_idx"
+      "CREATE INDEX reports_uuid_expr_idx ON reports(CAST(transaction_uuid AS text))"
```
ajroetker: I think we'll probably want to make tests for these in the rollup using @senior's migration diff'ing, to make sure they don't just disappear!

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1465,6 +1465,31 @@
...
+      "CREATE INDEX resources_hash_expr_idx ON catalog_resources(trim(leading '\\x' from resource::text))"
+      "CREATE INDEX rpc_hash_expr_idx ON resource_params_cache(trim(leading '\\x' from resource::text))"
+
+      "CREATE INDEX resource_params_hash_expr_idx ON resource_params(trim(leading '\\x' from resource::text))"
+      "DROP INDEX idx_resources_params_resource"
+
+      "CREATE UNIQUE INDEX catalogs_hash_expr_idx ON catalogs(trim(leading '\\x' from hash::text))"
+      "ALTER TABLE catalogs DROP CONSTRAINT catalogs_hash_key"
+      "DROP INDEX idx_catalogs_transaction_uuid"
+      "CREATE INDEX catalogs_uuid_expr_idx ON catalogs(CAST(transaction_uuid AS text))"
```
ajroetker: I think we should probably name this catalogs_transaction_uuid... instead of just catalogs_uuid, since in 4.0.0 we'll also have an idx on catalogs_catalog_uuid_expr_idx

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1465,6 +1465,31 @@
       (format "ALTER TABLE reports ADD COLUMN logs %s DEFAULT NULL" jsonb-type)
       (format "ALTER TABLE reports ADD COLUMN resources %s DEFAULT NULL" jsonb-type))))
 
+(defn add-expression-indexes-for-bytea-queries
+  []
+  (if (sutils/postgres?)
+    (jdbc/do-commands
+      "CREATE UNIQUE INDEX reports_hash_expr_idx ON reports(trim(leading '\\x' from hash::text))"
```
mullr: should be a bit faster to use 'decode' here.

PR #1762, opened 2015-11-23
(PDB-2112) fix benchmark -A flag

wkalt:
This removes some unwarranted slashes from our benchmark code, and implements a
test for the -A flag.


wkalt:
transient failure ^

PR #1764, opened 2015-11-27
(PDB-2122) New 'root' single query endpoint and 'from' operator

kbarber:
This patch introduces the new 'root' endpoint that allows for querying any
entity inside a single query by providing a context query.

Here I've included this new feature all the way back to produce-streaming-body,
stream-query-result and the query API. This patch means that all the HTTP
endpoints are required to wrap their queries in from, so that this concept
of a single query for all endpoints exists all the way back to the query engine
API.

For cleanup, I've modified the produce-streaming-body and stream-query-results
to take an options map instead of a list of ordered arguments as we seem to change
this frequently and some validation is desired. Plus this simplifies argument
passing as most of the data is already available in `globals`.

As a side-effect of adding the from operator, I've added a new format for
subqueries that uses the same syntax at this level, as it does on the top-level.
This is provided for consistency, and it will hopefully remove the need for
the select_<entity> operator which was always a bit special.

### DIFF
--- a/documentation/_puppetdb_nav.html
+++ b/documentation/_puppetdb_nav.html
@@ -58,8 +58,10 @@
     <ul>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/upgrading-from-v3.html">Upgrading from Version 3</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/query.html">Query Structure</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/entities.html">Entities</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/operators.html">Available Operators</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/paging.html">Query Paging</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/index.html">Root Endpoint (Experimental)</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/nodes.html">Nodes Endpoint</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/environments.html">Environments Endpoint</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/factsets.html">Factsets Endpoint</a></li>
--- a/documentation/api/query/v4/entities.markdown
+++ b/documentation/api/query/v4/entities.markdown
@@ -1,0 +1,53 @@
+---
+title: "PuppetDB 3.2 » API » v4 » Entities"
+layout: default
+canonical: "/puppetdb/latest/api/query/v4/entities.html"
+---
+
+[aggregate-event-counts]: ./aggregate-event-counts.html
+[catalogs]: ./catalogs.html
+[edges]: ./edges.html
+[environments]: ./environments.html
+[event-counts]: ./event-counts.html
+[events]: ./events.html
+[fact-names]: ./fact-names.html
+[facts]: ./facts.html
+[fact-contents]: ./fact-contents.html
+[fact-paths]: ./fact-paths.html
+[nodes]: ./nodes.html
+[query]: ./query.html
+[reports]: ./reports.html
+[resources]: ./resources.html
+[root]: ./index.html
+[from]: ./operators.html#context-operators
+[subquery]: ./operators.html#subquery-operators
+
+The PuppetDB API provides access to a series of data entities that map to the Puppet eco-system
+and the data that PuppetDB stores.
+
+## Entity Types
+
+The following table shows the list of entities available, and their respective REST endpoints for
+direct querying.
+
+The entity name is utilised within queries themselves, in particular within the [`from`][from]
+and [`subquery`][subquery] operators.
+
+Entity Name                                        | REST Endpoint
+---------------------------------------------------|---------------------------------------------------------------
+[`aggregate_event_counts`][aggregate-event-counts] | [/pdb/query/v4/aggregate-event-counts][aggregate-event-counts]
+[`catalogs`][catalogs]                             | [/pdb/query/v4/catalogs][catalogs]
+[`edges`][edges]                                   | [/pdb/query/v4/edges][edges]
+[`environments`][environments]                     | [/pdb/query/v4/environments][environments]
+[`event_counts`][event-counts]                     | [/pdb/query/v4/event-counts][event-counts]
+[`events`][events]                                 | [/pdb/query/v4/events][events]
+[`facts`][facts]                                   | [/pdb/query/v4/facts][facts]
+[`fact_contents`][fact-contents]                   | [/pdb/query/v4/fact-contents][fact-contents]
+[`fact_names`][fact-names]                         | [/pdb/query/v4/fact-names][fact-names]
+[`fact_paths`][fact-paths]                         | [/pdb/query/v4/fact-paths][fact-paths]
+[`nodes`][nodes]                                   | [/pdb/query/v4/nodes][nodes]
+[`reports`][reports]                               | [/pdb/query/v4/reports][reports]
+[`resources`][resources]                           | [/pdb/query/v4/resources][resources]
+
+You can also query a particular entity using the [root endpoint][root] by using a top-level [`from`][from]
+operator.
--- a/documentation/api/query/v4/index.markdown
+++ b/documentation/api/query/v4/index.markdown
@@ -1,0 +1,63 @@
+---
+title: "PuppetDB 3.2 » API » v4 » Root Endpoint"
+layout: default
+canonical: "/puppetdb/latest/api/query/v4/index.html"
+---
+
+[curl]: ../curl.html#using-curl-from-localhost-non-sslhttp
+[paging]: ./paging.html
+[query]: ./query.html
+[from]: ./operators.html#context-operators
+[entities]: ./entities.html
+
+*Note:* This endpoint is experimental. It may be altered or removed in a future release.
+
+The root query endpoint can be used to retrieve any known entities, from a
+single endpoint.
+
+## `/pdb/query/v4`
+
+This will return any known entity based on the required `query` field. Unlike
+other endpoints, the [entity][entities] must be supplied using a query with the [`from`][from]
+operator.
+
+### URL Parameters
+
+* `query`: Required. A JSON array containing the query in prefix notation
+(`["from", "<ENTITY>", ["<OPERATOR>", "<FIELD>", "<VALUE>"]]`). Unlike other endpoints,
+a query with a [`from`][from] is required to choose the [entity][entities] to query for. For
+general info about queries, see [the page on query structure.][query]
+
+### Response Format
+
+The response will be in `application/json`, and will contain a list of JSON
+object results based on the [entity][entities] provided in the top-level [`from`][from] query.
+
+### Examples
+
+[Using `curl` from localhost][curl]:
+
+    curl -X GET http://localhost:8080/pdb/query/v4 --data-urlencode 'query=["from","nodes",["=","certname","macbook-pro.local"]]'
+
+    [
+      {
+        "catalog_environment": "production",
+        "catalog_timestamp": "2015-11-23T19:25:25.561Z",
+        "certname": "macbook-pro.local",
+        "deactivated": null,
+        "expired": null,
+        "facts_environment": "production",
+        "facts_timestamp": "2015-11-23T19:25:25.079Z",
+        "latest_report_hash": "0b2aa3bbb1deb71a5328c1d934eadbba5f52d733",
+        "latest_report_status": "unchanged",
+        "report_environment": "production",
+        "report_timestamp": "2015-11-23T19:25:23.394Z"
+      }
+    ]
+
+## Paging
+
+This query endpoint supports paged results via the common PuppetDB paging
+URL parameters.  For more information, please see the documentation
+on [paging][paging].
+
--- a/documentation/api/query/v4/operators.markdown
+++ b/documentation/api/query/v4/operators.markdown
@@ -4,6 +4,7 @@ layout: default
 canonical: "/puppetdb/latest/api/query/v4/operators.html"
 ---
 
+[root]: ./index.html
 [catalogs]: ./catalogs.html
 [edges]: ./edges.html
 [environments]: ./environments.html
@@ -15,6 +16,7 @@ canonical: "/puppetdb/latest/api/query/v4/operators.html"
 [query]: ./query.html
 [reports]: ./reports.html
 [resources]: ./resources.html
+[entities]: ./entities.html
 
 PuppetDB's [query strings][query] can use several common operators.
 
@@ -176,6 +178,32 @@ To get the average uptime for your nodes,
 
     ["extract", [["function", "avg", "value"]], ["=", "name", "uptime_seconds"]]
 
+## Context Operators
+
+*Note:* Context setting support is new and experimental. Setting the context at
+the top of the query is only supported on the [root] endpoint.
+
+Setting context in a query allows you to choose the entity you are querying
+on. This augments the endpoint support we have today, whereby the endpoint
+decides the context. For example `/pdb/query/v4/nodes` sets the context of the query
+to `nodes`.
+
+
+### `from`
+
+The `from` operator allows one to choose the [entity][entities] that you want to query and
+provide an optional query clause for filtering those results. This operator can
+be used at the top-level context of a query like so:
+
+    ["from", "nodes", ["=", "certname", "myserver"]]
+
+The `from` operator can also be used in a subquery for setting the context when
+using the [`in` operator](#subquery-operators).
+
+When querying a particular endpoint, such as `/pdb/query/v4/nodes` the endpoint provides
+the context for the query. While querying the [root] endpoint requires specifying a
+context explicitly.
+
 ## Subquery Operators
 
 Subqueries allow you to correlate data from multiple sources or multiple
@@ -197,20 +225,10 @@ related entity and the query to use:
 
     ["subquery", "<ENTITY>", <SUBQUERY STATEMENT>]
 
-The entity is a mapping to the particular endpoint or data type you desire, however not
-all entities are supported from all endpoints, as not every relationship makes sense.
-Consult the endpoint document for details on what implicit relationships are supported:
-
-* [`catalogs`][catalogs]
-* [`edges`][edges]
-* [`environments`][environments]
-* [`events`][events]
-* [`facts`][facts]
-* [`fact_contents`][fact-contents]
-* [`fact_paths`][fact-paths]
-* [`nodes`][nodes]
-* [`reports`][reports]
-* [`resources`][resources]
+The [`<ENTITY>`][entities] is the particular entity you are subquerying on, however not
+all entities are implicitly relatable to all other entities, as not every relationship makes sense.
+Consult the documentation for the chosen [`<ENTITY>`][entities] for details on what
+implicit relationships are supported.
 
 Internal to PuppetDB, we keep a mapping of how different entities relate to each
 other, and so no other data beyond the entity is needed in this case. This is
@@ -247,9 +265,15 @@ mapped internally. For these more advanced subqueries, you need to specify exact
 a subquery should join on. This is where an explicit subquery can be useful.
 
 Explicit subqueries are unlike the other operators listed above. They always appear
-together in the following form:
+together in one of the following forms:
 
     ["in", ["<FIELDS>"], ["extract", ["<FIELDS>"], <SUBQUERY STATEMENT>] ]
+
+The second new methodology uses `from` to set the context, and now looks like this:
+
+    ["in", ["<FIELDS>"], ["from", <ENTITY>, ["extract", ["<FIELDS>"], <SUBQUERY>] ] ]
+
+*Note:* This new format is experimental and may change or be removed in the future.
 
 That is:
 
@@ -275,21 +299,37 @@ An `in` statement constitutes a full query string, which can be used alone or as
 
 "In" statements are **non-transitive** and take two arguments:
 
-* The first argument **must** consist of one or more **fields** for the endpoint **being queried.**. This is a string or vector of strings.
-* The second argument **must** be an **`extract` statement,** which acts as a list of possible values for the fields.
+* The first argument **must** consist of one or more **fields** for the endpoint or entity **being queried.**. This is a string or vector of strings.
+* The second argument **must** be either:
+** an **`extract` statement,** which acts as a list of fields to extract during the subquery for matching against the **fields** in the `in` clause.
+** a **`from` statement,** which sets the context, and allows for an extract statement to be provided. *Note:* this syntax is new and experimental.
 
-**Matches if:** the field values are included in the list of values created by the `extract` statement.
+**Matches if:** the field values are included in the list of values created by the `extract` or `from` statement.
+
+#### `from`
+
+*Note:* the use of `from` in a subquery is experimental. It may change or be removed in the future.
+
+This statement works like the top-level [`from`](#context-operators) operator, and expects an [entity][entities] as the first argument and an optional query in the second
+argument, however when used within an `in` clause an `extract` statement is expected to choose the fields like so:
+
+    ["in", "certname",
+     ["from", "facts",
+      ["extract", "certname",
+       [<QUERY>]]]]
 
 #### `extract`
 
 "Extract" statements are **non-transitive** and take two arguments:
 
 * The first argument **must** be a valid set of **fields** for the endpoint **being subqueried** (see second argument). This is a string or vector of strings.
-* The second argument **must** be a **subquery statement.**
+* The second argument:
+** **must** contain a **subquery statement**
+** or when used with the new `from` operator, **may** contain an optional query.
 
 As the second argument of an `in` statement, an `extract` statement acts as a list of possible values. This list is compiled by extracting the value of the requested field from every result of the subquery.
 
-#### Subquery Statements
+#### `select_<ENTITY>` Subquery Statements
 
 A subquery statement **does not** constitute a full query string. It may only be used as the second argument of an `extract` statement.
 
@@ -299,8 +339,6 @@ Subquery statements are **non-transitive** and take two arguments:
 * The second argument **must** be a **full query string** that makes sense for the endpoint being subqueried.
 
 As the second argument of an `extract` statement, a subquery statement acts as a collection of PuppetDB objects. Each of the objects returned by the subquery has many fields; the `extract` statement takes the value of one field from each of those objects, and passes that list of values to the `in` statement that contains it.
-
-#### Available Subqueries
 
 Each subquery acts as a normal query to one of the PuppetDB endpoints. For info on constructing useful queries, see the docs page for the endpoint matching the subquery:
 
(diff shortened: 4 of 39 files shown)


kbarber:
There is a bit of a chicken and egg scenario for merge tests to pass for this PR. When we merge this pe-puppetdb-extensions will fail, so we'll need this PR also merged: https://github.com/puppetlabs/pe-puppetdb-extensions/pull/144

documentation/_puppetdb_nav.html
```diff
@@ -60,6 +60,7 @@ <h2 id="puppetdb">PuppetDB {{ puppetdb_version }}</h2>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/query.html">Query Structure</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/operators.html">Available Operators</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/paging.html">Query Paging</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/api/query/v4/root.html">Root Endpoint (Experimental)</a></li>
```
kbarber: I couldn't think of a better name for the endpoint, ideas welcome.

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -5,15 +5,16 @@
 
 (defn routes
   [version optional-handlers]
-  (let [param-spec {:required ["summarize_by"]
+  (let [handlers (cons (partial http-q/restrict-query-to-entity "aggregate_event_counts")
```
kbarber: endpoints need to wrap their queries in a from as you can see here now, since the from is expected in the query when using the query engine functions.

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -290,31 +299,14 @@
...
-   `optional-handlers` with the middleware function that executes the
-   query."
-  [entity version param-spec & optional-handlers]
+  [version param-spec & optional-handlers]
   (let [handlers (or optional-handlers [identity])]
     (app
-      (extract-query param-spec)
-      (apply comp
-             (fn [{:keys [params globals puppetdb-query]}]
-               (warn-experimental entity (:warn-experimental globals))
```
kbarber: I pushed experimental warning on entities into the query engine, as it feels like a 'lower-down' problem.

src/puppetlabs/puppetdb/http/server.clj
```diff
@@ -11,18 +11,6 @@
             [net.cgrand.moustache :refer [app]]
             [ring.util.response :as rr]))
 
-(defn deprecated-app
```
kbarber: I've re-used experimental-warning elsewhere, so I moved these both to http.clj for general use.

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -101,55 +106,77 @@
 
       (munge-fn version url-prefix))))
 
-(defn stream-query-result
+(def query-options-schema
```
kbarber: produce-streaming-body and stream-query-results now takes an options hash that gets validated and has defaults. We keep moving around these parameters and this feels like a step in a better direction to manage these, at least produce-streaming-body has a contract with pe-puppetdb-extensions for example.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1539,6 +1574,39 @@
...
+
+(defn warn-experimental
+  "Show a warning if the endpoint is experimental."
+  [entity]
+  (when (contains? experimental-entities entity)
+    (log/warn (format
+                "The %s entity is experimental and may be altered or removed in the future."
+                (name entity)))))
+
+(defn parse-query-context
```
kbarber: This magic extracts the 'from' and returns the remaining query and entity. In a way its a compatibility layer, later on we want to push the 'from' further down into the query engine, but munging still happens at the query_engine.clj level, so we still need this extraction for now ...

documentation/_puppetdb_nav.html
ajroetker: Would using the route `/pdb/query/v4` be crazy?
If so what about `/pdb/query/v4/context`  or `/pdb/query/v4/from` to match it with the operator?

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
```diff
@@ -5,15 +5,16 @@
...
+                       optional-handlers)
+        param-spec {:required ["summarize_by"]
                     :optional ["query" "counts_filter" "count_by"
                                "distinct_resources" "distinct_start_time"
                                "distinct_end_time"]}
-        query-route #(apply (partial http-q/query-route :aggregate-event-counts
-                                     version param-spec) %)]
+        query-route #(apply (partial http-q/query-route version param-spec) %)]
```
ajroetker: Should we just make a wrapper around `http-q/query-route` which takes a list like this interface expects?

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -56,7 +58,10 @@
 
 (defn get-munge
   [entity]
-  (get-in @entity-fn-idx [entity :munge]))
+  (let [munge-result (get-in @entity-fn-idx [entity :munge])]
+    (if (nil? munge-result)
```
ajroetker:
```clojure
(if-let [munge-result (...)]
 munge-result
 (throw ...))
```

kbarber: Ta :-).

documentation/_puppetdb_nav.html
kbarber: It actually is /pdb/query/v4 ... I'm pointing out 'Root Endpoint' here as this is just the name in documentation. I could have called it 'Index Endpoint' instead, and make the docs appear in index.html, but I just wasn't sure.

src/puppetlabs/puppetdb/http/aggregate_event_counts.clj
kbarber: Good thinking, doing that now. Plus, I've made a query-route-from wrapper that automatically puts in the restriction for you, saves on some loc at least :-).

documentation/_puppetdb_nav.html
ajroetker: Oh nice sorry, misunderstood. Is there a way we could redirect `/puppetdb/{{ puppetdb_version }}/api/query/v4` to `/puppetdb/{{ puppetdb_version }}/api/query/v4/root.html`? Otherwise the name is :+1:  from me.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -318,10 +318,10 @@
                               (format "Attempt to set url-prefix to %s when it's already been set to %s" url-prefix @old-url-prefix)))))
   (shared-globals [this]
                   (:shared-globals (service-context this)))
-  (query [this query-obj version query-expr paging-options row-callback-fn]
-         (let [{db :scf-read-db} (get (service-context this) :shared-globals)
-               url-prefix @(get (service-context this) :url-prefix)]
-           (qeng/stream-query-result query-obj version query-expr paging-options db url-prefix row-callback-fn))))
+  (query [this version query-expr paging-options row-callback-fn]
+         (let [query-options (-> (select-keys (get (service-context this) :shared-globals) [:scf-read-db :warn-experimental])
```
ajroetker: I might pull out the `(get (service-context this) :shared-gloabls)` to be the first argument of the `->`

ajroetker: Or even pull out the `(service-context this)` to a named variable in the `let`

src/puppetlabs/puppetdb/http/edges.clj
```diff
@@ -12,11 +12,10 @@
                    http-q/restrict-query-to-active-nodes
                    identity)
          handlers (cons handler optional-handlers)
-         param-spec {:optional paging/query-params}
-         query-route #(apply (partial http-q/query-route :edges version param-spec) %)]
+         param-spec {:optional paging/query-params}]
      (app
        [""]
-       (query-route handlers)))))
+       (http-q/query-route-from "edges" version param-spec handlers)))))
```
ajroetker: Nice :dancer:, so much easier to read!

src/puppetlabs/puppetdb/http/v4.clj
```diff
@@ -13,12 +14,24 @@
...
             [net.cgrand.moustache :as moustache]))
 
 (def version :v4)
 
+(defn experimental-index-app
+  [version]
+  (fn [request]
+    (http/experimental-warning
+     (index/index-app version)
+     "The index endpoint is experimental"
```
ajroetker: Should call this `root` to match the docs?

ajroetker:
Tested out bunch of queries using this patch and it all lgtm. One note that is a preexisting issue, invalid operators will fail silently. For instance I tried `-d 'query="["from","reports",["exadsacnsldc","certname"]]' and PDB returned the unextracted reports.

ajroetker:
Also with the addition of this endpoint, it would be really nice if we had some kind of discovery endpoint or doc which graphed out all the valid entities for the `from` operators. Something like `/pdb/query` returning:
```json
{"/v4" : ["reports","catalogs",...,"report_logs"]} 
```
We could even end up adding the valid implicit `subquery` relationships or some such.

src/puppetlabs/puppetdb/cli/services.clj
kbarber: Cool.

src/puppetlabs/puppetdb/http/v4.clj
kbarber: Yep.

src/puppetlabs/puppetdb/http/edges.clj
kbarber: Agreed

documentation/_puppetdb_nav.html
kbarber: Well, I can just make it index.html. Saves any hassle that way.

kbarber:
@ajroetker for the discovery/doc side of things: I think discovery is someone we can circle back on, its a complex topic - let me ponder it some more. I do notice however, I fail to link to the available entities properly in docs, I've got some ideas on this and I'll submit something soon.

I'll take a quick look into the operator issue, if its easy I'll fix otherwise I'll create a ticket.

kbarber:
@ajroetker I've added the new entities.markdown doc now and updated some other docs related to this problem, that splits out the available entities. Seems we were using it all over the place, so I've modded those places to reference this, actually simplifying our docs somewhat. I think _this_ is the place we should have our relationship diagram also ... which somewhat answers that question I've had in the back of my head also.

documentation/api/query/v4/entities.markdown
```diff
@@ -0,0 +1,53 @@
+---
+title: "PuppetDB 3.2 » API » v4 » Entities"
```
kbarber: I think this is where our relationship diagram could eventually live.

kbarber:
Those last two failures are because I cancelled those tests ^^. I did this on purpose, to only leave the last one running.

documentation/api/query/v4/entities.markdown
```diff
@@ -0,0 +1,53 @@
...
+The PuppetDB API provides access to a series of data entities that map to the Puppet eco-system
+and the data that PuppetDB stores.
+
+## Entity Types
+
+The following table shows the list of entities available, and their respective REST endpoints for
+direct querying.
+
+The entity name is utilised within queries themselves, in particular within the [`from`][from]
+and [`subquery`][subquery] operators.
```
ajroetker: This is great, I've been wanting this for a long time.

kbarber: Yeah, its basically replacing select_<entity> ... and providing upper and lower-level syntax consistency in doing so. My decision to move the 'from' part to the outside though, is more that even in things like SQL, the table context is the mandatory part, extract is not mandatory in principle (ie. select * from foo).

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -290,31 +299,25 @@
...
+
+(defn query-route-from
+  "Convenience wrapper for query-route, which automatically wraps the query in a
+  `from` to set context."
+  ([entity version param-spec]
+   (query-route-from entity version param-spec [identity]))
+  ([entity version param-spec optional-handlers]
+   (let [handlers (cons (partial restrict-query-to-entity entity)
+                        optional-handlers)]
+     (query-route version param-spec handlers))))
```
ajroetker: Any way that query-route and query-route-from could be merged, or could the `([version param-spec] (query-route version param-spec [identity]))` case of query-route be removed, just a little bothered by the duplication of the optional-handlers defaulting interface?

src/puppetlabs/puppetdb/http/nodes.clj
```diff
@@ -13,14 +13,12 @@
...
 (defn node-status
   "Produce a response body for a single environment."
-  [api-version node db url-prefix]
+  [api-version node options]
   (let [status (first
-                (eng/stream-query-result :nodes
-                                         api-version
-                                         ["=" "certname" node]
+                (eng/stream-query-result api-version
+                                         ["from" "nodes" ["=" "certname" node]]
```
ajroetker: Just making a note that anything that runs `stream-query-result` directly might not handle database streaming errors correctly, and also won't pretty-print.

ajroetker: A pre-existing problem*

src/puppetlabs/puppetdb/http/index.clj
```diff
@@ -0,0 +1,18 @@
...
+  (:require [net.cgrand.moustache :refer [app]]
+            [puppetlabs.puppetdb.http.query :as http-q]
+            [puppetlabs.puppetdb.middleware :refer [wrap-with-paging-options]]
+            [puppetlabs.puppetdb.query.paging :as paging]))
+
+(defn routes
+  [version optional-handlers]
+  (let [param-spec {:optional paging/query-params
+                    :required ["query"]}]
+    (app
```
ajroetker: I'd prefer us using compojure instead of moustache, like we do for `src/puppetlabs/puppetdb/metrics.clj`.

ajroetker:
Just some small notes, otherwise definitely :+1: again!

src/puppetlabs/puppetdb/http/index.clj
kbarber: All the other query http endpoints use this methodology, this feels like a refactor wishlist for later.

src/puppetlabs/puppetdb/http/query.clj
kbarber: Done, second one.

kbarber:
@ajroetker new patch up again.

documentation/api/query/v4/operators.markdown
```diff
@@ -176,6 +178,30 @@ To get the average uptime for your nodes,
...
     ["extract", [["function", "avg", "value"]], ["=", "name", "uptime_seconds"]]
 
+## Context Operators
+
+*Note:* Context setting support is new and experimental. Setting the context at
+the top of the query is only supported on the [root] endpoint.
+
+Setting context in a query allows you to choose the entity you are querying
+on. This augments the endpoint support we have today, whereby the endpoint
+decides the context by allowing a single endpoint to contain all the information
```
ajroetker: I think you can cut off this sentence and just say "whereby the endpoint decides the context. For example `/pdb/query/v4/nodes` sets the context of the query to `nodes`.

documentation/api/query/v4/operators.markdown
```diff
@@ -176,6 +178,30 @@ To get the average uptime for your nodes,
...
+required to show results.
+
+### `from`
+
+The `from` operator allows one to choose the entity that you want to query and
```
ajroetker: Could `entity` here have a link to the "entities" doc? Channeling my inner Wyatt w.r.t. entity.

documentation/api/query/v4/operators.markdown
```diff
@@ -176,6 +178,30 @@ To get the average uptime for your nodes,
...
+provide an optional query clause for filtering those results. This operator can
+be used at the top-level context of a query like so:
+
+    ["from", "nodes", ["=", "certname", "myserver"]]
+
+The `from` operator can also be used in a subquery for setting the context when
+using the [`in` operator](#subquery-operators).
+
+When providing a query to a particular endpoint such as `/nodes` the endpoint decides
+the context. With the new [root] endpoint, a query must contain the context.
```
ajroetker: > When querying a particular endpoint, such as `/nodes`, the endpoint provides the context for the query. While querying the [root] endpoint requires specifying a context explicitly.

documentation/api/query/v4/operators.markdown
```diff
@@ -197,20 +223,10 @@ related entity and the query to use:
...
-* [`environments`][environments]
-* [`events`][events]
-* [`facts`][facts]
-* [`fact_contents`][fact-contents]
-* [`fact_paths`][fact-paths]
-* [`nodes`][nodes]
-* [`reports`][reports]
-* [`resources`][resources]
+The [`<ENTITY>`][entities] is the particular entity you are subquerying on, however not
+all entities are relateable to other entities, as not every relationship makes sense.
```
ajroetker: > not all entities are implicitly relatable to all other entities

documentation/api/query/v4/operators.markdown
```diff
@@ -395,3 +431,50 @@ For the previous query, we also allow the shorthand
...
+    ["and",
+      ["=", "name", "ipaddress"],
+      ["in", "certname",
+        ["from", "resources",
+          ["extract", "certname",
+            ["and",
+              ["=", "type", "Class"],
+              ["=", "title", "Apache"]]]]]]
+
+This queries on the `/facts` endpoint filters for uptime_hours of all nodes with
```
ajroetker: > Executing this query on the `/facts` endpoint would filter for `uptime_hours` for all nodes with `facts_environment` set to `production`:

documentation/api/query/v4/operators.markdown
```diff
@@ -395,3 +431,50 @@ For the previous query, we also allow the shorthand
...
+
+    ["and",
+      ["=", "name", "uptime_hours"],
+      ["in", "certname",
+        ["from", "nodes",
+          ["extract", "certname",
+            ["=", "facts_environment", "production"]]]]]
+
+To find node information for a host that has a macaddress of `aa:bb:cc:dd:ee:00` as
+its first macaddress on the interface `eth0`, you could use this query on '/nodes':
```
ajroetker: '/nodes' -> `/nodes`

documentation/api/query/v4/query.markdown
```diff
@@ -35,6 +36,12 @@ That is, most queries will look like a GET request to a URL that resembles the f
 
     https://puppetdb:8081/pdb/query/v4/<ENDPOINT>?query=<QUERY STRING>
 
+Alternatively, you can provide the entity instead of using the `<ENDPOINT>` suffix with the following:
```
ajroetker: > Alternatively, you can provide the context/entity in a query, instead of using the `<ENPOINT>` suffix, with the following:

ajroetker:
Proofread the documentation updates @kbarber, pretty minor stuff. I've been through the code and docs as much as I can! If someone else /cc @rbrw @mullr @senior wants to just give a separate proofread through the code and docs, I'm ready to merge this in.

kbarber:
@ajroetker updated from your review.

ajroetker:
@kbarber I noticed that `pretty=true` is not a valid query param to this endpoint. It might be nice to add that as an option.

documentation/api/query/v4/index.markdown
```diff
@@ -0,0 +1,66 @@
...
+
+### URL Parameters
+
+* `query`: Required. A JSON array containing the query in prefix notation
+(`["from", "<ENTITY>", ["<OPERATOR>", "<FIELD>", "<VALUE>"]]`). See the
+sections below for the supported operators and fields. Unlike other endpoints,
+a query with a `from` is required to choose the entity to query for. For
+general info about queries, see [the page on query structure.][query]
+
+If a query parameter is not provided, all results will be returned.
```
senior: Is this a copy and past error? Above talks about requiring a query parameter.

documentation/api/query/v4/index.markdown
```diff
@@ -0,0 +1,66 @@
...
+
+### Response Format
+
+The response will be in `application/json`, and will contain a list of JSON
+object results based on the query provided.
```
senior: I think we need to say more about the response format here. Maybe with links that go to the from entity and the correct response format?

documentation/api/query/v4/index.markdown
```diff
@@ -0,0 +1,66 @@
...
+The root query endpoint can be used to retrieve any known entities, from a
+single endpoint.
+
+## `/pdb/query/v4`
+
+This will return any known entity based on the required `query` field. Unlike
+other endpoints, the [entity][entities] must be supplied using a query with the [`from`][from]
+operator.
+
+### URL Parameters
```
senior: We should also have links here to the entities and their available query fields.

src/puppetlabs/puppetdb/http/fact_names.clj
```diff
@@ -14,27 +14,26 @@
...
-  [entity version param-spec & optional-handlers]
-  (app
-    (http-q/extract-query param-spec)
-    (apply comp
-           (fn [{:keys [params globals puppetdb-query]}]
-             (let [puppetdb-query (assoc-when puppetdb-query :order_by [[:name :ascending]])]
-               (produce-streaming-body
-                 entity
+  [version param-spec & optional-handlers]
+  (let [handlers (cons (partial http-q/restrict-query-to-entity "fact_names")
```
senior: Why not omit the let (and the cons) and just add the partial and optional handlers in to the comp below?

src/puppetlabs/puppetdb/http/index.clj
senior: Meh, I think new endpoints have been compojure. That's certainly where we want to go. I think this would be pretty simple to make compojure as well, just adding work for ourselves later if we don't.

src/puppetlabs/puppetdb/http/nodes.clj
senior: @ajroetker can you check on this and create a ticket if you find anything?

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -179,6 +175,19 @@
   (restrict-query ["=" "title" title]
                   req))
 
+(defn wrap-in-from
```
senior: fwiw I expected to see this function use an in clause and a from clause until seeing the implementation.

src/puppetlabs/puppetdb/http/index.clj
kbarber: Its more a matter of, 'might as well do them all' since all the existing patterns are the same today, instead of progressing in piecemeal and having the endpoints look different, yet again.

src/puppetlabs/puppetdb/http/server.clj
senior: :+1:

src/puppetlabs/puppetdb/http/query.clj
kbarber: or reading the note below? I can change it to wrap-with-from or something.

documentation/api/query/v4/index.markdown
kbarber: Yes it is.

documentation/api/query/v4/index.markdown
kbarber: I'll link to the entities page a little better. I'm avoiding repeating myself here, since that list will become out of date. Sometimes, we need the ability to insert sub-documents, rather than link to them I think.

documentation/api/query/v4/index.markdown
kbarber: Same as above, I'll add more linking to the entities page.

src/puppetlabs/puppetdb/http/fact_names.clj
kbarber: +1

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -101,55 +105,77 @@
...
-   (let [munge-fn (get-munge-fn entity version paging-options url-prefix)]
-     (jdbc/with-transacted-connection db
+   (stream-query-result version query paging-options options doall))
+  ([version :- s/Keyword
+    query
+    paging-options
+    options :- query-options-schema
+    row-fn]
+   (let [{:keys [scf-read-db url-prefix warn-experimental pretty-print]
+          :or {warn-experimental true
```
senior: I'm not sure I'm understanding `warn-experimental` and that it defaults to true. Is this saying by default we're output a warning for any call to stream-query-result that doesn't set it to false?

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -101,55 +105,77 @@
...
+               pretty-print false
+               url-prefix "/pdb"}} options
```
senior: do we not have "/pdb" someplace as a top-level def?

src/puppetlabs/puppetdb/query_eng.clj
kbarber: The concept is that we didn't want to warn that those few "distinct" entities were now experimental in PE I believe. By default we always throw warnings for experimental entities, unless its switched off like in PE. This applies to event_counts and aggregate_event_counts.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1543,6 +1578,39 @@
...
+  (cm/match
+   query
+   ["from" (entity-str :guard #(string? %)) & remaining-query]
+   (let [remaining-query (cm/match
+                          remaining-query
+                          [(q :guard #(vector? %))] q
+                          [] []
+                          :else (throw (IllegalArgumentException. "Your `from` query accepts an optional query only as a second argument. Check your query and try again.")))
+         entity (keyword (utils/underscores->dashes entity-str))]
+     (when warn
```
senior: oh, warn-experimental is for experimental entities only? not for things like from?

kbarber:
@ajroetker pretty=true works, do you mean its not in the docs?

src/puppetlabs/puppetdb/query_eng/engine.clj
kbarber: It was created I believe to stop warnings just for those endpoints that we know PE uses regularly (at least for now). For the 'from' ... well, its experimental to everyone really, not just FOSS. I think its two different cases generally.

senior:
:+1: just some minor comments

src/puppetlabs/puppetdb/query_eng.clj
kbarber: Well, we have it defaulted in config, so on second thoughts I'll just stop defaulting it and make it mandatory. The decision to do this anyway was fairly arbitrary, it just helped 2-3 test cases, but not really necessary. I'm going to remove the defaulting and make it mandatory - since it did help detect a few cases I noticed where we weren't setting it.

kbarber:
@senior suggested changes now up.

kbarber:
@senior @ajroetker green now, anything I've missed?

src/puppetlabs/puppetdb/http/index.clj
senior: Discussed with the group, it does add some work to move to compojure (or some other lib) but we're going to proceed with it and get the routing lib and refactor in the next sprint or two.

PR #1771, opened 2015-12-03
(PDB-992) Move to HikariCP instead of BoneCP

ajroetker:
This commit moves our connection pooling to use HikariCP instead of
BoneCP. This means we lose our pool metrics for now, until we update our
codehale metrics dependency.

### DIFF
--- a/documentation/api/metrics/v1/mbeans.markdown
+++ b/documentation/api/metrics/v1/mbeans.markdown
@@ -57,9 +57,9 @@ Responses return a JSON Object mapping strings to (strings/numbers/booleans).
 
 ### Database metrics
 
-* `com.jolbox.bonecp:type=BoneCP`: Database connection pool
-  metrics. How long it takes to get a free connection, average
-  execution times, number of free connections, etc.
+* See
+  [the HikariCP documentation](https://github.com/brettwooldridge/HikariCP/wiki/Dropwizard-Metrics)
+  for the metric names and descriptions.
 
 ### Command-processing metrics
 
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -211,9 +211,7 @@
   (with-open [init-db-pool (jdbc/make-connection-pool
                             (assoc write-db-config
                                    ;; Block waiting to grab a connection
-                                   :connection-timeout 0
-                                   ;; Only allocate connections when needed
-                                   :pool-availability-threshold 0))]
+                                   :connection-timeout 0))]
     (let [db-pool-map {:datasource init-db-pool}]
       (initialize-schema db-pool-map config))))
 
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -208,12 +208,20 @@
   being fully started when PuppetDB starts. This connection pool will
   be opened and closed within the body of this function."
   [write-db-config config]
-  (with-open [init-db-pool (jdbc/make-connection-pool
-                            (assoc write-db-config
-                                   ;; Block waiting to grab a connection
-                                   :connection-timeout 0))]
-    (let [db-pool-map {:datasource init-db-pool}]
-      (initialize-schema db-pool-map config))))
+  (loop [db-spec (assoc write-db-config
+                        ;; Block waiting to grab a connection
+                        :connection-timeout 15000
+                        :pool-name "PDBMigrationsPool")]
+    (if-let [result
+             (try
+               (with-open [init-db-pool (jdbc/make-connection-pool db-spec)]
+                 (let [db-pool-map {:datasource init-db-pool}]
+                   (initialize-schema db-pool-map config)
+                   ::success))
+               (catch java.sql.SQLTransientConnectionException e
+                 (log/error e "Error while attempting to create connection pool")))]
+      result
+      (recur db-spec))))
 
 (defn start-puppetdb
   [context config service get-registered-endpoints]
@@ -230,8 +238,8 @@
         {:keys [dlo-compression-threshold]} command-processing
         {:keys [disable-update-checking]} puppetdb
 
-        write-db (jdbc/pooled-datasource database)
-        read-db (jdbc/pooled-datasource (assoc read-database :read-only? true))
+        write-db (jdbc/pooled-datasource (assoc database :pool-name "PDBWritePool"))
+        read-db (jdbc/pooled-datasource (assoc read-database :read-only? true :pool-name "PDBReadPool"))
         discard-dir (io/file (conf/mq-discard-dir config))]
 
     (when-let [v (version/version)]
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -54,6 +54,7 @@
      :conn-max-age (pls/defaulted-maybe s/Int 60)
      :conn-keep-alive (pls/defaulted-maybe s/Int 45)
      :conn-lifetime (s/maybe s/Int)
+     :maximum-pool-size (pls/defaulted-maybe s/Int 10)
      :classname (pls/defaulted-maybe String "org.postgresql.Driver")
      :subprotocol (pls/defaulted-maybe String "postgresql")
      :subname (s/maybe String)
@@ -96,6 +97,7 @@
    :log-statements Boolean
    :statements-cache-size s/Int
    :connection-timeout s/Int
+   :maximum-pool-size s/Int
    (s/optional-key :conn-lifetime) (s/maybe Minutes)
    (s/optional-key :username) String
    (s/optional-key :user) String
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -68,7 +68,7 @@
      :stats (pls/defaulted-maybe String "true")
      :log-statements (pls/defaulted-maybe String "true")
      :statements-cache-size (pls/defaulted-maybe s/Int 1000)
-     :connection-timeout (pls/defaulted-maybe s/Int 500)}))
+     :connection-timeout (pls/defaulted-maybe s/Int 3000)}))
 
 (def write-database-config-in
   "Includes the common database config params, also the write-db specific ones"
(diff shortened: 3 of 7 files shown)


kbarber:
Did you look into using the shared pooling code here: https://github.com/puppetlabs/jdbc-util ?

kbarber:
Just to note, whoever ends up testing this on its own, we should put this through some destructive testing to be sure it behaves as expected: especially doing things like, killing PG or restarting PG while running - that sort of thing. We should find the edges with the timeouts as well ... in the past PDB would start up, and just keep trying, I'm pretty sure that might change now. Anyway ... need to find the failure behaviours and see if they match our users expectations.

ajroetker:
@kbarber I actually started out trying to use the shared pooling code, and update it for what we need. I found myself running into a lack of documentation for how to run the tests there. Related is that the repo still claims it's a work in progress. Also we would have to release a new version of the utils for us to use the changes. I thought all that work might be saved for another ticket once we transition to HikariCP, this will allow us to review the whole patch in one PR and hopefully catch any edges cases we might miss in the transition.

ajroetker:
@kbarber I can't speak to edge cases that's what I've been looking for myself. Though a note on:
> We should find the edges with the timeouts as well ... in the past PDB would start up, and just keep trying, I'm pretty sure that might change now.
HikariCP has the same behavior as BoneCP here if I understand correctly, in that you can pass a `connection-lifetime = 0` and it will wait forever for a connection, the only difference here is that HikariCP doesn't document this (I had to look in the source code).

kbarber:
@ajroetker in regards to jdbc-utils, I concur with your thinking, which is why I wasn't too worried. Feels like we need to create at least 2 follow up tickets of this work: sync up jdbc-utils with this update & get metrics updated/consider shared metrics library etc.

I think the README is just old, fwiw. But I hear your concerns about stability. I guess we would apply some cleanups like this for our patch perhaps, to take it up a notch :-).

Also, thanks for considering the timeout behaviours, this kind of thing really is my concern. I think though, there were some opinions expressed about this before, let me find the thread.

ajroetker:
Verified that if PuppetDB goes up before the database has gone up, that HikariCP will keep trying to connect until it succeed.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -68,7 +68,7 @@
      :stats (pls/defaulted-maybe String "true")
      :log-statements (pls/defaulted-maybe String "true")
      :statements-cache-size (pls/defaulted-maybe s/Int 1000)
-     :connection-timeout (pls/defaulted-maybe s/Int 500)}))
+     :connection-timeout (pls/defaulted-maybe s/Int 3000)}))
```
ajroetker: The connection timeout for HikariCP cannot be less that `1000ms`.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -360,44 +361,32 @@
...
+           read-only?]
+    :as db-spec}]
+  (let [config (HikariConfig.)]
+    (doto config
+      (.setJdbcUrl (str "jdbc:" subprotocol ":" subname))
+      (.setAutoCommit false)
+      (.setInitializationFailFast false))
+    (some->> connection-timeout (.setConnectionTimeout config))
+    (some->> conn-max-age pl-time/to-millis (.setIdleTimeout config))
+    (some->> conn-lifetime pl-time/to-millis (.setMaxLifetime config))
```
ajroetker: I moved these to being `some->>` because it will make it easier to abstract this code to the puppetlabs/jdbc-util repo.

src/puppetlabs/puppetdb/jdbc/internal.clj
```diff
@@ -2,56 +2,7 @@
...
-   the configure timeout."
-  [conn stmt sql params time-elapsed log-statements? query-execution-limit]
-  (log/warnf (str "Query slower than %ss threshold:  "
-                  "actual execution time: %.4f seconds; Query: %s; "
-                  (query-params->str log-statements? params))
-             query-execution-limit
-             (/ time-elapsed 1000000000.0)
-             sql))
-
-(defn connection-hook
```
ajroetker: This might be where our logging of slow statements disappeared @wkalt, I couldn't see where this code we getting used. Looks like it might have got unhooked at some point.

documentation/api/metrics/v1/mbeans.markdown
```diff
@@ -57,9 +57,9 @@ Responses return a JSON Object mapping strings to (strings/numbers/booleans).
 
 ### Database metrics
 
-* `com.jolbox.bonecp:type=BoneCP`: Database connection pool
-  metrics. How long it takes to get a free connection, average
-  execution times, number of free connections, etc.
+* See
```
kbarber: Well, not true really, with just this patch ;-).

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -360,44 +360,32 @@
...
-           connection-timeout pool-availability-threshold]
-    :as db}]
-  (let [;; Load the database driver class explicitly, to avoid jar load ordering
-        ;; issues.
-        _ (Class/forName classname)
-        log-slow-statements-duration (pl-time/to-seconds log-slow-statements)
-        config          (doto (new BoneCPConfig)
-                          (.setDefaultAutoCommit false)
-                          (.setLazyInit true)
-                          (.setMinConnectionsPerPartition partition-conn-min)
```
kbarber: These extra configuration items, some are no longer needed, should we throw an error, deprecate & retire etc.? We should be making it clear they aren't going to work at least somewhere. log-slow-statements is an example. I think the partition config was _never_ documented.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -360,44 +360,32 @@
...
-                          (.setMaxConnectionsPerPartition partition-conn-max)
```
kbarber: Actually if I'm not mistaken, we now have no pool control options without the `partition-*` options. Perhaps we need to introduce maximum-pool-size, hikari recommends not adjusting minimum pool idle, and maintaining a fixed pool, so perhaps the 1 option is all that is needed.

documentation/api/metrics/v1/mbeans.markdown
ajroetker: Yeah I'll update this section after I add in the metrics-registry integration, this is mostly filler for that PR.

src/puppetlabs/puppetdb/jdbc.clj
ajroetker: I'll deprecate them similar to the way we deprecate classname etc.

src/puppetlabs/puppetdb/jdbc.clj
ajroetker: Noted, I'll add that option in, should we document it as well?

kbarber:
@ajroetker I think the service never enters an 'up' state, when you start without a database, leave it for a few minutes and then start it later on. I curl'd in-between time also, not sure that makes a difference. I couldn't replicate it when starting PG immediately after, the delay seems to be a factor.

src/puppetlabs/puppetdb/jdbc.clj
kbarber: Absolutely, we've been meaning to document the partitioning side of things anyway, now we only want to expose 1 parameter, its easier.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -1,6 +1,6 @@
 (ns puppetlabs.puppetdb.jdbc
```
kbarber: As mentioned in chat, the retry code put in to work around the BoneCP bug for not handling 57P01 can be removed now most probably: https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/jdbc.clj#L300-L305

By having it still there, my testing is under question now, since it might not have been doing the right thing in the pooler, but I believe I've tested a 57P01 case under HikariCP and it works just fine.

ajroetker:
@kbarber Comments should all be addressed now!

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -208,14 +208,20 @@
...
+                        :connection-timeout 15000
+                        :pool-name "PDBMigrationsPool")]
+    (if-let [result
+             (try
+               (with-open [init-db-pool (jdbc/make-connection-pool db-spec)]
+                 (let [db-pool-map {:datasource init-db-pool}]
+                   (initialize-schema db-pool-map config)
+                   ::success))
+               (catch java.sql.SQLTransientConnectionException e
+                 nil))]
```
kbarber: So we only really see this logged at INFO level when it goes wrong:

```
2015-12-09 13:40:12,434 INFO  [main] [c.z.h.HikariDataSource] PDBMigrationsPool - is starting.
2015-12-09 13:38:22,302 INFO  [main] [c.z.h.p.HikariPool] PDBMigrationsPool - is closing down.
```

Unless the user is running DEBUG, then they won't see the persistent retries (which is a good thing) but they also won't see the failure reason ever, probably doesn't leave enough clues for the user. Perhaps we need to catch the exception and log it here at ERROR or something like that?

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -208,14 +208,20 @@
...
-                            (assoc write-db-config
-                                   ;; Block waiting to grab a connection
-                                   :connection-timeout 0
-                                   ;; Only allocate connections when needed
-                                   :pool-availability-threshold 0))]
-    (let [db-pool-map {:datasource init-db-pool}]
-      (initialize-schema db-pool-map config))))
+  (loop [db-spec (assoc write-db-config
+                        ;; Block waiting to grab a connection
+                        :connection-timeout 15000
```
kbarber: This retry behaviour seems pretty good now.

src/puppetlabs/puppetdb/cli/services.clj
ajroetker: Yeah that's reasonable, I'll add that in here.

ajroetker:
@kbarber addressed your comments!

PR #1775, opened 2015-12-04
(PDB-2234) Upgrade our metrics-clojure dependency (to Dropwizard version)

ajroetker:
This commit updates our metrics-clojure dependency from `0.7.0` to
`2.6.0` which uses the dropwizard java metrics library. This update
means that we register our metrics with a metrics registry, which is the
object we will need in order to register HikariCP metrics. This commit
does just enough to upgrade and does not refactor our metrics code.
Notable future changes that should be done in a follow-up patch, would
be to register all our metrics on start up, and have functions which use
those metrics, fetch the metric object from the registry rather than
storing the metric object in the atoms we have today as well as the
registries.

This commits also adds an explicit registering of the MetricsRegistry
with the JMXReporter from Dropwizard metrics, required by their API in
the newest version. We start the reporting during the trapperkeeper init
phase and stop the reporting during the trapperkeeper stop phase.

This commit also cleans up the `puppetlabs.puppetdb.metrics.core`
namespace, to separate concerns with the http handlers which use the
namespace in `puppetlabs.puppetdb.metrics`. This allows us to keep the
`metrics-registry` and `jmx-metrics-reporter` objects in a sane
location, i.e. no cyclic dependencies between the middleware and metrics
namespaces.

**Note** This patch is independent of the HikariCP patch. They may both be merged separately. After they are merged, we'll need to update HikariCP to `.setMetricsRegistry` to the `metrics-registry` in this patch and `.setPoolName` to something whatever we want because the metrics will all be prefixed with the `pool-name`.


ajroetker:
Note I have chosen not to use the `default-registry` provided by the `metrics-clojure` library for 2 reasons. The first is that if we want to move to `trapperkeeper-metrics` eventually, we'll need to pass the registry object as an argument. The second is that the object is explicitly created in our code and we can track the object to its point of origin in our codebase, rather that the mysterious `default-registry` which gets created in the `metrics-clojure` code and would look quite mysterious when we passed it through to HikariCP.

src/puppetlabs/puppetdb/metrics/core.clj
```diff
@@ -30,60 +31,21 @@
...
-       (linkify-names)
-       (into (sorted-map))
-       (http/json-response)))
+       (map #(vector % (format "/mbeans/%s" (url-encode %))))
+       (into (sorted-map))))
 
 (defn get-mbean
   "Returns the attributes of a given MBean"
   [name]
-  (if ((all-mbean-names) name)
```
ajroetker: This refactoring wasn't all 100% necessary, but I was already refactoring parts of it to remove some of the namespace cyclic dependencies I was getting by putting the metrics-registry in this namespace, so I thought it couldn't hurt to tidy this code up a bit.

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -247,7 +248,7 @@
                       (doseq [metric-root metric-roots
                               :let [meter-key [:meters metric-root status]]
                               :when (not (get-in @storage meter-key))]
-                        (swap! storage assoc-in meter-key (meter [prefix metric-root (str status)] "reqs/s"))
```
ajroetker: Not sure where this was surfaced before but note that we no longer specify `reqs/s` in the new `meter` interface.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -167,34 +168,35 @@
...
+   :gc-environments    (timer metrics-registry [ns-str "default" "gc-environments-time"])
+   :gc-report-statuses (timer metrics-registry [ns-str "default" "gc-report-statuses"])
+   :gc-fact-paths  (timer metrics-registry [ns-str "default" "gc-fact-paths"])
+
+   :updated-catalog    (counter metrics-registry [ns-str "default" "new-catalogs"])
+   :duplicate-catalog  (counter metrics-registry [ns-str "default" "duplicate-catalogs"])
+   :duplicate-pct      (gauge-fn metrics-registry [ns-str "default" "duplicate-pct"]
+                                 (fn []
+                                   (let [dupes (value (:duplicate-catalog performance-metrics))
+                                         new   (value (:updated-catalog performance-metrics))]
```
ajroetker: I just would like to note that this self-referential code bothered me. :smile: When I updated the library the first error I got was from this causing a `NullPointerException`.

test/puppetlabs/puppetdb/command/dlo_test.clj
```diff
@@ -111,8 +113,8 @@
           (is (empty? (archives subdir)))
           (is (empty? (archives other-subdir)))
 
-          (is (= 0 (.count compression)))
-          (is (= 0 (.count failures))))
+          (is (= 0 (timers/number-recorded compression)))
+          (is (= 0.0 (meters/rate-one failures))))
```
ajroetker: There is no `.count` method on the object or equivalent retrieval function in the `metrics-clojure` library anymore, so float comparison using the 1 minute rate was the best I could come up with.

kbarber:
@ajroetker I haven't had the chance to review this yet, I'm wondering though - is this entangled with the HikariCP patch in any way? Can it be reviewed out-of-band from that patch?

ajroetker:
@kbarber Put a comment in the PR description, but yeah they are independent.

senior:
I'll take a look at this

src/puppetlabs/puppetdb/scf/storage.clj
senior: Should we just split it into two? This seems like it will cause problems.

senior:
We should talk about this. This changes the names of all of our mbeans. As an example, the old mbean was called:

```
puppetlabs.puppetdb.query.population:type=default,name=num-nodes
```

New one is:

```
metrics:name=puppetlabs.puppetdb.query.population.default.num-nodes
```

In the past we've talked about having our own mbeans follow semver, which would mean we'd want to deprecate the old ones? Not sure it's worth it, but we should discuss it. Also because of these name changes, most of the mbeans on the dashboard are broken

ajroetker:
## problem

Dropwizard doesn't allow us to set custom JMX attributes, such as `group=`. Also we have to make a whole new `MetricsRegistry` and  `JMXReporter` object (which we need to start and stop alongside PuppetDB) for every new domain we wish to create, i.e. classname in `<classname>:group=...,name=...`.

## today

### jconsole

- [message-queue, http, population and storage](https://www.dropbox.com/s/d0h3fn652yrohd3/Screenshot%202015-12-14%2016.42.09.png?dl=0)
- [message-queue and http](https://www.dropbox.com/s/3sbvnjcbo68y9qy/Screenshot%202015-12-14%2016.42.32.png?dl=0)
- [http and population](https://www.dropbox.com/s/m6pavrblgpahaqo/Screenshot%202015-12-14%2016.43.09.png?dl=0)
- [storage and population](https://www.dropbox.com/s/mfwxs2rr2wgn3yr/Screenshot%202015-12-14%2016.43.30.png?dl=0)

### structure

- HTTP:
    - `<namespace>:group=<endpoint>,name=<status>`
    - `<namespace>:group=<endpoint>,name=<misc>`
- DLO:
    - `puppetlabs.puppetdb.command.dlo:type=<directory>,name=<misc>`
    - `puppetlabs.puppetdb.command.dlo:type=<subdirectory>,name=<misc>`
- MQ:
    - `puppetlabs.puppetdb.command:type=<command>.<version>,name=<misc>`
    - `puppetlabs.puppetdb.command:type=global,name=<misc>`
- POPULATION:
    - `puppetlabs.puppetdb.query.population:type=default,name=<misc>`
- STORAGE:
    - `puppetlabs.puppetdb.scf.storage:type=default,name=<misc>`

## proposed

### jconsole

- [database and http](https://www.dropbox.com/s/a5sa3aym60fgizh/Screenshot%202015-12-14%2016.13.44.png?dl=0)
- [message-queue, population and storage](https://www.dropbox.com/s/5kh8bflkwudqycr/Screenshot%202015-12-14%2016.21.20.png?dl=0)
- [population and storage](https://www.dropbox.com/s/eui2cndmj88vr19/Screenshot%202015-12-14%2016.22.04.png?dl=0)

### structure
- DB:
    - `puppetlabs.puppetdb.database:PDBWritePool.<HikariCP metric>`
    - `puppetlabs.puppetdb.database:PDBReadPool.<HikariCP metric>`
- HTTP: **Jconsole grouping affected**: This one is more significant than the others, it used to be that every namespace which called `wrap-with-metrics` got a new top level `<namespace>:type=<endpoint>`. We could make decide to make a registry (which corresponds to the `<namespace>` domain) for each API we have (i.e. `/pdb/cmd`, `/pdb/query`, `/pdb/admin`, `/pdb/ext`....).
    - `puppetlabs.puppetdb.http:name=<endpoint e.g. /pdb/query/v4>.<misc e.g. 200>`  
or
   - `puppetlabs.puppetdb.http.cmd:...`
   - `puppetlabs.puppetdb.http.query:...`
   - `puppetlabs.puppetdb.http.admin:...`
- DLO: **Jconsole grouping affected**
    - `puppetlabs.puppetdb.dlo:name=<directory>.<misc>`
- MQ: **Jconsole grouping affected** Similar for HTTP metrics we could decide to make a registry for each of the commands.
    - `puppetlabs.puppetdb.mq:name=<command>.<version>.<misc>`
    - `puppetlabs.puppetdb.mq:name=global.<misc>`  
or  
    - `puppetlabs.puppetdb.mq.replace-catalog:name=<version>.<misc>`
    - `puppetlabs.puppetdb.mq.global:name=<misc>`  
    - `puppetlabs.puppetdb.mq.replace-facts:name=<version>.<misc>`
    - `puppetlabs.puppetdb.mq.store-report:name=<version>.<misc>`  
- POPULATION:
    - `puppetlabs.puppetdb.population:name=<misc>`
- STORAGE:
    - `puppetlabs.puppetdb.storage:name=<misc>`

ajroetker:
ping @wkalt @senior @kbarber I've posted a really long comment here detailing what JConsole looks like before and after this patch, I also give alternatives for what the API could look like. I think if we're breaking the metrics API anyways we might as well decide on a good naming scheme for our metrics.

wkalt:
@ajroetker leaving aside the question of a compatibility layer, I think my preference for the http endpoints would be to continue embedding the full query path in the names.

For the DLO, I'm not sure what you mean by <directory> and the names you've listed as existing don't match what I see in the output of /metrics/v1/mbeans, but I think we should continue to call it something general like "discarded" so that the same queries are applicable across installations with different vardirs. Good chance I'm misunderstanding the proposal here though.

For MQ, my preference would be for the first of your two options, mainly because it seems more consistent with my preference on http. Population and storage seem fine to me as proposed.

Also tests are failing on this now -- looks like we probably failed to remove ns-str somewhere.

senior:
I agree with @wkalt on the first MQ option you had there, seems good to me since our number of commands is fairly small

+1 on not including the vardir in the mbean names for the DLO

ajroetker:
@wkalt @senior Updated the tests, dashboard, and documentation. Should be all ready for review now.

PR #1777, opened 2015-12-08
(PDB-2239) Use encode for handling bytea data

ajroetker:
This commit changes our handling of bytea from using `trim` (which
relies implicitly on `bytea_output` being set to `'hex'` in PostgreSQL)
to instead use `encode(...,'hex')`. This allows our `bytea` usage to be
database setting independent.

This commit also cleans up some of the `fact_values` code in storage.clj
which was using `parse-db-hash` which is also not `bytea_output`
independent.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -948,50 +948,49 @@
 
 (defn existing-row-ids
   "Returns a map from value to id for each value that's already in the
-   named database column.
-   `column-transform` is used to modify the sql for the values"
-  [table column values column-transform java-type sql-type]
+   named database column."
+  [table column values]
   (into {}
         (for [{:keys [value id]}
               (if (sutils/postgres?)
-                (query-to-vec
-                  (format
-                    "WITH values AS (SELECT unnest(?) as vals)
-                     SELECT %s AS value, id FROM %s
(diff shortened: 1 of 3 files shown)


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -948,50 +948,49 @@
...
+                (query-to-vec (format
+                               "WITH values AS (SELECT unnest(?) as vals)
+                                SELECT %s AS value, id FROM %s
+                                INNER JOIN values ON values.vals = %s"
+                               column table column)
+                              (sutils/array-to-param "text" String values))
                 (apply query-to-vec
                        (format "SELECT %s AS value, id FROM %s WHERE %s %s"
-                               (column-transform column) (name table) column (jdbc/in-clause values))
+                               column table column (jdbc/in-clause values))
```
wkalt: this is definitely a quibble, but why remove (name table)? Seemed like that allowed consistency with all the other jdbc functions.

wkalt:
+1

src/puppetlabs/puppetdb/scf/storage.clj
ajroetker: I was mostly trying to figure out what each of these arguments was doing and saw that we didn't actually use the fact it was a keyword anywhere :ship: I just tried to strip everything to the bare minimum.

PR #1779, opened 2015-12-08
(maint) Cleanup our acceptance tests

ajroetker:
This commit removes some embedded database testing remanents and changes
some of our existing beaker code to use proper beaker helpers.


rbrw:
Plus or minus my understanding of Ruby, etc., this looks pretty good to me.  Unless we want someone with a bit more experience there to take a look, I'd likely merge it.

PR #1781, opened 2015-12-10
(PDB-163) Add the `in`-`array` operator to the query language

ajroetker:
This commit adds the `in`-`array` operator to the query language.

For example:

```json

        ["from", "nodes",
            ["subquery", "resources",
                ["in", "type",
                    ["array",
                      ["File",
                       "Apache::Mode",
                       "Apt::Pin"]]]]]

```

### DIFF
--- a/documentation/api/query/v4/operators.markdown
+++ b/documentation/api/query/v4/operators.markdown
@@ -306,6 +306,8 @@ An `in` statement constitutes a full query string, which can be used alone or as
    the subquery for matching against the **fields** in the `in` clause.
 ** a **`from` statement,** which sets the context, and allows for an extract
    statement to be provided. *Note:* this syntax is new and experimental.
+** an **`array` statement,** which acts as a list of values to match against the
+   **field** in the `in` clause.
 
 **Matches if:** the field values are included in the list of values created by the `extract` or `from` statement.
 
@@ -332,8 +334,8 @@ which is equivalent to the following query:
      ["=","certname","bar.local"],
      ["=","certname","baz.local"]]
 
-The in operator support much of the same syntax as the `=` operator. For
-example, the following query on the `/nodes` endpoint is valid:
+The `in`-`array` operators support much of the same syntax as the `=` operator.
+For example, the following query on the `/nodes` endpoint is valid:
 
     ["in", ["fact", "uptime_seconds"], 
      ["array",
--- a/documentation/api/query/v4/operators.markdown
+++ b/documentation/api/query/v4/operators.markdown
@@ -299,19 +299,57 @@ An `in` statement constitutes a full query string, which can be used alone or as
 
 "In" statements are **non-transitive** and take two arguments:
 
-* The first argument **must** consist of one or more **fields** for the endpoint or entity **being queried.**. This is a string or vector of strings.
+* The first argument **must** consist of one or more **fields** for the endpoint
+  or entity **being queried.**. This is a string or vector of strings.
 * The second argument **must** be either:
-** an **`extract` statement,** which acts as a list of fields to extract during the subquery for matching against the **fields** in the `in` clause.
-** a **`from` statement,** which sets the context, and allows for an extract statement to be provided. *Note:* this syntax is new and experimental.
+** an **`extract` statement,** which acts as a list of fields to extract during
+   the subquery for matching against the **fields** in the `in` clause.
+** a **`from` statement,** which sets the context, and allows for an extract
+   statement to be provided. *Note:* this syntax is new and experimental.
 
 **Matches if:** the field values are included in the list of values created by the `extract` or `from` statement.
 
+##### `array`
+
+An `in` statement also accepts an `array` statement as a second argument.
+
+"Array" statements take a single vector argument of values to match the first
+argument of `in` against.
+
+The following query filters for the nodes, `foo.local`, `bar.local`, and
+`baz.local`:
+
+    ["in", "certname",
+     ["array",
+      ["foo.local",
+       "bar.local",
+       "baz.local"]]]
+
+which is equivalent to the following query:
+
+    ["or",
+     ["=","certname","foo.local"],
+     ["=","certname","bar.local"],
+     ["=","certname","baz.local"]]
+
+The in operator support much of the same syntax as the `=` operator. For
+example, the following query on the `/nodes` endpoint is valid:
+
+    ["in", ["fact", "uptime_seconds"], 
+     ["array",
+      [20000.0,
+       150.0,
+       300000]]]
+
 #### `from`
 
-*Note:* the use of `from` in a subquery is experimental. It may change or be removed in the future.
+*Note:* the use of `from` in a subquery is experimental. It may change or be
+ removed in the future.
 
-This statement works like the top-level [`from`](#context-operators) operator, and expects an [entity][entities] as the first argument and an optional query in the second
-argument, however when used within an `in` clause an `extract` statement is expected to choose the fields like so:
+This statement works like the top-level [`from`](#context-operators) operator,
+and expects an [entity][entities] as the first argument and an optional query in
+the second argument, however when used within an `in` clause an `extract`
+statement is expected to choose the fields like so:
 
     ["in", "certname",
      ["from", "facts",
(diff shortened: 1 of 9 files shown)


project.clj
```diff
@@ -79,7 +79,7 @@
                  [com.novemberain/pantomime "2.1.0"]
                  [fast-zip-visit "1.0.2"]
                  [robert/hooke "1.3.0"]
-                 [honeysql "0.5.2"]
+                 [honeysql "0.6.2"]
```
ajroetker: I originally updated `honeysql` for the `array` function added in `0.5.3` but I didn't need it. Doesn't hurt to leave the upgrade though.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -112,8 +112,7 @@
   [q]
   (and
    (vector? q)
-   (string? (first q))
-   (every? (complement coll?) (rest q))))
```
ajroetker: `params` to the query, (i.e. `(rest q)`), can be an array now

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -218,11 +219,11 @@
                              "environment" {:type :string
                                             :queryable? true
                                             :field :env.environment}
-                             "value_integer" {:type :number
+                             "value_integer" {:type :integer
                                               :query-only? true
                                               :queryable? false
                                               :field :fv.value_integer}
-                             "value_float" {:type :number
+                             "value_float" {:type :float
```
ajroetker: We handle `floats` and `integers` differently in array construction, so we need to differentiate now.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1546,19 +1609,19 @@
              :state state}
             :else {:node node :state state}))
 
-(pls/defn-validated op-to-string
-  "Convert an operator to a lowercase string or vector of such."
-  [op]
-  (if (vector? op)
-    (mapv string/lower-case op)
```
ajroetker: Removed this because I couldn't find anywhere it was used (it would be used for something like `[["not=", "or", "and"], "certname", "foo"]`....)

documentation/api/query/v4/operators.markdown
```diff
@@ -299,19 +299,57 @@ An `in` statement constitutes a full query string, which can be used alone or as
...
+** an **`extract` statement,** which acts as a list of fields to extract during
+   the subquery for matching against the **fields** in the `in` clause.
+** a **`from` statement,** which sets the context, and allows for an extract
+   statement to be provided. *Note:* this syntax is new and experimental.
 
 **Matches if:** the field values are included in the list of values created by the `extract` or `from` statement.
 
+##### `array`
+
+An `in` statement also accepts an `array` statement as a second argument.
```
wkalt: the section for "in" above this contains some must language around the second argument that needs to be updated. Doing that may make this redundant.

documentation/api/query/v4/operators.markdown
```diff
@@ -299,19 +299,57 @@ An `in` statement constitutes a full query string, which can be used alone or as
...
+       "baz.local"]]]
+
+which is equivalent to the following query:
+
+    ["or",
+     ["=","certname","foo.local"],
+     ["=","certname","bar.local"],
+     ["=","certname","baz.local"]]
+
+The in operator support much of the same syntax as the `=` operator. For
```
wkalt: supports

wkalt:
this seems not to work for arrays of hashes

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1080,6 +1086,29 @@
...
+                           :else (throw (IllegalArgumentException.
+                                         "All values in 'array' must be the same type.")))]
+              ["in" "certname"
+               ["extract" "certname"
+                ["select_facts"
+                 ["and"
+                  ["=" "name" fact-name]
+                  clause]]]])
+
+
```
wkalt: extra line here

wkalt:
putting a nonexistent column after in when used with an array produces an NPE

ajroetker:
Looks like UUID columns have the same problem, I'll take a look and add some more tests.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -187,10 +188,15 @@
...
   then that satisfies the match."
-  [orig-table _ column]
+  [column]
   (hcore/raw
    (format "EXISTS(SELECT 1 FROM UNNEST(%s) WHERE UNNEST ~ ?)" (name column))))
 
+(defn sql-in-array
+  [column]
+  (hcore/raw
+   (format "EXISTS(SELECT 1 from UNNEST(?) WHERE unnest = %s)" (name column))))
```
wkalt: Would this be better implemented as
```
(format "%s = ANY(?)" (name column))
```
?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1523,6 +1583,9 @@
                 {:node node
                  :state (conj state column-validation-message)}))
 
+            [["in" _ ["array" _]]]
+            nil
```
wkalt: seems like the  field validation should go here

src/puppetlabs/puppetdb/scf/storage_utils.clj
ajroetker: I thought we didn't want to use that because ANY didn't use indices:

```bash
puppetdb=# explain select * from reports where exists(select * from unnest(ARRAY[1,2]) inner join reports on unnest.unnest = reports.id);
                                               QUERY PLAN
---------------------------------------------------------------------------------------------------------
 Result  (cost=1.46..11.96 rows=50 width=1347)
   One-Time Filter: $1
   InitPlan 1 (returns $1)
     ->  Nested Loop  (cost=0.14..66.00 rows=50 width=0)
           ->  Function Scan on unnest  (cost=0.00..1.00 rows=100 width=4)
           ->  Index Only Scan using reports_pkey on reports reports_1  (cost=0.14..0.64 rows=1 width=8)
                 Index Cond: (id = unnest.unnest)
   ->  Seq Scan on reports  (cost=0.00..10.50 rows=50 width=1347)
(8 rows)

Time: 1.117 ms
puppetdb=# explain select * from reports where reports.id = ANY(ARRAY[1,2]);
                        QUERY PLAN
-----------------------------------------------------------
 Seq Scan on reports  (cost=0.00..10.62 rows=2 width=1347)
   Filter: (id = ANY ('{1,2}'::integer[]))
(2 rows)
```

wkalt: no, I don't think that's true at all
```
puppetdb_5000_672_310=# explain analyze select * from reports where certname = ANY(ARRAY['host-129']);
                                                            QUERY PLAN                                                            
----------------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on reports  (cost=17.52..2592.95 rows=657 width=212) (actual time=0.119..0.480 rows=672 loops=1)
   Recheck Cond: (certname = ANY ('{host-129}'::text[]))
   Heap Blocks: exact=672
   ->  Bitmap Index Scan on reports_certname_idx  (cost=0.00..17.36 rows=657 width=0) (actual time=0.068..0.068 rows=672 loops=1)
         Index Cond: (certname = ANY ('{host-129}'::text[]))
 Planning time: 0.062 ms
 Execution time: 0.539 ms
(7 rows)
```

ajroetker:
Addressed your comments @wkalt

wkalt:
+1

documentation/api/query/v4/operators.markdown
```diff
@@ -299,19 +299,59 @@ An `in` statement constitutes a full query string, which can be used alone or as
...
+An `in` statement also accepts an `array` statement as a second argument.
+
+"Array" statements take a single vector argument of values to match the first
+argument of `in` against.
+
+The following query filters for the nodes, `foo.local`, `bar.local`, and
+`baz.local`:
+
+    ["in", "certname",
+     ["array",
```
kbarber: So how much have you extrapolated this new operator/declaration to other scopes, for example, `containment_path` in `reports` is an array, and equality matching there could apply with this new 'literal' declaration. Going down another similar road, would one use 'map' to declare a map, for equality matching as well, for example? All I'm stating here is that for consistency, we would end up using this literal declarator everywhere to indicate an array, and I just want to make sure this is the correct thing.

ajroetker: So I think there's some precedence for the literal decorator with `["fact", "..."]`. Also I think without the decorator could cause problems, i.e. `[in ["fact" "name"] ["extract", ...` becomes ambiguous, is `extract` a keyword or is it a member of an array literal? we could use `["array" &rest]`, such as `[array foo bar`] which means the literal `[foo bar]`, sort of like the clojure vector constructor, but then this means a `map` would start to look like `[map [foo baz] [bar buzz]`, instead of `{foo baz ...}`. I thought of `array` and `fact`, as casting operators, converting something that looks ambiguous. What would be your concerns with a literal declarator?

kbarber: I think a literal decorator is fine, if its required to remove ambiguity. We'll want to use the syntax elsewhere though, if we need to provide a literal - the main case I can think of is for equality TBH. Anyway - just wanted to ponder it, I think this is fine.

PR #1782, opened 2015-12-10
(PDB-2251) Document unchanged resources for PE 2015.3.0 (Ankeny)

ajroetker:
This commit documents unchanged resources information for PE Ankeny in
the FOSS documentation.

### DIFF
--- a/documentation/_puppetdb_nav.html
+++ b/documentation/_puppetdb_nav.html
@@ -39,6 +39,7 @@
       <li><a href="/puppetdb/{{ puppetdb_version }}/scaling_recommendations.html">Scaling Recommendations</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/repl.html">Debugging with Remote REPL</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/load_testing_tool.html">Load Testing</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/resource_statuses.html">Resource Statuses (PE only)</a></li>
     </ul>
   </li>
   <li><strong>Troubleshooting</strong>
--- a/documentation/api/query/v4/reports.markdown
+++ b/documentation/api/query/v4/reports.markdown
@@ -102,9 +102,17 @@ is of the form:
       "configuration_version": <catalog identifier>,
       "certname": <node name>,
       "resource_events": <expanded resource events>,
+      "resources" : <expanded resources (PE only)>
       "metrics" : <expanded metrics>,
       "logs" : <expanded logs>
     }
+
+> **Note: The `resources` field is Puppet Enterprise (PE) only**
+>
+> The response format in PE contains an additional field, `resources`, which
+> contains all the resource statuses for the Puppet run corresponding to the
+> `report`, both changed and unchanged resources.
+
 
 > **Note: Expansion and the `data` field**
 >
@@ -129,6 +137,36 @@ The `<expanded resource events>` object is of the following form:
         "containment_path": <containment heirarchy of resource within catalog>
       } ... ]
     }
+
+The `<expanded resources>` object is of the following form:
+
+    {
+      "href": <url>,
+      "data": [ {
+        "timestamp": <timestamp (from agent) at which event occurred>,
+        "resource_type": <type of resource event occurred on>,
+        "resource_title": <title of resource event occurred on>,
+        "containment_path": <containment heirarchy of resource within catalog>
+        "skipped" : <boolean for whether or not the resource was skipped>, 
+        "events" : [<event> ...]
+      } ... ]
+    }
+
+where an `<event>` object is of the form:
+
+    {
+            "timestamp": <timestamp (from agent) at which event occurred>,
+            "property": <property/parameter of resource on which event occurred>,
+            "new_value": <new value for resource property>,
+            "old_value": <old value of resource property>,
+            "status": <status of event (`success`, `failure`, or `noop`)>,
+            "message": <description of what happened during event>
+    }
+
+> **Note: On `resources` versus `resource_events`**
+>
+> Unchanged resources are accessed through the `resources` field, the
+> `resource_events` field does not contain this information.
 
 The `<expanded metrics>` object is as follows:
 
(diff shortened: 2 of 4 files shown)


documentation/puppetdb_connection.markdown
```diff
@@ -64,6 +64,21 @@ This setting can let the Puppet master stay partially available during a PuppetD
 
 The default value is false.
 
+## Differences in Puppet Enterprise
```
kbarber: Doesn't this bundle server & port into the 'differences in puppet enterprise'?

documentation/pe_puppetdb_notes.markdown
```diff
@@ -0,0 +1,16 @@
+---
```
kbarber: Is this release notes? Or rather, what is your overall goal for this file?

We should include an index change for this in _puppetdb_nav.html somewhere, so people can find it directly right?

documentation/pe_puppetdb_notes.markdown
```diff
@@ -0,0 +1,16 @@
...
+title: "PuppetDB 3.2 » PE PuppetDB Notes"
+layout: default
+canonical: "/pe-puppetdb-notes"
```
kbarber: /puppetdb/latest/pe-puppetdb-notes.html is what its meant to look like right?

documentation/pe_puppetdb_notes.markdown
ajroetker: PE specific notes, more like a directory of PE features/differences for users to use for navigation, but yeah I imagine this will end up mostly looking like release notes.

documentation/puppetdb_connection.markdown
```diff
@@ -64,6 +64,21 @@ This setting can let the Puppet master stay partially available during a PuppetD
...
+
+### `include_unchanged_resources`
```
kbarber: Another way to do this, have the setting on its own, but say (PE Only) in the title. Just thinking out loud.

documentation/api/query/v4/reports.markdown
```diff
@@ -310,3 +310,79 @@ this route.
 This query endpoint supports paged results via the common PuppetDB paging
 URL parameters.  For more information, please see the documentation
 on [paging][paging].
+
+## Differences in Puppet Enterprise
+
+### Response format
+
+The response format in PE contains an additional field, `resources`, which
+contains all the unchanged resources for the Puppet run corresponding to the
```
kbarber: Wait, isn't it all resource states? Not just unchanged resources? I mean, resources that change will also be represented here won't they?

documentation/puppetdb_connection.markdown
ajroetker: I'm not sure, I went back and forth on that, which is why I made PE notes, I think we want both a "All the PE things" doc and we want pe features to be in places where a user would expect to find them (like in the reports doc).

ajroetker: For clarification the PE notes doc is intended to be a `PE only` document.

documentation/api/query/v4/reports.markdown
ajroetker: They will, just a typo

documentation/pe_puppetdb_notes.markdown
```diff
@@ -0,0 +1,16 @@
...
+---
+
+[reports]: ./api/query/v4/reports.html
+[terminus]: ./puppetdb_connection.markdown#include_unchanged_resources
+
+## 2015.3.0
```
kbarber: Hmmm .... a user has already navigated to PDB 3.2 somehow, which yes applies to 2015.3.0. I dunno, this feels like release notes to me, but PE specific ... is a different file really the correct way to do this? I mean, if its chronological, then its release notes ... we can easily add the PE version to release notes perhaps reference PE versions. See what I mean?

If its a list of differences between PE and FOSS, then that makes sense, but then its not as useful chronologically, its more useful as a list of features or differences or something.

kbarber: I mean, do we want two release notes, PE notes already go into our release notes ... we can make them bold, put them in a section, whatever. Having two files means two files to update.

yeah ... I think a release note format is probably the wrong approach anyway for these kind of notes, expressing differences between FOSS and PE should be a new layout of some kind. If its required at all. I can see expressing large-set 'features' that require multiple configuration items to be enabled is not a bad thing.

kbarber: Hmm, there are other large-set features that might span multiple config areas that aren't PE btw. Like structured facts, trusted nodes. Just saying, these concepts of 'change these 3 things to turn on this featureset' aligns better with my thinking, and having a PE only section isn't bad either. I dunno I dunno, I need to ponder this some more. I think I know what you're trying to achieve though.

ajroetker: I'm not really opinionated here, we just need somewhere to tell users about unchanged resources as a feature and how to turn it off, to address the concern of CS folks. Wherever that lives for now is cool. Maybe I'll just remove this file and leave the other 2 changes I have, then we can make a different ticket to address the PE docs problem in general?

documentation/puppetdb_connection.markdown
kbarber: So to a PE user, all configuration items are equal. They aren't special, FOSS is, by not having/being able to use the feature. See what I mean? We don't want PE users to have to look in two sections, just one if we can. FOSS users can mentally tick off what they lack by a section header marker ... or a bold "Note:" in the config section (although thats less obvious).

ajroetker:
@kbarber updated per our discussion!

documentation/_puppetdb_nav.html
```diff
@@ -39,6 +39,7 @@ <h2 id="puppetdb">PuppetDB {{ puppetdb_version }}</h2>
       <li><a href="/puppetdb/{{ puppetdb_version }}/scaling_recommendations.html">Scaling Recommendations</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/repl.html">Debugging with Remote REPL</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/load_testing_tool.html">Load Testing</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/resource_statuses.html">Resource Statuses **PE FEATURE**</a></li>
```
kbarber: So you've followed this schema elsewhere also, and I think the positioning is correct, I'm just worried that bold + caps will be like shouting. I was thinking literally: Resources Statuses (PE only) <- same concept a bit more subtle.

documentation/api/query/v4/reports.markdown
```diff
@@ -102,10 +102,18 @@ is of the form:
       "configuration_version": <catalog identifier>,
       "certname": <node name>,
       "resource_events": <expanded resource events>,
+      "resources" : <expanded resources **PE FEATURE**>
```
kbarber: This is a pre-tag btw, the bold markers won't work.

documentation/_puppetdb_nav.html
wkalt: +1 to this, don't like the caps. I'd also wonder to what extent "PE" is mainly an internal acronym. My preference would be to spell out Puppet Enterprise in docs, but it's probably a docs question.

documentation/puppetdb_connection.markdown
```diff
@@ -64,7 +64,20 @@ This setting can let the Puppet master stay partially available during a PuppetD
 
 The default value is false.
 
-### `server` and `port`
+### `include_unchanged_resources` **PE FEATURE**
+
+> **Warning:** This setting is for PE use only, using this setting in the
```
wkalt: comma should be a period

documentation/_puppetdb_nav.html
ajroetker: Sounds good :dancer:

wkalt: docs stance is to use Puppet Enterprise on first use in a document. After that PE is ok.

ajroetker:
@kbarber @wkalt Updated with your guys' input and the input from docs!

PR #1784, opened 2015-12-11
(PDB-156) reject-large-commands should be enabled by default

senior:
This commit enables `reject-large-commands` by default. It can still be
disabled and the size of commands to reject can still be set via
`max-command-size`.


kbarber:
Looks good, master is still running, I'll merge when its all green.

PR #1785, opened 2015-12-11
(PDB-2255) Make notable-pdb-event? configurable

rbrw:
Originally notable-pdb-event? wasn't set up to be customized, but now we
want that (i.e. PDB-2215), so make it configurable by
adding *notable-pdb-event?* to testutils.services.


rbrw:
Named the dynamic "*notable-log-event?*" given the differing context, but don't feel strongly about it.

PR #1786, opened 2015-12-15
(maint) log command sizes at debug level

wkalt:
This changes our logging of command sizes from info to debug level.


wkalt:
I only briefly discussed this with @senior, so if we think it's unreasonable it's definitely up for discussion.

senior:
I'm fine with it at debug or info, as long as we can get at it when needed. :+1:

PR #1787, opened 2015-12-16
(PDB-1927) PQL translation to AST

kbarber:
This patch introduces the new Puppet query language to PuppetDB. It provides a basic
parse for this capability, documentation for the new language and a hook in the
index endpoint for the HTTP API side.

As this work is fairly new and extensive, some of the expected work has been
broken into future tickets. For now, this is marked as experimental until we are
happy with the language.


kbarber:
Finalising this work is mainly part of: https://tickets.puppetlabs.com/browse/PDB-1927

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -244,8 +245,12 @@
 (defn post-req->query
   "Takes a POST body and parses the JSON to create a pdb query map"
   [req]
-  (with-open [reader (-> req :body clojure.java.io/reader)]
-    (json/parse-stream reader true)))
+  (-> (with-open [reader (-> req :body clojure.java.io/reader)]
+        (json/parse-stream reader true))
+      (update-in [:query] (fn [query]
```
ajroetker: `update :query`

src/puppetlabs/puppetdb/pql.clj
```diff
@@ -0,0 +1,69 @@
...
+                  (for [r partial-reasons]
+                    (format "%s\n" (print-reason r))))]
+    (join [opening expected freasons preasons])))
+
+(defn parse-json-or-pql-to-ast
+  "Attempt to parse a string as JSON first, else fall back
+  to parsing it as PQL."
+  [query]
+  (try
+    (json/parse-strict-string query true)
```
ajroetker: Could you not just check whether the string starts with "\s*["?

ajroetker: Just seems like a user could easily type `["extract", "certname"` and end up with a pql error.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
```diff
@@ -0,0 +1,139 @@
...
+<lbracket> = '[';
+<rbracket> = ']';
+
+(* Booleans *)
+boolean = true | false;
+true = <"true">;
+false = <"false">;
+
+(* Numbers *)
+integer    = [negative], digits;
```
wkalt: we need to allow for integers expressed in e-notation per json. Currently a comparison on 10e5 will fail but 10.0e5 will succeed.

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -0,0 +1,102 @@
...
+
+(defn transform-extract
+  [& args]
+  ["extract" (apply vector args)])
+
+(defn transform-expr4
+  ;; Single arg? collapse
+  ([data] data)
+  ;; Multiple args? turn it into an or statement
+  ([data & args] (vec (concat ["or"] [data] args))))
```
ajroetker: It looks like this could just be `(concat ["or" data] args`. Similarly below.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
wkalt: numbers of the form 10e+5 and 10.0e+5 should also be legal

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
```diff
@@ -0,0 +1,139 @@
...
+   - aggregate functions
+   - ~>
+   - in <array> *)
+
+(* TODO: the following features are awaiting PDB AST changes first, but
+   generally have support within PDB in one form or another:
+
+   - order by
+   - limit, offset
+   - include_total *)
```
wkalt: not sure if there's a better place to note this, but the aggregate endpoints and their query params belong in one of these lists too

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
```diff
@@ -0,0 +1,139 @@
...
+
+   This file is used by Mark Engelberg's 'Instaparse' parser in clojure, and
+   follows its own form of EBNF, so is not strictly 'EBNF' but an augmentation. *)
+
+(* TODO: the following items are just not supported in the syntax yet:
+
+   - group_by
+   - aggregate functions
+   - ~>
+   - in <array> *)
```
wkalt: also explicit subqueries; can't remember where we landed on that

wkalt: my mistake, looks like it's supported here

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
```diff
@@ -0,0 +1,139 @@
...
+
+(* Boolean operators *)
+<and> = <'and'>;
+<or>  = <'or'>;
+not   = <'!'>;
+
+(* Regexp strings *)
+regexp = <regexpquote>, stringwithoutregexpquote, <regexpquote>
+<stringwithoutregexpquote> = #'(?:[^/\\]|\\.)*';
+<regexpquote> = "/"
```
wkalt: I don't have a problem with it necessarily, but what's the reason for different quoting between PQL and AST for regex?

wkalt:
A few minor comments, but looks good to me generally with respect to code and functional testing aside from AJ's comment on throwing a PQL error on invalid json. I suspect there may be more stuff that comes to light when we document it. Nice work.

ajroetker:
Fwiw also functionally tested this and looks nice to me as well.

src/puppetlabs/puppetdb/http/query.clj
kbarber: I hate that there are two methodologies, btw. It just makes it a question of taste, which is an icky place to be.

src/puppetlabs/puppetdb/pql.clj
kbarber: +1 ... and more so, this would make the old query lang the default, which is a good place to be while this is experimental. Eventually, I could see this reverting in time perhaps.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
kbarber: Oh I thought I had done integers, let me fix. Indeed I'll look into the other issue also.

Still, aren't you chuffed I added e-notation support? :-).

src/puppetlabs/puppetdb/pql/transform.clj
kbarber: There was a dumb dumb reason for it, let me dig in.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
kbarber: You mean aggregate-event-counts? Why and in what way? If you mean we should provide mechanisms that provide aggregate like a capabilities I'm all ears, but making them work in PQL as they do today seems like folly.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
kbarber: Well, I don't know to be honest. There is no difference technically, just that some languages use this syntax (ie. ~ /foo/), and its familiar to some people. Forgetting any technical cost, thats literally where my head was at. I think though, I wouldn't mind so much to drop this, I'm not beholden.

On a similar angle, one could ask a similar related question about single quotes and double quotes ...  right know we have no variable interpolation for example, but we might.

These are the questions that keep me up at night :-). Anyway, I don't mind removing the // ... I think TBH, its probably more common in query languages to use quotes, maybe I'll take a quick survey for arguments sake.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
```diff
@@ -0,0 +1,139 @@
...
+(* TODO: the follow items need fundamental changes to PDB first:
+
+   - descending into hashes, like facts & resource parameters conveniently
+   - modifying child data returned (ie. document aware projection) *)
+
+(* Top level query, this is the entry point for queries *)
+<query> = from;
+
+(* From - Setting context and providing filtering and projection *)
+from = entity, [<whitespace>], where, [[<whitespace>], extract];
```
kbarber: I was considering moving the field list for projection to the front, like so:

```
nodes[certname] { certname = 'foo' }
```

As opposed to:

```
nodes { certname = 'foo'} [certname]
```

Just feels better after typing some examples over and over, and keeps the column names next to the entity type instead of way at the end after the filter.

Any opinions?

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
kbarber: I've fixed support for the + modifier no problem.

But there is a gotcha with support integer e notation, in that the Java Integer parseInt support doesn't support it. Hilarious really, but then I started to realise something ... what is 1E-5? Its a float ... but ... 1E5, its an integer. By trying to 'figure out' its still an integer and doing something magical, this can create some confusion, or we could just presume anything with E notation is a real. 

```
puppetlabs.puppetdb.core=> (Double. "1e-1")
0.1
puppetlabs.puppetdb.core=> (Integer. "1e-1")
NumberFormatException For input string: "1e-1"  java.lang.NumberFormatException.forInputString (NumberFormatException.java:65)

puppetlabs.puppetdb.core=> (Double. "1e1")
10.0
puppetlabs.puppetdb.core=> (Integer. "1e1")
NumberFormatException For input string: "1e1"  java.lang.NumberFormatException.forInputString (NumberFormatException.java:65)
```

kbarber:
@wkalt re: the integer E notation problem, I'm tempted to bump it for now, since it can be added later if we want ... that is, if you think its important to add E notation for integers I might just create a small ticket for its support for later. What do you think? My feeling at the moment is that we have real E notation working properly, that covers most cases, integer support is additive and well - I'm tempted to wait for people to ask explicitly for it.

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,339 @@
...
+layout: default
+canonical: "/puppetdb/latest/api/query/v4/pql.html"
+---
+
+[entities]: ./entities.html
+[subquery]: #subqueries
+
+> **Experimental Feature**: This featureset is experimental, and it may be altered or removed in
+> future release.
+
```
bodepd: something at the top to specify how this language is going to be used would be helpful. Just reading this, I thought it might be a Puppet language feature.

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,339 @@
...
+Puppet Query Language (PQL) is a query language designed with PuppetDB and Puppet data in mind. It
+provides a string based query language as an alternative to the legacy query language PuppetDB has
+always supported.
+
+## Query Structure
+
+A PQL query has the following structure:
+
+    <entity> [<projection>] { <filter> <modifiers> }
+
```
bodepd: looking at these high level terms, it is not clear what projection is. I'm not sure if I have a better suggestion though, maybe fields?

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,339 @@
...
+As an example, if you wanted to query for `nodes` related data, but only see the `certnames` field
+for each node, with a regex filter across certname:
+
+    nodes[certname] { certname ~ '^web' }
```
bodepd: it would be great to clarify here out of "certname ~ '^web'", if its a filter or a filter and a modifier, it's probably explained further down, but it would be great to clarify the meanings of the term above with an example near their first usage

documentation/api/query/v4/pql.markdown
kbarber: I've broken this down below, immediately in short lines instead of leaving it until the section headers later.

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+### Boolean operators
+
+Boolean operators are used within PQL filters to join conditons together to perform the filtering
+test within PuppetDB.
+
+There are only 3 boolean operators today, in order of natural precedence:
+
+* `!` - performs a logical `not` or negation
+* `and` - performs a logical `and` or conjuction
+* `or` - performs a logical `or` or disjunction
```
senior: Currently this is case sensitive. Is that on purpose?

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+The entity context can also be used within a subquery, see the [subquery] section for more details.
+
+## Projection
+
+The projection part of a query, provides a mechanism to choose a subset of fields that are returned
+or to modify the way those fields are displayed with the usage of functions.
+
+The projection, lives within the brackets of a query:
+
+    nodes[<projection>] {}
```
senior: This is space sensitive right now, which can be unintuitive:

```
nodes[ certname]{ certname in ['host-0','host-1']}
      ^

Expected one of:

max
min
sum
avg
count
#"[a-zA-Z_]+"
```

```
curl -X GET http://localhost:8080/pdb/query/v4\?pretty\=true --data-urlencode "query=nodes[certname ]{ certname in ['host-0','host-1']}"
PQL parse error at line 1, column 16:

nodes[certname ]{ certname in ['host-0','host-1']}
               ^

Expected:
,
```

```
 curl -X GET http://localhost:8080/pdb/query/v4\?pretty\=true --data-urlencode "query=nodes[]{ certname in ['host-0','host-1']}"
PQL parse error at line 1, column 7:

nodes[]{ certname in ['host-0','host-1']}
      ^

Expected one of:

max
min
sum
avg
count
#"[a-zA-Z_]+"
```
Funny, it's tolerated between fields in the projection:

```
curl -X GET http://localhost:8080/pdb/query/v4\?pretty\=true --data-urlencode "query=nodes[deactivated , certname]{ certname in ['host-0','host-1']}"
```

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+defines the results returned. For example, if you wanted to see all node information, you could
+provide a query as follows:
+
+    nodes {}
+
+And it would be enough to return all node data, without filtering or pagination.
+
+The entity context can also be used within a subquery, see the [subquery] section for more details.
+
+## Projection
```
senior: How would you feel about an empty projection being the same as a missing projection?

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+
+Puppet Query Language (PQL) is a query language designed with PuppetDB and Puppet data in mind. It
+provides a string based query language as an alternative to the [AST query language][ast] PuppetDB has
+always supported.
+
+## Query Structure
+
+A PQL query has the following structure:
+
+    <entity> [<projection>] { <filter> <modifiers> }
```
senior: Should be tolerant of a leading space as well.

documentation/api/query/v4/index.markdown
```diff
@@ -55,6 +57,26 @@ object results based on the [entity][entities] provided in the top-level [`from`
       }
     ]
 
+The same query can also be executed using [PQL][pql]:
+
+    curl -X GET http://localhost:8080/pdb/query/v4 --data-urlencode 'query=nodes { certname = "macbook-pro.local" }'
```
senior: This is the only place in the docs I have found that tell you how to execute a PQL query. Seems like we should at least link to it from the reference.

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+
+For PQL queries, the entity context is the minimal amount of information one must provide, as it
+defines the results returned. For example, if you wanted to see all node information, you could
+provide a query as follows:
+
+    nodes {}
+
+And it would be enough to return all node data, without filtering or pagination.
+
+The entity context can also be used within a subquery, see the [subquery] section for more details.
```
senior: Might be good to be more specific about the reponse formats here since we don't provide a prolog now.

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+    facts { value < 4 }
+
+The operator will only work on numbers however, any other types will return errors.
+
+#### Regexp: `~`
+
+For strings you can match using a regular expression pattern, by using the `~` operator and a valid
+regular expression:
+
+    nodes { certname ~ "foo.*" }
```
senior: I think this is the only place in this doc that you're using double quotes

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+
+The operator can also be used to ensure the values of a field match the fields returned, from a subquery,
+which has the form of a nested PQL query within the filter:
+
+    nodes { certname in facts[certname] { value = 'foo' } }
+
+With the subquery form, we can even match on multiple fields, as long as the fields being matched, and the
+subqueries projection fields match:
+
+    facts { [name, value] in facts[name, value] { value ~ 'foo' } }
```
senior: nice

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+Each field for an entity supports matching using a conditional against a provided literal value.
+
+#### Strings
+
+PQL supports legal UTF-8 strings as literals for comparison.
+
+There are two types of literal strings:
+
+* single quoted - no escaping, just straight text
+* double quoted - supports escape characters
```
senior: Why have both?

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+* `\t` - tab
+* `\uXXXX` - unicode character
+
+For example to match on a string with a newline in it:
+
+    facts { value = "first line\nsecondline" }
+
+#### Booleans
+
+Booleans are represented by using the bar words: `true` or `false`.
```
senior: bare

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+      resources { type = 'Package' and title = 'tomcat' } and
+      certname ~ '^web'
+    }
+
+## Modifiers
+
+Modifiers to the query provide extra facilities beyond a filter for limiting and changing the resultsets
+returned in special ways.
+
+### `group by`
```
senior: modifier doesn't seem like a good term for group by. Maybe it should be combined with functions in an aggregation section?

senior:
I've tested this manually and the functionality looks good. I'll take a look at the code in the morning

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -236,40 +237,53 @@
...
 (defn post-req->query
   "Takes a POST body and parses the JSON to create a pdb query map"
-  [req]
-  (with-open [reader (-> req :body clojure.java.io/reader)]
-    (json/parse-stream reader true)))
+  [req parse-fn]
+  (-> (with-open [reader (-> req :body clojure.java.io/reader)]
+        (json/parse-stream reader true))
+      (update :query (fn [query]
+                            (if (vector? query)
```
senior: I think the spacing is off here

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -236,40 +237,53 @@
...
   (fn [{:keys [request-method body params puppetdb-query] :as req}]
     (handler
      (if puppetdb-query
        req
-       (assoc req :puppetdb-query (create-query-map req param-spec))))))
+       (assoc req :puppetdb-query (create-query-map req param-spec parse-fn))))))
+
+(defn extract-query
+  [handler param-spec]
+  (extract-query* handler param-spec #(json/parse-strict-string % true)))
```
senior: Any reason not to make this multi-arity with this parse-fn the default?

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -0,0 +1,123 @@
...
+  ([entity]
+   ["subquery" entity])
+  ([entity arg2]
+   ["subquery" entity arg2]))
+
+(defn transform-extract
+  [& args]
+  ["extract" (apply vector args)])
+
+(defn transform-expr4
```
senior: seems like transform-or would be a better name than transform-expr4

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -0,0 +1,123 @@
...
+  (:require [clojure.string :as str]
+            [puppetlabs.puppetdb.cheshire :as json]))
+
+(defn transform-from
+  ([entity]
+   ["from" entity])
+  ([entity arg2]
+   ["from" entity arg2])
+  ([entity extract expression]
+   ["from" entity (apply vector (concat extract [expression]))])
```
senior: I think you just want vec here, rather than apply vector

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -0,0 +1,123 @@
...
+  ([entity extract expression modifier]
+   ["from" entity (apply vector (concat extract [expression modifier]))]))
```
senior: Same as above

documentation/api/query/v4/pql.markdown
kbarber: Good question! I thought about this, and I'm almost certain we don't want tokens that represent things, like fields and entities case-insensitive ...

So I think beyond the boolean operators: that leaves keywords like, group by, pagination stuff like limit, offset, order by etc. that could be case-insensitive. Also, there is functions.

I think at least whatever we choose it should be consistent. For me, I'm leaning towards case-sensitive, because it forces a style, however on the flip side, for SQL familiarity case-insensitive might be less surprising, but still - I feel like a firmer style decision isn't a bad thing.

One thing to note, leaving it case-sensitive, means we can change our minds easier later on, if we get demand. Its harder to revert case-insensitivity ...

documentation/api/query/v4/pql.markdown
kbarber: Got it, will fix.

documentation/api/query/v4/pql.markdown
kbarber: I feel good about this, +1

documentation/api/query/v4/pql.markdown
kbarber: +1

documentation/api/query/v4/index.markdown
kbarber: So I've missed out the tutorial part of this patch, let me work on that.

documentation/api/query/v4/pql.markdown
kbarber: So ... a few reasons.

One of the things I notice about languages is that you often want the reverse of the quoting system you're using on the outside of a query string (like in curl) than you do on the inside. So think ... scripts ... bash ... that kind of thing. Having the ability to do that to avoid thinking too hard about escaping is sometime nice. In puppet for example, we support both, and sometimes users might wrap the string in single-quotes, but want a string inside the query, so they use double quotes:

    puppetdb::query("nodes { certname = 'foo' }")
    puppetdb::query('nodes { certname = "foo" }')

The other side of it, is for escaping versus non escaping behaviours. Obviously we should support escaping, as thats quite commonly required, things like carriage returns come to mind ... but sometimes you don't want escaping, single quotes will do.

I guess this feels like a normal/predictable things for me to do, but I can be convinced to drop one of them (perhaps the single quote) if you thought it was designing to far ahead.

documentation/api/query/v4/pql.markdown
kbarber: So, I am very very torn about where 'modifiers' live, maybe we could have a chat about this.

src/puppetlabs/puppetdb/pql/transform.clj
kbarber: So, I did have this once before, and I can go back to it ... I'd also want to update the EBNF to reflect that also. +1.

documentation/api/query/v4/pql.markdown
mullr: Have we had any discussion about a missing filter being the same as an empty filter as well?  So instead of `nodes {}` I could just do `nodes`. Don't really know if there are any further implications of this, just my gut reaction when playing with it.

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+The projection part of a query, provides a mechanism to choose a subset of fields that are returned
+or to modify the way those fields are displayed with the usage of functions.
+
+The projection, lives within the brackets of a query:
+
+    nodes[<projection>] {}
+
+And is a comma seperated list of fields and functions:
+
+    facts[name, count()] { group by name }
```
mullr:
```
bin/puppet-db query "facts [name, count()] { group by name }"
<html>
<head>
<title>Error 500 </title>
</head>
<body>
<h2>HTTP ERROR: 500</h2>
<p>Problem accessing /pdb/query/v4. Reason:
<pre>    org.postgresql.util.PSQLException: ERROR: column "fp.name" must appear in the GROUP BY clause or be used in an aggregate function
  Position: 8</pre></p>
</body>
</html>
error: No error
```

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+be be inefficient for both the database and the network to do this. By providing an entity field
+however, you can reduce what fields are returned.
+
+The entity field section of a query, can contain a number of field names separated by a comma:
+
+    entity[field1, field2, field3] {}
+
+As an example, to return only `certnames` for all nodes:
+
+    nodes[certnames] {}
```
mullr: s/certnames/certname/

documentation/api/query/v4/pql.markdown
```diff
@@ -0,0 +1,361 @@
...
+
+#### `avg(<field>)`
+
+Returns the average value for the values held in the `<field>` argument.
+
+#### `sum(<field>)`
+
+Returns the sum of values for the values held in the `<field>` argument.
+
+#### `min(<field>)`
```
mullr: Maybe this is a known issue, but I can `bin/puppet-db query "nodes [report_timestamp] { }"` where I can't `bin/puppet-db query "nodes [min(report_timestamp)] { }"`

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
wkalt: in practice this may only matter for float comparisons against integer columns like "line", and it seems pretty unlikely those would be expressed this way. In the case of numeric fact values we split it into an or against either column anyway, so the result will be the same up to fp error. To summarize I guess I'm okay with that assumption.

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -0,0 +1,123 @@
...
+
+(defn transform-subquery
+  ([entity]
+   ["subquery" entity])
+  ([entity arg2]
+   ["subquery" entity arg2]))
+
+(defn transform-extract
+  [& args]
+  ["extract" (apply vector args)])
```
ajroetker: Similar to Ryan's comment, this should also be `vec` instead of `apply vector`

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -0,0 +1,123 @@
...
+(defn transform-expr2
+  ;; Single arg? Just collapse the :expr2 and pass back the data,
+  ;; closing the nesting.
+  ([data] data)
+  ;; Two args? This means :not [data] so convert it into a "not"
+  ([_ data] ["not" data]))
+
+(defn transform-function
+  [entity args]
+  (concat ["function" entity] args))
```
ajroetker: should this be `vec (concat`?

ajroetker: On that note a `(def concatv (comp vec concat))` would make a good helper function for this file.

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -0,0 +1,123 @@
...
+(defn transform-condexpnull
+  [entity type]
+  ["null?" entity
+   (case (first type)
+     :condisnull true
+     :condisnotnull false)])
+
+(defn transform-groupedlist
+  [& args]
+  args)
```
ajroetker: should this be `vec` args for consistency with what you have elsewhere?

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -0,0 +1,123 @@
...
+
+(defn transform-groupedliterallist
+  [& args]
+  ["array" args])
```
ajroetker: similarly here?

documentation/api/query/v4/pql.markdown
kbarber: @mullr so, I have thought about it re: bare entities. I kind of feel like a bare empty would feel a bit unintuitive, and originally I was considering always ensuring entities had some other marker. Puppet uses a capitalized convention in some parts (ie. Resources, Nodes) but for resource pointers ... for example ... and that made me thing using that might make more sense, then everything that is capitalized is definitely an entity, not something else like a keyword. Outside of Puppet though, this is pretty weird.

One point - An empty extract is quite common, an empty filter is very uncommon I would presume, at least for a lot of people. Even those who want the house, will want to paginate for example ... and right now the pagination marker is planned to be within the {} (I guess, I'm ... so unsure about this though).

We could go the other way, and require nothing to be optional, and that makes some symmetry:

```nodes[]{}```

Kind of looks ugly though :-). Maybe the extraction should be mandatory, filter optional:

```nodes[]```

Hohum.

mullr: What's the value in making these required? Does it clean up the syntax somehow are reduce ambiguity? I find it pretty clear to say "write the entity you want, write filters in {} and write select in []", allowing it in any particular order (entity first, probably) and allowing you to put in the parts needed. 

I'm not sure I agree that empty filter is uncommon; for interactive use especially, I can easily imagine people starting out with an empty filter and moving on from there. That's certainly what I did. I would expect whatever tool I'm using (pdb-cli, the console, whatever) to automatically apply paging parameters as appropriate.

documentation/api/query/v4/pql.markdown
kbarber: fixed

src/puppetlabs/puppetdb/http/query.clj
kbarber: fixed

src/puppetlabs/puppetdb/http/query.clj
kbarber: fixed

src/puppetlabs/puppetdb/pql/transform.clj
kbarber: fixed

src/puppetlabs/puppetdb/pql/transform.clj
kbarber: fixed

src/puppetlabs/puppetdb/pql/transform.clj
kbarber: fixed

documentation/api/query/v4/pql.markdown
senior: :+1:

documentation/api/query/v4/pql.markdown
senior: Doesn't look like it...

documentation/api/query/v4/pql.markdown
kbarber: As discussed, this work and discussion has moved to here: https://tickets.puppetlabs.com/browse/PDB-2357 and the mailing list discussion

documentation/api/query/v4/pql.markdown
kbarber: @mullr I think you make some good points, I'm moving this conversation into another ticket for now: https://tickets.puppetlabs.com/browse/PDB-2358 ... so we can debate it on its own. This shouldn't be hard to change either way, if we decide its a good idea. I think this will pan out to be a team discussion, then execution on the details we decide. Thanks mate.

documentation/api/query/v4/pql.markdown
kbarber: okay, figured this out - its all about vec sensitivity in the AST side of things, @ajroetker had already spotted it below.

documentation/api/query/v4/pql.markdown
kbarber: This is a not-known issue, and pre-existing also for AST. I'm going to raise a bug on it.

kbarber: Raised it here: https://tickets.puppetlabs.com/browse/PDB-2360.

src/puppetlabs/puppetdb/pql/transform.clj
kbarber: I don't believe it needs to be? It gets passed up as a vec.

kbarber:
That was me cancelling a broken job ^^. Another job is coming through soon.

kbarber:

kbarber:
Last failure due to cancelled an old job ^^

mullr:
+1

PR #1788, opened 2015-12-17
(PDB-2262) Don't assume group-by needs function calls

ajroetker:
This commit fixes an issue in the query language where the group-by
operator assumed a function operator in the columns of the extract.
This was causing `NullPointerException`s for valid queries. This commit
also adds some more tests around this behavior.


senior:
:+1:

PR #1789, opened 2015-12-17
(PDB-2256) Fix performance regression in puppetdb-termini

vzctl:
Show only the first index of non-utf8 characters in the warning.

Performance of the all_indexes_of_char drops non-linearly with larger
strings. The benchmark is available at:

https://gist.github.com/vzctl/3f86e72feab5bca98b06

### DIFF
--- a/puppet/lib/puppet/util/puppetdb/char_encoding.rb
+++ b/puppet/lib/puppet/util/puppetdb/char_encoding.rb
@@ -34,11 +34,6 @@ module CharEncoding
   DEFAULT_INVALID_CHAR = "\ufffd"
 
   # @api private
-  def self.all_indexes_of_char(str, char)
-    (0..str.length).find_all{ |i| str[i] == char}
-  end
-
-  # @api private
   #
   # Takes an array and returns a sub-array without the last element
   #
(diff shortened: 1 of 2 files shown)


olivierHa:
If the non-utf8 characters are at the end of the "analysed" string/catalog, the whole string will be parsed and performance regression will still be present no ?

vzctl:
The problem is not that the whole string gets parsed, it's the way it is parsed.
I included str.index in the benchmark and there is no performance impact even with non-utf8 bytes at the end of the string.

senior:
I started looking into this yesterday and started working up a fix. I think @vzctl is on the right track. One thing that this PR does though is lose information. I think the way we can make this bug go away is to not try to find ALL bad chars, but just the first section of them. I think giving the user a bit of context around the bad characters (before and after) is useful (this PR drops the end char). I think the issue we've got now is catalogs that have a lot of bad characters (such as loading up a lot of binary data) get hung up trying to find them all. I'll put up a PR once I'm done and ping @vzctl and @olivierHa on it. I also want to get this on stable as master won't be released for a while and I'd like to get out a 3.2.3 with this fix in it sooner.

vzctl:
In my use-case we have catalogs with hundreds (thousands?) of non-utf8 characters and I would argue that showing all of them in the warnings is a bad idea.

Honestly, I think showing the first occurrence is the simplest and fastest way to solve the problem, not sure if it makes sense to worry about losing information :)

senior:
@vzctl In your case, I agree the context information isn't very helpful. We don't actually encourage storing binary data in the catalog like you're doing. Generally we recommend that the data be stored outside ofthe catalog using something like [this](https://docs.puppetlabs.com/guides/file_serving.html).

The code's purpose is helping users that have unrecognized character data. Pror to PDB-135, all non-ASCII characters were treated as invalid. It's still possible in the current code to have valid character data that can't get convert and it gets pretty frustrating trying to track down the source of the data.

I just pushed up https://github.com/puppetlabs/puppetdb/pull/1792 which should fix your problem and have no performance impact in your situation and allow users that have the character encoding problem to still get debugging information if they want it. I'll have more discussion on the performance over on that PR.

Thanks for the PR and investigating this issue! We'll get this in the next (3.2.3) release.

PR #1790, opened 2015-12-17
(maint) Remove paging-option parsing duplication

ajroetker:
This commit removes the `wrap-with-paging-options` middleware from
PuppetDB which was duplicating code from the `http/query.clj` namespace
and whose only function was to catch `IllegalArgumentException`s and
return an `error-response`. This commit adds this exception catching to
it's own `wrap-with-illegal-arugment-catch` middleware. This commit also
moves the illegal-argument-catch and the verify-accepts-json middleware
to wrap the whole of the Query-API app which greatly simplifies the code
in each of the endpoints.


src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -112,26 +114,38 @@
...
-  "Given a list of order-by clauses that conform to the PuppetDB
-  paging API, validate and convert the order-by to our internal
-  format. Incoming is of the form:
+(defn parse-order-by
+  "Given a list of order-by clauses that conform to the PuppetDB paging API,
+  validate and convert the order-by to our internal format. Incoming is of the
+  form:
 
-  {:field String (s/optional-key :order) String}
+  ~~~clojure
```
senior: Why the ~~~ and the extra spaces?

ajroetker: I thought our docstrings were in markdown?

PR #1791, opened 2015-12-17
(PDB-2264) Add setting for toggling historical-catalogs storage

ajroetker:
This commit adds a setting for toggling the storage of
historical-catalogs (to be used in PE) to the `puppetdb` section of the
PuppetDB config.


senior:
:+1:

PR #1792, opened 2015-12-18
(PDB-2256) Fix terminus bug with large binary catalog data

senior:
The terminus includes code to help users diagnose the source of data
that cannot be converted to UTF-8. There are several sources of this,
but one is having an incorrect (or known) character set for portions of
the catalog. When this invalid character data is large, it would cause
the terminus to hang trying to include debugging information.

This patch changes the terminus to only look for the first instance of
bad data. It will also avoid doing the extra calculations needed for the
error context unless debug mode is enabled. When not in debug mode there
should be no impact on performance.

### DIFF
--- a/puppet/lib/puppet/util/puppetdb/char_encoding.rb
+++ b/puppet/lib/puppet/util/puppetdb/char_encoding.rb
@@ -32,66 +32,48 @@ module CharEncoding
   Utf8ReplacementChar = [ 0xEF, 0xBF, 0xBD ].pack("c*")
 
   DEFAULT_INVALID_CHAR = "\ufffd"
-
-  # @api private
-  def self.all_indexes_of_char(str, char)
-    (0..str.length).find_all{ |i| str[i] == char}
-  end
+  NOT_INVALID_REGEX = Regexp.new( "[^" + DEFAULT_INVALID_CHAR + "]" )
 
   # @api private
   #
-  # Takes an array and returns a sub-array without the last element
+  # Finds the beginning and ending index of the first block of invalid
+  # characters.
   #
-  # @return [Object]
-  def self.drop_last(array)
-    array[0..-2]
-  end
+  # @param str string to scan for invalid characters
+  # @return Range
+  def self.first_invalid_char_range(str)
+    begin_bad_chars_idx = str.index(DEFAULT_INVALID_CHAR)
 
-  # @api private
-  #
-  # Takes an array of increasing integers and collapses the sequential
-  # integers into ranges
-  #
-  # @param index_array an array of sorted integers
-  # @return [Range]
-  def self.collapse_ranges(index_array)
-    ranges = index_array.each.inject([]) do |spans, n|
-      if spans.empty? || spans.last.end != n - 1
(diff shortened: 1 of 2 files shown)


puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -104,13 +86,14 @@ def self.error_char_context(str, bad_char_indexes)
   # @param error_context_str information about where this string came from for use in error messages
   # @return String
   def self.warn_if_invalid_chars(str, error_context_str)
-    bad_char_indexes = all_indexes_of_char(str, DEFAULT_INVALID_CHAR)
-    if bad_char_indexes.empty?
+
+    if str.index(DEFAULT_INVALID_CHAR).nil?
```
senior: This new code only checks the index of the first invalid character, just like https://github.com/puppetlabs/puppetdb/pull/1789/files#diff-a0488ae759c41269703997e51a97b388R102

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -104,13 +86,14 @@ def self.error_char_context(str, bad_char_indexes)
...
       str
     else
       Puppet.warning "#{error_context_str} ignoring invalid UTF-8 byte sequences in data to be sent to PuppetDB, see debug logging for more info"
+
       if Puppet.settings[:log_level] == "debug"
-        Puppet.debug error_context_str + "\n" + error_char_context(str, bad_char_indexes).join("\n")
+        Puppet.debug error_context_str + "\n" + error_char_context(str, first_invalid_char_range(str))
```
senior: The new code only outputs debugging info for the first chunk of invalid characters. This overhead will only occur when debug mode is enabled.

puppet/lib/puppet/util/puppetdb/char_encoding.rb
```diff
@@ -32,66 +32,48 @@ module CharEncoding
...
-  end
-
-  # @api private
-  #
-  # Takes an array of increasing integers and collapses the sequential
-  # integers into ranges
-  #
-  # @param index_array an array of sorted integers
-  # @return [Range]
-  def self.collapse_ranges(index_array)
```
senior: This code can be removed because we're no longer outputting information on all chunks of bad data, only the first, so there will only ever be 1 range.

senior:
@vzctl @olivierHa can you try this out and see if it fixes your problem? I used a manifest like the one described [here](https://tickets.puppetlabs.com/browse/PDB-2256) and it's working fine for me.

vzctl:
Yes, it fixes the problem, there are no performance issues with and without debug output. Thanks @senior for fixing it up!

ajroetker:
Testing this now :dancer:

ajroetker:
Works for me using the catalogs for testing the original patch as well as some new ones generated per the ticket description :+1:, any idea what's going on with travis @senior?

ajroetker:
Travis looks to be an unrelated timeout, gonna merge this in

PR #1794, opened 2015-12-22
(PDB-2159) Augment the command processed message

mullr:
... so we can do better cache invalidation by observing response-mult.


senior:
:+1:

PR #1795, opened 2015-12-23
(MAINT) Upgrade to data.xml 0.1.0-beta1

senior:


mullr:
I don't think we actually *use* this anymore, can we just ditch the dep?

senior:
@mullr just checked, you're right, I'll close this an open a PR to remove it.

PR #1802, opened 2016-01-04
(PDB-2286) Add ability to always store catalog jsonb fields, (PDB-2290) Munge edges and resources jsonb for storage

ajroetker:
This PR fixes a bug with historical catalogs where a user could turn
off historical catalogs and break some of the PE views because the JSONB
column storage was also turned off. This new atom will always be set to
true in PE (as opposed to the historical-catalogs' atom which is config
dependent).

This PR aligns the edges and resources jsonb format with the return
format from querying the catalogs endpoint, this means we do not have to
parse and munge jsonb in export, the new endpoints and sync.

### DIFF
--- a/src/puppetlabs/puppetdb/catalogs.clj
+++ b/src/puppetlabs/puppetdb/catalogs.clj
@@ -158,7 +158,9 @@
    :file (s/maybe s/Str)
    :line (s/maybe s/Int)
    :parameters {s/Any s/Any}
-   :resource s/Str
+   ;; On historical-catalogs this is an optional-key but this is required on
+   ;; "normal" catalogs
+   (s/optional-key :resource) s/Str
    :tags [(s/maybe s/Str)]
    :title s/Str
    :type s/Str})
--- a/src/puppetlabs/puppetdb/export.clj
+++ b/src/puppetlabs/puppetdb/export.clj
@@ -2,6 +2,7 @@
   (:require [clj-time.core :refer [now]]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
+            [puppetlabs.puppetdb.scf.storage :as storage]
             [puppetlabs.puppetdb.scf.storage-utils :as sutils]
             [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.anonymizer :as anon]
@@ -49,9 +50,7 @@
           "catalogs" ["catalogs" (str (:certname datum) ".json")]
           "reports" ["reports" (export-report-filename datum)])]
     {:file-suffix file-suffix
-     :contents (if (= entity :reports)
-                 (-> datum (dissoc :hash) json/generate-pretty-string)
-                 (json/generate-pretty-string datum))}))
+     :contents (json/generate-pretty-string datum)}))
 
 (defn export-data->tar-items
   [entity data]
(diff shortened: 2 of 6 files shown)


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -351,16 +351,32 @@
    (comp first sql/result-set-seq)))
 
 (def store-catalogs-historically? (atom false))
+(def store-catalogs-jsonb-columns? (atom false))
```
mullr: Can you put a comment here saying the same thing you did in your commit message about how these two are different?

ajroetker:
@mullr added the comments suggested by your comments :boom:

src/puppetlabs/puppetdb/export.clj
```diff
@@ -92,8 +91,14 @@
...
                                            (maybe-anonymize anonymize-fn anon-config)
                                            (export-data->tar-items entity)
-                                           (add-tar-entries tar-writer)))]]
-      (query-fn query-api-version ["from" entity] nil query-callback-fn))))
+                                           (add-tar-entries tar-writer)))
+                  entity* (if (and (= entity "catalogs")
+                                   @storage/store-catalogs-jsonb-columns?)
+                            ;; *Warning* this can only be used in PE so it
+                            ;; *cannot be tested against in the FOSS repo
+                            "historical_catalogs"
```
senior: I guess what this needs is a plug point for PE?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -350,17 +350,43 @@
     certname]
    (comp first sql/result-set-seq)))
 
+;; `store-catalogs-historically?` is used for toggling historical catalog
+;; storage, the ability to toggle this variable is in the PE extensions repo.
```
senior: Seems like we can say this is PE only and configurable and leave out repo information

senior:
:+1: minor comment on a comment

src/puppetlabs/puppetdb/export.clj
ajroetker: Yeah I don't exactly what we'd want to do here, maybe make an atom that PE can use to change the query entities or such eventually.

src/puppetlabs/puppetdb/scf/storage.clj
ajroetker: :+1:

ajroetker:
@senior fixed the comment!

src/puppetlabs/puppetdb/catalogs.clj
```diff
@@ -158,7 +158,9 @@
    :file (s/maybe s/Str)
    :line (s/maybe s/Int)
    :parameters {s/Any s/Any}
-   :resource s/Str
+   ;; On historical-catalogs this is an optional-key but this is required on
+   ;; "normal" catalogs
+   (s/optional-key :resource) s/Str
```
wkalt: I'd rather this be addressed by us storing resource hashes. I think you could find uses for them and I think not having them introduces unneeded inconsistency between the historical and present day views.

ajroetker: Either way I'd like to address it in another patch, it requires some nontrivial changes to the storage code.

test/puppetlabs/puppetdb/acceptance/cli.clj
```diff
@@ -49,7 +49,7 @@
 
       (svc-utils/call-with-single-quiet-pdb-instance
        (let [anon-out-map (tar->map anon-out-file)
-             anon-certname (-> anon-out-map (get "reports") first val (get "certname"))]
+             anon-certname (some-> anon-out-map (get "reports") first val (get "certname"))]
```
wkalt: why is this needed?

src/puppetlabs/puppetdb/export.clj
wkalt: will the export that results from this be importable?

wkalt: ah got it.  they don't overwrite in PE

PR #1803, opened 2016-01-05
(PDB-2290) Align edges and resources formats with query formats

ajroetker:
This commit aligns the edges and resources jsonb format with the return
format from querying the catalogs endpoint, this means we do not have to
parse and munge jsonb in export, the new endpoints and sync.


ajroetker:
Closed in favor of https://github.com/puppetlabs/puppetdb/pull/1802

PR #1804, opened 2016-01-06
(PDB-2230) support multipart/form-data command posting

wkalt:
This will allow more granular logging at the wire without parsing the
command payload itself.

### DIFF
--- a/documentation/api/command/v1/commands.markdown
+++ b/documentation/api/command/v1/commands.markdown
@@ -32,19 +32,25 @@ The entire command MUST be encoded as UTF-8.
 
 ## Command submission
 
-Commands are submitted via HTTP to the `/pdb/cmd/v1` URL and must
-conform to the following rules:
+Commands must be submitted via HTTP to the `/pdb/cmd/v1` endpoint via one of
+two mechanisms:
 
-* A `POST` is used
-* The `POST` body must contain the JSON payload.
-* There is an `Accept` header that matches `application/json`.
-* The content-type is `application/json`.
+### Multipart/form-data POST
 
-Optionally, there may be a query parameter, `checksum`, that contains a SHA-1 hash of
-the payload which will be used for verification.
+When POSTing a command with Content-Type multipart/form-data, separate
+forms are used for certname, command name, command version, payload, and
+optionally checksum.
 
-When a command is successfully submitted, the submitter will
-receive the following:
+### Application/json POST (deprecated)
+
+When using Content-Type application/json, the POST body is a JSON payload
+containing values for certname, command name, command version, and payload.
+This method requires Accept and Content-Type headers set to application/json. A
+checksum may optionally be provided as a separate parameter.
+
+In either case, the checksum should contain a SHA-1 hash of the payload which
+will be used for content verification with the server. When a command is
+successfully submitted, the submitter will receive the following:
 
 * A response code of 200
 * A content-type of `application/json`
@@ -52,7 +58,8 @@ receive the following:
   value is a UUID corresponding to the submitted command. This can be used, for example, by
   clients to correlate submitted commands with server-side logs.
 
-The PuppetDB termini for puppet masters use this command API to update facts, catalogs, and reports for nodes.
+The PuppetDB termini for Puppet masters use this command API to update facts,
+catalogs, and reports for nodes, and will always include the checksum.
 
 ### Blocking command submission (EXPERIMENTAL)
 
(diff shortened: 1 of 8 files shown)


documentation/api/command/v1/commands.markdown
```diff
@@ -32,19 +32,25 @@ The entire command MUST be encoded as UTF-8.
...
-* The `POST` body must contain the JSON payload.
-* There is an `Accept` header that matches `application/json`.
-* The content-type is `application/json`.
+### Multipart/form-data POST
 
-Optionally, there may be a query parameter, `checksum`, that contains a SHA-1 hash of
-the payload which will be used for verification.
+When POSTing a command with Content-Type multipart/form-data, separate
+forms are used for certname, command name, command version, payload, and
+optionally checksum.
```
mullr: should note that the payload should come last

wkalt: good catch

puppet/lib/puppet/util/puppetdb/command.rb
```diff
@@ -44,22 +56,26 @@ def initialize(command, version, certname, payload)
       # Puppet 4.1.0. We need a better answer to non-utf8 data end-to-end.
       }.to_pson, "Error encoding a '#{command}' command for host '#{certname}'")
     end
+    @command = Puppet::Util::Puppetdb::CharEncoding.coerce_to_utf8(command)
+    @version = version
+    @certname = Puppet::Util::Puppetdb::CharEncoding.coerce_to_utf8(certname)
```
wkalt: is this necessary on certname and command? need to ensure the checksums match but I'm not sure a certname can contain nonascii anyway

documentation/api/command/v1/commands.markdown
```diff
@@ -32,19 +32,25 @@ The entire command MUST be encoded as UTF-8.
...
 
-When a command is successfully submitted, the submitter will
-receive the following:
+### Application/json POST (deprecated)
```
wkalt: these headers render badly. need to fix

documentation/api/command/v1/commands.markdown
```diff
@@ -52,7 +58,8 @@ receive the following:
   value is a UUID corresponding to the submitted command. This can be used, for example, by
   clients to correlate submitted commands with server-side logs.
 
-The PuppetDB termini for puppet masters use this command API to update facts, catalogs, and reports for nodes.
+The PuppetDB termini for Puppet masters use this command API to update facts,
```
mullr: termini and use need to agree in quantity

src/puppetlabs/puppetdb/cheshire.clj
```diff
@@ -19,23 +19,29 @@
             [clojure.java.io :as io]
             [clojure.set :as set]))
 
+(defrecord EncodedPayload [data])
```
mullr: would like a comment here as to EncodedPayload's raison d'être.

mullr: Maybe it should be called RawJsonString or something like that?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -21,10 +22,17 @@
 
 (def valid-commands-str (str/join ", " (sort (vals command-names))))
 
+(defn maybe-str->number
+  [s]
+  (if (number? s) s (edn/read-string s)))
```
mullr: the edn reader is an awfully heavy hammer to bring in here. Would Integer/parseInt do?

mullr: Also, I tend to think of functions of these shape as coercions, so maybe `coerce-to-int`?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -21,10 +22,17 @@
...
+
 (defn validate-command-version
   [app]
-  (fn [{:keys [body-string] :as req}]
-    (let [{:keys [command version]} (json/parse-string body-string true)
+  (fn [{:keys [body-string multipart-params] :as req}]
+    (let [{:strs [command version]} (if multipart-params
+                                      (select-keys multipart-params ["command" "version"])
+                                      (json/parse-string body-string))
```
mullr: I think we should log a warning when this happens, since we're deprecating it.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -310,6 +312,12 @@
                 #(update % 0 underscores->dashes)
                 m))
 
+(defn synthesize-body-str
+  [{:strs [command version certname payload]}]
+  (json/generate-string
+    {:command command :version (edn/read-string version)
```
mullr: Here as well, probably don't want edn. Maybe move the coersion fn into this file and use it?

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -310,6 +312,12 @@
...
+     :certname certname :payload (EncodedPayload. payload)}))
```
mullr: I'd consider making a regular clojure fn for this so we don't need to mess around with the import. Probably not that important though.

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -297,6 +298,13 @@
           (if (nil? body-string)
             (http/error-response (str "Empty body for application/json submission"))
             (app (assoc req :body-string body-string))))
+        "multipart/form-data"
+        (app (let [mp-request (mp/multipart-params-request req)
```
mullr: It appears that this isn't actually streaming. Rather, it'll put the data into a file, which we don't really want to do. See https://github.com/ring-clojure/ring/blob/1.4.0/ring-core/src/ring/middleware/multipart_params.clj#L79 and https://github.com/ring-clojure/ring/blob/1.4.0/ring-core/src/ring/middleware/multipart_params/temp_file.clj 

The out-of-the-box alternative loads into a byte array, which is at least better, but still not great.

PR #1807, opened 2016-01-08
(PDB-2230) support submission of certname/command/version via post params

wkalt:
This allows us to do more granular logging at the wire without parsing the
command payload itself.

### DIFF
--- a/documentation/api/command/v1/commands.markdown
+++ b/documentation/api/command/v1/commands.markdown
@@ -32,19 +32,24 @@ The entire command MUST be encoded as UTF-8.
 
 ## Command submission
 
-Commands are submitted via HTTP to the `/pdb/cmd/v1` URL and must
-conform to the following rules:
+Commands must be submitted via HTTP to the `/pdb/cmd/v1` endpoint via one of
+two mechanisms:
 
-* A `POST` is used
-* The `POST` body must contain the JSON payload.
-* There is an `Accept` header that matches `application/json`.
-* The content-type is `application/json`.
+* Payload and parameters: This method entails POSTing the certname, command name,
+command version, and optionally the checksum as parameters, with the POST body
+containing the given command's wire format. This mechanism allows PuppetDB to
+provide better validation and feedback at time of POSTing without inspecting the
+command payload itself, and should be preferred over the alternative due to
+lower memory consumption.
 
-Optionally, there may be a query parameter, `checksum`, that contains a SHA-1 hash of
-the payload which will be used for verification.
+* Payload only (deprecated): This method entails POSTing a single JSON body
+containing the certname, command name, and command version along side a
+`payload` key valued with the given command's wire format. The checksum is
+optionally provided as a parameter.
 
-When a command is successfully submitted, the submitter will
-receive the following:
+In either case, the checksum should contain a SHA-1 hash of the payload which
+will be used for content verification with the server. When a command is
+successfully submitted, the submitter will receive the following:
 
 * A response code of 200
 * A content-type of `application/json`
@@ -52,7 +57,8 @@ receive the following:
   value is a UUID corresponding to the submitted command. This can be used, for example, by
   clients to correlate submitted commands with server-side logs.
 
-The PuppetDB termini for puppet masters use this command API to update facts, catalogs, and reports for nodes.
+The PuppetDB termini for Puppet masters use this command API to update facts,
+catalogs, and reports for nodes, and will always include the checksum.
 
 ### Blocking command submission (EXPERIMENTAL)
 
@@ -158,12 +164,28 @@ Puppet resources. It is structured as a JSON object, conforming to the
 To post a `replace facts` command you can use the following curl command:
 
     curl -X POST \
+      -H 'Content-Type:application/json' \
+      -H 'Accept:application/json' \
+      -d '{"certname":"test1","environment":"DEV","values":{"myfact":"myvalue"},"producer_timestamp":"2015-01-01"}' \
+      "http://localhost:8080/pdb/cmd/v1?command=replace-facts&version=4&certname=test1"
+
+or equivalently (with the deprecated mechanism):
+
+    curl -X POST \
       -H "Accept: application/json" \
       -H "Content-Type: application/json" \
       -d '{"command":"replace facts","version":4,"payload":{"certname":"test1","environment":"DEV","values":{"myfact":"myvalue"},"producer_timestamp":"2015-01-01"}}' \
       http://localhost:8080/pdb/cmd/v1
 
 An example of `deactivate node`:
+
+    curl -X POST \
+      -H 'Content-Type:application/json' \
+      -H 'Accept:application/json' \
+      -d '{"certname":"test1","producer_timestamp":"2015-01-01"}' \
+      "http://localhost:8080/pdb/cmd/v1?certname=test1&command=deactivate-node&version=3"
+
+or equivalently:
 
     curl -X POST \
       -H "Accept: application/json" \
--- a/puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
+++ b/puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
@@ -38,15 +38,10 @@ describe Puppet::Resource::Catalog::Puppetdb do
     end
 
     it "should POST the catalog command as a JSON string" do
-      command_payload = subject.munge_catalog(catalog, options)
-      payload = {
-        :command => Puppet::Util::Puppetdb::CommandNames::CommandReplaceCatalog,
-        :version => 7,
-        :payload => command_payload,
-      }.to_json
+      command_payload = subject.munge_catalog(catalog, options).to_json
 
       http.expects(:post).with do |uri, body, headers|
-        expect(body).to eq(payload)
+        expect(body).to eq(command_payload)
       end.returns response
 
       save
--- a/puppet/spec/unit/indirector/facts/puppetdb_spec.rb
+++ b/puppet/spec/unit/indirector/facts/puppetdb_spec.rb
@@ -43,17 +43,11 @@ describe Puppet::Node::Facts::Puppetdb do
       trusted_data = {"foo" => "foobar", "certname" => "testing_posting"}
       subject.stubs(:get_trusted_info).returns trusted_data
 
-      f = {
+      payload = {
         "certname" => facts.name,
         "values" => facts.values.merge({"trusted" => trusted_data}),
         "environment" => "my_environment",
         "producer_timestamp" => "a test",
-      }
-
-      payload = {
-        :command => CommandReplaceFacts,
-        :version => 4,
-        :payload => f,
       }.to_pson
 
       http.expects(:post).with do |uri, body, headers|
@@ -75,11 +69,10 @@ describe Puppet::Node::Facts::Puppetdb do
       save
 
       message = JSON.parse(sent_payload)
-      sent_facts = message['payload']
 
       # We shouldn't modify the original instance
       facts.values['something'].should == 100
-      sent_facts['values']['something'].should == 100
+      message['values']['something'].should == 100
     end
   end
 
--- a/src/puppetlabs/puppetdb/cheshire.clj
+++ b/src/puppetlabs/puppetdb/cheshire.clj
@@ -19,23 +19,29 @@
             [clojure.java.io :as io]
             [clojure.set :as set]))
 
+(defrecord RawJsonString [data])
+
 ;; Alias coerce/to-string to avoid reflection
 (def ^String to-string coerce/to-string)
 (defn add-common-json-encoders!*
   "Non-memoize version of add-common-json-encoders!"
   []
   (generate/add-encoder
-   org.postgresql.util.PGobject
-   (fn [^PGobject data ^JsonGenerator jsonGenerator]
-     ;; The .getPrettyPrinter method on the Jackson jsonGenerator will return
-     ;; nil if `:pretty` is not set as a cheshire option
-     (if (.getPrettyPrinter jsonGenerator)
-       (generate/encode-map (core/parse-string (.getValue data)) jsonGenerator)
-       (.writeRawValue jsonGenerator (.getValue data)))))
+    org.postgresql.util.PGobject
+    (fn [^PGobject data ^JsonGenerator jsonGenerator]
+      ;; The .getPrettyPrinter method on the Jackson jsonGenerator will return
+      ;; nil if `:pretty` is not set as a cheshire option
+      (if (.getPrettyPrinter jsonGenerator)
+        (generate/encode-map (core/parse-string (.getValue data)) jsonGenerator)
+        (.writeRawValue jsonGenerator (.getValue data)))))
   (generate/add-encoder
-   org.joda.time.DateTime
-   (fn [data ^JsonGenerator jsonGenerator]
-     (.writeString jsonGenerator (to-string data)))))
+    org.joda.time.DateTime
+    (fn [data ^JsonGenerator jsonGenerator]
+      (.writeString jsonGenerator (to-string data))))
+  (generate/add-encoder
+    RawJsonString
+    (fn [^String data ^JsonGenerator jsonGenerator]
+      (.writeRawValue jsonGenerator (:data data)))))
 
 (def
   ^{:doc "Registers some common encoders for cheshire JSON encoding.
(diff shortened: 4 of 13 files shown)


documentation/api/command/v1/commands.markdown
```diff
@@ -166,6 +180,14 @@ To post a `replace facts` command you can use the following curl command:
 An example of `deactivate node`:
 
     curl -X POST \
+      -H 'Content-Type:application/json' \
+      -H 'Accept:application/json' \
+      -d '{"certname":"test1","producer_timestamp":"2015-01-01"}' \
```
ajroetker: Can we not make `certname` in the command body optional?

puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
```diff
@@ -39,11 +39,7 @@ def save
 
     it "should POST the catalog command as a JSON string" do
       command_payload = subject.munge_catalog(catalog, options)
-      payload = {
-        :command => Puppet::Util::Puppetdb::CommandNames::CommandReplaceCatalog,
-        :version => 7,
-        :payload => command_payload,
-      }.to_json
+      payload = command_payload.to_json
```
ajroetker: what the need for the extra step? can we just `to_json` the line above?

puppet/spec/unit/indirector/facts/puppetdb_spec.rb
```diff
@@ -50,11 +50,7 @@ def save
         "producer_timestamp" => "a test",
       }
 
-      payload = {
-        :command => CommandReplaceFacts,
-        :version => 4,
-        :payload => f,
-      }.to_pson
+      payload = f.to_pson
```
ajroetker: Similarly here, do we need `f`?

src/puppetlabs/puppetdb/cheshire.clj
```diff
@@ -19,23 +19,29 @@
...
+  (generate/add-encoder
+    org.joda.time.DateTime
+    (fn [data ^JsonGenerator jsonGenerator]
+      (.writeString jsonGenerator (to-string data))))
   (generate/add-encoder
-   org.joda.time.DateTime
-   (fn [data ^JsonGenerator jsonGenerator]
-     (.writeString jsonGenerator (to-string data)))))
+    RawJsonString
+    (fn [data ^JsonGenerator jsonGenerator]
```
ajroetker: @senior do we have to worry about reflection with defrecord?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -21,21 +21,31 @@
 
 (def valid-commands-str (str/join ", " (sort (vals command-names))))
 
+(defn coerce-to-int
+  [s]
+  (if (number? s) s (Integer. s)))
```
ajroetker: Can you use https://github.com/puppetlabs/clj-kitchensink/blob/master/src/puppetlabs/kitchensink/core.clj#L207 here or parse-number?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -21,21 +21,31 @@
...
+
 (defn validate-command-version
   [app]
-  (fn [{:keys [body-string] :as req}]
-    (let [{:keys [command version]} (json/parse-string body-string true)
+  (fn [{:keys [body-string params param-post?] :as req}]
+    (let [{:strs [command version]} (if param-post?
+                                      params
+                                      (json/parse-string body-string))
+          command (str/replace command "-" " ")
```
ajroetker: So it seems kind of silly to go back and forth between `-` and `" "`, would we have to bump versions of anything to just always use `-` in our commands?

ajroetker: I.e. would this be a good patch to do that in?

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -13,7 +13,8 @@
             [slingshot.slingshot :refer [try+ throw+]]
             [com.rpl.specter :as sp])
   (:import [java.net MalformedURLException URISyntaxException URL]
-           [org.postgresql.util PGobject]))
+           [org.postgresql.util PGobject]
+           [puppetlabs.puppetdb.cheshire RawJsonString]))
```
ajroetker: I think the spaces in these imports are unnecessary, `[puppetlabs.puppetdb.cheshire RawJsonString]` means import `puppetlabs.puppetdb.cheshire` and `puppetlabs.puppetdb.cheshire.RawJsonString`, I think you just want `[puppetlabs.puppetdb.cheshire.RawJsonString]`.

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -293,10 +293,19 @@
...
+              reconstructed-body (assoc params "payload" body-string)
+              param-post? (boolean (and (get params "certname")
+                                        (get params "version")
+                                        (get params "command")))]
           (if (nil? body-string)
             (http/error-response (str "Empty body for application/json submission"))
-            (app (assoc req :body-string body-string))))
+            (let [submission-body (if param-post?
+                                    (-> params
+                                        (assoc "payload" body-string)
```
ajroetker: looks like you could just use `reconstructed-body` here

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -293,10 +293,19 @@
           (app (assoc req :body-string payload))
           (http/error-response (str "Missing required parameter 'payload'")))
         "application/json"
-        (let [body-string (request/body-string req)]
+        (let [body-string (request/body-string req)
+              reconstructed-body (assoc params "payload" body-string)
+              param-post? (boolean (and (get params "certname")
+                                        (get params "version")
+                                        (get params "command")))]
```
ajroetker: It feels like if they specify any one of these, that we should error if they don't have all of them.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -310,6 +311,12 @@
                 #(update % 0 underscores->dashes)
                 m))
 
+(defn synthesize-body-str
```
ajroetker: Could we call this `cmd-params->json-str` or something more specific?

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -310,6 +311,12 @@
...
+  [{:strs [command version certname payload]}]
+  (json/generate-string
+    {:command (string/replace command "-" " ") :version (Integer. version)
```
ajroetker: It looks like we're doing this twice, in `validate-command...` and here, is it possible to consolidate?

documentation/api/command/v1/commands.markdown
```diff
@@ -32,19 +32,24 @@ The entire command MUST be encoded as UTF-8.
...
+two mechanisms:
 
-* A `POST` is used
-* The `POST` body must contain the JSON payload.
-* There is an `Accept` header that matches `application/json`.
-* The content-type is `application/json`.
+* Payload and parameters: This method entails POSTing the certname, command name,
+command version, and optionally the checksum as parameters, with the POST body
+containing the given command's wire format. This mechanism allows PuppetDB to
+provide some information at the wire without inspecting the command payload
```
ajroetker: This mechanism allows PuppetDB to provide better validation and feedback at the time of POSTing without...

src/puppetlabs/puppetdb/cheshire.clj
mullr: @ajroetker The question should be, do we have to worry about it with 'extend'. My gut tells me no, since the compiler should statically know that the type of the `this` param (data in this case) is RawJsonString. But the Clojure compiler often doesn't understand things that we'd consider obvious. The best way to find out is to set `*warn-on-reflection*` and see what happens.

documentation/api/command/v1/commands.markdown
wkalt: I think we should avoid this for now. We'll end up having to write that it's optional in one case but not the other, which starts to feel like a separate version. I'd vote for leaving it until we can bump the command and retire the old mechanism.

puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
wkalt: will fix

puppet/spec/unit/indirector/facts/puppetdb_spec.rb
wkalt: nope

src/puppetlabs/puppetdb/cheshire.clj
wkalt: I'll add a type hint

src/puppetlabs/puppetdb/http/command.clj
wkalt: sure, I don't mind

src/puppetlabs/puppetdb/http/command.clj
wkalt: I'd be on board with this change, but I feel like it's enough of a docs/code change for its own patch. It would break cross-version imports from 3->4 (which we don't guarantee) without some kind of replace call like this one -- don't know that we care.

src/puppetlabs/puppetdb/utils.clj
wkalt: this doesn't seem to work

test/puppetlabs/puppetdb/middleware_test.clj
```diff
@@ -117,13 +118,26 @@
...
     (testing "url encoded payload should populate body-string"
       (let [test-req {:params {"payload" test-content}
                       :headers {"content-type" "application/x-www-form-urlencoded"}}]
         (is (= (wrapped-fn test-req)
-               (assoc test-req :body-string test-content)))))))
+               (assoc test-req :body-string test-content)))))
+
+    (testing "param-post? is true when params are included"
+      (let [test-req {:params {"certname" "foo.com"
+                               "command" "fix drywall"
```
mullr: :+1:

src/puppetlabs/puppetdb/utils.clj
ajroetker: sorry, leave out the `[]` it should just be `puppetlabs.puppetdb.cheshire.RawJsonString`, https://clojuredocs.org/clojure.core/import

src/puppetlabs/puppetdb/http/command.clj
ajroetker: :+1: we can make a ticket and put it in another patch

ajroetker: Ticket made PDB-2311, but a note from the channel... Should these be underscores instead of dashes?

ajroetker:

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -310,6 +311,12 @@
                 #(update % 0 underscores->dashes)
                 m))
 
+(defn cmd-params->json-str
+  [{:strs [command version certname payload]}]
+  (json/generate-string
+    {:command (string/replace command "-" " ") :version (Integer. version)
```
ajroetker: looks like this `string/replace` is unnecessary now

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -286,10 +286,30 @@
...
+            (nil? body-string)
             (http/error-response (str "Empty body for application/json submission"))
-            (app (assoc req :body-string body-string))))
+
+            :else
+            (if param-post?
+              (let [command' (s/replace command "_" " ")
+                    submission-body (-> params
+                                        (assoc "payload" body-string)
+                                        (update "command" s/replace "_" " ")
```
ajroetker: can this just be `assoc "command" command'`?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -21,21 +21,30 @@
 
 (def valid-commands-str (str/join ", " (sort (vals command-names))))
 
+(defn coerce-to-int
+  [s]
+  (if (number? s) s (Integer/parseInt s)))
```
ajroetker: looks like my other comment got deleted, but I think `puppetlabs.kitchensink.core/parse-int` does basically the same thing as what we want here

wkalt: no, it requires a string as input

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -286,10 +286,30 @@
           (app (assoc req :body-string payload))
           (http/error-response (str "Missing required parameter 'payload'")))
         "application/json"
```
ajroetker: It seems like this function and https://github.com/puppetlabs/puppetdb/pull/1807/files#diff-d82b3486b4cd22472cd542b7773086e9L24 could/should probably be combined into 1 function

src/puppetlabs/puppetdb/http/command.clj
ajroetker: It looks like below you're using `Integer.` for this, could you the same thing be done in both places?

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -286,10 +286,30 @@
...
-        (let [body-string (request/body-string req)]
-          (if (nil? body-string)
+        (let [{:strs [certname version command]} params
```
rbrw: If I'm not misunderstanding, we already have a command-app handler that ensures the incoming request will be application/jjson, so I think we might be able to considerably simplify this function and the related test function.

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -286,10 +286,30 @@
...
+              body-string (request/body-string req)
+              param-post? (boolean (and certname version command))]
+
+          (cond
+            (and (not param-post?) (or certname version command))
+            (http/error-response (str "Command submission via POST parameters requires certname, version, and command."))
```
rbrw: Possibly vestigial "str" here and in the other error-response.

rbrw: Also wondered about a possibly more direct message like "Commands submitted via POST require certname, version, and command parameters." or similar.

src/puppetlabs/puppetdb/http/command.clj
wkalt: just to tie this off, neither of those works here. I also let payload-to-body-string be with respect to your suggested refactor because rob has some code on top of this that removes the function anyway.

wkalt:
@ajroetker comments addressed

ajroetker:
I'm :+1: everything lgtm and works

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -284,15 +284,36 @@
...
+              (let [command' (s/replace command "_" " ")
+                    submission-body (-> params
+                                        (assoc "payload" body-string)
+                                        (assoc "command" command')
+                                        utils/cmd-params->json-str)]
+                (-> req
+                    (assoc-in [:params "command"] command')
+                    (assoc :body-string submission-body :param-post? param-post?)
+                    app))
+              (app (assoc req :body-string body-string :param-post? param-post?)))))
```
senior: This seems like a good candidate for being a handler, not middleware, it should be addressed as a separate issue though.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -13,7 +13,8 @@
             [slingshot.slingshot :refer [try+ throw+]]
             [com.rpl.specter :as sp])
   (:import [java.net MalformedURLException URISyntaxException URL]
-           [org.postgresql.util PGobject]))
+           [org.postgresql.util PGobject]
+           puppetlabs.puppetdb.cheshire.RawJsonString))
```
senior: cosmetic, by why have this import line look different from the others?

senior:
:+1:

src/puppetlabs/puppetdb/utils.clj
ajroetker: @senior `[org.postgresql.util PGobject]` will import `org.postgresql.util` and `PGobject`, here we just want `RawJsonString`

senior: Now that I look closer, this isn't needed at all. That line can be removed and line 318 can be changed to (json/->RawJsonString payload)

PR #1808, opened 2016-01-08
(PDB-2234) Add ability for sync tests to disable metrics

ajroetker:
This commit adds a similar mechanism to HikariCP that we have for
ActiveMQ to disable JMX reporting. This is because JMX is per JVM which
means our sync tests' metrics will collide by trying to start 2 PDB's in
the same JVM.


senior:
:+1:

PR #1811, opened 2016-01-12
(maint) Fix pretty-printing for PGObjects which are lists

ajroetker:
This commit fixes an issue with the PGObject JSON encoder (when
pretty-printing) which assumed the object was a map.

### DIFF
--- a/src/puppetlabs/puppetdb/cheshire.clj
+++ b/src/puppetlabs/puppetdb/cheshire.clj
@@ -14,7 +14,7 @@
            [org.postgresql.util PGobject]
            [java.io Writer])
   (:require [cheshire.core :as core]
-            [cheshire.generate :as generate]
+            [cheshire.generate :refer [add-encoder encode-map encode-seq]]
             [clj-time.coerce :as coerce]
             [clojure.java.io :as io]
             [clojure.set :as set]))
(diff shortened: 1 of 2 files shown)


wkalt:
+1

wkalt:
@ajroetker this needs a rebase

ajroetker:
@wkalt fixed up with a rebase

src/puppetlabs/puppetdb/cheshire.clj
```diff
@@ -26,19 +26,27 @@
...
     (fn [^PGobject data ^JsonGenerator jsonGenerator]
       ;; The .getPrettyPrinter method on the Jackson jsonGenerator will return
       ;; nil if `:pretty` is not set as a cheshire option
       (if (.getPrettyPrinter jsonGenerator)
-        (generate/encode-map (core/parse-string (.getValue data)) jsonGenerator)
+        (let [obj (core/parse-string (.getValue data))
+              encode-fn (condp instance? obj
+                          clojure.lang.IPersistentMap encode-map
+                          clojure.lang.IPersistentVector encode-seq
+                          clojure.lang.IPersistentSet encode-seq
```
senior: It's not possible for this to come back from the DB is it? I would think that's true of most of the things in this list.

kbarber:
@ajroetker is this fixing a real issue with pretty printing to a user? I get what this means, but I'm trying to understand where users see this problem. It's just that if its a bug, some tests to capture this would be nice.

PR #1812, opened 2016-01-13
(PDB-2252) Try to avoid parsing a POSTed cmd

rbrw:
Don't parse the incoming command body when it has command and version
parameters (which should now be the common case).  Instead, validate the
parameters, and pass the body directly to the queue, uninterpreted.
Otherwise, warn if body has to be parsed into RAM.

Attach the command, version, and if available, certname, to the queue
message as JMS properties, extracting them from the body (after
parsing), if they weren't available in the incoming request.

As before, reject incoming commands that don't have both command and
version when they have either.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -112,11 +112,14 @@
   [mq-connection :- mq/connection-schema
    mq-endpoint :- s/Str
    raw-command :- s/Str
-   uuid :- (s/maybe s/Str)]
+   uuid :- (s/maybe s/Str)
+   properties :- (s/maybe {s/Str s/Str})] ;; For now stick with str -> str
   (let [uuid (or uuid (kitchensink/uuid))]
-    (mq/send-message! mq-connection mq-endpoint
-                      raw-command
-                      {"received" (kitchensink/timestamp) "id" uuid})
+    (mq/send-message! mq-connection mq-endpoint raw-command
+                      ;; Until/unless we require that all callers
+                      ;; include received, etc.
+                      (merge {"received" (kitchensink/timestamp) "id" uuid}
+                             properties))
     uuid))
 
 (defn-validated ^:private do-enqueue-command :- s/Str
@@ -127,11 +130,15 @@
    command :- s/Str
    version :- s/Int
    payload
-   uuid :- (s/maybe s/Str)]
+   uuid :- (s/maybe s/Str)
+   properties]
   (let [command-map {:command command
                      :version version
                      :payload payload}]
-    (do-enqueue-raw-command mq-connection mq-endpoint (json/generate-string command-map) uuid)))
+    (do-enqueue-raw-command mq-connection mq-endpoint
+                            (json/generate-string command-map)
+                            uuid
+                            properties)))
 
 ;; ## Command processing exception classes
 
@@ -286,10 +293,11 @@
   (enqueue-command
     [this command version payload]
     [this command version payload uuid]
+    [this command version payload uuid properties]
     "Annotates the command via annotate-command, submits it for
     processing, and then returns its unique id.")
 
-  (enqueue-raw-command [this raw-command uuid]
+  (enqueue-raw-command [this raw-command uuid properties]
     "Submits the raw-command for processing and returns the command's
     unique id.")
 
@@ -376,21 +384,25 @@
     (enqueue-command this command version payload nil))
 
   (enqueue-command [this command version payload uuid]
+    (enqueue-command this command version payload uuid nil))
+
+  (enqueue-command [this command version payload uuid properties]
     (let [config (get-config)
           connection (:connection (service-context this))
           endpoint (get-in config [:command-processing :mq :endpoint])
           command (if (string? command) command (command-names command))
           result (do-enqueue-command connection endpoint
-                                     command version payload uuid)]
+                                      command version payload uuid properties)]
       ;; Obviously assumes that if do-* doesn't throw, msg is in
       (swap! (:stats (service-context this)) update :received-commands inc)
       result))
 
-  (enqueue-raw-command [this raw-command uuid]
+  (enqueue-raw-command [this raw-command uuid properties]
     (let [config (get-config)
           connection (:connection (service-context this))
           endpoint (get-in config [:command-processing :mq :endpoint])
-          result (do-enqueue-raw-command connection endpoint raw-command uuid)]
+          result (do-enqueue-raw-command connection endpoint
+                                          raw-command uuid properties)]
       ;; Obviously assumes that if do-* doesn't throw, msg is in
       (swap! (:stats (service-context this)) update :received-commands inc)
       result))
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -96,7 +96,7 @@
           (string? (:command %))
           (number? (:version %))
           (map? (:annotations %))]}
-  (let [message (if (= (class body) (class (byte-array 0)))
+  (let [message (if (instance? utils/byte-array-class body)
                       (json/parse-string (String. body "UTF-8") true)
                       (json/parse-string body true))
         default-annotations {:received (kitchensink/timestamp)
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -77,6 +77,21 @@
             [clojure.set :as set]
             [clojure.core.async :as async]))
 
+(defn fatality
+  "Create an object representing a fatal command-processing exception
+
+  cause - object representing the cause of the failure
+  "
+  [cause]
+  {:fatal true :cause cause})
+
+(defmacro upon-error-throw-fatality
+  [& body]
+  `(try+
+    ~@body
+    (catch Throwable e#
+      (throw+ (fatality e#)))))
+
 (defn version-range [min-version max-version]
   (set (range min-version (inc max-version))))
 
@@ -86,23 +101,102 @@
    "store report" (version-range 3 6)
    "deactivate node" (version-range 1 3)})
 
-;; ## Command parsing
+(defn- die-on-header-payload-mismatch
+  [name in-header in-body]
+  (when (and in-header in-body (not= in-header in-body))
+    (throw+
+     (fatality
+      (Exception.
+       (format "%s mismatch between message properties and payload (%s != %s)"
+               name in-header in-body))))))
+
+(def new-message-schema
+  {:headers {:command s/Str
+             :version s/Str
+             :certname s/Str
+             :received s/Str
+             :id s/Str}
+   :body (s/cond-pre s/Str utils/byte-array-class)})
+
+(def queue-message-schema
+  ;; Perhaps eventually require :id and :received
+  {(s/optional-key :headers) {(s/optional-key :id) s/Str
+                              (s/optional-key :received) s/Str
+                              (s/optional-key :scheduledJobId) s/Str
+                              (s/optional-key :JMSXDeliveryCount) s/Str
+                              s/Any s/Any}
+   :body (s/cond-pre s/Str utils/byte-array-class)})
+
+(def new-command-schema ;; This could probably be stricter
+  {:command s/Str
+   :version s/Int
+   :certname s/Str
+   :payload {s/Any s/Any}
+   :annotations
+   (s/pred
+    (fn [x]
+      (and (map? x)
+           (some-> x :id string?)
+           (some-> x :received string?)
+           (empty? (select-keys x [:command :version :certname])))))})
+
+(def queue-command-schema
+  ;; Created to match existing test behavior, but is that
+  ;; comprehensive? i.e. if we're wrong, and this is too strict, then
+  ;; it could break on outside queue content on release.  Adding a
+  ;; schema here could change our effectively public queue API.
+  (-> new-command-schema
+      (dissoc :certname)
+      (assoc (s/optional-key :certname) s/Str)
+      (dissoc :payload)
+      (assoc (s/optional-key :payload) (s/cond-pre s/Str {s/Any s/Any}))))
+
+(defn-validated ^:private parse-new-command :- new-command-schema
+  "Parses a new-format queue command and returns it in the traditional
+  format (see parse-queue-command).  New-style messages must have all
+  5 headers (command, version, certname, received, and id), and the
+  body must be the bare command content (i.e. the payload).  So for a
+  \"deactivate node\" command, a string like
+  {\"certname\":\"test1\",\"producer_timestamp\":\"2015-01-01\"}."
+  [{:keys [headers body]} :- new-message-schema]
+  (let [{:keys [command version certname received id]} headers
+        version (Integer/parseInt version)]
+    (let [message (json/parse body true)]
+      (die-on-header-payload-mismatch "certname"
+                                      certname
+                                      (get-in message ["payload" "certname"]))
+      ;; Since new commands aren't the queue retry format, we expect
+      ;; no annotations.
+      (assert (not (:annotations message)))
+      {:command command
+       :version version
+       :certname certname ;; Duplicated with respect to payload for now.
+       :payload message
+       :annotations (dissoc headers :command :version :certname)})))
+
+(defn-validated ^:private parse-queue-command :- queue-command-schema
+  "Parses a traditional queue command (also the current format for
+  failed, re-queued messages).  See parse-new-command for the handling
+  of newly received messages.  These messages must not have command,
+  version, or certname headers, and the body must be a JSON map like
+  this: {\"command\":\"deactivate node\",\"version\":3,\"payload\":{\"certname\":...}}."
+  [{:keys [headers body]} :- queue-message-schema]
+  (let [{:keys [command version certname]} headers]
+    (update (json/parse-strict body true) :annotations merge
+            {:received (kitchensink/timestamp) :id (kitchensink/uuid)}
+            headers)))
 
 (defn parse-command
-  "Take a wire-format command and parse it into a command object."
-  [{:keys [headers body]}]
+  "Parses a queue message and returns it as a command map."
+  [{:keys [headers body] :as message}]
   {:post [(map? %)
           (:payload %)
           (string? (:command %))
           (number? (:version %))
           (map? (:annotations %))]}
-  (let [message (if (instance? utils/byte-array-class body)
-                      (json/parse-string (String. body "UTF-8") true)
-                      (json/parse-string body true))
-        default-annotations {:received (kitchensink/timestamp)
-                             :id (kitchensink/uuid)}]
-    (update message :annotations
-            merge default-annotations headers)))
+  (if (:command headers)
+    (parse-new-command message)
+    (parse-queue-command message)))
 
 ;; ## Command submission
 
@@ -111,7 +205,7 @@
   its id."
   [mq-connection :- mq/connection-schema
    mq-endpoint :- s/Str
-   raw-command :- s/Str
+   raw-command :- (s/cond-pre s/Str utils/byte-array-class)
    uuid :- (s/maybe s/Str)
    properties :- (s/maybe {s/Str s/Str})] ;; For now stick with str -> str
   (let [uuid (or uuid (kitchensink/uuid))]
@@ -139,23 +233,6 @@
                             (json/generate-string command-map)
                             uuid
                             properties)))
-
-;; ## Command processing exception classes
-
-(defn fatality
-  "Create an object representing a fatal command-processing exception
-
-  cause - object representing the cause of the failure
-  "
-  [cause]
-  {:fatal true :cause cause})
-
-(defmacro upon-error-throw-fatality
-  [& body]
-  `(try+
-    ~@body
-    (catch Throwable e#
-      (throw+ (fatality e#)))))
 
 ;; Catalog replacement
 
--- a/src/puppetlabs/puppetdb/http/command.clj
+++ b/src/puppetlabs/puppetdb/http/command.clj
@@ -107,6 +107,11 @@
           (do-submit)
           (http/json-response {:uuid uuid}))))))
 
+(defn- add-received-param
+  [handle]
+  (fn [req]
+    (handle (assoc-in req [:params "received"] (kitchensink/timestamp)))))
+
 (defn routes [enqueue-fn get-response-pub]
   (cmdi/context "/v1"
                 (cmdi/ANY "" []
@@ -126,6 +131,7 @@
       (mid/fail-when-payload-too-large reject-large-commands? max-command-size)
       mid/verify-accepts-json
       mid/verify-checksum
+      add-received-param ;; must be (temporally) after validate-query-params
       (mid/validate-query-params {:optional ["checksum" "secondsToWaitForCompletion"
                                              "certname" "command" "version"]})
       mid/payload-to-body-string
--- a/src/puppetlabs/puppetdb/http/command.clj
+++ b/src/puppetlabs/puppetdb/http/command.clj
@@ -128,12 +128,12 @@
   (-> (routes enqueue-fn get-response-pub)
       mid/make-pdb-handler
       validate-command-version
-      mid/verify-accepts-json
       add-received-param ;; must be (temporally) after validate-query-params
       ;; The checksum here is vestigial.  It is no longer checked
       (mid/validate-query-params {:optional ["checksum" "secondsToWaitForCompletion"
                                              "certname" "command" "version"]})
       mid/payload-to-body-string
+      mid/verify-accepts-json
       (mid/verify-content-type ["application/json"])
       (mid/fail-when-payload-too-large reject-large-commands? max-command-size)
       (mid/wrap-with-metrics (atom {}) http/leading-uris)
--- a/src/puppetlabs/puppetdb/http/command.clj
+++ b/src/puppetlabs/puppetdb/http/command.clj
@@ -1,5 +1,6 @@
 (ns puppetlabs.puppetdb.http.command
-  (:require [puppetlabs.puppetdb.command.constants :refer [command-names]]
+  (:require [clojure.set :as set]
+            [puppetlabs.puppetdb.command.constants :refer [command-names]]
             [puppetlabs.trapperkeeper.core :refer [defservice]]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
@@ -8,9 +9,16 @@
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.config :as conf]
             [puppetlabs.puppetdb.middleware :as mid]
+            [puppetlabs.puppetdb.schema :refer [defn-validated]]
+            [puppetlabs.puppetdb.utils :as utils]
             [clojure.core.async :as async]
             [puppetlabs.kitchensink.core :as kitchensink]
-            [puppetlabs.comidi :as cmdi]))
+            [puppetlabs.comidi :as cmdi]
+            [ring.util.request :as request]
+            [schema.core :as s]
+            [slingshot.slingshot :refer [try+ throw+]])
+  (:import [org.apache.commons.io IOUtils]
+           [org.apache.commons.fileupload.util LimitedInputStream]))
 
 (def min-supported-commands
   {"replace catalog" 6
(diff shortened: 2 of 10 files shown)


rbrw:
This depends on (is currently on top of) https://github.com/puppetlabs/puppetdb/pull/1807

rbrw:
You might want to look at the individual commits via gitk, magit, or whatever.

ajroetker:
@rbrw looks like this needs a rebase after the comidi patch

src/puppetlabs/puppetdb/command.clj
```diff
@@ -77,6 +77,21 @@
             [clojure.set :as set]
             [clojure.core.async :as async]))
 
+(defn fatality
```
mullr: FINISH HIM

src/puppetlabs/puppetdb/command.clj
```diff
@@ -111,12 +181,15 @@
...
+   raw-command :- (s/cond-pre s/Str (Class/forName "[B"))
+   uuid :- (s/maybe s/Str)
+   properties :- (s/maybe {s/Str s/Str})] ;; For now stick with str -> str
   (let [uuid (or uuid (kitchensink/uuid))]
-    (mq/send-message! mq-connection mq-endpoint
-                      raw-command
-                      {"received" (kitchensink/timestamp) "id" uuid})
+    (mq/send-message! mq-connection mq-endpoint raw-command
+                      ;; Until/unless we require that all callers
+                      ;; include received, etc.
```
rbrw: https://tickets.puppetlabs.com/browse/PDB-2315

mullr:
First commit message has `""File`, double quotes.

src/puppetlabs/puppetdb/mq.clj
```diff
@@ -164,10 +164,15 @@
    (let [^TextMessage text-message message]
      (.getText text-message))
    (instance? javax.jms.BytesMessage message)
-   (let [^BytesMessage bytes-message message]
-     (.readUTF8 bytes-message))
+   (let [^BytesMessage bytes-message message
+         len (.getBodyLength message)
+         buf (byte-array len)]
+     (assert (= len (.readBytes bytes-message buf)))
```
mullr: Should throw something better than an assertion error if this check fails; at least it should include a message.

src/puppetlabs/puppetdb/mq.clj
```diff
@@ -164,10 +164,15 @@
...
+     (String. buf "UTF-8"))
    :else
-   (throw (ex-info (str "Expected a text message, instead found: " (class message)) {}))))
+   (throw (ex-info (str "Expected a text message, instead found: "
```
mullr: text or bytes now

mullr:
> (PDB-2252) Add properties argument to enqueue cmds
> So that we can providing additional properties (e.g. eventually command,
> version, and certname).

s/providing/provide/

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +101,78 @@
    "store report" (version-range 3 6)
    "deactivate node" (version-range 1 3)})
 
-;; ## Command parsing
+(defn- read-command-body
+  [body]
+  (if (= (class body) (class (byte-array 0)))
```
mullr: would be nice to factor this out into a `byte-array?` fn, maybe something like...

```clojure
(def byte-array?
  (let [t (Class/forName "[B")]
    (fn [x]
      (instance? t x))))
```

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +101,78 @@
...
+    (json/parse-string (String. body "UTF-8") true)
```
mullr: Should be able to a stick a ByteArrayInputStream on top of body and use `parse-stream` to avoid the intermediate `String`.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +101,78 @@
...
+     (fatality
+      (Exception.
+       (format "%s mismatch between message properties and payload (%s != %s)"
+               name in-header in-body))))))
+
+(defn parse-new-command
+  "Parses a new-format queue command and returns it in the traditional
+  format."
+  [{:keys [headers body]}]
+  ;; New-style messages must have all 5 headers, and the body must be
```
mullr: I think this comment wants to go in the docstring

src/puppetlabs/puppetdb/command.clj
rbrw: Sounds plausible (I'd just kept the original).

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +101,78 @@
...
+  ;; the bare command content (i.e. the payload).  So for
+  ;; a "deactivate node" command,
+  ;; a string like {"certname":"test1","producer_timestamp":"2015-01-01"}.
+  (let [{:keys [command version certname received id]} headers
+        version (if-let [v (:version headers)] (Integer/parseInt v) nil)]
```
mullr: `(some-> headers :version ...)`

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +101,78 @@
...
+    (when-not (and command version certname received id)
+      (throw+
+       (fatality (Exception. (str "Missing command message properties:"
```
mullr: Do we ever call `fatality` without calling `Exception.` for its first parameter? May be worth doing that inside of `fatality`.

mullr: Would this constraint be better expressed using a schema?

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +101,78 @@
...
+       :payload message
+       :annotations (dissoc headers :command :version :certname)})))
+
+(defn- parse-queue-command
+  [{:keys [headers body]}]
+  ;; Queue messages must not have command, version, or certname headers,
+  ;; and the body must be a JSON map like this:
+  ;; {"command":"deactivate node","version":3, "payload":{"certname":...}}
+  (let [{:keys [command version certname]} headers]
+    (when (or command version certname)
```
mullr: Similarly, a schema could do this job well.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +101,78 @@
...
+      (throw+
+       (fatality (Exception. (str "Unexpected command message properties:"
+                                  (when command " command")
+                                  (when version " version")
+                                  (when certname " certname"))))))
+    (update (read-command-body body)
+            :annotations merge
```
mullr: The way this is line wrapped makes it hard to read... would be easier if `:annotations` was on the previous line, I think.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +101,78 @@
...
+            {:received (kitchensink/timestamp) :id (kitchensink/uuid)}
+            headers)))
 
 (defn parse-command
-  "Take a wire-format command and parse it into a command object."
-  [{:keys [headers body]}]
+  "Parses a queue message and returns it as a command object."
```
mullr: s/object/map/?

rbrw: Fine with me if no one objects - seemed unusual to me too.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -111,12 +181,15 @@
   its id."
   [mq-connection :- mq/connection-schema
    mq-endpoint :- s/Str
-   raw-command :- s/Str
-   uuid :- (s/maybe s/Str)]
+   raw-command :- (s/cond-pre s/Str (Class/forName "[B"))
```
mullr: Why do we want `s/cond-pre` instead of `s/either`? The latter is easier to read, and I don't think we need the ordered matching that `s/cond-pre` claims to provide.

mullr: ... as @rbrw points out, `s/either` is deprecated. :whale:

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+  \"certname\" entries, and that the entries are in the correct
+  format, i.e. spaces instead of underscores in the command name,
+  integer instead of string for the version, etc.  Ensures that
+  the :body only contains the \"payload\", as either a string or
+  stream.  The :body will be a stream unless reading the body stream
+  is unavoidable (i.e. old-style, non-param POST)."
+  [handle]
+  (fn [{:keys [body params] :as req}]
+    (let [req (if-let [cmd (params "command")]
+                (assoc-in req [:params "command"] (str/replace cmd "_" " "))
```
mullr: `(update-in req [:params "command"] str/replace "_" " ")`

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+                req)
+          req (if-let [ver (params "version")]
+                (assoc-in req [:params "version"] (Integer/parseInt ver))
```
mullr: ditto

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+                req)
+          {:strs [command version certname]} params
```
mullr: woah, TIL `:strs`

mullr: could move this up to the top and use it to simplify the if-lets

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+          missing-items (fn [kind command version certname]
+                          (http/error-response
+                           (str "Missing POST " kind ":"
```
mullr: schema me!

mullr: ... or maybe not here example. On the params map, perahps?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+        (or command version certname)
+        (missing-items "parameters" command version certname)
+        
+        :else        
+        (do
+          (log/warn
+           (str "Unable to stream command posted without parameters"
+                " (loading into RAM)"))
+          (if-let [body-str (request/body-string req)]
+            (let [body (json/parse-string body-str)
```
mullr: as above, should be able to `parse-stream` this sucker

src/puppetlabs/puppetdb/http/command.clj
rbrw: Hmm, what did you have in mind -- the intention was for them to have the "final" values, after any fixup.

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+        
+        :else        
+        (do
+          (log/warn
+           (str "Unable to stream command posted without parameters"
+                " (loading into RAM)"))
+          (if-let [body-str (request/body-string req)]
+            (let [body (json/parse-string body-str)                    
+                  required-fields #{"command" "version" "payload"}]
+              (if-not (= required-fields (set (keys body)))
```
mullr: schema schema

src/puppetlabs/puppetdb/http/command.clj
mullr: oh, duh, n/m

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+                                    (when certname {"certname" certname}))
+                      {:strs [command version certname]} params]
+                  ;; Enforce certname too? (currently breaks tests)
+                  (if-not (and command version)
+                    (missing-items "elements" command version true)
+                    (handle
+                     (assoc req
+                            :body (json/generate-string (body "payload"))
+                            :params (assoc params "version"
+                                           (Integer. version))))))))
```
mullr: `Integer/parseInt`? (slightly more explicit)

rbrw: I *think* I just copied what was in the final version of @wkalt's patch (that this is on top of), as I thought there'd been discussion there.

mullr: heh ok

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+    ;; For now body will be in-memory, but eventually may be a stream.
+    (try+
+     (let [uuid (kitchensink/uuid)
+           completion-timeout-ms (some-> params
+                                         (get "secondsToWaitForCompletion")
+                                         Double/parseDouble
+                                         (* 1000))
+           submit-params (select-keys params ["certname" "command" "version"])
+           submit-params (if-let [v (submit-params "version")]
+                           (assoc submit-params "version" (str v))
```
mullr: update-in

mullr: or jus t udpate

src/puppetlabs/puppetdb/http/command.clj
rbrw: ...which it looks like doesn't apply here.

test/puppetlabs/puppetdb/http/command_test.clj
```diff
@@ -139,3 +136,110 @@
...
+  (let [normalize (#'tgt/normalize-request identity)]
+
+    ;; Check for an error if some but not all params are present
+    (doseq [items (apply concat
+                         (map #(combinations ["command" "version" "certname"] %)
+                              [1 2]))]
+      (let [result (normalize {:body ::ignored
+                               :params (into {} (for [k items] [k "1"]))})]
+        (is (= 400 (:status result)))
+        (is (= (apply str
```
mullr: Feels like this test should be a bit /less/ specific, maybe just grapa for the things you think are supposed to be there.

test/puppetlabs/puppetdb/http/command_test.clj
```diff
@@ -139,3 +136,110 @@
         (let [timestamp (get-in good-msg [:headers :received])]
           (time/parse (time/formatters :date-time) timestamp))))))
 
+(deftest normalize-request
```
mullr: I think this is actually a handful of different tests. I don't subscribe to the "one assertion per test" rule, but past a couple it gets hard to track down failures.

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -64,6 +65,20 @@
...
+  "Returns a stream that will throw ::body-stream-overflow and drain
+  the rest of the stream if more than max-size-data is read."
+  ;; The drain is because ruby.  i.e. if we closed the connection
+  ;; without that, one of the ruby clients wouldn't handle the broken
+  ;; pipe in a friendly way.
+  (proxy [LimitedInputStream] [stream max-size]
+    (raiseError [max-size count]
+      (loop []
+        (when-not (= -1 (.read stream))
+          (.skip stream (* 1024 1024))
```
mullr: is there any reason not to use MAXINT or something here?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
           timeout-chan (http/json-response {:uuid uuid
                                             :processed false
                                             :timed_out true}
-                                           503)
+                                           http/status-unavailable)
```
mullr: thank you

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+  (if-not reject-large-commands?
+    (if (instance? java.io.InputStream body)
+      (IOUtils/toByteArray body)
+      body)
+    (cond
+      (instance? java.io.InputStream body)
+      (IOUtils/toByteArray
+       (restrained-drained-stream body (long max-command-size)))
+
+      (string? body) ;; Assuming UCS-2
```
mullr: I get what you're doing here, but maybe we should say it a bit more clearly in the comment

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
-                                        (get "secondsToWaitForCompletion")
-                                        Double/parseDouble
-                                        (* 1000))
-          do-submit #(enqueue-fn body-string uuid)]
-      (if (some-> completion-timeout-ms pos?)
-        (blocking-submit-command do-submit (get-response-pub) uuid completion-timeout-ms)
-        (do
-          (do-submit)
-          (http/json-response {:uuid uuid}))))))
+  [enqueue-fn get-response-pub reject-large-commands? max-command-size]
```
mullr: can we imply `reject-large-commands?` with a `nil` `max-command-size`? Not sure if that's good or bad, but at least it's one fewer thing to flow around.

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +92,145 @@
...
+                                  submit-params)]
+       (if (some-> completion-timeout-ms pos?)
+         (blocking-submit-command do-submit (get-response-pub)
+                                  uuid
+                                  completion-timeout-ms)
+         (do
+           (do-submit)
+           (http/json-response {:uuid uuid}))))
+     (catch (= ::body-stream-overflow %) _
+       (-> "Command rejected due to size exceeding max-command-size"
```
mullr: This is better in theory, but I'm not sure I like it any more than the map.

mullr:
@rbrw ok, finished my pass through this, let me know if you have any questions.

src/puppetlabs/puppetdb/http/command.clj
rbrw: ...perhaps not, just being cautious (could even be MAXLONG, API-wise), but that reminds me, I should double-check this with @senior  He and I dealt with some corner cases in his patch that I don't recall clearly offhand.

src/puppetlabs/puppetdb/command.clj
rbrw: Agreed, though since we also refer to the class in a couple of other contexts, I just added a utils/byte-array-class and used that everywhere.

src/puppetlabs/puppetdb/command.clj
rbrw: With respect to fatality, yes, it's used to forward an existing exception in upon-error-throw-fatality.

src/puppetlabs/puppetdb/http/command.clj
rbrw: I think that may pass nil to str/replace when there's no command.

rbrw: Oh, wait, you probably meant instead of the assoc-in, not the whole if-let.

test/puppetlabs/puppetdb/http/command_test.clj
rbrw: Hmm, well I wanted the test to make sure that the string actually matches the error case.  I suppose I could have chopped off the bit after the colon and turned that into a set for comparison.  Would that be preferable?

src/puppetlabs/puppetdb/http/command.clj
rbrw: Should be error-response - was just mislead by some existing code.

test/puppetlabs/puppetdb/http/command_test.clj
mullr: My concern is that exact string matching on error messages is very fragile. Parsing the message would be pretty fragile as well. Saying it contains some word or another is about as far as I'd want to go.

rbrw:
@mullr In addition to the top commit, the "replace coerce-to-json" and "byte-array-class" patches are new.  I think that should be the majority of the changes since you last looked.

test/puppetlabs/puppetdb/http/command_test.clj
rbrw: I look at it as a tradeoff (in the absence of more discriminating exceptions), if you're not specific enough, then you of course risk false negatives, and false positives mean you have to update the tests with the code, which doesn't bother me too much as long as it's not too frequent.

rbrw:
This will need a final squash before merge.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -56,6 +56,7 @@
    In either case, the command itself, once string-ified, must be a
    JSON-formatted string with the aforementioned structure."
   (:require [clojure.tools.logging :as log]
+            [clojure.java.io :as io]
```
mullr: unused

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +103,110 @@
...
+(def new-message-schema
+  {:headers {:command s/Str
+             :version s/Str
+             :certname s/Str
+             :received s/Str
+             :id s/Str}
+   :body (s/cond-pre s/Str utils/byte-array-class)})
+
+(def queue-message-schema
+  ;; FIXME: do we want to (and can we) be so strict here, i.e. we want
```
mullr: since there's a ticket for this, we can take out the FIXME

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +103,110 @@
...
+(defn-validated parse-new-command :- new-command-schema
+  "Parses a new-format queue command and returns it in the traditional
+  format (see parse-queue-command).  New-style messages must have all
+  5 headers (command, version, certname, received, and id), and the
+  body must be the bare command content (i.e. the payload).  So for a
+  \"deactivate node\" command, a string like
+  {\"certname\":\"test1\",\"producer_timestamp\":\"2015-01-01\"}."
+  [{:keys [headers body]} :- new-message-schema]
+  (let [{:keys [command version certname received id]} headers
+        version (some-> headers :version Integer/parseInt)]
```
mullr: this immediately shadows `version` above, and the use of `some->` contradicts the assertion in the docstring that it's required.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +103,110 @@
...
+(defn-validated parse-queue-command :- queue-command-schema
+  "Parses a traditional queue command (also the current format for
+  failed, re-queued messages).  See parse-new-command for the handling
+  of newly received messages.  These messages must not have command,
+  version, or certname headers, and the body must be a JSON map like
+  this: {\"command\":\"deactivate node\",\"version\":3,\"payload\":{\"certname\":...}}."
+  [{:keys [headers body]} :- queue-message-schema]
+  (let [{:keys [command version certname]} headers]
+    (update (json/parse-strict body true) :annotations merge
+            ;; cf. PDB-2315
```
mullr: not a fan of issue comments in the code... if it's important enough to put here, let's put the ticket higher in the backlog and fix it soon

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +103,110 @@
...
+            {:received (kitchensink/timestamp) :id (kitchensink/uuid)}
+            headers)))
+
+(defn-validated parse-command
```
mullr: This won't validate anything unless you use `:- my-schema` somewhere, I think

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +97,142 @@
...
+   s/Any s/Any})
+
+(defn-validated normalize-new-request
+  [{:keys [params body] :as req} :- new-request-schema]
+  (-> req
+      (update-in [:params "command"] str/replace "_" " ")
+      (update-in [:params "version"] #(Integer/parseInt %))))
+
+(def old-request-schema
+  (s/conditional
```
mullr: A 1-armed conditional schema is a little funny... What happens if it's not a map?

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -77,40 +97,142 @@
...
+  [{:keys [params body] :as req} :- old-request-schema]
+  (log/warn (str "Unable to stream command posted without parameters"
+                 " (loading into RAM)"))
+  (if-not body
+    (http/error-response "Empty application/json POST body")
+    (let [body (json/parse-strict (:body req))]
+      (if (empty? body)
+        (http/error-response "Empty application/json POST body")
+        (do
+          ;; Enforce certname too? (currently breaks tests)
```
mullr: ticket me as tech debt!

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -133,6 +133,17 @@
              (query->sql remaining-query entity version paging-options)]
          (jdbc/with-query-results-cursor results-query (comp row-fn munge-fn)))))))
 
+;; Do we still need this, i.e. do we need the pass-through, and the
+;; strict selectivity in the caller below?
```
mullr: errrr, I don't know. I don't know why we would want to parse-strict when it's a string though, that doesn't make much sense to me.

mullr:
Just a few small comments, looking much better overall

src/puppetlabs/puppetdb/command.clj
rbrw: @mullr fair enough - thought it might make things a bit easier on the person working on the ticket.

src/puppetlabs/puppetdb/http/command.clj
rbrw: I thought it was an error (i.e. thought the docs said it was like cond, except that fallthough was a schema error, like case).

src/puppetlabs/puppetdb/command.clj
rbrw: Thanks (missed that in the function split).

src/puppetlabs/puppetdb/command.clj
rbrw: If I understand you, that's intentional since the validation's already been handled by the two old/new sub-calls, and I didn't want to duplicate the work.

src/puppetlabs/puppetdb/http/command.clj
rbrw: https://tickets.puppetlabs.com/browse/PDB-2347

src/puppetlabs/puppetdb/query_eng.clj
rbrw: I suppose it could be a lot more efficient to parse 1M top-level vector items one at a time (assuming a very large string), as long as you can avoid holding on to the head.  In any case, at least in the new calls, I just used strict everywhere.

src/puppetlabs/puppetdb/command.clj
mullr: that's fine; in that case, we should just say `defn`

src/puppetlabs/puppetdb/command.clj
```diff
@@ -86,23 +103,110 @@
...
+           (empty? (select-keys x [:command :version :certname])))))})
+
+(def queue-command-schema
+  ;; Created to match existing test behavior, but is that
+  ;; comprehensive? i.e. if we're wrong, and this is too strict, then
+  ;; it could break on outside queue content on release.  Adding a
+  ;; schema here could change our effectively public queue API.
+  (-> new-command-schema
+      ;;(assoc :command (s/cond-pre s/Int s/Str))
+      ;;(assoc :version (s/cond-pre s/Int s/Str))
```
wkalt: what's the story with these lines? Seems like maybe the alternative cases have been ruled out?

src/puppetlabs/puppetdb/http/command.clj
mullr: ughhh the schema docs recommend this style don't they. Ah well.

src/puppetlabs/puppetdb/query_eng.clj
mullr: hm. well, I imagine we'll refine our view on this as we move into streaming more.

src/puppetlabs/puppetdb/command.clj
rbrw: Ahh, indeed - so we should.

test/puppetlabs/puppetdb/http/command_test.clj
```diff
@@ -139,3 +137,111 @@
...
+    ;; Make sure that when all three params are present, the body is passed on.
+    (let [before-params {"certname" "x" "command" "y_z" "version" "1"
+                         "received" (kitchensink/timestamp)}
+          before {:params before-params
+                  :body (ByteArrayInputStream. (byte-array 0))}
+          after (normalize before)
+          after-params (:params after)]
+      ;; Accomodate the fact that command and version will be transformed.
+      (is (identical? (:body before) (:body after)))
+      (is (= (dissoc before :params) (dissoc after :params)))
```
wkalt: is this testing the same thing as the line above?

src/puppetlabs/puppetdb/query_eng.clj
rbrw: Yeah, though with some luck we eventually might not care about the parse-string functions at all.

src/puppetlabs/puppetdb/command.clj
rbrw: Doh, I meant to kill those before I pushed.

test/puppetlabs/puppetdb/http/command_test.clj
rbrw: The identical? is testing that the function passes the *exact* pointer straight through, i.e. we hopefully don't do any work on it (but we're just indirectly checking that by seeing that it's passed through - so not a definitive test).  the latter is checking that excepting params, before and after (as a whole) are the same, wrt "=", though not identical?.  i.e. no new map entries are inserted, etc.

wkalt:
can we pull out the checksum code in the terminus now that the server is dropping it?

test/puppetlabs/puppetdb/http/command_test.clj
wkalt: got it, thanks

wkalt:
generally +1

rbrw:
Good point wrt terminus: https://tickets.puppetlabs.com/browse/PDB-2348

PR #1813, opened 2016-01-13
(PDB-2182) support paging operators in query body

wkalt:
This patch implements support for paging operators within a "from" clause.

### DIFF
--- a/documentation/api/query/v4/operators.markdown
+++ b/documentation/api/query/v4/operators.markdown
@@ -191,7 +191,7 @@ to `nodes`.
 ### `from`
 
 The `from` operator allows you to choose the [entity][entities] that you want to query and
-provide an optional query clause for filtering those results. This operator can
+provide optional query and paging clauses to filter those results. This operator can
 be used at the top-level context of a query:
 
     ["from", "nodes", ["=", "certname", "myserver"]]
@@ -202,6 +202,58 @@ using the [`in` operator](#subquery-operators).
 When querying a particular endpoint, such as `/pdb/query/v4/nodes`, the endpoint provides
 the context for the query. Querying the [root] endpoint requires specifying a
 context explicitly.
+
+## Paging operators (`limit`, `offset`, `order_by`)
+
+PuppetDB allows specification of paging clauses within a "from" clause in a
+query or subquery. The `limit` and `offset` operators both accept an
+integer-valued argument, and `order_by` accepts a vector of either column names
+or vector pairs containing a column name and an ordering of "asc" or "desc".
+For example,
+
+    ["limit", 1]
+
+    ["offset", 1]
+
+    ["order_by", ["certname"]]
+
+    ["order_by", ["certname", ["timestamp", "desc"]]]
+
+When no ordering is explicitly specified, as in the case of "certname" in the
+example above, ascending order is assumed. Here are a few examples of queries
+using paging operators:
+
+Return the most recent ten reports for a certname:
+
+    ["from", "reports",
+      ["=", "certname", "myserver"],
+      ["order_by", [["timestamp", "desc"]]],
+      ["limit", 10]]
+
+Return the next page of ten reports:
+
+    ["from", "reports",
+      ["=", "certname", "myserver"],
+      ["order_by", [["timestamp", "desc"]]],
+      ["limit", 10],
+      ["offset", 10]]
+
+Return the most recent ten reports for any certname:
+
+    ["from", "reports",
+      ["order_by", [["timestamp", "desc"]]],
+      ["limit", 10]]
+
+Return the nodes represented in the ten most recent reports:
+
+    ["from", "nodes",
+      ["in", "certname",
+        ["from", "reports",
+          ["extract", "certname"],
+          ["limit", 10],
+          ["order_by", [["certname", "desc"]]]]]]
+
+The order in which paging operators are supplied does not matter.
 
 ## Subquery operators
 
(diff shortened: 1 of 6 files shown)


wkalt:
This does not yet support queries of the form

```
[from [from reports [= certname foo] [limit 100]] [extract [[function count] status] [group_by status]]]
```
I'm still looking at ways to get support for this, but given that there's no planned PQL analog and it'll require some fairly invasive changes to from/extract, it's probably worth splitting out if it's wanted.

wkalt:

documentation/api/query/v4/operators.markdown
```diff
@@ -204,6 +204,38 @@ When querying a particular endpoint, such as `/pdb/query/v4/nodes` the endpoint
...
+
+    ["from", "reports", ["=", "certname", "myserver"], ["order_by", [["timestamp", "desc"]]], ["limit", 10], ["offset", 10]]
+
+Return the most recent ten reports for any certname:
+
+    ["from", "reports", ["order_by", [["timestamp", "desc"]]], ["limit", 10]]
+
+Return the nodes represented in the ten most recent reports:
+
+    ["from", "nodes", ["in", "certname", ["from", "reports", ["extract", "certname"], ["limit", 10], ["order_by", [["certname", "desc"]]]]]]
```
kbarber: Examples don't wrap in the docs website, maybe these few examples would read better if they were multi-line?

documentation/api/query/v4/operators.markdown
```diff
@@ -204,6 +204,38 @@ When querying a particular endpoint, such as `/pdb/query/v4/nodes` the endpoint
...
 
+## Paging Operators (`limit`, `offset`, `order_by`)
+
+PuppetDB allows specification of paging clauses within a "from" clause in a
+query or subquery. The `limit` and `offset` operators both accept an
+integer-valued argument, and `order_by` accepts a vector of either column names
+or vector pairs containing a column name and an ordering of "asc" or "desc".
+For example,
+
+    ["order_by", ["certname", ["timestamp", "desc"]]]
```
kbarber: Breaking down this paragraph above, showing an example like this for each might be nice. Like "The limit and offset operators both accept an integer valued argument: <insert examples>. The 'order_by' ... blah blah: <insert examples>"

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1267,19 +1267,48 @@
...
+  [clause clauses]
+  (when-let [candidates (seq (filter #(= (first %) clause) clauses))]
+    (if (> (count candidates) 1)
+      (throw (IllegalArgumentException.
+               (format "Multiple %s clauses are not permitted" clause)))
+      (second (first candidates)))))
+
+(defn process-order-by
+  [clauses]
+  (when clauses
```
kbarber: validation question: should this be a valid query if there are no clauses? For me it silently works.

kbarber:
I'm +1, this is looking nice, I'm glad we got the subquery and outer query case, the symmetry looks nice.

I do worry about validation, and generally how sucky we are at validation - because I can see numerous edge cases that should throw errors but don't, but this is no different to our existing code ... the problem being that validation has to occur at the leaves of our code, rather than something which throws on the shape of the data at a higher level, hohum. I honestly think this needs a higher level rethink though.

wkalt:
@KenBarber addressed your comments

documentation/api/query/v4/operators.markdown
```diff
@@ -203,6 +203,58 @@ When querying a particular endpoint, such as `/pdb/query/v4/nodes`, the endpoint
 the context for the query. Querying the [root] endpoint requires specifying a
 context explicitly.
 
+## Paging operators (`limit`, `offset`, `order_by`)
+
+PuppetDB allows specification of paging clauses within a "from" clause in a
+query or subquery. The `limit` and `offset` operators both accept an
```
mullr: s/operators/options/  ?

mullr: I guess this is in the "operators" doc, so that's just what we call them. Not sure how I feel about that, but whatever.

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -18,6 +18,13 @@
 (def query-params ["query" "limit" "offset" "order_by" "include_total"])
 (def count-header "X-Records")
 
+(defn munge-query-ordering
+  [clauses]
+  (for [c clauses]
+    (if (vector? c)
+      (assoc c 1 ((second c) {:asc :ascending :desc :descending}))
```
mullr: `(update c 1 {:asc :ascending :desc :descending})`

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1270,19 +1270,54 @@
...
+
+      :else (second (first candidates))))
+
+(defn process-order-by
+  [clauses]
+  (when clauses
+    (for [clause clauses]
+      (if (vector? clause) (mapv keyword clause) (keyword clause)))))
+
+(defn create-paging-map
```
mullr: I'd like to see a hint in here somewhere, with a schema or docstring or something, of what the data looks like coming into and going out of this function.

mullr:
Generally +1, just some doc and style notes.

documentation/api/query/v4/operators.markdown
kbarber: I proposed a rename of the doc actually in my PQL patch to 'AST query language' and making it more about the AST language, instead of just about operators and such. I'd be happy to rethink our nomenclature.

mullr: It may be worth a few brain cells. `operator` in particular seems wrong here, since it's not, well, operating on anything. But if that's the least of our problems, we're in good shape.

PR #1814, opened 2016-01-14
(maint) stop deleting unassociated report statuses

wkalt:
This removes a gc operation that can become expensive on large environments.
Since report_statuses has 3 rows max, there's no reason to clean the table.


wkalt:
Any reason not to do this? @npwalker timed this query at 138735 ms on the scale testing environment.

npwalker:
Seems like if we change a report status it would come from puppet and we could clean out the unused status as a migration to a newer version of puppetdb.  

Totally agree we should remove this code. 

```
pe-puppetdb=# select * from report_statuses;
 id |  status
----+-----------
  1 | failed
  2 | changed
  3 | unchanged
(3 rows)
```

PR #1815, opened 2016-01-15
(tests) Don't install rubygems package on Ubuntu 14.04 Trusty

kurtwall:
On trusty (and later?), the gem command is packaged with the ruby package instead. Work done in support of PDB-997.


kbarber:
@kurtwall this is targeted at master, not stable. Want to rebase/reopen it?

PR #1817, opened 2016-01-15
(maint) Fix nodes query for historical catalogs

ajroetker:
This commit fixes the nodes query to left join on only the latest
catalog and not all (historical catalogs) and fixes the documentation
for resource-graphs.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -117,8 +117,11 @@
                                "resources" {:columns ["certname"]}}
 
                :selection {:from [:certnames]
-                           :left-join [:catalogs
-                                       [:= :certnames.certname :catalogs.certname]
+                           :left-join [:latest_catalogs
+                                       [:= :latest_catalogs.certname_id :certnames.id]
+
+                                       :catalogs
+                                       [:= :catalogs.id :latest_catalogs.catalog_id]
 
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -117,17 +117,17 @@
...
-                                       [:= :certnames.certname :fs.certname]
+                                       :catalogs
+                                       [:= :catalogs.id :latest_catalogs.catalog_id]
 
                                        :reports
-                                       [:and
-                                        [:= :certnames.certname :reports.certname]
-                                        [:in :reports.id {:select [:latest_report_id]
-                                                          :from [:certnames]}]]
+                                       [:= :certnames.latest_report_id :reports.id]
```
ajroetker: Fixed this up too while i was in here, is this what you had in mind?

ajroetker: ping @wkalt

wkalt: yeah, exactly

wkalt: we may want this change in stable though @ajroetker

ajroetker:
@wkalt fixed up for your comment by moving the reports change to another PR

wkalt:
works for me

PR #1818, opened 2016-01-15
(maint) Use join on `id` for nodes-query with reports

ajroetker:
This commit changes the nodes query to use a "regular" left join from
certnames to reports using the `latest_report_id` instead of using an
`in` expression.


waynr:
test this please

senior:
@ajroetker This is to fix a currently broken query right? If so, I'd really like to see a test that covers us here in case we break it again.

ajroetker:
@senior this query isn't broken, just inefficient, I think you're think of a fix that was already merged

PR #1820, opened 2016-01-15
(PDB-2271) support relative path to vardir

cprice404:
This commit modifies the configuration validation to support
specifying a relative path, which is necessary in order to
support a portable development environment where users
can run the system "out-of-the-box" (e.g., by including
a developer config file that has a hard-coded vardir path
of "./target/puppetdb-vardir".)

This is a requirement for the PE-AIO dev environment and tests
to work properly, at least based on their current implementation.

### DIFF
--- a/src/puppetlabs/puppetdb/config.clj
+++ b/src/puppetlabs/puppetdb/config.clj
@@ -245,7 +245,10 @@
   "Checks that `vardir` is specified, exists, and is writeable, throwing
   appropriate exceptions if any condition is unmet."
   [config]
-  (let [vardir (io/file (get-in config [:global :vardir]))]
+  (let [vardir (some-> (get-in config [:global :vardir])
+                       io/file
+                       fs/absolute-path
+                       fs/file)]
     (cond
      (nil? vardir)
      (throw (IllegalArgumentException.
--- a/test/puppetlabs/puppetdb/config_test.clj
+++ b/test/puppetlabs/puppetdb/config_test.clj
@@ -143,10 +143,6 @@
     (is (thrown-with-msg? IllegalArgumentException #"is not specified"
                           (validate-vardir {:global {:vardir nil}}))))
 
-  (testing "should fail if it's not an absolute path"
-    (is (thrown-with-msg? IllegalArgumentException #"must be an absolute path"
-                          (validate-vardir (vardir "foo/bar/baz")))))
-
   (testing "should fail if it doesn't exist"
     (is (thrown-with-msg? java.io.FileNotFoundException #"does not exist"
                           (validate-vardir (vardir "/abc/def/ghi")))))
@@ -172,6 +168,15 @@
                      (.delete)
                      (.mkdir)
                      (.setWritable true))]
+      (is (= (validate-vardir (vardir filename))
+             (vardir filename)))))
+
+  (testing "should support relative paths"
+    (let [filename "./target/totally/okay"]
+      (doto (fs/file filename)
+        (.deleteOnExit)
+        (.mkdirs)
+        (.setWritable true))
       (is (= (validate-vardir (vardir filename))
              (vardir filename))))))
 


cprice404:
Interested in hearing what you guys think of this; we can potentially discuss other options if this is undesirable for some reason that I'm not thinking of.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -245,7 +245,10 @@
   "Checks that `vardir` is specified, exists, and is writeable, throwing
   appropriate exceptions if any condition is unmet."
   [config]
-  (let [vardir (io/file (get-in config [:global :vardir]))]
+  (let [vardir (some-> (get-in config [:global :vardir])
+                       io/file
```
ajroetker: Is this `io/file` necessary? https://github.com/Raynes/fs/blob/master/src/me/raynes/fs.clj#L58

ajroetker: Sorry that should have been a link to here, https://github.com/Raynes/fs/blob/master/src/me/raynes/fs.clj#L106

src/puppetlabs/puppetdb/config.clj
```diff
@@ -245,7 +245,10 @@
...
+                       fs/absolute-path
+                       fs/file)]
```
ajroetker: It looks like this is redundant as well as `fs/absolute-path` calls `fs/file`

ajroetker:
@cprice404 We also have some validation https://github.com/cprice404/puppetdb/blob/feature/master/PDB-2271-support-relative-vardir/src/puppetlabs/puppetdb/config.clj#L257 for checking the path is absolute

test/puppetlabs/puppetdb/config_test.clj
```diff
@@ -173,6 +169,15 @@
                      (.mkdir)
                      (.setWritable true))]
       (is (= (validate-vardir (vardir filename))
+             (vardir filename)))))
+
+  (testing "should support relative paths"
+    (let [filename "./target/totally/okay"]
+      (doto (fs/file filename)
+        (.deleteOnExit)
+        (.mkdirs)
```
ajroetker: You cold probably use our testutils function for doing some of this https://github.com/puppetlabs/puppetdb/blob/master/test/puppetlabs/puppetdb/testutils.clj#L254

cprice404: Perhaps, but in this case it's important that the path be relative, and at a glance it looks like that fn would return an absolute path.

In any case I'm more than happy to refactor this PR however the team feels is appropriate, assuming you guys are OK with the general spirit of the PR.

ajroetker: I more meant you could use `(temp-dir "foo")` and use that for testing out relativity from that temp-dir. I.e. `/foo/../foo/./bar` or some such.

ajroetker:
@cprice404 are there any relative path in puppet-server land? Such as the certs or ruby paths?

test/puppetlabs/puppetdb/config_test.clj
cprice404: I think that anything that begins with a leading slash is considered an absolute path by the various java/clojure APIs.

src/puppetlabs/puppetdb/config.clj
cprice404: You are probably correct about both of the above, thanks.  Will fix, assuming you guys are OK with the behavior change.

cprice404:
> @cprice404 are there any relative paths in puppet-server land? Such as the certs or ruby paths?

The other projects in PE all *tolerate* relative paths.  The PE modules will manage the config files using absolute paths, but the services will work OK with relative paths, and we do rely on that for a fair amount of stuff in dev environments and tests in Puppet Server.

If you have a look at this file:

https://github.com/puppetlabs/pe-aio/blob/master/dev/pe-puppet-server.conf.sample

you can see lots of examples of places where the config for the various other PE services works OK with relative paths.

waynr:
test this please

senior:
@cprice404 you're not thinking we'll support this for end users right? This is just something we're putting in to make dev easier with an all in one JVM PE?

cprice404:
@senior not "support", per se, but perhaps not actively prevent?

senior:
@cprice404 to be more precise, we'll not document it, not actively encourage it's use, but will be a useful escape hatch for testing?

cprice404:
@senior that sounds accurate to me, yes.

cprice404:
@senior thanks for merging this!  If you guys would like for me to go back and follow-up on some of @ajroetker 's suggestions, just let me know.

PR #1821, opened 2016-01-19
(PDB-2186) Add PuppetDB utility function for querying PuppetDB

ajroetker:
This commit adds a PuppetDB terminus utility function to query PuppetDB at the `root`
endpoint.

### DIFF
--- a/puppet/lib/puppet/util/puppetdb.rb
+++ b/puppet/lib/puppet/util/puppetdb.rb
@@ -6,6 +6,7 @@ require 'puppet/util/puppetdb/command'
 require 'puppet/util/puppetdb/config'
 require 'digest/sha1'
 require 'time'
+require 'json'
 require 'fileutils'
 
 module Puppet::Util::Puppetdb
@@ -61,6 +62,21 @@ module Puppet::Util::Puppetdb
             [:puppetdb, :command, :submit, command_name, version]) do
       command = Puppet::Util::Puppetdb::Command.new(command_name, version, certname, payload)
       command.submit
+    end
+  end
+
+  # Query PuppetDB.
+  #
+  # @param query [String, Array] The PQL or AST query for PuppetDB
+  # @return [Array<Hash>]
+  def query_puppetdb(query)
+    profile("Submitted query '#{query}'", [:puppetdb, :query, query]) do
+      headers = { "Accept" => "application/json",
+                  "Content-Type" => "application/json; charset=UTF-8" }
+      response = Puppet::Util::Puppetdb::Http.action("/pdb/query/v4", :query) do |http_instance, path|
+        http_instance.post(path, { 'query' => query }.to_json, headers)
+      end
+      JSON.parse(response.body)
     end
   end
 
--- a/puppet/lib/puppet/util/puppetdb/http.rb
+++ b/puppet/lib/puppet/util/puppetdb/http.rb
@@ -147,7 +147,7 @@ module Puppet::Util::Puppetdb
         request_exception = with_http_error_logging(server_url, route) {
           http = Puppet::Network::HttpPool.http_instance(server_url.host, server_url.port)
 
-          response = timeout(config.server_url_timeout) do
+          response = Timeout.timeout(config.server_url_timeout) do
             http_callback.call(http, route)
           end
         }
@@ -182,7 +182,7 @@ module Puppet::Util::Puppetdb
         request_exception = with_http_error_logging(server_url, route) {
           http = Puppet::Network::HttpPool.http_instance(server_url.host, server_url.port)
 
-          response = timeout(config.server_url_timeout) do
+          response = Timeout.timeout(config.server_url_timeout) do
             http_callback.call(http, route)
           end
         }
--- a/puppet/spec/unit/util/puppetdb_spec.rb
+++ b/puppet/spec/unit/util/puppetdb_spec.rb
@@ -5,27 +5,47 @@ require 'spec_helper'
 require 'digest/sha1'
 require 'puppet/util/puppetdb'
 require 'puppet/util/puppetdb/command_names'
+require 'puppet/util/puppetdb/http'
+require 'json'
 
-Command                 = Puppet::Util::Puppetdb::Command
 
+class FakeHttpResponse
+  def initialize(body)
+    @body = body
+  end
+  attr_reader :body
+end
 
 describe Puppet::Util::Puppetdb do
   subject { Object.new.extend described_class }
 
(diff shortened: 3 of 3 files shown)


puppet/lib/puppet/util/puppetdb.rb
```diff
@@ -64,6 +66,26 @@ def submit_command(certname, payload, command_name, version)
...
+  # @param version [Number] version number of command
+  # @return [Hash <String, String>]
+  def query_puppetdb(query)
+    profile("Submitted query '#{query}'",
+            [:puppetdb, :query, query]) do
+      headers = { "Accept" => "application/json",
+                  "Content-Type" => "application/json; charset=UTF-8" }
+      response = Puppet::Util::Puppetdb::Http.action("/pdb/query/v4?query=#{CGI.escape(query)}",
+                                                     :query) do |http_instance, path|
+        http_instance.get(path, headers)
```
mullr: Should we use the POST query api here, just as insurance against monster queries?

ajroetker: Can we add post params with our http code?

mullr: don't need them. Just a json body:

```
{ "query": query }
```

mullr: (assuming that actually works on this endpoint)

puppet/lib/puppet/util/puppetdb.rb
```diff
@@ -64,6 +66,25 @@ def submit_command(certname, payload, command_name, version)
     end
   end
 
+  # Query PuppetDB.
+  #
+  # @param certname [String] The certname this command operates on
+  # @param payload [String] payload
+  # @param command_name [String] name of command
+  # @param version [Number] version number of command
+  # @return [Hash <String, String>]
```
mullr: I think these `@` comments go with something else

puppet/lib/puppet/util/puppetdb.rb
```diff
@@ -64,6 +66,25 @@ def submit_command(certname, payload, command_name, version)
...
+  def query_puppetdb(query)
+    profile("Submitted query '#{query}'",
+            [:puppetdb, :query, query]) do
+      headers = { "Accept" => "application/json",
+                  "Content-Type" => "application/json; charset=UTF-8" }
+      response = Puppet::Util::Puppetdb::Http.action("/pdb/query/v4", :query) do |http_instance, path|
+        http_instance.post(path, { 'query' => query }.to_json, headers)
```
mullr: do we need to tell to_json to use UTF-8, or is that the default?

ajroetker: We shouldn't need to worry about encoding issues here since it already goes through the puppet compiler to get here.

puppet/lib/puppet/util/puppetdb.rb
```diff
@@ -64,6 +66,22 @@ def submit_command(certname, payload, command_name, version)
     end
   end
 
+  # Query PuppetDB.
+  #
+  # @param query [String] The PQL or AST query for PuppetDB
+  # @return [Hash]
```
kbarber: `[Array<Hash>]` isn't it?

puppet/lib/puppet/util/puppetdb.rb
```diff
@@ -64,6 +66,22 @@ def submit_command(certname, payload, command_name, version)
     end
   end
 
+  # Query PuppetDB.
+  #
+  # @param query [String] The PQL or AST query for PuppetDB
```
kbarber: `[String, Array]` ... right? I mean, it takes an array for AST, and a string for PQL, so it could be either?

puppet/lib/puppet/parser/functions/query.rb
```diff
@@ -0,0 +1,45 @@
+Puppet::Parser::Functions.newfunction(:query, :type => :rvalue, :arity => -3, :doc => <<-EOT
```
kbarber: Hmm. Acceptance test perhaps? With the function in a manifest? I can't think of any other way to test this without changing up what we do in some majorish way. Hrm.

puppet/lib/puppet/parser/functions/query.rb
```diff
@@ -0,0 +1,45 @@
...
+  query(["from", "resources",
+         ["and", ["=", "type", "Class"],
+          ["=", "title", "Puppet_enterprise::Profile::Puppetdb"],
+          ["not", ["=", "certname", $certname]]]])
+  [{"certname" => "foo.bar.com"
+    "environment" => "production",
+    "parameters" => {"master_certname" => "master.foo.lan"
+                     ...}} ...]
+
+  query(["from", "resources",
```
kbarber: Hmm, can this sort of pivot, be outside of the baked in functionality for now? I just feel like a pivot functionality on an array of hashes is fairly generic, and is a modification one can make to the resulting data later on.

kbarber: Also, perhaps we're designing too far ahead :-).

puppet/lib/puppet/parser/functions/query.rb
```diff
@@ -0,0 +1,45 @@
...
+         ["and", ["=", "type", "Class"],
+          ["=", "title", "Puppet_enterprise::Profile::Puppetdb"],
+          ["not", ["=", "certname", $certname]]]],
+        'certname')
+  {"foo.bar.com" => {"certname" => "foo.bar.com"
+                     "parameters" => {"master_certname" => "master.foo.lan"
+                                      ...}}}
+
+  query(["from", "resources",
```
kbarber: Same here, just feels like we should keep what the function does for now very simple, and add this later on when demand hits, or we feel its necessary. Again, this kind of thing should be a generic function also, probably in stdlib or something?

puppet/lib/puppet/parser/functions/query.rb
```diff
@@ -0,0 +1,45 @@
...
+         ["and", ["=", "type", "Class"],
+          ["=", "title", "Puppet_enterprise::Profile::Puppetdb"],
+          ["not", ["=", "certname", $certname]]]]'
+        'certname',
+        true)
+  ["foo.bar.com", "baz.buzz.net", ...]
+EOT
+                                     ) do |args|
```
kbarber: I think a unit test, that mocks the final call would be a good thing here, anything that at least checks the code syntax would be nice, right now without any unit test even basic ruby validation doesn't occur I think.

PR #1823, opened 2016-01-20
(maint) use table stats to compute the number of resources

wkalt:
This commit changes one of our population metrics to use table stats to compute
the number of resources. It also eliminates a restriction to active nodes for
the metric, and splits the num-nodes metric into num-active-nodes and
num-inactive-nodes in the dashboard to make things clearer. num-nodes is left
around for backward compatibility. Lastly it adds a mechanism for independent
specification of chart options for our metrics, and changes the
percent-resource-duplication query to run every 30 seconds instead of 5. This
may or may not be the right number.

### DIFF
--- a/resources/public/dashboard/index.html
+++ b/resources/public/dashboard/index.html
@@ -108,6 +108,19 @@ body {
     };
   };
 
+  function merge(a, b){
+      // merge objects a and b, giving b's contents precedence and returning
+      // a new object
+      var result = {};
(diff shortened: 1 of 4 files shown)


kbarber:
@wkalt looks like you still have a reference to pop/num-nodes in population_test.clj

wkalt:

wkalt:
reopened on master

PR #1825, opened 2016-01-22
(maint) use table stats to compute the number of resources

wkalt:
This commit changes one of our population metrics to use table stats to compute
the number of resources. It also eliminates a restriction to active nodes for
the metric, and splits the num-nodes metric into num-active-nodes and
num-inactive-nodes in the dashboard to make things clearer. num-nodes is left
around for backward compatibility. Lastly it adds a mechanism for independent
specification of chart options for our metrics, and changes the
percent-resource-duplication query to run every 30 seconds instead of 5. This
may or may not be the right number.


src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -223,6 +223,14 @@
   [column]
   (format "encode(%s::bytea, 'hex')" column))
 
+(defn vacuum-analyze
+  "Must call within a db connection"
+  [db]
+  (sql/with-db-connection [conn db]
+    (-> (doto (:connection conn) (.setAutoCommit true))
+        .createStatement
+        (.execute "vacuum (analyze)"))))
```
ajroetker: There's a jdbc function for 231/232, http://clojure.github.io/java.jdbc/#clojure.java.jdbc/execute!, is this any different from what you have there?

ajroetker:
Looks good to me, I'm ok with the overall design and the dashboard and new options merge looks good to me as well.

resources/public/dashboard/index.html
```diff
@@ -108,6 +108,19 @@
     };
   };
 
+  function merge(a, b){
+      // merge objects a and b, giving b's contents precedence and returning
+      // a new object
+      var result = {};
+      for (item in a) {
```
mullr: `var item`

src/puppetlabs/puppetdb/query/population.clj
```diff
@@ -11,61 +11,54 @@
...
-                                 where-nodes-are-active
-                                 (hh/select :catalog_resources.resource)
-                                 (hh/modifiers :distinct))
-        num-unique (-> (hh/from [distinct-resources-q :r])
-                       select-count)
-        num-total (num-resources)]
-    (quotient (- num-total num-unique) num-total)))
+  (let [nresources (num-resources)
+        ndistinct (-> "select count(*) as c from
+                       (select distinct resource from catalog_resources) dist"
```
mullr: This is the expensive query, right? Throttling the client (the dashboard) should indeed prevent it from knocking over the system. But since this is exposed via jmx, any jmx client could do the same thing. This includes an overzealous (misconfigured?) monitoring system. For this reason, we should consider throttling this on the server side instead / as well. (memoization with a timeout or something similarly simple would do nicely, I think)

test/puppetlabs/puppetdb/query/population_test.clj
```diff
@@ -30,86 +32,76 @@
...
-        (jdbc/insert! :latest_catalogs {:catalog_id 2 :certname_id 2})
         (jdbc/insert!
-         :resource_params_cache
-         {:resource (sutils/munge-hash-for-storage "01") :parameters nil}
-         {:resource (sutils/munge-hash-for-storage "02") :parameters nil}
-         {:resource (sutils/munge-hash-for-storage "03") :parameters nil})
+          :resource_params_cache
+          {:resource (sutils/munge-hash-for-storage "01") :parameters nil}
+          {:resource (sutils/munge-hash-for-storage "02") :parameters nil}
+          {:resource (sutils/munge-hash-for-storage "03") :parameters nil})
```
mullr: why? standard style is
```

(fooo
 bar
 baz)
```

test/puppetlabs/puppetdb/query/population_test.clj
```diff
@@ -30,86 +32,76 @@
...
-         {:certname_id 2 :resource (sutils/munge-hash-for-storage "01") :type "Foo" :title "Baz" :exported true :tags (to-jdbc-varchar-array [])}
-         {:certname_id 1 :resource (sutils/munge-hash-for-storage "02") :type "Foo" :title "Boo" :exported true :tags (to-jdbc-varchar-array [])}
-         {:certname_id 1 :resource (sutils/munge-hash-for-storage "03") :type "Foo" :title "Goo" :exported true :tags (to-jdbc-varchar-array [])})
+          :catalog_resources
+          {:certname_id 1 :resource (sutils/munge-hash-for-storage "01") :type "Foo" :title "Bar" :exported true :tags (to-jdbc-varchar-array [])}
+          ;; c2's resource shouldn'sutils/munge-hash-for-storage t be counted, as they don't correspond to an active node
+          {:certname_id 2 :resource (sutils/munge-hash-for-storage "01") :type "Foo" :title "Baz" :exported true :tags (to-jdbc-varchar-array [])}
+          {:certname_id 1 :resource (sutils/munge-hash-for-storage "02") :type "Foo" :title "Boo" :exported true :tags (to-jdbc-varchar-array [])}
+          {:certname_id 1 :resource (sutils/munge-hash-for-storage "03") :type "Foo" :title "Goo" :exported true :tags (to-jdbc-varchar-array [])})
```
mullr: ditto

mullr:
Generally lgtm. I do think we should consider server-side throttling of this. Noted a few whitespace issues; there's more of the same further down.

src/puppetlabs/puppetdb/query/population.clj
wkalt: I memoized it with a 30 second ttl -- might be worth some thought as to whether that's sufficiently high

mullr: We were seeing 5-10s queries on the big db, right? If that's correct, I'd jack it up to about a minute.

mullr: (which is also a really easy number to communicate)

src/puppetlabs/puppetdb/query/population.clj
```diff
@@ -6,66 +6,63 @@
...
+                      (first-query-result :c))]
+    (if (zero? nresources)
+      0
+      (/ (- nresources ndistinct) nresources))))
+
+(def pct-resource-duplication
+  (mem/ttl pct-resource-duplication* :ttl/threshold 30000))
```
mullr: `clojure.core.memoize`, how do I love thee? Let me recursively enumerate the ways.

mullr:
tested locally, appears to work as advertised

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -223,6 +223,12 @@
   [column]
   (format "encode(%s::bytea, 'hex')" column))
 
+(defn vacuum-analyze
+  "Must call within a db connection"
```
rbrw: Is that directed at the caller?

wkalt: it was, but I guess it's not true

PR #1827, opened 2016-01-26
(PDB-2214) anonymize words independently

wkalt:
which meant that the tool produced unmanagably large tarballs on big installs.


wkalt:
I still need to test and compare this on a production dataset

PR #1829, opened 2016-01-26
(PDB-2214) anonymize words as ?????

wkalt:
which meant that the tool produced unmanagebly large tarballs on big installs.

With this patch a text blob is replaced by a string of question marks of the
same length.

### DIFF
--- a/src/puppetlabs/puppetdb/anonymizer.clj
+++ b/src/puppetlabs/puppetdb/anonymizer.clj
@@ -109,6 +109,11 @@
       (rule-match? x context) (get x "anonymize")
       :else (recur (rest xs)))))
 
+(defn anonymize-text
+  "This is for anonymizing text data where we care only about size, not content."
+  [text]
+  (str/join "" (repeat (count text) "?")))
+
 ;; Functions for anonymizing the final leaf data
 (defn anonymize-leaf-value
   "Based on the input value, return an appropriate random replacement"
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -109,6 +109,11 @@
       (rule-match? x context) (get x "anonymize")
       :else (recur (rest xs)))))
 
+(defn anonymize-text
+  "This is for anonymizing text data where we care only about size, not content."
+  [text]
+  (str/join "" (repeat (count text) "?")))
+
```
rbrw: Suspect this may be a lot more efficient: https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/StringUtils.html#repeat%28char,%20int%29  Or could probably use java.util.Arrays fill (though requiring (local) mutation).

test/puppetlabs/puppetdb/anonymizer_test.clj
```diff
@@ -162,7 +162,9 @@
 (deftest test-anonymize-leaf-message
   (testing "should return a string of equal length"
     (is (string? (anonymize-leaf-memoize :message "good old string")))
-    (is (= 15 (count (anonymize-leaf-memoize :message "good old string"))))))
+    (is (= 15 (count (anonymize-leaf-memoize :message "good old string"))))
+    (is (string? (anonymize-leaf-memoize :text "good old string")))
+    (is (= 15 (count (anonymize-leaf-memoize :text "good old string"))))))
```
rbrw: Think I'd go ahead and verify that we get exactly what we expect here, i.e. "???????????????".

wkalt:
reopening on stable

PR #1830, opened 2016-01-26
(PDB-2214) anonymize words as ?????

wkalt:
which meant that the tool produced unmanagebly large tarballs on big installs.

With this patch a text blob is replaced by a string of question marks of the
same length.

### DIFF
--- a/src/puppetlabs/puppetdb/anonymizer.clj
+++ b/src/puppetlabs/puppetdb/anonymizer.clj
@@ -5,7 +5,8 @@
             [puppetlabs.puppetdb.random :refer :all]
             [puppetlabs.puppetdb.reports :as reports]
             [puppetlabs.puppetdb.schema :as pls]
-            [schema.core :as s]))
+            [schema.core :as s])
+  (:import [org.apache.commons.lang3 StringUtils]))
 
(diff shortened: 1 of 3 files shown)


src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -109,6 +110,12 @@
       (rule-match? x context) (get x "anonymize")
       :else (recur (rest xs)))))
 
+(defn anonymize-text
+  "This is for anonymizing text data where we care only about size, not content."
+  [text]
+  (when text
+    (StringUtils/repeat "?" (count text))))
```
ajroetker: Why not just use `clojure.string` instead of adding a dependency? `(clojure.string/join (repeat (count text) "?")`

mullr: Could totally imagine this being really slow, not sure though

wkalt: StringUtils is about 100x faster than that approach @ajroetker

ajroetker: Which one @mullr?

PR #1831, opened 2016-01-27
(PDB-2133) Add [developer] configuration section with option to pretty print

wkalt:
This adds a [developer] section in configuration, along with an option to
pretty print API responses by default.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -205,7 +205,6 @@ make external HTTP requests. In this case you can configure a proxy
 server to send requests to the `updates.puppetlabs.com` URL and
 override this setting to point to your proxy server.
 
-
 `[puppetdb]` settings
 -----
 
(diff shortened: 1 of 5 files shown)


ajroetker:
lgtm +1

documentation/configure.markdown
```diff
@@ -881,3 +880,17 @@ default this is `127.0.0.1` only. As this is an insecure channel this
...
 
 If you wish to listen on all interfaces, you can specify `0.0.0.0`, for example, although this is generally not recommended for production. 
+
+`[developer]` settings
+-----
+
+The `[developer]` section contains configuration items that may be useful to
+users developing against the PuppetDB API. These should be used with care in
+production, as they may impede performance.
+
```
rbrw: Wondered if there's any reason we might want to be a little more conservative here to start, i.e. possibly along the "only use if you know what you're doing" lines.   I suppose it depends on what we think we might do with it in the future.

documentation/configure.markdown
```diff
@@ -881,3 +880,17 @@ default this is `127.0.0.1` only. As this is an insecure channel this
...
+### `pretty-print`
+
+Setting this to true will enable default pretty-printing of API responses. This
+entails a performance penalty both in the speed of data transfer and the size
+of the transferred data. As such it is probably unsuitable for production use.
```
rbrw: Perhaps even just "speed and size of the data transfer"?

rbrw:
Commit summary line: ENAMETOOLONG

wkalt:
@rbrw I did a bit of rewording -- see what you think

PR #1832, opened 2016-01-28
(PE-13485) Implement a consistent request logging format

npwalker:
Prior to this commit puppetdb used a different request
logging format than console-services and puppetserver.

After this commit the request logging will be of the suggested
formatting for all trapper-keeper projects.

### DIFF
--- a/resources/ext/config/request-logging.xml
+++ b/resources/ext/config/request-logging.xml
@@ -2,13 +2,7 @@
     <appender name="FILE" class="ch.qos.logback.core.FileAppender">
         <file>/var/log/puppetlabs/puppetdb/puppetdb-access.log</file>
         <encoder>
-            <pattern>combined</pattern>
-            <!-- To have the same "combined" pattern with elapsedTime ('%D')
-                 appended use the following line:
-
             <pattern>%h %l %u [%t] "%r" %s %b "%i{Referer}" "%i{User-Agent}" %D</pattern>
-
-            -->
         </encoder>
     </appender>
     <appender-ref ref="FILE" />


wkalt:
+1 looks like this is the same as before with the elapsed time added

pcarlisle:
should fix the comment just below it

npwalker:
@pcarlisle valid what should the comment say?  

Honestly all of the request-logging.xml files should probably have the same comment too

cprice404:
@npwalker I have a ticket for this, and @Sharpie was suggesting that we should standardize in the other direction...

cprice404:
The ticket is PE-13485, and in there @Sharpie points out that PDB is the only team in compliance with the "Apache Combined log format", and there are tools out there that already recognize that format.  Therefore it seems like it would make sense to take @Sharpie 's suggestions and change console and Puppet Server instead of changing puppetdb.  I have been meaning to send an email to the dev list to discuss this but just haven't gotten there yet.

cprice404:
agh, sorry, i saw this PR before I saw that you'd already added comments to PE-13485.  I'll shut up now and assume you are working towards @charlie's recommendations.

npwalker:
@cprice404 no worries.  The only change was to add duration to the end which still means that it conforms to common log format but also gives us additional performance data.  

It was actually his second recommendation in the description and I submitted a PR to puppet server as well.  :)

cprice404:
@wkalt @senior is there another copy of this file in PE-puppetdb that needs love?  Or are you guys able to avoid that since you're not composing with any other services for PE?

ajroetker:
@cprice404 yeah we have one https://github.com/puppetlabs/pe-puppetdb-extensions/blob/master/resources/logback.xml

ajroetker:
@cprice404 actually it looks like we don't have another request logging config so this must indeed be the one that makes its way into PE

senior:
Is this ready to go? Or is it still waiting on something?

kbarber:
Taking a last minute look now, sorry for the delay, we've had some backlog.

PR #1833, opened 2016-01-29
(PDB-2126) Add catalog_uuid etc. to reports and catalogs

ajroetker:


ajroetker:
The associated PE PR is https://github.com/puppetlabs/pe-puppetdb-extensions/pull/161

ajroetker:

puppet/lib/puppet/indirector/catalog/puppetdb.rb
```diff
@@ -119,11 +119,22 @@ def add_transaction_uuid(hash, transaction_uuid)
   # Include code_id in hash, returning the complete hash.
   #
   # @param hash [Hash] original data hash
-  # @param code_id [String] code_id
-  # @return [Hash] returns original hash augmented with transaction_uuid
+  # @return [Hash] returns original hash with a gaurunteed code_id key
+  # @api private
+  def add_code_id_if_missing(hash)
+    hash['code_id'] = hash['code_id']
```
senior: What's this code doing?

puppet/lib/puppet/reports/puppetdb.rb
```diff
@@ -40,21 +40,28 @@ def report_to_hash
       resources = build_resources_list
       is_noop = resources.any? { |rs| has_noop_event?(rs) } and resources.none? { |rs| has_failed_event?(rs) }
 
+
+      defaulted_catalog_uuid = defined?(catalog_uuid) ? catalog_uuid : transaction_uuid
```
senior: I think this and the same code in catalogs should be refactored to happen in the same place, that way our defaults can't drift apart and we're defaulting in the same way.

puppet/lib/puppet/indirector/catalog/puppetdb.rb
ajroetker: ensuring that the `original hash` has a `code_id` key, even if the value is nil.

puppet/lib/puppet/reports/puppetdb.rb
ajroetker: Hm I think that would require us to refactor report handling in the terminus quite a bit, right now on catalogs we call `to_data_hash` and handle the catalog like we would any ordinary hash. For reports we use the instance variables on the Puppet class to access the fields. If we wanted to do that, we should probably do that in another ticket.

puppet/lib/puppet/indirector/catalog/puppetdb.rb
mullr: I think we talked about this, but seeing as how we're not usually ruby slingers, this at the very least needs a clarifying comment. A more verbose but more obvious version of the same thing is ok too.

senior:
@ajroetker needs a rebase

kbarber:
@ajroetker I think we missed the docs update here for wire formats etc.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -97,8 +97,8 @@
 
 (def supported-command-versions
   {"replace facts" (version-range 2 4)
-   "replace catalog" (version-range 4 7)
-   "store report" (version-range 3 6)
+   "replace catalog" (version-range 4 8)
```
kbarber: Hmm, have we lost any form of deprecation warnings for old commands now? I can see warn-deprecated, but its not used. Anyway, I think we don't drop deprecation messages anymore for older stuff ...

ajroetker: I think that might have maybe been related to the comidi move? I didn't touch anything related to the deprecations.

kbarber: I think it was probably removed when we retired all the old commands, when we put them back again we didn't put back the deprecated handling: https://github.com/puppetlabs/puppetdb/blob/2.3.x/src/com/puppetlabs/puppetdb/command.clj#L323

ajroetker:
@kbarber added documentation for the new wire-formats

src/puppetlabs/puppetdb/command.clj
kbarber: Either way, we should be stating things are deprecated if we can, either fixed in this patch or another, otherwise we'll find people might miss the message.

ajroetker: Made a ticket for this https://tickets.puppetlabs.com/browse/PDB-2433

kbarber:
@ajroetker sorry mate, this is still missing the documentation updates for the query side of things, those new fields are added to catalogs & reports, but not documented atm it would seem.

ajroetker:
@kbarber added the api documentation

documentation/api/query/v4/reports.markdown
```diff
@@ -70,6 +70,15 @@ The following fields are allowed as filter criteria and are returned in all resp
 
 * `transaction_uuid` (string): a string used to identify a Puppet run.
 
+* `catalog_uuid` (string): a string used to tie a catalog to a report to the
+  catalog used from that Puppet run.
+
+* `code_id` (string): a string used to tie a catalog to the Puppet code which generated the catalog.
+
+* `cached_catalog_status` (string): a string used to identify whether the Puppet
```
kbarber: So the possible values: explicitly_requested, on_failure, not_used or null ... I know its up to puppet, but perhaps its worth mentioning what the possible values might be.

kbarber: I can't find an authoritative document to point this at. It would seem to me that the report processor format document might contain this.

ajroetker: Updated below :)

kbarber:
Last error was due to the PR changing underneath etc. not a real problem.

kbarber:
I'm going to live dangerously and merge this, since the last update was only a docs update.

PR #1836, opened 2016-02-02
(PDB-2353) support upgrades from 2.3.2

wkalt:
This reverts some of the combined migration to allow upgrades from 2.3.2 and
later.

### DIFF
--- a/test/puppetlabs/puppetdb/scf/migrate_test.clj
+++ b/test/puppetlabs/puppetdb/scf/migrate_test.clj
@@ -52,7 +52,7 @@
     (testing "should return missing migrations if the *db* is partially migrated"
       (jdbc/with-db-connection *db*
         (clear-db-for-testing!)
-        (let [applied '(34 35 37)]
+        (let [applied '(28 29 31)]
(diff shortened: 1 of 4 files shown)


senior:
It looks like there were no database migrations from 2.3.2 through 2.3.7, is that why we're supporting back to 2.3.2?

wkalt:
@senior the earliest PDB in PE 3.8.z is 2.3.2, so 2.3.2 is the actual requirement because we need to support upgrades from all z's according to the email. You're right that it doesn't make a practical difference though.

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -358,27 +452,6 @@
                   (map #(dissoc % :same))
                   set))))))
 
-(deftest test-migrate-from-unsupported-version
```
ajroetker: Should we still have this test for old 1.x migrations?

ajroetker:
Works for me (database upgrades fine) just the one comment about the test and then I think we're good to merge

PR #1837, opened 2016-02-02
(PDB-2380) Limit the storage of historical-catalogs

ajroetker:
This commit limits the storage of historical catalogs to only the latest
three catalogs for a node. This commit limits the storage because for
our benchmarking data with 5000 nodes (~3 million catalogs) we were
seeing the catalogs table take up around 140GB of space.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -225,13 +225,14 @@ If not supplied, PuppetDB uses standard HTTPS without any additional
 authorization. All HTTPS clients must still supply valid, verifiable
 SSL client certificates.
 
-### `include-historical-catalogs` (PE only)
+### `historical-catalogs-limit` (PE only)
 
 > **Warning:** This setting is intended for use only in Puppet Enterprise (PE).
 > This setting will be ignored without a PE PuppetDB package.
 
-Setting this to `false` disables the storage of historical catalogs. Defaults to
-`true`.
+This settings controls how many historical catalogs are stored in PuppetDB per
+node. Setting this to `0` disables the storage of historical catalogs. Defaults
+to `3`.
 
 ### `disable-update-checking`
 
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -350,9 +350,9 @@
     certname]
    (comp first sql/result-set-seq)))
 
-;; `store-catalogs-historically?` is used for toggling historical catalog
+;; `historical-catalogs-limit` is used for toggling historical catalog
 ;; storage, this is configurable and PE only.
-(def store-catalogs-historically? (atom false))
+(def historical-catalogs-limit (atom 0))
 
 ;; `store-catalogs-jsonb-columns?` is used for toggling storage of the resources
 ;; and edges jsonb blobs for catalogs. These blobs are used in PE only and this
@@ -747,7 +747,8 @@
   "Persist the supplied catalog in the database, returning its
    similarity hash."
   [{:keys [producer_timestamp resources certname] :as catalog} :- catalog-schema
-   received-timestamp :- pls/Timestamp]
+   received-timestamp :- pls/Timestamp
+   historical-catalogs-limit]
   (time! (:replace-catalog performance-metrics)
          (jdbc/with-db-transaction []
            (let [hash (time! (:catalog-hash performance-metrics)
@@ -765,7 +766,13 @@
                           (if-not latest-producer-timestamp
                             (jdbc/insert! :latest_catalogs {:certname_id certname-id :catalog_id catalog-id})
                             (jdbc/update! :latest_catalogs {:catalog_id catalog-id} ["certname_id=?" certname-id]))
-                          (update-catalog-associations! certname-id catalog refs-to-hashes)))))
+                          (update-catalog-associations! certname-id catalog refs-to-hashes)))
+                      (jdbc/delete! :catalogs
+                                    [(str "id NOT IN (SELECT id FROM catalogs "
+                                          "           WHERE certname=?"
+                                          "           ORDER BY producer_timestamp DESC LIMIT ?)")
+                                     certname
+                                     historical-catalogs-limit])))
              hash))))
 
 (pls/defn-validated replace-catalog!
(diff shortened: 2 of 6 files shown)


senior:
Marking as WIP until we get feedback from the advisory group

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -763,9 +763,15 @@
                         (let [refs-to-hashes (time! (:resource-hashes performance-metrics)
                                                     (kitchensink/mapvals shash/resource-identity-hash resources))]
                           (if-not latest-producer-timestamp
-                            (jdbc/insert! :latest_catalogs {:certname_id certname-id :catalog_id catalog-id})
+                            (do (jdbc/insert! :latest_catalogs {:certname_id certname-id :catalog_id catalog-id})
+                                )
```
wkalt: I don't think this do is needed

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -763,9 +763,15 @@
...
                             (jdbc/update! :latest_catalogs {:catalog_id catalog-id} ["certname_id=?" certname-id]))
-                          (update-catalog-associations! certname-id catalog refs-to-hashes)))))
+                          (update-catalog-associations! certname-id catalog refs-to-hashes)))
+                      (jdbc/delete! :catalogs
+                                    [(str "id NOT IN (SELECT id FROM catalogs "
+                                          "           WHERE certname=?"
+                                          "           ORDER BY producer_timestamp DESC LIMIT 3)")
```
wkalt: seems like this should only run when historical catalogs are being stored?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1261,22 +1267,6 @@
...
-(defn delete-old-unassociated-catalogs!
-  [time]
-  {:pre [(kitchensink/datetime? time)]}
-  (when @store-catalogs-historically?
-    (let [time (to-timestamp time)]
-      (jdbc/do-prepared
-       (str "DELETE FROM catalogs oc "
-            "       WHERE oc.producer_timestamp < ? "
-            "       AND NOT EXISTS (SELECT 1 FROM catalogs c "
-            "                       JOIN reports r ON c.catalog_uuid = r.catalog_uuid "
```
wkalt: it seems like by eliminating consideration of reports we must be changing behavior when report-ttl is < 3 days. I don't understand why this function isn't needed anymore.

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: also we should parameterize the 3 instead of hardcoding it.

ajroetker: This is only run when historical-catalogs is on already

src/puppetlabs/puppetdb/scf/storage.clj
ajroetker: Do we want to delete catalogs that have no reports? What's the problem with always keeping 3 catalogs?

ajroetker: With respect to how this is surfaced in the GUI, always having 3 catalogs means that the resource-graph view will always have historical data

src/puppetlabs/puppetdb/scf/storage.clj
ajroetker: Parametrize it from where?

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: I'm just saying it's a behavior change that's not implied by the ticket -- I don't know that there is a problem with it, it's just surprising.

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: ah, see that now

src/puppetlabs/puppetdb/scf/storage.clj
ajroetker: I just figured we don't really need the burden of the GC task anymore, but I can add it back if you guys think it's necessary.

ajroetker:
Updated the config option @wkalt

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -511,17 +511,11 @@
...
-  (testing "garbage collection leaves the newest catalogs"
-    (jdbc/with-transacted-connection *db*
-      (delete-old-unassociated-catalogs! (now)))
-    (is (= [{:count 1}]
-           (query-to-vec ["SELECT COUNT(*) FROM catalogs"])))))
+  (testing "only stores up to three catalogs a certname"
+    (store-catalog! (assoc catalog :producer_timestamp (-> 3 days ago)) (now))
+    (store-catalog! (assoc catalog :producer_timestamp (-> 4 days ago)) (now))
+    (store-catalog! (assoc catalog :producer_timestamp (-> 5 days ago)) (now))
+    (is (= [{:count 3}] (query-to-vec ["SELECT COUNT(*) FROM catalogs"])))))
```
mullr: we should have a test case for what happens when we add a fourth. ... unless I've misread this one and that's what it's actually doing.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -765,7 +766,13 @@
                           (if-not latest-producer-timestamp
                             (jdbc/insert! :latest_catalogs {:certname_id certname-id :catalog_id catalog-id})
                             (jdbc/update! :latest_catalogs {:catalog_id catalog-id} ["certname_id=?" certname-id]))
-                          (update-catalog-associations! certname-id catalog refs-to-hashes)))))
+                          (update-catalog-associations! certname-id catalog refs-to-hashes)))
+                      (jdbc/delete! :catalogs
+                                    [(format (str "id NOT IN (SELECT id FROM catalogs "
+                                                  "           WHERE certname=?"
+                                                  "           ORDER BY producer_timestamp DESC LIMIT %s)")
```
mullr: sql string manipulation bad! query parameters good! :monkey:

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -801,15 +808,16 @@
 
 (pls/defn-validated store-catalog!
   "Persist the supplied catalog in the database, returning its similarity hash.
-   On PE we alter the atom 'store-catalogs-historically' to be true and we will
+   On PE we alter the atom 'historical-catalog-limi' to be >0 and we will
```
mullr: s/limi/limit/

src/puppetlabs/puppetdb/scf/storage.clj
mullr: I think it's best to be either timestamp-based or count-based, not both; thus, I like this change.

mullr:
generally +1, but fix yer sql injection!

test/puppetlabs/puppetdb/scf/storage_test.clj
ajroetker: Yep this is actually adding a 4th, 5th, and 6th.

ajroetker:
@mullr addressed ur comments :)

kbarber:
@ajroetker you've missed the documentation item for this setting.

kbarber:
This requires this PR to be merged also: https://github.com/puppetlabs/pe-puppetdb-extensions/pull/162

ajroetker:
@kbarber I was under the impression we were leaving this undocumented

kbarber:
@ajroetker umm, fine, the old setting is still documented however.

documentation/configure.markdown
```diff
@@ -225,13 +225,14 @@ If not supplied, PuppetDB uses standard HTTPS without any additional
 authorization. All HTTPS clients must still supply valid, verifiable
 SSL client certificates.
 
-### `include-historical-catalogs` (PE only)
+### `historical-catalogs-limit` (PE only)
```
ajroetker: @kbarber updated the docs to let users know how to turn this off

PR #1839, opened 2016-02-04
(PDB-2319) paging support in PQL

wkalt:
This enables support for limit, offset, and order by clauses in PQL.

### DIFF
--- a/documentation/api/query/v4/pql.markdown
+++ b/documentation/api/query/v4/pql.markdown
@@ -9,7 +9,7 @@ canonical: "/puppetdb/latest/api/query/v4/pql.html"
 [ast]: ./ast.html
 
 > **Experimental Feature**: This featureset is experimental, and it may be altered or removed in
-> future release.
+> a future release.
 
 Puppet Query Language (PQL) is a query language designed with PuppetDB and Puppet data in mind. It
 provides a string based query language as an alternative to the [AST query language][ast] PuppetDB has
@@ -39,8 +39,9 @@ A PQL query has the following structure:
 
 Which is broken up into the following parts:
 
-* `entity`: Rquired. The entity context that this query executes on.
-* `projection`: Optional. Modifies the output of how the `entity` fields are returned.
+* `entity`: Required. The entity context that this query executes on.
+* `projection`: Optional. Restricts the output to a selection of fields or
+  function results.
 * `filter`: Optional. The filter to match on for this `entity`.
 * `modifiers`: Optional. Contains modifiers for the query.
 
@@ -69,14 +70,14 @@ The entity context can also be used within a subquery, see the [subquery] sectio
 
 ## Projection
 
-The projection part of a query, provides a mechanism to choose a subset of fields that are returned
+The projection part of a query provides a mechanism to choose a subset of fields that are returned
 or to modify the way those fields are displayed with the usage of functions.
 
-The projection, lives within the brackets of a query:
+The projection lives within the brackets of a query:
 
     nodes[<projection>] {}
 
-And is a comma seperated list of fields and functions:
+And is a comma separated list of fields and functions:
 
     facts[name, count()] { group by name }
 
@@ -364,14 +365,9 @@ query combined everything before, but with a `resource` subquery for `Package[to
       certname ~ '^web'
     }
 
-## Modifiers
+## Group By
 
-Modifiers to the query provide extra facilities beyond a filter for limiting and changing the resultsets
-returned in special ways.
-
-### `group by`
-
-`group by` will condense into a single row all selected rows that share the same values for the grouped expressions.
+A `group by` clause will condense into a single row all selected rows that share the same values for the grouped expressions.
 
 For example, to only show a list of fact names, you can group by the `name` field:
 
@@ -381,3 +377,35 @@ Combined with aggregate functions, you can effectively rollup of results and agg
 up values, for example to return how many facts exist for each fact name across all facts:
 
     facts[name, count(value)] { certname ~ 'web.*' group by name }
+
+## Paging
+PQL supports restriction of the result set via the SQL-like paging clauses
+`limit`, `offset`, and `order by`.
+
+### `limit` and `offset`
+
+Limit and offset clauses are supplied with integer arguments and may appear in
+any order within the braced section of a PQL query. Offset always takes
+precedence over limit:
+
+    reports {certname = 'foo.com' limit 10}
+
+    reports {certname = 'foo.com' limit 10 offset 10}
+
+    reports {certname = 'foo.com' offset 10 limit 10}
+
+Note that since there is no default ordering for results returned by PuppetDB,
+`limit` and `offset` are generally only useful in combination with `order by`.
+
+### `order by`
+
+An `order by` clause will order the result set on a selection of columns in
+ascending or descending order. The argument to an `order by` is a
+comma-separated list of fields, each optionally appended with a keyword `asc`
+or `desc`. If no keyword is supplied, `asc` is assumed:
+
+    reports {certname = 'foo.com' order by receive_time}
+
+    reports {certname ~ 'web.*' order by receive_time, certname desc}
+
+    reports {certname ~ 'web.*' order by receive_time desc, certname desc limit 10}
--- a/resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
+++ b/resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
@@ -9,8 +9,6 @@
 (* Note: the following features are awaiting PDB AST changes first, but
    generally have support within PDB in one form or another:
 
-   - order by
-   - limit, offset
    - include_total
 
    Also, the following items need fundamental changes to PDB first:
@@ -30,7 +28,7 @@ extract = <lbracket>, [<whitespace>], [extractfields], [<whitespace>], <rbracket
 <extractfields> = fieldlist;
 
 (* Filtering *)
-<where> = <lbrace>, [<whitespace>], [expression], [<whitespace>], [modifier], [<whitespace>], <rbrace>;
+<where> = <lbrace>, [<whitespace>], [expression], [<whitespace>], [groupbyclause | {pagingclause, [<whitespace>]}], {<whitespace>}, <rbrace>;
 
 (* Static list of entity types *)
 <entity> = 'facts' |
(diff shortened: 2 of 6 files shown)


documentation/api/query/v4/pql.markdown
```diff
@@ -39,8 +39,8 @@ A PQL query has the following structure:
 
 Which is broken up into the following parts:
 
-* `entity`: Rquired. The entity context that this query executes on.
-* `projection`: Optional. Modifies the output of how the `entity` fields are returned.
+* `entity`: Required. The entity context that this query executes on.
+* `projection`: Optional. Restricts the output to a selection of fields.
```
kbarber: The new wording doesn't quite describe that it covers functions also, and later with DAP (document aware projection) we'll need to extend it again. I know you're better at describing things than I am, perhaps a different wording here? Maybe just saying 'a selection of fields or functions' or something like that?

Tangent/Unrelated: Also fwiw, lots of people hate the word 'projection' even if it is correct. I've had this feedback directly from a large number of people. @nfagerlund, not sure if you have any ideas around this also.

documentation/api/query/v4/pql.markdown
```diff
@@ -69,14 +69,14 @@ The entity context can also be used within a subquery, see the [subquery] sectio
...
+The projection part of a query provides a mechanism to choose a subset of fields that are returned
 or to modify the way those fields are displayed with the usage of functions.
 
-The projection, lives within the brackets of a query:
+The projection lives within the brackets of a query:
 
     nodes[<projection>] {}
 
-And is a comma seperated list of fields and functions:
+And is a comma separated list of fields and functions:
```
kbarber: I always do this.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
```diff
@@ -30,7 +28,7 @@ extract = <lbracket>, [<whitespace>], [extractfields], [<whitespace>], <rbracket
 <extractfields> = fieldlist;
 
 (* Filtering *)
-<where> = <lbrace>, [<whitespace>], [expression], [<whitespace>], [modifier], [<whitespace>], <rbrace>;
+<where> = <lbrace>, [<whitespace>], [expression], [<whitespace>], {modifier | pagingclause, [<whitespace>]}, [<whitespace>], <rbrace>;
```
kbarber: It's an organisational nitpick but I lumped pagination with grouping all as 'modifiers' logically in my head, but with the group by potential change I guess this goes away. If we think we're wanting to separate this into two places, perhaps modifier should just become group by directly and we collapse one layer.

documentation/api/query/v4/pql.markdown
```diff
@@ -381,3 +381,32 @@ Combined with aggregate functions, you can effectively rollup of results and agg
 up values, for example to return how many facts exist for each fact name across all facts:
 
     facts[name, count(value)] { certname ~ 'web.*' group by name }
+
+## Paging
+PQL supports restriction of the result set via the standard SQL paging clauses
```
kbarber: Might be worth re-iterating somewhere that an order and offset, doesn't make much sense without an order by. We state this elsewhere, I think in the AST doc, but since we've split this up now might be worth mentioning again.

test/puppetlabs/puppetdb/pql_test.clj
```diff
@@ -216,4 +216,49 @@
...
+
+    "reports[certname]{certname = 'foo' limit 10 order by certname, receive_time desc}"
+    ["from" "reports"
+     ["extract" ["certname"] ["=" "certname" "foo"]]
+     ["limit" 10] ["order_by" ["certname" ["receive_time" "desc"]]]]
+
+    "reports[certname]{certname = 'foo' order by certname desc, receive_time asc limit 10}"
+    ["from" "reports"
+     ["extract" ["certname"] ["=" "certname" "foo"]]
+     ["order_by" [["certname" "desc"] ["receive_time" "asc"]]] ["limit" 10]]))
```
kbarber: We should probably add a test that mixes pagination with group by, because they are logically OR'd in the EBNF.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
```diff
@@ -125,6 +123,15 @@ sqstring = (<singlequote>, stringwithoutsinglequotes, <singlequote>);
 <modifier> = groupby;
 groupby  = <'group by '>, fieldlist;
 
+(* Paging *)
```
kbarber: I can't see any parser tests for this new block. Probably worth adding since parts of this are squirrely.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
kbarber: Oh here is something else, what happens when someone adds two modifiers/pagination parameters? Can we get a test and a guard against this possibility? Perhaps ... the trick is to force an order to stop this, but sounds like we should at least throw an error somewhere.

kbarber:
It's looking good mate :-). Nice to see someone else hacking on the parser :-).

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
wkalt: I went back and forth on this, but settled on splitting them out because they end up in different spots in the final query. Group by is always inside an extract, but the paging clauses go outside the extract and inside the surrounding from. I'd be fine with making group by its own thing or whatever we settle on.

I'm not sure what you mean by two modifiers/pagination parameters. Two paging parameters will work, but I think that this query

```
reports[status, count()]{group by status limit 1]
```

is unfortunately ambiguous under the language. Erroring for now would probably be best in my opinion, happy to put that in.

documentation/api/query/v4/pql.markdown
wkalt: +1

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
wkalt: ah thanks, missed that ns

test/puppetlabs/puppetdb/pql_test.clj
wkalt: let's talk about this one in standup -- I think we've got to figure out a better approach for mixed group by / paging. It could be the proposed syntax changes get us out of it though.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
wkalt: Maybe we can actually resolve by requiring paging clauses to come before group by clauses when specified together, but I'm not sure the AST support is there for the combo yet.

documentation/api/query/v4/pql.markdown
nfagerlund: Not off the top of my head, I don't. 

So as I understand it, uh...

* A "filter" is a filter that controls which subset of entities the query will return. 
* A "projection" is one of the following:
    * A filter that modifies each entity in that subset.
    * Instructions on how to aggregate data in the subset of entities into a single value.

...? Right? This terminology is baffling to me as well. Maybe if you could explain the metaphor step by step? What's being projected, onto what surface?

documentation/api/query/v4/pql.markdown
mullr: Might want to refer to these as 'sql-like', since this is in no way an implementation of the sql standard.

resources/puppetlabs/puppetdb/pql/pql-grammar-experimental.ebnf
```diff
@@ -121,10 +119,19 @@ sqstring = (<singlequote>, stringwithoutsinglequotes, <singlequote>);
...
-<modifier> = groupby;
+(* Group By *)
+<groupbyclause> = groupby;
 groupby  = <'group by '>, fieldlist;
 
+(* Paging *)
+<pagingclause> = limit | offset | orderby;
+limit = <'limit '>, integer;
+offset = <'offset '>, integer;
+orderby = <'order by '>, orderparam, [orderparamlist];
```
mullr: these three will only match exactly the space character; it would be better to use the `<whitespace>` non-terminal below.

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -2,15 +2,21 @@
   (:require [clojure.string :as str]
             [puppetlabs.puppetdb.cheshire :as json]))
 
+(defn paging-clause?
```
mullr: Is there some way to get this from instaparse, as these are all part of the <pagingclause> rule? Wouldn't be surprised if we could put this as metadata on the AST.

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -2,15 +2,21 @@
...
+  [v]
+  (contains? #{"limit" "offset" "order_by"} (first v)))
+
+(defn slurp-expr->extract
+  [clauses]
+  (let [paging-clauses (filter paging-clause? clauses)
```
mullr: `clojure.core/group-by`

test/puppetlabs/puppetdb/pql/parser_test.clj
```diff
@@ -742,6 +742,26 @@
...
+    (are [in expected] (= (parse in :start :pagingclause) expected)
+         "order by name" [[:orderby [:orderparam "name"]]]
+         "order by name, value" [[:orderby [:orderparam "name"] [:orderparam "value"]]]
+         "order by name asc, value desc" [[:orderby
+                                           [:orderparam "name" "asc"]
+                                           [:orderparam "value" "desc"]]]
+         "order by name desc, value" [[:orderby
+                                       [:orderparam "name" "desc"]
+                                       [:orderparam "value"]]])))
+
```
mullr: would like to see tests for combining limit and offset, in both orders.

mullr: oh, you have that below. n/m

mullr:
brief style notes plus some whitespace notes in the grammar.

src/puppetlabs/puppetdb/pql/transform.clj
wkalt: I don't think this would be done in instaparse -- we'd have to tag the return values of the transform functions in this ns with metadata. IMO doing this just for paging clauses would be a bit weird, and doing it for all the transform return values would be unnecessary.

wkalt:
@mullr addressed your comments, have a look

src/puppetlabs/puppetdb/pql/transform.clj
mullr: yeah, instaparse doesn't appear to support that. What I'm reacting to here is that we're building some logic to rebuild something that's already present in the structure of the grammar. Another way to handle this would be to change `<paging-clause>` to `paging-clause`, then have our transformer look for `[:paging-clause ...]` structures. But that may may be sufficiently different from the existing code that we don't want to go there.

PR #1841, opened 2016-02-04
(PDB-2289) Teach add-criteria about ["from" ...]

mullr:
This is needed so the sync code can use the newer query syntax.


ajroetker:
:+1: tested with the PE PR and works fine

PR #1843, opened 2016-02-05
(maint) maybe fix upgrade tests

wkalt:
This will cause the terminus to be installed with package 'puppetdb-terminus'
when upgrading from 2.3.x. The puppetdb-terminus is a dummy package that points
to the latest puppetdb-termini, but it's needed to obsolete the legacy terminus
package.

### DIFF
--- a/acceptance/setup/pre_suite/70_install_released_puppetdb.rb
+++ b/acceptance/setup/pre_suite/70_install_released_puppetdb.rb
@@ -2,13 +2,19 @@
 OLDEST_SUPPORTED_UPGRADE="2.3.8"
 
 if ([:upgrade_oldest, :upgrade_latest].include? test_config[:install_mode] and not test_config[:skip_presuite_provisioning])
-  install_target = test_config[:install_mode] == :upgrade_latest ? 'latest' : OLDEST_SUPPORTED_UPGRADE
+  if test_config[:install_mode] == :upgrade_latest
+    install_target = 'latest'
+    terminus_package = 'puppetdb-termini'
+  else
+    install_target = OLDEST_SUPPORTED_UPGRADE
+    terminus_package = 'puppetdb-terminus'
+  end
   step "Install most recent released PuppetDB on the PuppetDB server for upgrade test" do
     databases.each do |database|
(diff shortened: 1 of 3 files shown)


acceptance/setup/pre_suite/70_install_released_puppetdb.rb
```diff
@@ -2,13 +2,19 @@
 OLDEST_SUPPORTED_UPGRADE="2.3.8"
 
 if ([:upgrade_oldest, :upgrade_latest].include? test_config[:install_mode] and not test_config[:skip_presuite_provisioning])
-  install_target = test_config[:install_mode] == :upgrade_latest ? 'latest' : OLDEST_SUPPORTED_UPGRADE
+  if test_config[:install_mode] == :upgrade_latest
+    install_target = 'latest'
+    terminus_package = 'puppetdb-termini'
+  else
```
mullr: is this the `:upgrade_oldest` branch? if so, I'd like this better as an explicit test followed by a bailout on an unexpected value.

acceptance/setup/pre_suite/90_install_devel_puppetdb.rb
```diff
@@ -41,6 +40,11 @@
   when :git
     install_puppetdb_termini_via_rake(master, databases)
   when :package
-    install_puppetdb_termini(master, databases)
+    os = PuppetDBExtensions.config[:os_families][master.name]
+    if test_config[:install_mode] == :upgrade_oldest && ([:redhat, :fedora].include? os)
+      install_puppetdb_termini(master, databases, nil, 'puppetdb-terminus')
+    else
```
mullr: ditto here

wkalt: we don't want to bail under any circumstances here, since the possible install_modes are validated in helper.rb, but 'm going to hold off on any changes to this since I think https://github.com/puppetlabs/puppetlabs-puppetdb/pull/220 is a better way to fix our issue once I fix some ruby 1.8.7 issue I just noticed.

wkalt:
closing -- this solution doesn't work

PR #1844, opened 2016-02-08
(maint) Update library dependencies

kbarber:

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -9,10 +9,10 @@
    :password :env/nexus_jenkins_password
    :sign-releases false})
 
-(def tk-version "1.1.1")
-(def tk-jetty9-version "1.3.1")
-(def ks-version "1.1.0")
-(def tk-status-version "0.2.1")
+(def tk-version "1.2.0")
+(def tk-jetty9-version "1.5.0")
+(def ks-version "1.2.0")
+(def tk-status-version "0.3.1")
 
 (def pdb-jvm-opts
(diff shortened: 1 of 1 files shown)


mullr:
LGTM

project.clj
```diff
@@ -9,10 +9,10 @@
    :password :env/nexus_jenkins_password
    :sign-releases false})
 
-(def tk-version "1.1.1")
-(def tk-jetty9-version "1.3.1")
-(def ks-version "1.1.0")
-(def tk-status-version "0.2.1")
+(def tk-version "1.2.0")
```
ajroetker: This should be `1.3.0` now

project.clj
```diff
@@ -9,10 +9,10 @@
...
+(def tk-jetty9-version "1.5.0")
+(def ks-version "1.2.0")
```
ajroetker: This should also be `1.3.0` now :)

project.clj
mullr: don't think we should upgrade to 1.3 until we're ready for HUP

kbarber: Not only that, there are some dependency complexities when I do this that may or may not take time to resolve, I say lets hold off on it. This patch was about plucking the majority of the low hanging fruit, I think we can dive into the individual troublesome pieces that are left as separate work.

project.clj
kbarber: Tied to the above bump complexity, lets circle back on these bumps.

PR #1846, opened 2016-02-09
(PDB-2231) Add dump-heap to save mem info via jmap

rbrw:


project.clj
```diff
@@ -119,6 +119,9 @@
   ;; code that we have.
   :classifiers  [["test" :testutils]]
 
+  ;; Inovke "lein test :all" to run all tests
+  :test-selectors {:default (fn [m] (not (:slow m)))}
```
kbarber: Can we reflect this now in all the test scripts in ext ? Since I presume we want to do the :all case in CI.

kbarber:
@rbrw wasn't most of the work for max-command-size done under stable, not master?

test/puppetlabs/puppetdb/testutils/mem.clj
```diff
@@ -0,0 +1,27 @@
...
+  (let [pid (atom nil)]
+    (fn []
+      (if-let [p @pid]
+        p
+        (reset! pid (->> (ManagementFactory/getRuntimeMXBean)
+                         .getName
+                         (re-find #"[0-9]+")
+                         Integer/parseInt))))))
+
+(defn dump-heap [filename]
```
kbarber: Your usage of this is commented out later on in your test.

kbarber:
@rbrw some pending comments on this PR ^^ if you missed them :-).

test/puppetlabs/puppetdb/testutils/mem.clj
rbrw: Yes, that's intentional - the commented commands just show how we can use dump-heap (when we're trying to diagnose problems), but we don't need it for the test, which is just checking to see that PDB doesn't die if we try to send something approaching max-command-size through.

project.clj
rbrw: Hopefully this is what you meant: https://github.com/puppetlabs/pe-puppetdb-extensions/pull/166

rbrw:
@kbarber indeed, the initial max-command-size work was done on stable, though the UCS-2 computation that's fixed in this series only exists on master (related to the more recent POST rework).  Though I suppose I could see about porting at least the test to stable if we think that's desirable.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -434,7 +434,7 @@
     ;; amq send thread
     (future
       (try
-        (let [s (.createSession connection true Session/SESSION_TRANSACTED)
+        (let [s (.createSession connection true 0)
```
rbrw: For reference: http://activemq.apache.org/maven/apidocs/org/apache/activemq/ActiveMQConnection.html#createSession%28boolean,%20int%29

project.clj
kbarber: No, I meant, our scripts in ext/ need to now use 'lein test :all' for FOSS. If we merge this in, these tests won't run in CI any more, I'm guessing we don't want that.

https://github.com/puppetlabs/puppetdb/blob/master/ext/jenkins/lein-test.sh

src/puppetlabs/puppetdb/cli/benchmark.clj
kbarber: I'm not quite across why we can't use the const symbols here, and are instead using the numbers. Session/SESSION_TRANSACTED translates to an integer of 0 doesn't it?

project.clj
rbrw: Oh, right - yes, we'll have to decide how often we want :slow to run -- though I suppose after this morning's discussion there might be some question about whether the test is as interesting now, aside from the fact that it makes it easy to perform some simple command memory processing dumps.

test/puppetlabs/puppetdb/testutils/mem.clj
rbrw: Test is gone now.

kbarber:
@rbrw I'm kind of not seeing the point of this patch now we've removed the test ... As a general rule, I'm not a fan of utility code lying around that isn't used ... we'll forget about it/its not tested/it will rot etc. etc. true, it might be useful under some circumstances, but I can find other ways of getting a memory dump during development, like by attaching yourkit or something ...

senior:
@kbarber I suggested that @rbrw reform the patch like this so that we could talk about it in standup. I agree with you on the dead code thing, I'm usually opposed to it. I do think the heap dump code is useful and different from what we do with yourkit though.

PR #1855, opened 2016-02-17
(maint) don't validate package version as semver in status service

wkalt:
This was causing issues in the extensions tests due to our snapshot versioning.


project.clj
```diff
@@ -78,7 +78,8 @@
                  [puppetlabs/trapperkeeper ~tk-version]
                  [puppetlabs/trapperkeeper-webserver-jetty9 ~tk-jetty9-version]
                  [prismatic/schema "1.0.4"]
-                 [puppetlabs/trapperkeeper-status ~tk-status-version :exclusions [trptcolin/versioneer]]
+                 [puppetlabs/trapperkeeper-status ~tk-status-version]
+                 [trptcolin/versioneer "0.2.0"]
```
ajroetker: might want to put this above the status service since it's a dependency of status

ajroetker:
@cprice404 does this seem reasonable or should we be fixing the status service to accept SNAPSHOTs?

src/puppetlabs/puppetdb/pdb_routing.clj
```diff
@@ -93,6 +94,16 @@
                    true
                    @maint-mode-atom))))
 
+(defn get-artifact-version
```
ajroetker: I'd make a note in a docstring about how this needs to be fixed upstream in `clj-semver`

PR #1856, opened 2016-02-18
(PDB-2435) Use read-db for population metrics

ajroetker:
This commit changes our population metrics to use the read-db and not
the write-db so that we leverage read replica's when they are available.


daenney:
Thank you! <3

PR #1858, opened 2016-02-19
(maint) Simplify core.async/AMQ use in benchmark

mullr:
Wrap the temp AMQ instance in a custom core.async Buffer,
allowing us to use a single channel with the custom buffer instead
of two separate channels.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -34,7 +34,7 @@
 
   :dependencies [[org.clojure/clojure "1.8.0"]
                  [cheshire "5.5.0"]
-                 [org.clojure/core.match "0.2.2"]
+                 [org.clojure/core.match "0.3.0-alpha4" :exclusions [org.clojure/tools.analyzer.jvm]]
                  [org.clojure/math.combinatorics "0.1.1"]
                  [org.clojure/math.numeric-tower "0.0.4"]
                  [org.clojure/tools.logging "0.3.1"]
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -531,28 +508,35 @@
...
-            (mapv close! [rate-limited-mq-out mq-in mq-out rate-monitor-ch command-send-ch])))
+            (mapv close! [rate-limited-mq-out mq rate-monitor-ch command-send-ch])))
         close-to-stop-ch))))
 
 (defn chan? [x]
   (satisfies? clojure.core.async.impl.protocols/Channel x))
 
+
 (defn -main [& args]
+  #_(let [c (chan (amq-broker-buffer "test-mq" "test-endpoint"))]
```
senior: I think you missed this?

project.clj
```diff
@@ -34,7 +34,7 @@
 
   :dependencies [[org.clojure/clojure "1.8.0"]
                  [cheshire "5.5.0"]
-                 [org.clojure/core.match "0.2.2"]
+                 [org.clojure/core.match "0.3.0-alpha4" :exclusions [org.clojure/tools.analyzer.jvm]]
```
senior: What was the alpha feature you needed here?

mullr: The `close-buf!` method on Buffer

mullr: ... sorry, I misread this. I needed a new version of core.async, which needs a newer tools.analyzer.jvm. Had to pull in a newer core.match as well to make that dep work.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -531,28 +508,31 @@
...
 (defn -main [& args]
   (let [x (apply benchmark-main args)]
-   (when (chan? x)
-     (println "Press ctrl-c to stop.")
-     (<!! x))))
+    (when (chan? x)
+      (println "Press ctrl-c to stop.")
+      (<!! x))))
+
+
```
wkalt: extra lines

wkalt:
this looks good to me. works fine too.

project.clj
kbarber: We would have to do this for upgrading to latest tk etc. anyway, so I agree with what @mullr has done here. Annoying though :-).

PR #1859, opened 2016-02-19
(PDB-2425) use table stats to compute the number of resources

senior:
This is a stable backport of
1b79d8158698a01cefe791f8f8cc0031e62950e3. This switches the metrics
calls that query for number of nodes and resources per node to use table
stats rather than a full table scan for each call. This isn't an exact
backport as some things are different in master right now (java.jdbc
usage, metrics, fixtures are different etc). This does preserve the
behavior of caching the resources per node calls though. Number of
inactive nodes was not backported as it's a new metric and will need a
non-bugfix release.


wkalt:
+1

PR #1862, opened 2016-02-22
(doc) Update paths in code examples for Puppet 4.

gguillotte:
Several code examples still use /etc/puppet/ paths for SSL certificates
and Puppet code manifests. Update these paths.

Resolves DOC-2560.


gguillotte:
Reissuing against stable per @wkalt.

PR #1864, opened 2016-02-22
(PDB-2441) Don't submit TextMessages (only UTF-8) 

rbrw:


rbrw:
Depends on https://github.com/puppetlabs/puppetdb/pull/1846

kbarber:
So the util code here, I have the same comments about it as per here: https://github.com/puppetlabs/puppetdb/pull/1846

src/puppetlabs/puppetdb/mq.clj
```diff
@@ -205,7 +205,8 @@
 (extend String
   ToJmsMessage
   {:-to-jms-message (fn [x properties session]
-                      (let [msg (.createTextMessage session x)]
+                      (let [msg (.createBytesMessage session)]
+                        (.writeBytes msg (.getBytes x "UTF-8"))
```
senior: This call and the createBytesMessage above it use reflection, probably something we should get cleaned up.

senior:
I think the session transacted changes you pulled in are conflicting with Russ' AMQ benchmark change that was just merged

PR #1866, opened 2016-02-24
(maint) PQL fixes & corrections from Scott Walker

kbarber:
* Convert examples to double quotes, so they can be cut & paste using the curl example
* Show better path and array regexp examples that actually work
* Use a better explicit subquery example for multiple fields
* Allow question marks in a field
* Add more tests for the above bugs/problems

### DIFF
--- a/src/puppetlabs/puppetdb/pql/transform.clj
+++ b/src/puppetlabs/puppetdb/pql/transform.clj
@@ -65,7 +65,7 @@
 
 (defn transform-groupedlist
   [& args]
-  args)
+  (vec args))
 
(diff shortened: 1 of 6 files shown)


src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -65,7 +65,7 @@
 
 (defn transform-groupedlist
   [& args]
-  args)
+  (vec args))
```
senior: Why this change?

senior:
:+1: from me, just one question

PR #1868, opened 2016-02-24
(PDB-2424) Fix delete-edges! to hit unique index

senior:
Currently delete-edges coerces the column and compares it with the data
passed in. This cause ALL source/target column values to be
decoded. This is slow and doesn't hit the unique index that is in
place. Instead this patch converts the source/target being deleted to
bytea and compares it with what is stored. This hits the index and will
be much faster as only 1 conversion is done, not one for each row + column.

### DIFF
--- a/test/puppetlabs/puppetdb/scf/storage_test.clj
+++ b/test/puppetlabs/puppetdb/scf/storage_test.clj
@@ -608,14 +608,21 @@
                        "before"] nil})))
 
           (testing "should only delete the 1 edge"
-            (is (= [[:edges [(format "certname=? and %s=? and %s=? and type=?"
-                                     (sutils/sql-hash-as-str "source")
-                                     (sutils/sql-hash-as-str "target"))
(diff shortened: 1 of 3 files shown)


test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -608,14 +608,21 @@
                        "before"] nil})))
 
           (testing "should only delete the 1 edge"
```
mullr: I wish there was some way we could test "this query does do a table scan" or similar, but I have no reasonable idea of how to implement taht.

ajroetker: we could do a explain analyze and regex the output maybe?

mullr: "reasonable idea" :)

mullr:
tested locally, appears to do the right thing.

PR #1869, opened 2016-02-25
(PDB-2454) reject-large-commands false by default

senior:
Attempting to set a command size that is "too large" is difficult as
it's very dependent on environment. Having this enabled by default was
causing commands that were able to be processed before to be rejected as
they were considered too large by this setting. With the potential
impact of these false positives, the setting should be false by default.

Users wanting to set a max-command-size still can. We also output debug
logging to help users get a feel for their current command sizes, using
that users should better be able to determine their typical command
sizes and find a better default than our current (overly conservative)
estimate.


rbrw:
lgtm

PR #1870, opened 2016-02-25
(PDB-2407) PQL- Allow inequalities on date fields

senior:
This patch allows inequality operators (>,>=,<,<=) to be used on date
fields similar to how they are allowed in the AST language. There is
currently no validation to ensure correct format of the date
strings (similar to AST).

### DIFF
--- a/test/puppetlabs/puppetdb/pql_test.clj
+++ b/test/puppetlabs/puppetdb/pql_test.clj
@@ -261,4 +261,21 @@
     "reports[certname]{certname = 'foo' order by certname desc, receive_time asc limit 10}"
     ["from" "reports"
      ["extract" ["certname"] ["=" "certname" "foo"]]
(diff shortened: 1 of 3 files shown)


test/puppetlabs/puppetdb/pql_test.clj
```diff
@@ -261,4 +261,21 @@
...
+    ["from" "reports"
+     [">=" "receive_time" "2016-02-07T08:45:42.170687300Z"]]
+
+    "reports{receive_time <= '2016-02-07T08:45:42.170687300Z'}"
+    ["from" "reports"
+     ["<=" "receive_time" "2016-02-07T08:45:42.170687300Z"]]
+
+    "reports{receive_time < '2016-02-07T08:45:42.170687300Z'}"
+    ["from" "reports"
+     ["<" "receive_time" "2016-02-07T08:45:42.170687300Z"]]))
```
ajroetker: Might wanna have some other date fields tested here, on resources and such.

PR #1872, opened 2016-02-25
(PDB-2438) Fix pretty-printing parameter for POST

ajroetker:
This commit fixes our `pretty` parameter for query POSTs, before this
commit we weren't properly passing the param from the POST body to the
produce-streaming-body function and weren't allowing the parameter
through validation.

### DIFF
--- a/src/puppetlabs/puppetdb/http/handlers.clj
+++ b/src/puppetlabs/puppetdb/http/handlers.clj
@@ -5,8 +5,7 @@
             [puppetlabs.puppetdb.query.paging :as paging]
             [puppetlabs.puppetdb.query-eng :refer [produce-streaming-body
                                                    stream-query-result]]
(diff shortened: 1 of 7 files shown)


senior:
I'll review/test this

senior:
running 
```
curl "http://localhost:8080/pdb/query/v4/nodes?pretty=true"
```
Doesn't give me formatted JSON and it doesn't work on master. Looks like it's broken and maybe needs a test to validate?

senior:
This works:
```
$  curl -X POST http://localhost:8080/pdb/query/v4/nodes \
  -H 'Content-Type:application/json' \
  -d '{"query":["~","certname",".*"], "pretty": true}''
```
so it might just affect GETs

ajroetker:
added a test!

senior:
@ajroetker looks like legit Travis failures

senior:
:+1: will merge when the pipeline is clear

PR #1873, opened 2016-02-26
(PDB-2311) Substitute spaces with underscores in command names

ajroetker:
This commit changes the names of the PuppetDB wire-format commands to
use underscores instead of spaces which requires a bump to the facts and
deactivate node commands.


ajroetker:
Breaks PE tests.... unless you merge https://github.com/puppetlabs/pe-puppetdb-extensions/pull/169 as well!!!

documentation/api/command/v1/commands.markdown
```diff
@@ -22,25 +24,25 @@ which have the following JSON wire format:
 
 `command` is a string identifying the command.
```
mullr: While I like the new way of line wrapping (M-q?), this makes it hard to see what changed and will break `git blame` down the road.

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -20,10 +20,17 @@
            [org.apache.commons.fileupload.util LimitedInputStream]))
 
 (def min-supported-commands
```
mullr: looks like this could be derived from supported-command-versions above? (yes, I know it's preexisting)

puppet/lib/puppet/util/puppetdb/command.rb
```diff
@@ -42,7 +42,7 @@ def initialize(command, version, certname, payload)
       # Puppet 4.1.0. We need a better answer to non-utf8 data end-to-end.
       }.to_pson, "Error encoding a '#{command}' command for host '#{certname}'")
     end
-    @command = Puppet::Util::Puppetdb::CharEncoding.coerce_to_utf8(command).gsub(" ", "_")
+    @command = Puppet::Util::Puppetdb::CharEncoding.coerce_to_utf8(command)
```
mullr: I don't understand this change... what did the `gsub` used to do? It looks to me like it already tried to submit commands with "_" instead of space.

src/puppetlabs/puppetdb/http/command.clj
```diff
@@ -164,7 +171,7 @@
   is unavoidable (i.e. old-style, non-param POST)."
   [handle]
   (fn [{:keys [params] :as req}]
-    (handle (if (params "command")
+    (handle (if (get params "command")
```
mullr: :+1:

mullr:
ok, generally looks ok. Main gripe is the formatting stuff.

puppet/lib/puppet/util/puppetdb/command.rb
ajroetker: Where are you looking?

mullr: at the red line and then the green line

src/puppetlabs/puppetdb/http/command.clj
ajroetker: That was intentional from what I recall, I had the same comment when the code originally went in

mullr: (ーー゛)

puppet/lib/puppet/util/puppetdb/command.rb
ajroetker: I mean where are you seeing that it already tried to submit commands with the underscore :P

mullr: That's my guess as to what the previous `gsub` was trying to do. Is that wrong? Didn't it try to replace spaces with underscores?

src/puppetlabs/puppetdb/http/command.clj
ajroetker: It had to do with needing them to be decoupled for future versions

mullr: ヽ(^。^)ノ

puppet/lib/puppet/util/puppetdb/command.rb
ajroetker: see hipchat feb29@4pm

wkalt:
My preference would be to keep space separating the command names in docs, not document this as a rename, and just say in the documentation on commands that the command payload contains the underscore-separated command name.

I'm not exactly sure how to handle this from a versioning perspective -- maybe it's more like a bump of the /cmd API?

PR #1874, opened 2016-02-26
(PDB-2360) Fix aliasing in aggregates and group-by

ajroetker:
This commit fixes a bug with aggregates and group-by where the fields
used as arguments to those operators (`function` and `group_by`) weren't
being converted to our `fields` (honeysql) constructs which have the
table disambiguation and correct names for certain query API fields
(e.g. queries to `/nodes` grouping by certname now work instead of
erroring with an ambiguous column message, and queries to `/reports`
with `["function","min","report_timestamp"]` will now properly rename
`report_timestamp` to `reports.end_time` in the SQL).

### DIFF
--- a/src/puppetlabs/puppetdb/http/query.clj
+++ b/src/puppetlabs/puppetdb/http/query.clj
@@ -86,7 +86,9 @@
     ["extract" columns crit]
 
     [["extract" columns subquery]]
-    ["extract" columns ["and" subquery crit]]
+    (if (= "group_by" (first subquery))
+      ["extract" columns crit subquery]
+      ["extract" columns ["and" subquery crit]])
 
     [["extract" columns subquery clauses]]
     ["extract" columns ["and" subquery crit] clauses]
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -86,7 +86,9 @@
     ["extract" columns crit]
 
     [["extract" columns subquery]]
-    ["extract" columns ["and" subquery crit]]
+    (if (= "group_by" (first subquery))
```
wkalt: bit of a nitpick but can't we just do this with match like in the rest of this function?

ajroetker: I had that but changed it to this, because it just leads to duplicating the clause and a guard. Thought this looked cleaner.

wkalt: I don't think it needs a guard, but I'm okay if you like this better

wkalt:
+1

PR #1875, opened 2016-02-29
(PDB-2329) Revise sticky terminus failover

mullr:
(This goes with https://github.com/puppetlabs/puppet-server/pull/931)

Use an injected java.util.concurrent.ConcurrentHashMap instance for
tracking the the state required for sticky query failover. This replaces
the previous behavior if trying to detect when we were running under
puppet server and reaching out for a static clojure.lang.Atom instance.


puppet/lib/puppet/util/puppetdb/concurrent_hash_map.rb
```diff
@@ -0,0 +1,32 @@
...
+    def put(key, value)
+      @mutex.synchronize {
+        @value[key] = value
+      }
+    end
+
+    def putIfAbsent(key, value)
+      @mutex.synchronize {
+        @value[key] = value unless @value[key]
+      }
```
cprice404: Why do you need all of this mutex stuff?

puppet/lib/puppet/util/puppetdb/shared_terminus_state.rb
```diff
@@ -0,0 +1,32 @@
...
+  # resource drift when master-puppetdb conections flicker.
+  #
+  # When running under puppet server, a real ConcurrentHashMap will be injected
+  # via the shared_terminus_state= class method which is shared between all
+  # local JRuby instances. When running under plain ruby, a local ruby
+  # implementation (called ConcurrentHashMap) is used.
+  class SharedTerminusState
+    @@shared_terminus_state = ConcurrentHashMap.new
+
+    @@MAP_KEY = :puppetdb_last_good_query_server_url_index
```
cprice404: Would be better to just use a constant here instead of a class variable; class variables are kind of evil

cprice404:
This looks mostly good to me, except I really don't think you need any of the mutex stuff.  Ruby doesn't actually have threads and you'll never be running in an environment where anyone would be trying to call those methods from multiple threads concurrently, so introducing a dependency on those constructs mostly feels like asking for trouble to me.

mullr:
@cprice404 all updated

cprice404:
+1 from me

kbarber:
@mullr I've been staring at this patch this morning, and it occurred to me it's hard to validate it is working. I know we can prove it doesn't break things, by looking at acceptance tests, but how do I tell its doing what I expect? It sounds like it should retain memory of sticky state across ruby instances, should I just recreate a failure using puppet server on my laptop, and ensure that after X amount of tests it still remembers that (now) that the other hostname is the primary service to call?

mullr:
@kbarber yes, one way to test this is to use a puppet server with my patch applied and hook it up to two puppetdbs. Don't configure replication between the two. Make sure puppetserver is running with two ruby instances. Then run puppet twice and see the messages go the the first puppetdb. Kill the first pdb, run puppet once, and check that the messages go to the second pdb. Then bring the first pdb back, waiting for it to start, and run puppet *twice* checking that all the messages continue to go to the second pdb.

kbarber:
For the shared cached code path this relies on this PR: https://github.com/puppetlabs/puppet-server/pull/931

kbarber:
Okay, my end-to-end tests with the corresponding puppet-server patch seem to say this does was it says on the tin. The only condition though where this matters, is when command_broadcast is true just to note, because if its false, it will update the cache in the current thread anyway :-). So really, this is about solving the sticky_read_failover and command_broadcast = true cases. Which it solves. So good.

kbarber:
@mullr I think we're still waiting on puppet-server, they wanted some tests. What do you want to do with this PR? Leave it, or we can close and re-open later if the other work is low on your priority list. Just wary of leaving this open indefinitely, 'tis all.

mullr:
This is still in limbo, as we figure out what to do with agent-side failover.

kbarber:
I'm going to close this for now since its in limbo, please re-open when we are ready to go.

PR #1876, opened 2016-03-01
(PDB-1960) This commit unblocks startup on the post migration vacuum

ajroetker:
This commit moves the post migration vacuum analyze to a future so that
we don't block startup on the analyze which can take a while on large
databases.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -57,7 +57,7 @@
                  ;; Database connectivity
                  [com.zaxxer/HikariCP "2.4.3"]
                  [org.clojure/java.jdbc "0.4.2"]
-                 [org.postgresql/postgresql "9.4-1206-jdbc4"]
+                 [org.postgresql/postgresql "9.4.1208"]
 
                  ;; MQ connectivity
                  [org.apache.activemq/activemq-broker "5.13.1" :exclusions [org.slf4j/slf4j-api]]
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1089,9 +1089,10 @@
     (into (sorted-map)
           (select-keys migrations pending))))
 
-(defn- sql-or-die [f]
+(defn- sql-or-die
   "Calls (f) and returns its result.  If f throws an SQLException,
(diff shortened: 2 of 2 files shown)


project.clj
```diff
@@ -57,7 +57,7 @@
                  ;; Database connectivity
                  [com.zaxxer/HikariCP "2.4.3"]
                  [org.clojure/java.jdbc "0.4.2"]
-                 [org.postgresql/postgresql "9.4-1206-jdbc4"]
+                 [org.postgresql/postgresql "9.4.1208"]
```
mullr: just housekeeping, or do we need something from the new lib?

ajroetker: Housekeeping, there was a ticket right next to this one for updating to 9.4, and I noticed we weren't using the jdbc driver versions that are on the postgres page

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1157,10 +1158,14 @@
...
         (assert (not (:connection db-connection-pool)))
-        (jdbc/with-db-connection db-connection-pool
-          (log/info "Analyzing database")
-          (sql-or-die
-           (fn [] (jdbc/do-commands-outside-txn "vacuum (analyze, verbose)")))))
+        ;; We don't want to stay in maintenance mode while we vaccum analyze,
+        ;; i.e. refuse http requests as this can take quite some time on large
+        ;; databases
+        (future
+          (jdbc/with-db-connection db-connection-pool
```
mullr: what happens if something in here throws?

ajroetker: Before this would have killed pdb on startup, now it would do nothing.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1157,10 +1158,14 @@
...
+           (log/info "Analyzing database")
+           (fn []
```
mullr: I don't think this actually ever gets called...

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: we need to at least log something

PR #1877, opened 2016-03-01
(PDB-1960) Don't block startup on the post migration vacuum

ajroetker:
This commit moves the post migration vacuum analyze to a future so that
we don't block startup on the analyze which can take a while on large
databases.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1594,9 +1594,10 @@
     (into (sorted-map)
           (select-keys migrations pending))))
 
-(defn- sql-or-die [f]
+(defn- sql-or-die
   "Calls (f) and returns its result.  If f throws an SQLException,
   logs the exception and its parent and then calls (System/exit 1)"
+  [f]
   ;; Here we've preserved existing behavior, but this may warrant
   ;; further consideration later.  If possible, we might want to
   ;; avoid System/exit in favor of careful exception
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -57,7 +57,7 @@
                  [com.jolbox/bonecp "0.7.1.RELEASE" :exclusions [org.slf4j/slf4j-api]]
                  [org.clojure/java.jdbc "0.3.7"]
                  [org.hsqldb/hsqldb "2.2.8"]
-                 [org.postgresql/postgresql "9.2-1003-jdbc4"]
+                 [org.postgresql/postgresql "9.4.1208"]
```
ajroetker: General upkeep but also fixes https://tickets.puppetlabs.com/browse/PDB-1937 for users sooner

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1634,13 +1634,21 @@
...
-                            .createStatement
-                            (.execute "vacuum (analyze, verbose)")))))))
-    (log/info "There are no pending migrations")))
+        ;; We don't want to stay in maintenance mode while we vaccum analyze,
+        ;; i.e. refuse http requests as this can take quite some time on large
+        ;; databases
+        (future
+          (jdbc/with-db-connection db-connection-pool
+            (log/info "Analyzing database")
+            (-> (doto (:connection (jdbc/db)) (.setAutoCommit true))
```
mullr: I think you may have intended this to be inside the `try` block below

ajroetker:

PR #1881, opened 2016-03-07
(maint) Remove unused terminus code and warning

mullr:
Remove an obsolete implementation of shared terminus state and eliminate
the associated warning message printed on startup.


ajroetker:
:+1: on another note we should have jruby terminus tests in travis :)

PR #1882, opened 2016-03-08
(PDB-1960) Remove race condition with initial vacuum analyze

ajroetker:
This commit move the asynchronous vacuum analyze to use the write-db
connection pool instead of the temporary one we use for migrations so
that we know the connection will still be around while we attempt to
execute the vacuum.


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -242,6 +242,10 @@
       (log/infof "PuppetDB version %s" v))
 
     (init-with-db database config)
+    (sutils/async-vacuum-analyze write-db)
```
kbarber: Won't this always execute, even if we don't do a migration?

ajroetker:
@kbarber fixed that up, should only run when there are migrations

PR #1887, opened 2016-03-08
(PDB-2390) Upgrade to ActiveMQ 5.13.2

mullr:
This fixes an issue with ever-growing scheduler logs for failed
commands.

Changelog for 5.13.1: https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12311210&version=12334251
Changelog for 5.13.2: https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12311210&version=12334774

I've been through these and nothing aside from the relevant fix looks to impact features we use.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -60,9 +60,9 @@
                  [org.postgresql/postgresql "9.2-1003-jdbc4"]
 
                  ;; MQ connectivity
-                 [org.apache.activemq/activemq-broker "5.13.0"]
-                 [org.apache.activemq/activemq-kahadb-store "5.13.0"]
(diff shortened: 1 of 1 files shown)


mullr:
I verified that this does indeed fix this issue. If you want to do the same, here's an easy way to submit a lot of failing commands: 

```sh
while true; do curl -X POST http://localhost:8080/pdb/cmd/v1 -H 'Content-Type:application/json' -d '{"command": "store report", "version": 7, "payload": 42}'; done
```

The bug is that db-*.log files would pile up in the `mq/localhost/scheduler` directory. With the fix, you should see the old ones (lower numbers) get cleaned up over time.

PR #1890, opened 2016-03-09
(PDB-2381) data collection script

wkalt:
Change a couple of our summary stats from character lengths to column sizes.
This will make the queries run faster.


ajroetker:
:+1:

PR #1893, opened 2016-03-09
(maint) Remove some more potential ordering issues from tests

ajroetker:
This commit removes some more potential ordering issues from tests.

### DIFF
--- a/test/puppetlabs/puppetdb/testutils/reports.clj
+++ b/test/puppetlabs/puppetdb/testutils/reports.clj
@@ -61,12 +61,22 @@
               (dissoc :environment :containing_class :certname)))
         resource-events)))
 
+(defn munge-resources-for-comparison [resources]
+  (let [resources-sort-fn (partial sort-by (juxt :resource_type :resource_title))
+        events-sort-fn (partial sort-by :property)]
+    (->> resources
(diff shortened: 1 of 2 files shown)


test/puppetlabs/puppetdb/testutils/reports.clj
```diff
@@ -61,12 +61,26 @@
               (dissoc :environment :containing_class :certname)))
         resource-events)))
 
+(defn munge-resources-for-comparison [resources]
+  (let [resources-sort-fn (partial sort-by
+                                   (fn [r]
+                                     (mapv r [:resource_type :resource_title :timestamp])))
```
kbarber: Isn't the uniqueness required here only for type and title?

test/puppetlabs/puppetdb/testutils/reports.clj
```diff
@@ -61,12 +61,26 @@
...
+        events-sort-fn (partial sort-by
+                                (fn [e]
+                                  (mapv e [:property :old_value :new_value])))]
```
kbarber: I thought the uniqueness here was for property only, I didn't think you could multiple changes on a property within one report.

PR #1894, opened 2016-03-10
(PDB-2491) Document the installation of the PuppetDB CLI

ajroetker:
This commit adds some preliminary documentation for the PuppetDB CLI to the
PuppetDB repo and briefly touches on some examples and configuration.


kbarber:
I'm really apprehensive to merge something like this, until we've shipped the CLI, or are just about to. I'm really apprehensive this will be late, I would hold off on merging this until we've got a package ready to go. Having it 'ready in docs' might be convenient, but if we ship 4.0 and no CLI, we'll have to revert these docs which will be more painful.

ajroetker:
@kbarber that's cool with me, no rush on my part to get this in.

kbarber:
This will need to be rebase onto stable later today, so I'm closing this for now. Later today 'stable' will represent 4.0.x, so thats where it should land.

PR #1897, opened 2016-03-15

rbrw:


rbrw:
This is ready for initial review, but I added "don't merge" because I'm working on one more test I may want to add, I might rewrite one operation as a reduce, and because I wanted to discuss one aspect with (I think) AJ.

rbrw:
I think this is ready for review.  I'll just want to squash the top commit before merge.

kbarber:
Since this is a bugfix on 4.0.0, this will need to be rebase onto stable later today, so I'm closing this for now. Later today 'stable' will represent 4.0.x, so thats where it should land.

PR #1902, opened 2016-03-17
(PDB-2491) Document the installation of the PuppetDB CLI

ajroetker:
This commit adds some preliminary documentation for the PuppetDB CLI to the
PuppetDB repo and briefly touches on some examples and configuration.


ajroetker:
docs only change

documentation/anonymization.markdown
```diff
@@ -4,20 +4,22 @@ layout: default
...
 
 Using the `export` command
 -----
 
-To create an anonymized PuppetDB archive directly, use the Puppet `db` subcommand
-from any node with puppet-client-tools installed:
+To create an anonymized PuppetDB archive directly, use the Puppet `db`
+subcommand from any node with puppet-client-tools installed:
 
-    $ puppet db export --outfile my-puppetdb-export.tar.gz --anonymization moderate
```
kbarber: I count 66 cases where we use $ in front of a cli command, and you've changed 5 of them in this PR, and added a whole bunch of bare ones as new ones. If we're going to change a convention, we should do it to the lot, or not at all.

documentation/_puppetdb_nav.html
```diff
@@ -15,7 +15,6 @@ <h2 id="puppetdb">PuppetDB {{ puppetdb_version }}</h2>
   </li>
   <li><strong>Installation</strong>
     <ul>
-      <li><a href="/puppetdb/{{ puppetdb_version }}/migrate.html">Migrating data</a></li>
```
kbarber: Nice catch

documentation/pdb_client_tools.markdown
```diff
@@ -0,0 +1,79 @@
+---
+title: "PuppetDB 4.0: PuppetDB CLI"
+layout: default
+canonical: "/puppetdb/latest/pdb_client_tools.html"
```
kbarber: Apparently, this is now optional: https://github.com/puppetlabs/puppetdb/commit/67606cbb07c91e45efbfbe8d774be028c938c666

documentation/pdb_client_tools.markdown
```diff
@@ -0,0 +1,79 @@
...
+If the node you installed the CLI on is not the same node as your PuppetDB
+server, you will need to add the CLI node's certname to the PuppetDB
+certificate-whitelist and specify the paths to the CLI node's cacert, cert, and
+private key when using the CLI either with flags or a configuration file.
+  
+To configure the PuppetDB CLI to talk to your PuppetDB with flags, add a
+configuration file at `$HOME/.puppetlabs/client-tools/puppetdb.conf`. For more
+details see the installed man page:
+
+  man /opt/puppetlabs/client/tools/share/man/man8/puppetdb_conf.8
```
kbarber: Oh god, this is horrible, can we get the man path updated? Or give instructions on adding this stuff to the users man path? Even 'MANPATH=/opt/puppetlabs/client/tools/share/man man puppetdb_conf' is nicer than this to drive home the manpath persistent point, but still horrible.

documentation/pdb_client_tools.markdown
```diff
@@ -0,0 +1,79 @@
...
+  
+### Step 4: Enjoy!
+
+Here are some examples of using the CLI.
+
+#### Using `puppet-query`
```
kbarber: Should we drop the hyphen in the title here and below?

documentation/puppetdb-faq.markdown
```diff
@@ -6,7 +6,7 @@ canonical: "/puppetdb/latest/puppetdb-faq.html"
 ---
 
 [trouble_kahadb]: ./trouble_kahadb_corruption.html
-[migrating]: ./migrate.html
+[migrating]: ./anonymization.html
```
kbarber: So anonymization.html, when you open it describes itself as instructions for anonymizing data, yes its also got some import/export commands, but doesn't break down a migrate like the old migrate.html did ... its examples are even inclusive of anonymization modes which isn't what a user would expect I guess. I think, either the anonymization doc needs a better migration (with no anon) change, or we should resurrect migrate.html but leave out the storeconfig stuff.

kbarber:
@ajroetker this needs a rebase, but also - I'm wary this anon doc doesn't cover the needs of the old migrate doc in this references we see. Perhaps we shouldn't have killed it completely, but just removed the storeconfigs section or something. Either that or repurpose anon ...

kbarber:
@ajroetker hey man, I think the scope of PDB-2491 also included changing the PQL examples to use the CLI also. Can we include that in this PR?

project.clj
```diff
@@ -1,4 +1,4 @@
-(def pdb-version "4.0.1-SNAPSHOT")
+(def pdb-version "4.1.0-SNAPSHOT")
```
kbarber: Probably carried over from a rebase ^^

documentation/pdb_client_tools.markdown
```diff
@@ -0,0 +1,88 @@
...
+### Step 2: Enable the Puppet Labs package repository
+
+If you didn't already use it to install Puppet, you will need to
+[enable the Puppet Labs package repository][repos] for your system.
+
+### Step 3: Install and configure the PuppetDB CLI
+
+Use Puppet to install the PuppetDB CLI:
+
+  $ puppet resource package puppet-client-tools ensure=latest
```
kbarber: Doesn't this require a 4 space indent?

kbarber: A few other examples below for this also.

documentation/api/query/tutorial-pql.markdown
```diff
@@ -30,7 +30,32 @@ and specifying a `query` URL parameter (in the GET case) or a JSON-valued
...
+command line by using the [PuppetDB CLI][cli_install] or using [curl][curl].
+
+### Querying with the PuppetDB CLI
+
+[See the PuppetDB CLI installation page for more information about using the PuppetDB CLI.][cli_install]
+
+**Without SSL:**
+
+    puppet query '<PQL query>' \
+      --urls http://puppetdb.example.com:8080/pdb/query/v4
```
kbarber: Having the user pass flags each time is annoying, but this document doesn't give the hint to the user on how to configure this permanently, perhaps a mention of the docs on how to do this (ie. `man puppetdb_conf`).

documentation/pdb_client_tools.markdown
```diff
@@ -0,0 +1,88 @@
...
+If the node you installed the CLI on is not the same node as your PuppetDB
+server, you will need to add the CLI node's certname to the PuppetDB
+certificate-whitelist and specify the paths to the CLI node's cacert, cert, and
+private key when using the CLI either with flags or a configuration file.
+  
+To configure the PuppetDB CLI to talk to your PuppetDB with flags, add a
+configuration file at `$HOME/.puppetlabs/client-tools/puppetdb.conf`. For more
+details see the installed man page:
+
+  $ man puppetdb_conf
```
kbarber: Probably should just hyphenate this, I know this is just how it works now though :-).

documentation/api/query/tutorial-pql.markdown
kbarber: Oh this is repeated in the pql reference, hohum - I had a feeling those two docs would overlap.

kbarber:
@ajroetker I think there are a few points left on this:

* The anonymization doc, its not angled towards documenting migration on its own, I'd say at least a title change to adapt it to both changes (you've changed the index, but not the canonical title of the document). And the document text is too far aligned with anonymization only, for example the intro just speaks of anonymization, if you got directly linked to this thinking you were looking for migration it would be confusing.
* The pql examples for running puppet query should emphasize how to persist configuration rather than always requiring the command line arguments, if I was looking at all those args in that example, I'd be wondering how I would persist them forever, so perhaps a link or another paragraph about this would be helpful.

If I'm not mistaken, this can't be merged until we ship the CLI right?

ajroetker:
@kbarber Does https://github.com/puppetlabs/puppetdb/pull/1902/files#diff-ea8d34bc660e869b518a99ae923bd654R60 not address the second bullet point in your list, I added that a bit ago.

ajroetker:
I'll make a separate ticket for aligning anonymization

ajroetker:
I think we'll want to ship the CLI before adding these docs.

kbarber:
@ajroetker yep, looks like that covers my concern for point 2, sorry I missed that.

ajroetker:
@kbarber I added links to our 3.2 docs for the `Migrating data`, since that page is for HSQL users primarily it seemed reasonable to link them to the earlier docs anyways.

kbarber:
@ajroetker I'm probably not explaining my concern well enough, let me have another stab, or perhaps we can chat about this today.

documentation/anonymization.markdown
```diff
@@ -1,10 +1,11 @@
 ---
 title: "PuppetDB 4.0: Data anonymization"
 layout: default
-canonical: "/puppetdb/latest/anonymization.html"
 ---
 
-The export tool has the ability to anonymize the PuppetDB archive before returning the archive. This is particularly useful when sharing PuppetDB data that contains sensitive items.
+The export tool has the ability to anonymize the PuppetDB archive before
```
kbarber: So this paragraph right here, and the title, is specific to data anonymization. At least its heavily directed that way.

However, in the new cli client tools docs, we link to it talking mainly about export/import ...

This doc by linking to it this way is becoming something that covers the export/import tools more generally, and anonymization now becomes an option one can use. I'd say change the title of the doc, maybe to 'Export, Import and Anonymization' or something like that, and perhaps have a better intro here.

documentation/pdb_client_tools.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+Or handle your PuppetDB imports:
+
+    $ puppet db import pdb-archive.tgz
+
+For more information on the `db` command:
+
+    $ man puppet-db
+  
+For more information about PuppetDB exports and imports [see][export].
```
kbarber: This is what I mean, its talking about exports/imports, but the document is called 'Data anonymization' its just mildly confusing.

documentation/_puppetdb_nav.html
```diff
@@ -13,7 +13,7 @@ <h2 id="puppetdb">PuppetDB {{ page.doc.version }}</h2>
   </li>
   <li><strong>Installation</strong>
     <ul>
-      <li><a href="{{puppetdb}}/migrate.html">Migrating data</a></li>
+      <li><a href="{{puppetdb}}/../3.2/migrate.html">Migrating data</a></li>
```
kbarber: The problem with linking this from here, is that it makes it sound like all the instructions in that doc will work on PDB 4, which it actually won't. I'm not quite sure what we're trying to achieve here by having it in the main index. If its upgrading or whatever help that people need in 3.2, perhaps a mention in the release notes is better ... or something like that.

ajroetker:
@kbarber updated this PR

ajroetker:
This relates to https://github.com/puppetlabs/puppetdb-cli/pull/34

PR #1904, opened 2016-03-18

rbrw:


rbrw:
Rebased to new master and ready for review - the "don't merge" is just there because I want to do a final squash (of currently just the last commit) before merge.

kbarber:
This is a bug, so it should be targeted at 'stable' not master.

rbrw:
Right, thanks.

PR #1905, opened 2016-03-21
(PDB-2565) Change =-after? to is-equal-after

kbarber:
This wraps the entire macro in an `is` assertion. Since all cases where this was
used did this anyway, it provides more convenience. However, the main reason was
to ensure that the differential between the old and new values actually gets
printed during testing instead of simply returning the expectation of being true.


kbarber:
Needs to be merged in-sync with this: https://github.com/puppetlabs/pe-puppetdb-extensions/pull/173

PR #1906, opened 2016-03-21

rbrw:


rbrw:
Rebased to new stable and ready for review - the "don't merge" is just there because I want to do a final squash (of currently just the last commit) before merge.

ajroetker:

ajroetker:
Fwiw the jenkins tests passed looks like the plugin just failed

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1473,9 +1556,13 @@
             [["not" expression]]
             (map->NotExpression {:clause (user-node->plan-node query-rec expression)})
 
-            [["in" column subquery-expression]]
-            (map->InExpression {:column (columns->fields query-rec (utils/vector-maybe column))
-                                :subquery (user-node->plan-node query-rec subquery-expression)})
+            [["in" column-s subquery-expr]]
```
kbarber: column or columns, lol ... I'd just call it columns, I can see that hyphen annoying people :-).

kbarber:
@rbrw seems to fix the bug :+1:

test/puppetlabs/puppetdb/query/ast_test.clj
```diff
@@ -0,0 +1,132 @@
...
+   [puppetlabs.kitchensink.core :refer [dissoc-in]]
+   [puppetlabs.puppetdb.cli.services :as cli-svcs]
+   [puppetlabs.puppetdb.command :refer [enqueue-command]]
+   [puppetlabs.puppetdb.examples :as examples]
+   [puppetlabs.puppetdb.testutils :refer [block-until-results]]
+   [puppetlabs.puppetdb.testutils.cli :refer [get-catalogs get-factsets]]
+   [puppetlabs.puppetdb.testutils.services :as svc-utils :refer [*server*]]
+   [puppetlabs.trapperkeeper.app :refer [get-service]]))
+
+(deftest ast-queries
```
kbarber: I'm not sure, but I keep thinking these tests should belong somewhere else, in that we've already got a place for this. Perhaps http/facts_test.clj?

test/puppetlabs/puppetdb/query/ast_test.clj
```diff
@@ -0,0 +1,132 @@
...
+  (svc-utils/with-single-quiet-pdb-instance
+    (let [dispatcher (get-service *server* :PuppetDBCommandDispatcher)
+          pdb (get-service *server* :PuppetDBServer)
+          query (fn [ast] (cli-svcs/query pdb :v4 ast {} set))]
+
+      (enqueue-command dispatcher :replace-facts 4
```
kbarber: I'm not quire sure also why this test setup needs to diverge that far from the way http/facts_test.clj does it, but perhaps you can enlighten me.

kbarber:
@rbrw except for the few comments I've made, +1 from me.

test/puppetlabs/puppetdb/query/ast_test.clj
rbrw: Happy to change it we prefer, but I did this initially because I didn't want an http test, i.e. that part of the stack's not relevant/desirable here -- these are intended to test AST forms as directly as possible, and because I was thinking about having additional tests eventually, which might or might not involve facts, and probably would cross resource types arbitrarily.

test/puppetlabs/puppetdb/query/ast_test.clj
rbrw: Perhaps my comment above covers it, but I wanted as direct and lightweight an AST test as possible.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -19,18 +19,72 @@
...
+                   "edges"
+                   "environments"
+                   "fact_paths"
+                   "fact_values"
+                   "facts"
+                   "factsets"
+                   "latest_report"
+                   "reports"
+                   "resource_events"
+                   "resource_params"})
```
ajroetker: This list is possibly different in PE, I don't think I'm a fan of adding this validation because it's mostly going to be something that fails whenever someone adds a new migration and they'll have to track down where the schema error is coming from.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -873,26 +920,20 @@
...
        (remove (comp :query-only? val))
-       keys
-       sort))
-
-(defn extract-fields
-  "Returns all fields from a projection.  Returns nil for fields which
-  are query-only? since these can't be projected either."
-  [[name {:keys [field]}] entity]
-  [field name])
+       keys))
```
ajroetker: Why'd you get rid of the sort here?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -873,26 +920,20 @@
   [{:keys [projections]}]
   (->> projections
        (remove (comp :query-only? val))
```
ajroetker: You changed this call to `(comp :query-only? second)` below, might want to change this one as well, but I actually like using `val` instead since it signals that the args are actually map kv pairs.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1254,17 +1313,23 @@
...
+                                   column-list))
+    (let [[subname & subexpr] expr
+          logobj (user-query->logical-obj subname)
+          projections (:projections logobj)]
+      (assoc logobj
+             :projected-fields (mapv (fn [name]
+                                       [name (projections name)])
+                                     column-list)
+             :where (when (seq subexpr)
+                      (user-node->plan-node logobj (first subexpr)))))))
```
ajroetker: `(some->> subexpr first (user-node->plan-node logobj))`

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1315,15 +1382,18 @@
         {:keys [limit offset order-by]} (create-paging-map clauses)]
     (if (extract-expression? expr)
       (let [[extract columns remaining-expr] expr
-            column-list (utils/vector-maybe columns)]
+            column-list (utils/vector-maybe columns)
+            projections (:projections query-rec)]
         (-> query-rec
-            (assoc :projected-fields column-list :where (user-node->plan-node query-rec remaining-expr))
+            (assoc :projected-fields (mapv (fn [name] [name (projections name)])
+                                           column-list)
```
ajroetker: Looks like you have a similar function above, might be worth pulling giving it a name and docstring :)

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1363,73 +1435,85 @@
...
 
-            [[(op :guard #{">" "<" ">=" "<="}) column value]]
-            (let [{:keys [type field]} (get-in query-rec [:projections column])]
-              (if (or (= :timestamp type) (and (or (= :float type)
-                                                   (= :integer type)) (number? value)))
+            [[(op :guard #{">" "<" ">=" "<="}) column-name value]]
+            (let [{:keys [type] :as cinfo} (get-in query-rec
+                                                   [:projections column-name])]
+              (if (or (= :timestamp type)
+                      (and (number? value) (#{:float :integer} type)))
```
ajroetker: `(get #{:float :integer} type)`

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1741,6 +1828,70 @@
...
+(defn- fix-in-expr-multi-comparisons
+  [node]
+  (let [columns (:column node)
+        projected-fields (get-in node [:subquery :projected-fields])]
+    (assert (= (count columns) (count projected-fields)))
+    (loop [cols columns
+           fields projected-fields
+           fixed-cols []
+           fixed-fields []]
+      (if-not (seq cols)
```
ajroetker: could we reverse the conditions and just use `if (seq`? `if-not (seq` always makes me double-take

src/puppetlabs/puppetdb/query_eng/engine.clj
rbrw: There's a (probably not elaborate enough) comment in the commit message for this change, but it's because I believe the engine relies on the correlation between the ordering of the projections and the columns, and so sorting was at best extra work, and at worst, might break the correlation unless both sides always sort exactly the same.  I think I had an example that would break, but now can't remember it.  (Or maybe I've just misunderstood the situation...)

src/puppetlabs/puppetdb/query_eng/engine.clj
rbrw: OK - I wondered about that, but figured it should be pretty obvious where you needed to add the string, and this made it very clear what Query expects.  However, I hadn't thought about PE.  If we like, I can remove the set, though I might want to at least add  a comment making it clear that source-table should be a string table name.

src/puppetlabs/puppetdb/query_eng/engine.clj
rbrw: It's actually not a kv pair in the projected-fields case.  The projected fields are a collection, not a map of two element vectors (i.e. projection-schemas), whose order in an in clause parallels the columns'.

src/puppetlabs/puppetdb/query_eng/engine.clj
rbrw: Oh, right.

src/puppetlabs/puppetdb/query_eng/engine.clj
rbrw: Sure, if it's confusing -- I suppose to some extent it's a habit; I sometimes tend to put the smaller and/or base case first, and the more complex case, or a recursion, last.

src/puppetlabs/puppetdb/query_eng/engine.clj
ajroetker: > I might want to at least add a comment making it clear that source-table should be a string table name.

I'd be :+1: for that.

src/puppetlabs/puppetdb/query_eng/engine.clj
ajroetker: Ah so this really *should* be second then?

rbrw: Right.

rbrw:
@ajroetker @kbarber Added new temp commits to hopefully address your concerns, leaving only the ast test location/type question.

test/puppetlabs/puppetdb/query/ast_test.clj
kbarber: So, http/index_test.clj sort of behaves this way - that is, most tests are generic and test cross-functionality, but has the PQL tests alongside their AST equivalent. However its full stack, involving the HTTP side of things also. So in a way, its doing a lot of the cross-entity conceptual testing already, like you have got here ... but its implementation is slightly different.

Perhaps we need to add to, improve and fix up that space rather than forking out a new test area on its own, otherwise I can just see indecision as to where tests should live becoming a thing with two generic coverage places existing. Also, the PQL coverage in the other file is already there, and this is important to ensure AST/PQL do equivalent things.

Maybe we could revise http/index_test.clj to be coverage for full-stack and direct also, who knows.

There is also room to add unit testing to certain things here, in query_eng_test.clj remember, like if you wanted to make sure certain SQL compilation/translation occurred outside of the full stack, that place is a good one. I noticed you've done some tiny fixes there, but there have been times doing unit tests there for complex query eng handling wasn't a bad idea.

Anyway, I'm not against the conceptual idea you have here, I'm just wary this is 'yet another place to add tests' and will create some ambiguity and duplication problems.

Also, query/ast_test.clj is kind of in the 'query' namespace, which we hope to remove in the future btw ... query.clj has a lot of legacy stuff in it, our goal is to collapse that functionality when we can, so probably this directory isn't go to add to atm :-).

test/puppetlabs/puppetdb/http/index_test.clj
```diff
@@ -241,3 +245,128 @@
...
+            (testing "that \"in\" works if field and subquery projection are multi"
+              (is (= #{{:name "foo" :value 1}
+                       {:name "bar" :value 2}
+                       {:name "baz" :value 3}
+                       {:name "match" :value "match"}}
+                     (query ["from" "facts"
+                             ["extract" ["name" "value"]
+                              ["in" ["value"]
+                               ["extract" ["value"]
+                                ["select_facts"]]]]])))))))))
```
wkalt: never knew this worked

PR #1907, opened 2016-03-21
(doc) Fix release note links.

gguillotte:
- Fix the broken `pqltutorial` link.
- Make the 3.2 `migrate` link a relative link.


kbarber:
Looks like this PR was raised against master, it should be rebased against stable. Closing, so you can re-open it against the correct branch.

PR #1908, opened 2016-03-21
(PDB-2482) Add ability to pass in token auth (RBAC auth) to PuppetDB

ajroetker:
This commit adds the ability to pass in token authorization functions to our PuppetDB middleware, these functions come from [clj-rbac-client](https://github.com/puppetlabs/clj-rbac-client) and will only be passed in to the authorization middleware in the PE
PuppetDB extensions repo, [see this PR](https://github.com/puppetlabs/pe-puppetdb-extensions/pull/175).

This commit allows us to pass in the rbac-client functions `valid-token->subject` and `is-permitted?`, [found here](https://github.com/puppetlabs/clj-rbac client/blob/master/src/puppetlabs/rbac_client/services/rbac.clj#L40) so that our cert-whitelist middleware can optionally fallback to checking for a [token](https://github.com/puppetlabs/clj-jwt) in a `X-Authentication` header in the request with which PuppetDB will use to talk to RBAC. If the subject of the token has the right permission, `puppetdb:use:*` (similar to what the Orchestrator uses for it's RBAC permission), the request is validated.

This will require us to use `client-auth: want` in our Jetty configuration for PuppetDB (by default Jetty uses `client-auth: need` which prevents requests which don't have a cert attached.

We do not cache the validity of the `token` because there is a hard requirement for revocation of permissions for tokens.

In summary:

1. Configuring PuppetDB to use a certificate-whitelist will add middleware to PuppetDB to validate requests.
2. Configuring `client-auth: want` in the Jetty config will allow SSL requests through to PuppetDB which do not have a `cert`.
3. PuppetDB will first attempt to authorize a request by checking if the ssl-client-cn is on the cert-whitelist.
4. If the ssl-client-cn is not on the whitelist, PuppetDB will check to see if the `X-Authentication` header has been set (i.e. the `token`), if it has been we query rbac for the subject of the `token`.
5. If the `token` corresponds to a valid subject, we query rbac again to see if the user has the permission `puppetdb:use:*`, and validate the request if they do.
6. If the `token` was invalid, the `subject` doesn't have the correct permissions, or no `token` was provided we respond with a `Permission denied`.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -57,7 +57,7 @@
                  ;; Database connectivity
                  [com.zaxxer/HikariCP "2.4.3"]
                  [org.clojure/java.jdbc "0.4.2"]
-                 [org.postgresql/postgresql "9.4.1208"]
+                 [org.postgresql/postgresql "9.4.1208.jre7"]
 
                  ;; MQ connectivity
                  [org.apache.activemq/activemq-broker "5.13.2" :exclusions [org.slf4j/slf4j-api]]
--- a/project.clj
+++ b/project.clj
@@ -57,7 +57,7 @@
                  ;; Database connectivity
                  [com.zaxxer/HikariCP "2.4.3"]
                  [org.clojure/java.jdbc "0.4.2"]
-                 [org.postgresql/postgresql "9.4-1206-jdbc4"]
+                 [org.postgresql/postgresql "9.4.1208"]
 
                  ;; MQ connectivity
                  [org.apache.activemq/activemq-broker "5.13.2" :exclusions [org.slf4j/slf4j-api]]
--- a/src/puppetlabs/puppetdb/middleware.clj
+++ b/src/puppetlabs/puppetdb/middleware.clj
@@ -19,6 +19,7 @@
             [metrics.meters :refer [meter mark!]]
             [clojure.walk :refer [keywordize-keys]]
             [puppetlabs.puppetdb.utils :as utils]
+            [slingshot.slingshot :refer [try+]]
             [bidi.bidi :as bidi]
             [bidi.ring :as bring]
             [bidi.schema :as bidi-schema]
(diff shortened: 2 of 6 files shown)


ajroetker:
ping @jonathannewman The new permission we are going to need is the `puppetdb:use:*` which is the same permission that the orchestrator uses.

jonathannewman:
@ajroetker that is a new permission entry.  The orchestrator uses `orchestration:use:*`  It implies a new `object-type` is added, with appropriate display name, with an new action (with display name), and bound to the appropriate roles by default.  https://github.com/puppetlabs/pe-rbac-service/blob/master/doc/permissions_and_default_roles.md#permissions  All of those things will have to be checked by UX.

jonathannewman:
@ajroetker what about Activity Logging?

ajroetker:
@jonathannewman 
1. Who handles that UX?
2. What's their process for requesting that UX?
3. Activity logging has been but a faint whisper in the wind. I have not heard anything about needing any activity service integration and I did not see it when I was investigating how the Orchestrator integrates with RBAC either. If you feel like we need activity service integration as well, would you mind putting a note in the ticket?

jonathannewman:
@ajroetker opened a ticket to track the work required in the ticket.  Department of redundancy department? :)

ajroetker:
@jonathannewman I'll set up a meeting to make sure we align priorities tracked by the tracking ticket. ;)

ajroetker:

project.clj
```diff
@@ -57,7 +57,7 @@
                  ;; Database connectivity
                  [com.zaxxer/HikariCP "2.4.3"]
                  [org.clojure/java.jdbc "0.4.2"]
-                 [org.postgresql/postgresql "9.4-1206-jdbc4"]
+                 [org.postgresql/postgresql "9.4.1208.jre7"]
```
ajroetker: There was a dependency conflict with the activity service so I thought I'd just bump it since it's an update anyways.

kbarber:

kbarber:

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -47,36 +48,78 @@
...
-        (do
-          (log/warnf "%s rejected by certificate whitelist %s" ssl-client-cn whitelist)
-          (format (str "The client certificate name (%s) doesn't "
-                       "appear in the certificate whitelist. Is your "
-                       "master's (or other PuppetDB client's) certname "
-                       "listed in PuppetDB's certificate-whitelist file?")
-                  ssl-client-cn))))))
+      (when-not (allowed? req)
+        (log/warnf "%s rejected by certificate whitelist %s" ssl-client-cn whitelist)
+        (http/json-response {:error (format (str "Permission denied: The client "
```
kbarber: Switching to json error messages is an API breaking change. There are ways and means for us to start utilizing JSON error messages, but we'd need some way to allow users to switch them on (thus deciding to break API on their own) or doing it in a v5.

kbarber:
@ajroetker I would expect some configuration being required to configure the RBAC client services, where does this live? We have no documentation for it, I presume its addendum to our existing conf.d directory stuff, I'm just wondering what documentation if any we need here.

ajroetker:
Closing and putting this into a design doc first.

PR #1909, opened 2016-03-22
(PDB-2548) prepared statement array in resource-exists?

wkalt:
This changes the resource-exists? function to use an array-valued prepared
statement parameter, rather than a list of parameters that could potentially
exceed max_short.


kbarber:
@wkalt not sure what we should do about rollup here.

PR #1912, opened 2016-03-25
(PDB-2494) add to_string function and ability to group by function results

wkalt:
This adds a to_string function that wraps PG's to_char function, and also
allows consumers to group by the results of a function, allowing users to get
counts of events by hour and similar.

### DIFF
--- a/documentation/api/query/v4/ast.markdown
+++ b/documentation/api/query/v4/ast.markdown
@@ -19,6 +19,7 @@ canonical: "/puppetdb/latest/api/query/v4/ast.html"
 [entities]: ./entities.html
 [pql]: ./pql.html
 [urlencode]: http://en.wikipedia.org/wiki/Percent-encoding
+[to-char]: http://www.postgresql.org/docs/9.4/static/functions-formatting.html
 
 ## Summary
 
@@ -182,10 +183,23 @@ or
 ### `function`
 
 The **function** operator is used to call a function on the result of a
-subquery. Supported functions are `count`, `avg`, `sum`, `min`, and `max`. The
-function operator is applied within the first argument of an extract, as in
-the examples above. The `avg`, `sum`, `min`, and `max` functions will ignore
-non-numeric fact values.
+subquery. Supported functions are described below.
+
+#### `avg`, `sum`, `min`, `max`
+These functions will operator on any numeric column and take the column name as
+an argument, as in the examples above.
+
+#### `count`
+The `count` function can be used with or without a column. When no column is
+supplied, it will return the number of results in the associated subquery.
+Using the function with a column will return the number of results where the
+specified column is not null.
+
+#### `to_string`
+The `to_string` function operates on timestamps and integers, allowing them to
+be formatted in a user-defined manner before being returned from puppetdb.
+Available  formats are the same as those documented for [PostgreSQL's `to_char`
+function][to-char].
 
 ### `group_by`
 
--- a/documentation/api/query/v4/pql.markdown
+++ b/documentation/api/query/v4/pql.markdown
@@ -429,6 +429,10 @@ each fact name across all facts, where the certame starts with `web`:
 
     facts[name, count(value)] { certname ~ "^web.*" group by name }
 
+Grouping on function results is also supported:
+
+    reports[count(), to_string(receive_time, "DAY")]{group by to_string(receive_time, "DAY")}
+
 ## Paging
 
 PQL supports restriction of the result set via the SQL-like paging clauses
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.puppetdb.query-eng.engine
   (:require [clojure.core.match :as cm]
             [clojure.string :as str]
+            [clojure.set :refer [map-invert]]
             [clojure.tools.logging :as log]
             [honeysql.core :as hcore]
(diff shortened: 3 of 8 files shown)


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -882,11 +890,24 @@
...
 
+(defn fnexpression->sql
+  ([expr]
+   (fnexpression->sql expr true))
+  ([{:keys [function column args]} alias?]
+   (let [qmarks (some->> (seq args)
+                         (str/join ", ")
+                         (str ", "))
+         column-alias (if alias?
+                        (format " AS %s" ((map-invert pdb-fns->pg-fns) function))
```
ajroetker: `(get (map-invert ..) function)`

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -882,11 +890,24 @@
...
+                        "")]
+     (hcore/raw (format "%s(%s%s)%s"
+                        function (name column) (str qmarks) column-alias)))))
```
ajroetker: Any reason we can't just do 
```clojure
(let [foo (apply hcore/call function column (seq args))]
  (if alias?
    [foo (get (map-invert pdb-fns->pg-fns) functions)]
    foo))
```

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1323,15 +1351,37 @@
...
+  [v]
+  (= "function" (first v)))
+
+(defn alias-columns
+  "Alias columns with their fully qualified names, and function expressions
+   with the function name."
+  [query-rec c]
+  (if (string? c)
+    (get-in query-rec [:projections c :field])
+    (keyword ((map-invert pdb-fns->pg-fns) (:function c)))))
```
ajroetker: Probably want a `(get` here as well

ajroetker:
This looks good to me code-wise and functionality-wise, @rbrw @wkalt might want watch out for merge conflicts with your two patches.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -34,9 +35,16 @@
...
 
 (def json-agg-row (comp h/json-agg h/row-to-json))
-(def supported-fns #{"sum" "avg" "min" "max" "count"})
+(def pdb-fns->pg-fns
+  {"sum" "sum"
+   "avg" "avg"
+   "min" "min"
+   "max" "max"
+   "count" "count"
+   "to_string" "to_char"})
```
ajroetker: I might just `def` the map-invert of this as well, with a `pg-fns->pdb-fns`, but not a big deal :)

wkalt:

wkalt:

wkalt:

documentation/api/query/v4/ast.markdown
```diff
@@ -182,10 +183,23 @@ or
...
+These functions will operator on any numeric column and take the column name as
+an argument, as in the examples above.
+
+#### `count`
+The `count` function can be used with or without a column. When no column is
+supplied, it will return the number of results in the associated subquery.
+Using the function with a column will return the number of results where the
+specified column is not null.
+
+#### `to_string`
```
kbarber: This needs to be added to the PQL reference as well.

kbarber:
@wkalt we're missing the necessary changes to have this work for PQL also, that is, EBNF changes, docs and some tests wouldn't hurt also.

wkalt:
@kbarber fixed

documentation/api/query/v4/pql.markdown
```diff
@@ -387,6 +387,10 @@ starts with `web`:
 
     facts[name, count(value)] { certname ~ "^web.*" group by name }
 
+Grouping on function results is also supported:
+
+    reports[count(), to_string(receive_time, DAY)]{group by to_string(receive_time, DAY)}
```
kbarber: ooh, bareword support for text args ... not sure I'm a fan ... right now unquoted-barewords in our language only represent functions, entities, fields ... this is none of those, it kind of changes the semantics a little. Up until now though our function args are pretty much only those things, so I get why this is easier but ... there are cases where users are going to have to provide spaces in their formatting ... so we'll need quotes anyway. Bare-word arbitrary string support in this place and not others is just going to be confusing also. I think this is a string argument, and should always be quoted basically.

test/puppetlabs/puppetdb/pql/parser_test.clj
```diff
@@ -722,8 +734,15 @@
...
     (are [in expected] (= (parse in :start :groupbyclause) expected)
-      "group by name" [[:groupby "name"]]
-      "group by name, value" [[:groupby "name" "value"]])
+         "group by name"
+         [[:groupby "name"]]
+
+         "group by name, value"
+         [[:groupby "name" "value"]]
+
+         "group by name, to_string(receive_time, \"HH24\")"
```
ajroetker: could we add a test for when `HH24` isn't quoted as well?

ajroetker: Nevermind, didn't see the one above

ajroetker:
👍 lgtm

wkalt:
@kbarber this will now restrict function calls to the form fn(column, "quoted string", "quoted string", ...). As discussed this will likely need refinement down the line.

kbarber:
@wkalt needs rebase

PR #1913, opened 2016-03-25
(PDB-2490) add latest_report_noop field to nodes

wkalt:
This adds a latest_report_noop flag to the nodes response to indicate whether
the latest puppet run for a node was a noop.


ajroetker:
Lgtm :+1:

PR #1915, opened 2016-03-29
(PDB-2576) Expire nodes in single SQL command

rbrw:
Instead of expiring one node at at time with a round-trip through PDB
for each, combine stale-nodes and expire-node! into expire-stale-nodes
which handles the entire process in a single SQL command, and returns
all of the expired certnames.

In addition to changing the expiration semantics a bit, this also alters
the reporting, logging "Auto-expired node foo" for each expired node
only after all of the nodes have been expired.  Previously an
"Auto-expiring node foo" was displayed for each node, just before that
node was expired.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -36,7 +36,7 @@
             [schema.core :as s]
             [puppetlabs.puppetdb.schema :as pls :refer [defn-validated]]
             [puppetlabs.puppetdb.utils :as utils]
-            [clj-time.core :refer [now]]
+            [clj-time.core :refer [ago now]]
             [puppetlabs.puppetdb.metrics.core :as metrics]
             [metrics.counters :refer [counter inc! value]]
             [metrics.gauges :refer [gauge-fn]]
@@ -45,7 +45,8 @@
             [puppetlabs.puppetdb.jdbc :refer [query-to-vec]]
             [puppetlabs.puppetdb.time :refer [to-timestamp]]
             [honeysql.core :as hcore])
-  (:import [org.postgresql.util PGobject]))
+  (:import [org.postgresql.util PGobject]
+           [org.joda.time Period]))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Schemas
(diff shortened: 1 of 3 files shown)


test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -1204,41 +1212,44 @@
...
-
-(deftest-db node-stale-reports
-  (testing "should not return a node with a recent report and nothing else"
+(deftest-db fresh-node-not-expired
+  (testing "fresh nodes are not expired"
+    (let [catalog (:empty catalogs)
+          certname (:certname catalog)]
+      (add-certname! certname)
+      (replace-catalog! (assoc catalog :producer_timestamp (now)) (now))
+      (is (= [] (expire-stale-nodes (-> 3 days .toPeriod))))
```
ajroetker: any reason not to just check `(empty?` here?

rbrw: Actually, I had that first, but then remembered that if you just check empty, you can't see what was wrong when it fails.

wkalt:
this looks fine to me, but @rbrw what's the semantic change you referenced in the commit message?

rbrw:
@wkalt I just meant that previously we logged expirations as we went, and performed the expirations incrementally, but now it's all or nothing, and you only find out what happened afterward.

wkalt:
@rbrw that seems fine to me

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1364,14 +1347,31 @@
...
+        expired-ts (to-timestamp (now))]
+    (map :certname
+         (jdbc/query-to-vec
+          (str
+           "update certnames set expired = ?"
+           "  where id in"
+           "    (select c.id from certnames c"
+           "         left outer join catalogs cats on c.certname=cats.certname"
+           "         left outer join factsets fs on c.certname=fs.certname"
+           "         left outer join reports r on c.certname=r.certname"
```
npwalker: It might be more efficient to join on certname.latest_report_id which will only return one row from reports instead of all of the reports.  

It's unfortunate that we need to join all the tables to determine inactivity.

ajroetker: Similar note for catalogs.

senior:
@rbrw @ajroetker @npwalker is there another change needed here? or is it ready for merge?

npwalker:
I believe the join to reports should be `left outer join reports r on c.latest_report_id = r.id`  

@ajroetker I don't see where we store the latest catalog id

rbrw:
This is on hold pending the resolution of https://tickets.puppetlabs.com/browse/PDB-2617

kbarber:
This needs to be merged first and rolled up to master: https://github.com/puppetlabs/puppetdb/pull/1929

PR #1917, opened 2016-03-31
(PDB-2268) Remove clj-http as a dependency from PuppetDB

ajroetker:
This PR includes removing clj-http as a dependnecy from PuppetDB, in favor of using the puppetlabs/clj-http-client.

The puppetlabs http-client includes all the SSL auth we need and removing clj-http removes version conflicts other projects get when trying to run PuppetDB in the same JVM (these conflicts come from clj-http and puppetlabs/clj-http-client).

This PR also include retiring the old import and export tools which were deprecated in PuppetDB 4.0.

__*Warning*__: There is also this PR on PE https://github.com/puppetlabs/pe-puppetdb-extensions/pull/178 which needs to be merged as well.


ajroetker:
transient ^

ajroetker:

wkalt:
this looks good to me. bummer that {:as :json} isn't supported but them's the breaks I guess.

cprice404:
> this looks good to me. bummer that {:as :json} isn't supported but them's the breaks I guess.

that'd be easy to add to clj-http-client if you guys need it.

acceptance/tests/commands/args-handling.rb
```diff
@@ -1,17 +1,7 @@
...
-
-    # We don't test ssl-setup here, because its actions without any arguments
-    # are tested elsewhere.
-    next if k == 'ssl-setup'
-    step "running puppetdb-#{k} with no arguments should not error out" do
-      on database, "#{bin_loc}/puppetdb #{k}", :acceptable_exit_codes => [1]
-      assert_match(/Missing required argument/, stdout)
-    end
+  step "running 'puppetdb ssl-setup -h' should not error out" do
+    on database, "#{bin_loc}/puppetdb ssl-setup -h"
```
rbrw: Perhaps Ruby expansion handles this somehow, or it's not something we want to pursue at the moment, but I'm wondering what happens if bin_loc has spaces in it, or something even unfriendlier.

ajroetker: I'd leave that for another patch, since this was preexisting :)

senior:
I'm not sure we should be retiring the CLI stuff. If we deprecated it in 4.0.0, don't we have to wait until 5.0.0 to drop them? I'm hoping we're still a good distance from 5.0.0

acceptance/tests/commands/args-handling.rb
rbrw: Right - fair enough.

src/puppetlabs/puppetdb/client.clj
```diff
@@ -2,7 +2,7 @@
   (:require [puppetlabs.puppetdb.http :as http]
             [clojure.tools.logging :as log]
             [puppetlabs.puppetdb.reports :as reports]
-            [clj-http.client :as http-client]
+            [puppetlabs.http.client.sync :as http-client]
```
rbrw: Looks like this might be aliased as both http-client and pl-http -- I wondered, if we're switching to primarily/only use this lib now, we might be able to just call it `http`, i.e. `http/get`, etc. or use some other fairly friendly, short name, while we're making changes.

src/puppetlabs/puppetdb/import.clj
```diff
@@ -58,11 +58,14 @@
...
-    (throw (IllegalStateException.
-            (format "Unable to find export metadata file '%s' in archive"
-                    metadata-path))))
-  (utils/read-json-content tar-reader true))
+  ;; *warning* archive/next-entry is side-effecting
+  (let [entry (archive/next-entry tar-reader)]
+    (when-not (= (.getName entry) metadata-path)
+      (throw (IllegalStateException.
+              (str "Unable to find export metadata '"
+                   metadata-path
```
rbrw: Not sure, but wondered about just using pr-str, which can't be fooled, to handle the quoting, i.e. `(pr-str metadata-path)`.

rbrw: Not sure, but wondered about just using pr-str, which can't be fooled, to handle the quoting, i.e. (pr-str metadata-path)

wkalt:
I think we need to change the puppetlabs client to accept ints on POST

PR #1920, opened 2016-04-01
(PDB-2579) Check hash duplication with historical catalogs

ajroetker:
This commit fixes an issue with historical catalogs where we were not
checking the for hash duplication and our catalog duplication rate
metric was always showing 0.


wkalt:
I think we'll want to hit the catalog-hash-match and catalog-hash-miss timers too

https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/scf/storage.clj#L741-741

also this PR should be named stable instead of master

PR #1921, opened 2016-04-01
(PDB-2579) Check hash duplication with historical catalogs

ajroetker:
This commit fixes an issue with historical catalogs where we were not
checking the for hash duplication and our catalog duplication rate
metric was always showing 0.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -766,27 +766,33 @@
            (let [hash (time! (:catalog-hash performance-metrics)
                              (shash/catalog-similarity-hash catalog))
                  {certname-id :certname_id
+                  stored-hash :catalog_hash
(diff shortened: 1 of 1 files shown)


senior:
I'll look at this one

senior:
This change looks ok to me, but am a little concerned that if it was to break tomorrow, we would miss it (again). We have historical catalogs tests. It seems like it would be pretty easy to add a before/after check of the changed metrics.

ajroetker:
@senior you'd think, but those tests don't spin up a whole PuppetDB just a database, I didn't see any tests for these metrics for non-historical catalogs either, not really sure if we should bulk add tests for all the storage.clj metrics or not.

senior:
@ajroetker is going to spin off a ticket for getting the metric initialization stuff in place so that we can start testing this stuff. People are using these metrics now and it's something we don't want to break in the future.

PR #1923, opened 2016-04-05
(maint) improve dotestseq macro

mruzicka:
Before this commit the ```dotestseq``` macro didn't show the values of the bindings per iteration in the test name, only the names of them.

    Testing case ([version endpoint] method)
    Testing case ([version endpoint] method)
    Testing case ([version endpoint] method)
    Testing case ([version endpoint] method)

This commit imoproves the macro such that it now produces better annotated test names like so:

    Testing case: [version endpoint]: [:v4 "/v4/catalogs/foo/edges"], method: :get
    Testing case: [version endpoint]: [:v4 "/v4/catalogs/foo/edges"], method: :post
    Testing case: [version endpoint]: [:v4 "/v4/catalogs/foo/resources"], method: :get
    Testing case: [version endpoint]: [:v4 "/v4/catalogs/foo/resources"], method: :post

### DIFF
--- a/test/puppetlabs/puppetdb/testutils.clj
+++ b/test/puppetlabs/puppetdb/testutils.clj
@@ -33,7 +33,11 @@
     `(do ~@body)
     (let [case-versions (remove keyword? (take-nth 2 bindings))]
       `(doseq ~bindings
-         (testing (str "Testing case " '~case-versions)
+         (testing (str "Testing case: " (clojure.string/join
+                                          ", "
+                                          (map #(str %1 ": " %2)
+                                               '~case-versions
+                                               (list ~@case-versions))))
            ~@body)))))
 
 (defmacro without-jmx


mruzicka:
Note that we've applied a similar fix in ```pcp-broker``` some time ago already: https://github.com/puppetlabs/pcp-broker/commit/839a77204ca043d25b5b3ae7f21c7cd42dc83a8e

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -33,7 +33,11 @@
     `(do ~@body)
     (let [case-versions (remove keyword? (take-nth 2 bindings))]
       `(doseq ~bindings
-         (testing (str "Testing case " '~case-versions)
+         (testing (str "Testing case: " (clojure.string/join
+                                          ", "
+                                          (map #(str %1 ": " %2)
+                                               '~case-versions
+                                               (list ~@case-versions))))
```
rbrw: Thanks, and seems like a good idea, though I might lean toward using normal let binding style for the values (produced via prn so there's no possibility of ambiguity), and making the prose chainable, since testing is intended to be nested.  So given
```clojure
(deftest some-test
  (testing "check something"
    (dotestseq [x [1 2 3]
                 y [:a :b]]
                (is (= 13 [x y])))))
```
we'd see something like this
```clojure
FAIL in (some-test) (core_test.clj:31)
check something with [[x 3] [y :b]]
expected: (= 13 [x y])
  actual: (not (= 13 [3 :b]))
```
given a definition like this
```clojure
(defmacro dotestseq [bindings & body]
  (if-not (seq bindings)
    `(do ~@body)
    (let [case-versions (remove keyword? (take-nth 2 bindings))]
      `(doseq ~bindings
         (testing (str "with " (pr-str (mapv vector
                                             '~case-versions
                                             ~(vec case-versions))))
           ~@body)))))
```
Though I'm not positive I got the expansion quite right, evaluation-wise.

mruzicka: :+1: on the chainable prose
but I'm not sure whether the let binding style is achievable, consider using destructuring:
```clojure
(deftest some-test
  (testing "check something"
    (dotestseq [{:keys [x y]} [{:x 10 :y 20}]
                z [:a :b]]
               (is (= 13 [x y z])))))
```
then you get something like this:
```
FAIL in (some-test) (testutils.clj:45)
check something with [[{:keys [a b]} {:keys [10 20]}] [[c d] [3 4]]]
expected: (= 13 [a b c d])
  actual: (not (= 13 [10 20 3 4]))
```
which doesn't really look like a let binding

mruzicka: perhaps the chainable prose with a ```(pr-str )``` hardened variant of the original implementation:
```clojure
(defmacro dotestseq [bindings & body]
  (if-not (seq bindings)
    `(do ~@body)
    (let [case-versions (remove keyword? (take-nth 2 bindings))]
      `(doseq ~bindings
         (testing (str "with " (clojure.string/join
                                 ", "
                                 (map #(str (pr-str %1) ": " (pr-str %2))
                                      '~case-versions
                                      (list ~@case-versions))))
           ~@body)))))
```
which produces this output:
```
FAIL in (some-test) (testutils.clj:47)
check something with {:keys [a b]}: {:keys [10 20]}, [c d]: [3 4]
expected: (= 13 [a b c d])
  actual: (not (= 13 [10 20 3 4]))
```

@rbrw what are your thoughts?

rbrw: @mruzicka Hmm, right, good point -- how about something like this?
```clojure
(defmacro dotestseq [bindings & body]
  (if-not (seq bindings)
    `(do ~@body)
    (let [binding-names (->> (take-nth 2 bindings)
                             flatten
                             (filter symbol?))]
      `(doseq ~bindings
         (testing (str "with " (pr-str (mapv vector
                                             '~binding-names
                                             ~(vec binding-names))))
           ~@body)))))
```

rbrw: @mruzicka hey, just wanted to ping you to see what you thought about the latest incarnation.

kbarber:
@mruzicka I think there are pending comments on this PR, can you take a look?

kbarber:
@mruzicka ping

kbarber:
This has been pending for quite a while with no update, I'm going to close it. Please re-open when ready.

PR #1927, opened 2016-04-08
(maint) Remove Debian 6 from pipelines

melissa:
Debian 6 has reached EOL and should be removed from all build/test
pipelines


senior:
@melissa can you target this at stable? I think we're good to go to merge it otherwise.

senior:
@melissa reopend as https://github.com/puppetlabs/puppetdb/pull/1932 against stable

melissa:
@senior thank you so much!

PR #1929, opened 2016-04-13
(PDB-2617) stale-nodes: use newest report/catalog

rbrw:
Don't consider a node stale just because any report/catalog is too old;
require that the newest report/catalog be too old.

### DIFF
--- a/test/puppetlabs/puppetdb/scf/storage_test.clj
+++ b/test/puppetlabs/puppetdb/scf/storage_test.clj
@@ -459,71 +459,82 @@
 (def certname (:certname catalog))
 (def current-time (str (now)))
 
+(defmacro with-historical-catalogs-enabled [limit & body]
+  `(let [orig-limit# @historical-catalogs-limit
+         orig-jsonb-setting# @store-catalogs-jsonb-columns?]
+     (try
+       (reset! historical-catalogs-limit ~limit)
+       (reset! store-catalogs-jsonb-columns? true)
+       ~@body
+       (finally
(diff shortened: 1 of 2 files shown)


wkalt:
It seems like this will never expire nodes for which no catalog exists?

senior:
@rbrw @wkalt I think what's there should work. Seems like we should have a test for the original bug and a test for this as well though

rbrw:
@senior agreed, I'll see about adding some.

wkalt:
@senior @rbrw yeah my comment was wrong. this looks fine.

ajroetker:
👍

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -1229,14 +1240,73 @@
...
-                     (assoc :environment "ENV2")
-                     (assoc :end_time (now))
-                     (assoc :producer_timestamp (now)))]
-      (store-example-report! report (now))
-      (is (= (stale-nodes (-> 1 days ago)) [])))))
+(deftest stale-nodes-behavior-for-reports
+  (let [report-at #(-> (:basic reports)
+                       (assoc :environment "ENV2")
+                       (assoc :end_time %)
+                       (assoc :producer_timestamp %))
```
ajroetker: Minor nitpick but you only need 1 `assoc` call here

rbrw: Uhh, right...

PR #1930, opened 2016-04-14
(PDB-2625) Upgrade raynes.fs to 1.4.6

nicklewis:
This harmonizes the version of raynes.fs with the one used in other
projects.

This requires upgrading kitchensink and trapperkeeper as well, because
raynes.fs 1.4.6 is backward-incompatible with 1.4.5. Specifically, the
absolute-path and normalized-path functions are removed. Those functions
have been added to kitchensink 1.3.0, so we now use the kitchensink
version. Because trapperkeeper 1.2.0 also depended on fs 1.4.5 (and is
thus broken by upgrading it), it's also now version 1.3.1.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -9,9 +9,9 @@
    :password :env/nexus_jenkins_password
    :sign-releases false})
 
-(def tk-version "1.2.0")
+(def tk-version "1.3.1")
 (def tk-jetty9-version "1.5.0")
-(def ks-version "1.2.0")
+(def ks-version "1.3.0")
 (def tk-status-version "0.3.1")
 (def i18n-version "0.2.1")
 
@@ -46,7 +46,7 @@
                  [metrics-clojure "2.6.1" :exclusions [org.clojure/clojure org.slf4j/slf4j-api]]
                  [clj-time "0.11.0"]
                  ;; Filesystem utilities
-                 [me.raynes/fs "1.4.5"]
+                 [me.raynes/fs "1.4.6"]
                  [org.apache.commons/commons-lang3 "3.3.1"]
                  ;; Version information
                  [puppetlabs/dujour-version-check "0.1.3"]
--- a/src/puppetlabs/puppetdb/cli/benchmark.clj
+++ b/src/puppetlabs/puppetdb/cli/benchmark.clj
@@ -412,7 +412,7 @@
   (count [this] @queue-count))
 
 (defn amq-broker-buffer [broker-name endpoint-name]
-  (let [dir (fs/absolute-path broker-name)
+  (let [dir (kitchensink/absolute-path broker-name)
         conn-str (str "vm://" broker-name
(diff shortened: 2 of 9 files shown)


nicklewis:
Should go along with puppetlabs/pe-puppetdb-extensions#180.

ajroetker:
@nicklewis looks like unit tests are broken https://travis-ci.org/puppetlabs/puppetdb/jobs/123131263

nicklewis:
Ughhh I guess there must be other breaking changes. Will look into it...

cprice404:
@nicklewis this might help:

https://github.com/puppetlabs/pe-file-sync/pull/207/files

The one that bit me there is that `list-dir` started returning file objects instead of strings, so I needed to add calls to `fs/base-name`, and/or remove calls to `fs/file`.

nicklewis:
@cprice404 Thanks. I was considering that change as a possible culprit, but that function doesn't appear to be used except in different tests than the ones that are failing. Even then, it's only checked for emptiness.

I can't find any other functions whose behavior seems different, so I'm wondering if it could be in some dependency that's using raynes.fs...

nicklewis:
Alright I figured it out and pushed up a fix.

`kitchensink/normalized-path` does *not* do the same thing that `fs/normalized-path` did, even though `kitchensink/absolute-path` *does* do the same thing that `fs/absolute-path` did. So the correct way to upgrade from fs 1.4.5 to fs 1.4.6 is to change `fs/absolute-path` to `kitchensink/absolute-path` and to change `fs/normalized-path` to `fs/normalized`.

cprice404:
@nicklewis ugh.  How does normalized-path differ between ks and old fs?

cprice404:
oh.  so in the old fs, `absolute-path` returned a String and `normalized-path` returned a `File`.  That seems dumb.

nicklewis:
@cprice404 Yeah. I think the new behaviors in fs and kitchensink make sense; it just means updating uses of the two functions is different.

cprice404:
gtk, will remember this in case it trips anyone else up.  thx

ajroetker:
We need to make sure we're HUPable before merging this.

kbarber:
@ajroetker I'm not sure I understand the HUP'able requirement, is this because we're bumping to latest TK in this patch also? What work needs to be done to get this past the line, do we need an extra patch or a set of tests, afaik we've done no work on HUP support as yet.

ajroetker:
@kbarber precisely, this bumps the TK version to HUP, so I thought we had agreed in standup that this patch was blocked by that ticket.

kbarber:
Cool, makes sense. This work is tracked in: https://tickets.puppetlabs.com/browse/PDB-2546.

wkalt:
I've merged this into my branch for HUP support in PDB-2546, since that depends on this bump. It'll get merged when my patch goes in.

PR #1931, opened 2016-04-15
(maint) Add puppet-agent libs to load path

melissa:
When we build puppetdb, we require access to the facter library.
Unfortunately, this is under the Ruby Lib path installed with
puppet-agent. System ruby, which we are currently using to build,
doesn't have access to this libdir. This appends the puppet-agent
vendor_lib path to the search path so system ruby can find Facter.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,8 @@
+# We need access to the ruby libraries installed via puppet-agent
+# More specifically, we need access to Facter to satisfy the
+# dependency on line 8
+$LOAD_PATH.push("/opt/puppetlabs/puppet/lib/ruby/vendor_ruby")
+
 require 'rake'
 require 'erb'
 require 'facter'


joshcooper:
You might run into gem incompatibility issues if system ruby tries to use gems installed by AIO ruby (because ruby doesn't guarantee gem compatibility between different ruby versions). Since the puppetdb `Gemfile` already contains `facter`, can you run the build task using `bundle exec` so that the system ruby can load facter?

melissa:
So, we don't need facter for building. I *think* it's just for the tests. I'll dig a little further and see what I can come up with

kbarber:
@melissa I think this: https://github.com/puppetlabs/puppetdb/blob/2.3.x/Rakefile and this: https://github.com/puppetlabs/puppetdb/blob/2.3.x/tasks/install.rake are consumers of facter data, its used to make decisions for installation and build afaik. For example libexec_dir is used in ext/templates in all kinds of places to hard code the libexec path. I think its fairly intermingled in build anyway ... BTW if you need my help just shout, I'm happy to try and unravel this for you, my memory of all this is just absent after living in ezbake bliss.

melissa:
It does look like we use facter during the build process. We're using to populate paths in the packaging artifacts

melissa:
I think https://github.com/puppetlabs/puppetlabs-modules/pull/5764 will be a better solution to this problem

PR #1933, opened 2016-04-20
(PDB-2620) Remove runtime validation of reports in storage

senior:
We currently validate reports in puppetdb/command.clj and then
re-validate those reports twice in puppetdb/scf/storage.clj. This second
and third validation is very CPU intensive. Some load testing has
revealed as much as 30% of the CPU usage was dedicated to this purpose.

This inefficientcy is mainly spent on unchanged resources and would
likely only be observed on catalogs that have a large number of
unchanged resources.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -132,7 +132,10 @@
                                   [org.clojure/test.check "0.5.9"]
                                   [environ "1.0.2"]
                                   [org.clojure/tools.cli "0.3.3"] ; prevents dependency clash caused by lein-cloverage
-                                  [riddley "0.1.12"]]}
+                                  [riddley "0.1.12"]]
+                   :injections [(do
+                                  (require 'schema.core)
+                                  (schema.core/set-fn-validation! true))]}
              :ezbake {:dependencies ^:replace [[puppetlabs/puppetdb ~pdb-version]
                                                [org.clojure/tools.nrepl "0.2.3"]]
                       :name "puppetdb"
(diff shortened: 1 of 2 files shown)


senior:
I think we have a similar problem with catalogs, but the solution will be a little bit different as we don't do upfront validation of the catalog like we do for reports. I created [PDB-2638](https://tickets.puppetlabs.com/browse/PDB-2638) to fix this.

kbarber:
@senior as discussed is there a better way to set this globally? Like using the set-fn-validation! function or something of that nature? I'd just be interested in enabling this globally for all tests without having to add the fixture to each namespace, seems like this would be easily overlooked.

project.clj
```diff
@@ -132,7 +132,10 @@
                                   [org.clojure/test.check "0.5.9"]
                                   [environ "1.0.2"]
                                   [org.clojure/tools.cli "0.3.3"] ; prevents dependency clash caused by lein-cloverage
-                                  [riddley "0.1.12"]]}
+                                  [riddley "0.1.12"]]
+                   :injections [(do
+                                  (require 'schema.core)
+                                  (schema.core/set-fn-validation! true))]}
```
senior: This is my hack to get function validation enabled on all test namespace. I'm not a fan of this approach, but it's better than requiring every test namespace to opt-in to validation.

senior:
@kbarber address your concerns in a second commit, not great, but it looks like it works for what we're wanting

ajroetker:
Might want to update https://github.com/senior/puppetdb/blob/ticket/stable/pdb-2620-remove-runtime-report-validation/src/puppetlabs/puppetdb/scf/storage.clj#L1235 as well, that also does schema validation.

PR #1934, opened 2016-04-21
(PDB-2527) allow latest-report? to be used in legacy events queries

wkalt:
* adds an index on certnames(latest_report_id)
* fixes a bug preventing latest_report? usage with the distinct_resources parameter
* moves query logic for unioned aggregate-event-counts queries from the
  aggregations to the CTE

### DIFF
--- a/src/puppetlabs/puppetdb/query/events.clj
+++ b/src/puppetlabs/puppetdb/query/events.clj
@@ -49,75 +49,77 @@
 
 (defn with-latest-events
   "CTE to wrap unioned queries when distinct_resources is used"
-  [query]
-  (str
-   "WITH latest_events AS
-      (SELECT certname,
-              configuration_version,
-              start_time as run_start_time,
-              end_time as run_end_time,
-              receive_time as report_receive_time,
-              hash,
-              status,
-              distinct_events.latest_timestamp AS timestamp,
-              distinct_events.resource_type AS resource_type,
-              distinct_events.resource_title AS resource_title,
-              distinct_events.property as property,
-              new_value,
-              old_value,
-              message,
-              file,
-              line,
-              containment_path,
-              containing_class,
-              environment
-      FROM
-      (SELECT certname_id,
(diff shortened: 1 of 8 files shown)


wkalt:
ping @kbrezina

test/puppetlabs/puppetdb/http/aggregate_event_counts_test.clj
```diff
@@ -91,7 +91,48 @@
...
+         ["<=" "run_end_time" current-time-str]))
+
+  (store-example-report! (:basic2 reports) (coerce/to-string (now)))
+
+  (are [result query] (= result
+                         (query-result method endpoint query
+                                       {:summarize_by "resource"
+                                        :distinct_resources true
+                                        :distinct_start_time 0
+                                        :distinct_end_time (now)}))
```
kbrezina: Do `:distinct_start_time` and `:distinct_end_time` parameters need to be provided even if `["=" "latest_report?" true]` query is used?

wkalt: @kbrezina yes

src/puppetlabs/puppetdb/query/events.clj
```diff
@@ -113,10 +122,8 @@
   (let [where-clause (if where (format "WHERE %s" where) "")]
```
kbarber: where-clause is no longer used

test/puppetlabs/puppetdb/http/aggregate_event_counts_test.clj
dmcpl: Hi @wkalt , if we are going to have to specify start/end times for latest_report? query, is it acceptable then just to use `0` and `now` as the time range as you have done here without fear of fiery explosion?

wkalt: @dmcpl absolutely, assuming you mean on the client side.

ajroetker:
:+1:

PR #1936, opened 2016-04-21
(maint) Update to latest ezbake release

underscorgan:


kbarber:
Hey @underscorgan, if we plan on having the changes in ezbake 0.3.24 apply to all versions 'still in flight and being built' then perhaps this should be targeted at 3.2.x and rolled up? Right now we still actively have jobs that build: 2.3.x, 3.2.x and 4.0.x - but 2.3.x is not ezbake.

underscorgan:
retargeted in https://github.com/puppetlabs/puppetdb/pull/1937

PR #1938, opened 2016-04-21
Add a basic checking of the "query" parameter in the function puppetdb_query()

flaf:
To be compatible with Puppet 4.0.0 where recursive types
are not possible, the type checked here is not accurate.
Ideally, the type should be:

  Variant[String[1], StringTree]

where StringTree is defined (recursively) by:

  StringTree = Array[Variant[String, StringTree]]


wkalt:
@pljenkinsro okay to test

wkalt:
@flaf thanks again!

PR #1939, opened 2016-04-25
(PDB-2546) add HUP support to PDB

wkalt:
Add HUP support to PDB

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -9,9 +9,9 @@
    :password :env/nexus_jenkins_password
    :sign-releases false})
 
-(def tk-version "1.2.0")
+(def tk-version "1.3.1")
 (def tk-jetty9-version "1.5.0")
-(def ks-version "1.2.0")
+(def ks-version "1.3.0")
 (def tk-status-version "0.3.1")
 (def i18n-version "0.2.2")
 
@@ -46,7 +46,7 @@
                  [metrics-clojure "2.6.1" :exclusions [org.clojure/clojure org.slf4j/slf4j-api]]
                  [clj-time "0.11.0"]
                  ;; Filesystem utilities
-                 [me.raynes/fs "1.4.5"]
+                 [me.raynes/fs "1.4.6"]
                  [org.apache.commons/commons-lang3 "3.3.1"]
                  ;; Version information
                  [puppetlabs/dujour-version-check "0.1.3"]
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -47,7 +47,7 @@
             [clojure.tools.logging :as log]
             [compojure.core :as compojure]
             [metrics.reporters.jmx :as jmx-reporter]
-            [overtone.at-at :refer [mk-pool interspaced]]
+            [overtone.at-at :refer [mk-pool interspaced stop-and-reset-pool!]]
             [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.command.constants :refer [command-names]]
@@ -174,6 +174,8 @@
     (.close ds))
   (when-let [ds (get-in context [:shared-globals :scf-read-db :datasource])]
     (.close ds))
+  (when-let [pool (:job-pool context)]
+    (stop-and-reset-pool! pool))
   context)
 
 (defn- transfer-old-messages! [mq-endpoint]
@@ -254,8 +256,7 @@
     (when-let [v (version/version)]
       (log/infof "PuppetDB version %s" v))
 
-    (when (init-with-db database config)
-      (sutils/async-vacuum-analyze write-db))
+    (init-with-db database config)
 
     (let [population-registry (get-in metrics/metrics-registries [:population :registry])]
       (pop/initialize-population-metrics! population-registry read-db))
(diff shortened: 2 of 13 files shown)


ajroetker:
@wkalt does this need a corresponding PR on PE as well (I know sync also uses at-at)?

ajroetker:
Lgtm other than the PE comment, tested with YourKit 2016.2 and `for i in {1..1000}; do kill -hup <pid>; sleep 1; done`

ajroetker:
Tested again while under load (from the benchmark tool) and it looks like there are some NPE's and then PDB crashes
```
java.lang.NullPointerException: null
	at org.apache.activemq.jms.pool.PooledConnection.createSession(PooledConnection.java:167) ~[activemq-jms-pool-5.13.2.jar:5.13.2]
	at sun.reflect.GeneratedMethodAccessor43.invoke(Unknown Source) ~[na:na]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_60]
	at java.lang.reflect.Method.invoke(Method.java:497) ~[na:1.8.0_60]
	at clojure.lang.Reflector.invokeMatchingMethod(Reflector.java:93) ~[clojure-1.8.0.jar:na]
	at clojure.lang.Reflector.invokeInstanceMethod(Reflector.java:28) ~[clojure-1.8.0.jar:na]
	at puppetlabs.puppetdb.mq$send_message_BANG_$fn__33418.invoke(mq.clj:251) ~[na:na]
	at puppetlabs.puppetdb.mq.proxy$java.lang.Object$Callable$7da976d4.call(Unknown Source) ~[na:na]
	at com.codahale.metrics.Timer.time(Timer.java:101) ~[metrics-core-3.1.2.jar:3.1.2]
	at puppetlabs.puppetdb.mq$send_message_BANG_.invokeStatic(mq.clj:250) ~[na:na]
	at puppetlabs.puppetdb.mq$send_message_BANG_.invoke(mq.clj:244) ~[na:na]
```

wkalt:
this still has an issue

project.clj
```diff
@@ -9,9 +9,9 @@
    :password :env/nexus_jenkins_password
    :sign-releases false})
 
-(def tk-version "1.2.0")
+(def tk-version "1.3.1")
```
ajroetker: Might as well bump these to 1.4 now that it's out

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -291,17 +291,8 @@
...
-        (jdbc/do-commands-outside-txn "vacuum (analyze, verbose)")
-        (catch java.sql.SQLException e
-          (log/error e "Caught SQLException during migration")
-          (when-let [next (.getNextException e)]
-            (log/error next "Unravelled exception")))))))
+(defn analyze-small-tables
+  [small-tables]
+  (log/info "Analyzing small tables")
+  (apply jdbc/do-commands-outside-txn
+         (map #(str "analyze " %) small-tables)))
```
rbrw: Tempted to change this to "analyze-tables", since it's not small-table specific, and put the log/info in migrate.clj above the call.

rbrw: Also, not critical, but I wondered if there might be a way to just probe pg for a list of the small tables, so we don't have to remember to maintain the value by hand.

wkalt: to your second question, not sure. the potential trouble would be that asking for an accurate count will be really expensive, whereas relying on statistics presupposes that the statistics are correct, which is what the analyze itself is supposed to ensure.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1153,8 +1154,8 @@
           (doseq [[version migration] pending]
             (log/infof "Applying database migration version %d" version)
             (sql-or-die (fn [] (migration) (record-migration! version)))))
-        true
-        )
+        (sutils/analyze-small-tables small-tables)
```
senior: How does this fit into the HUP change?

project.clj
senior: I'd hold off. I think there's enough variables here, seems useful to test this on it's own before another version bump. Also we need to make sure the other PE projects are bumping to 1.4.0 for the next release before we bump.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -174,6 +174,8 @@
     (.close ds))
   (when-let [ds (get-in context [:shared-globals :scf-read-db :datasource])]
     (.close ds))
+  (when-let [pool (:job-pool context)]
+    (stop-and-reset-pool! pool))
```
senior: I think this will probably need to be applied to the sync code in extensions as well

senior:
@wkalt is there anything we need to be doing to HUP the core.async threadpools?

src/puppetlabs/puppetdb/mq.clj
```diff
@@ -243,7 +243,7 @@
 
 (defn send-message!
   "Sends message with attributes via a new session (sessions are
-  single-threaded)."
+   single-threaded)."
```
rbrw: Hmm, do we have a convention here -- seems like our code has both, but looks like by default clojure-mode (at least) uses the first style (via M-q).

src/puppetlabs/puppetdb/scf/storage_utils.clj
rbrw: So *if* we pursued this, I'd wondered if we might just be able to ask for a list of tables, and then ask each table whether it has more than N rows (without any sorting) -- perhaps that'd be fast?  i.e. hand-waving "select count(*) from foo limit 50;"

PR #1942, opened 2016-04-26
(PDB-2477) Drop constraints while sweeping reports/nodes

rbrw:
Drop contraints while sweeping reports (and reestablish them afterward)
to avoid triggering per-report checks during delete.

This appears to substantially decrease the sweep time, and we believe it
should be safe because all of the operations involved should be covered
by PostgreSQL MVCC (i.e. guarded by the transaction).

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -23,6 +23,7 @@
             [puppetlabs.puppetdb.reports :as reports]
             [puppetlabs.puppetdb.facts :as facts :refer [facts-schema]]
             [puppetlabs.kitchensink.core :as kitchensink]
+            [puppetlabs.puppetdb.scf.migrate :as migrate]
             [puppetlabs.puppetdb.scf.storage-utils :as sutils]
             [puppetlabs.puppetdb.jdbc :as jdbc]
             [com.rpl.specter :as sp]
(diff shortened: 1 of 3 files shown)


npwalker:
@rbrw Did you get any numbers on how long it takes to recreate the constraints?  I assume this is faster but I think when you make a constraint it has to recheck the conditions.  

As for a possible performance improvement, if you broke this into 3 transactions I think it would help with blocking out user activity.  

You can grab the information into the temp table outside of a transaction and not block user activity.  Then, inside a transaction, remove the certnames constraint and update the latest_report field ( although I wonder if we should even be doing this at all ) then end the transaction.  

Do the same for resource_events and finally delete from reports.  

Since the delete from reports will take the longest I feel it's less than ideal to block activity on certnames and resource_events while that's happening.  Now if for some reason something went wrong we could get into a state where the first two deletes happened but not the one from reports.  I don't see anything wrong with that possibility.

rbrw:
@npwalker I didn't test the restore constraint times independently, but the overall sweep time is much faster with this approach.  For example, here on a big data set, the node sweep changes I just added take the duration of a sweep that deletes all nodes down by a factor of 3 to 4.

With respect to breaking up the transaction, I'm inclined to hold off on that until we see where these improvements leave us.  I suspect that some of the ways we might try to break this up might not improve performance (e.g. the initial select shouldn't be likely to cause significant conflicts in PostgreSQL since it's read-only?), and breaking up the transaction up is likely to add complexity/risk with respect to avoiding races, etc.

In any case, I'd first be inclined to finish adding the other bits we've planned, i.e. an endpoint trigger that provides external control over the sweep timing.

That said, if we *do* decide we want to attempt further improvements, I might be tempted to evaluate something a bit different, i.e. batching.  For example, we could probably allow the specification of a "batch size", and only select/process that many items per invocation.  That, when coupled with the endpoint trigger (perhaps with an optional batch size parameter), might be easier to trust.

senior:

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -235,13 +236,107 @@
    ["SELECT expired FROM certnames WHERE certname=?" certname]
    (comp :expired first sql/result-set-seq)))
 
-(defn purge-deactivated-and-expired-nodes!
+(defn purge-deactivated-and-expired-nodes-old!
```
wkalt: seems like this is unused now -- maybe go to "purge-deactivated-nodes!*" and "purge-deactivated-nodes!" ?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -235,13 +236,107 @@
...
+             on delete cascade"
+         "alter table latest_catalogs add constraint latest_catalogs_catalog_id_fkey
+           foreign key (catalog_id) references catalogs(id) on delete cascade"
+
+         "drop table obsolete_nodes"])))))
+
+(defn- full-sql-exception-msg [ex]
+  (apply str (take-while identity (iterate #(.getNextException %) ex))))
+
+(defn purge-deactivated-and-expired-nodes! [x]
```
wkalt: s/x/timestamp

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: Yeah, that's not even supposed to be in there - was using it for before/after testing.

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: This function shouldn't even be there either.

PR #1943, opened 2016-04-27
(PDB-2621) Force all textual errors to be text/plain

kbarber:
This patch fixes the cases where text errors were not returning the correct
text/plain Content-Type header by fixing error-response to do this, and forcing
all relevant code to use this function instead of trying to create errors
themeselves.

This also adds tests to ensure errors return the correct content-type.


senior:
I'll test this one

senior:
:+1:

PR #1944, opened 2016-04-27
(PDB-2597) Update release notes for 4.0.1 and 4.0.2

rbrw:


documentation/release_notes.markdown
```diff
@@ -14,6 +14,104 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+```
+
+* Nodes should no longer be incorrectly eligible for expiration in an
+  unlikely corner case.
+  ([PDB-2617](https://tickets.puppetlabs.com/browse/PDB-2617))
+
+* PuppetDB has migrated to ezbake 0.3.24.
+  ([PDB-2645](https://tickets.puppetlabs.com/browse/PDB-2645))
+
+* Support for Debian 6 is no longer being tested.
```
senior: Maybe we should instead say Debian 6 has been EOL'd and as such is no longer being tested?

documentation/release_notes.markdown
```diff
@@ -14,6 +14,104 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+
+* An unintentional limit on the number of resources in an internal
+  resources-exists? call has been fixed.  The previous arrangement
+  could cause a command involving a very large number of resources to
+  fail.  ([PDB-2548](https://tickets.puppetlabs.com/browse/PDB-2548))
+
+* The indirector tests now have access to the environment.
+  ([PDB-2577](https://tickets.puppetlabs.com/browse/PDB-2577))
+
+* The =-after? test utility has been changed to is-equal-after.
```
senior: Seems like we can remove this, it's a test only utility function that we don't ship

PR #1945, opened 2016-04-29
(maint) Add maintainers section to README.md

kbarber:
I also fixed the old puppetlabs.com address for the docs link.


README.md
```diff
@@ -1,7 +1,13 @@
...
-[docs]: https://docs.puppetlabs.com/puppetdb/latest
+[docs]: https://docs.puppet.com/puppetdb/latest
 
 PuppetDB is the fast, scalable, and reliable data warehouse for Puppet. It caches data generated by Puppet, and gives you advanced features at awesome speed with a powerful API.
 
 For documentation on this product, consult the [latest documentation][docs].
+
+## Maintenance
+
+Maintainers: Ryan Senior <ryan.senior@puppet.com>, Wyatt Alt <wyatt@puppet.com>, Rob Browning <rlb@puppet.com>, AJ Roetker <andrew.roetker@puppet.com>, Russell Mull <russell.mull@puppet.com>, Ken Barber <ken@puppet.com>
```
ajroetker: What about putting our alias here? puppetdb-team@puppet.com

kbarber: I think there was some debate about putting aliases in an internal email thread at the moment, we can wait to see where that goes or chime in I think ... also, I'm not sure how public-ready the alias is, we might need to ensure we get moderation for example (as a moderator of puppet-users, this becomes a thing eventually).

wkalt:
@kbarber should this be against stable?

README.md
senior: @lutter This doesn't seem like the right thing for a Puppet product vs. a shared library (public or internal). In product terms, people working on it is more fluid. Is puppet-users or some other generic public forum a better address for this?

lutter: Agreed - this is a bad fit for projects with well-established practices. How about you add a sentence to this section along the lines "For questions/problems/dev help email <our public mailing list>" and then as maintainers list one or two of the people working on PuppetDB. The reason for listing people is to give others a way to get in touch with a human if they can't get attention through other means. In that sense, the maintainer duties are mostly around redirecting and connecting people with the right devs for whatever their problems are.

senior: I talked with @lutter a bit on this. It seems like the best approach here is to point people to puppet-users and puppet-dev (with a short description of what discussions go where).

senior:
Handled in https://github.com/puppetlabs/puppetdb/pull/1959

PR #1946, opened 2016-04-29
(PDB-2501) Add the ability to test against puppetserver nightlies

melissa:


melissa:
I believe this is actually sufficient. We should be building puppetserver nightlies for all platforms, but on the off chance we don't, we can always fall back to the packages available in the public repos. Should I add a line that includes puppet-agent nightlies as well?

melissa:
`Get:12 http://nightlies.puppetlabs.com/puppetserver-latest/repos/apt/trusty/ trusty/PC1 puppetserver all 2.4.0.master-0.1SNAPSHOT.2016.04.28T0317puppetlabs1 [32.8 MB]`
^^ installing puppetserver from the enabled nightlies repo instead of apt.puppetlabs.com

melissa:
cc/ @wkalt @senior

wkalt:

acceptance/helper.rb
```diff
@@ -17,7 +17,7 @@ module PuppetDBExtensions
   # by default.
   LeinCommandPrefix = "cd #{GitReposDir}/puppetdb; LEIN_ROOT=true LEIN_SNAPSHOTS_IN_RELEASE=true PATH=/opt/puppetlabs/puppet/bin:$PATH"
 
-  def self.initialize_test_config(options, os_families)
+  def self.initialize_test_config(options, os_families, use_nightlies)
```
melissa: @wkalt I'm thinking this might make more sense if we add `use_nightlies` to the options hash. I'm digging more into that now

melissa:
@wkalt done! This uses an environment variable to decide if it's going to use nightlies or not. It defaults to not. All I need to do now is update ci-job-configs! But that change isn't blocking this

wkalt:

wkalt:
This looks good to me, will merge when the pipeline is clear.

PR #1947, opened 2016-05-03
(PDB-2663) Bump ezbake version to include xenial fixes

kbarber:
This bumpes ezbake so we can fix PuppetDB on xenial pulling in java 9.


kbarber:
Also see: https://github.com/puppetlabs/pe-puppetdb-extensions/pull/183 for the corresponding change in PE.

PR #1948, opened 2016-05-03
(PE-15026) Allow configuration of the ActiveMQ Broker's MemoryLimit

ajroetker:
the default 1GB.

### DIFF
--- a/config.sample.ini
+++ b/config.sample.ini
@@ -52,6 +52,9 @@ vardir = /var/lib/puppetdb
 # Maximum amount of disk space (in MB) to allow for ActiveMQ temporary message storage
 # temp-usage = 51200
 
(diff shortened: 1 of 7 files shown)


ajroetker:
@npwalker ping

npwalker:
@ajroetker is it within scope of this PR to make the default scale as a percentage of the total heap? 

Looks like you're putting this on 3.1 so maybe that's for the PR to master?

ajroetker:
@npwalker Yeah I think that would either be a PR for stable or we should hold off for our "future MQ" work planned for master. Also FWIW I would make it a function of Total Heap and Command-Processing Threads (Consumers).

ajroetker:
Retargeted at 3.2.x here https://github.com/puppetlabs/puppetdb/pull/1949

PR #1949, opened 2016-05-03
(PE-15026) Allow configuration of the ActiveMQ Broker's MemoryLimit

ajroetker:
the default 1GB.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -621,6 +621,17 @@ PuppetDB's ActiveMQ can use for persistent message storage.
 This setting sets the maximum amount of space in megabytes that
 PuppetDB's ActiveMQ can use for temporary message storage.
 
+### `memory-usage`
+
+This setting sets the maximum amount of memory in megabytes available for
+PuppetDB's ActiveMQ Broker.
+
+**Warning** Setting this value too high (such that memory-usage exceeds the size
+of the heap) can cause out of memory (OOM) errors. ActiveMQ does not treat this
+as a hard limit. In testing, we've seen it use up to `125%` of the specified
+value, and overall memory usage will also be affected by the `max-command-size`
+and `threads` parameters.
+
 ### `max-frame-size`
 
 This setting sets the maximum frame size for persisted activemq messages
--- a/src/puppetlabs/puppetdb/mq.clj
+++ b/src/puppetlabs/puppetdb/mq.clj
@@ -1,7 +1,7 @@
 (ns puppetlabs.puppetdb.mq
   (:import [org.apache.activemq.broker BrokerService]
            [org.apache.activemq ScheduledMessage]
-           [org.apache.activemq.usage SystemUsage]
+           [org.apache.activemq.usage SystemUsage MemoryUsage]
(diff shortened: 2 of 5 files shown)


ajroetker:
@rbrw @senior @wkalt added a note in the docs about setting this value too high and removed it from the example configs.

documentation/configure.markdown
```diff
@@ -621,6 +621,17 @@ PuppetDB's ActiveMQ can use for persistent message storage.
 This setting sets the maximum amount of space in megabytes that
 PuppetDB's ActiveMQ can use for temporary message storage.
 
+### `memory-usage`
+
+This setting sets the maximum amount of memory in megabytes available for
+PuppetDB's ActiveMQ Broker.
+
+**Warning** Setting this value to high (such that memory-usage exceeds the size
```
rbrw: "too"

documentation/configure.markdown
```diff
@@ -621,6 +621,17 @@ PuppetDB's ActiveMQ can use for persistent message storage.
...
+of the heap) can cause out of memory (OOM) errors. ActiveMQ's memory may exceed
```
rbrw: Wondered about something like "ActiveMQ also doesn't appear to treat this as a hard limit.  In testing, we've seen it use up to 125% of the specified value, and overall memory usage will also be affected by the `max-command-size` and `threads` parameters."

src/puppetlabs/puppetdb/mq.clj
```diff
@@ -45,6 +45,17 @@
...
 
+(defn- set-memory-usage!
+  "Configures the `MemoryUsage` setting for an instance of `BrokerService`.
+
+  `broker`     - the `BrokerService` to configure
+  `megabytes ` - the maximum amount of memory usage to allow for the BrokeService,
+  or `nil` to use the default value of 1GB.
+
+  Returns the (potentially modified) `broker` object."
+  [broker megabytes]
```
rbrw: Happy to leave this as is, though I've been tending away from javadoc-style docstrings in most cases, tending toward Clojure's style, i.e. tending toward strings like "Sets the maximum number of megabytes that the broker is allowed to use.  If megabytes is nil, then the broker will revert to its default limit (currently 1GB).  Returns the updated broker."

ajroetker: This is a copy paste from the docstring for temp-usage, store-usage above

documentation/configure.markdown
```diff
@@ -621,6 +621,17 @@ PuppetDB's ActiveMQ can use for persistent message storage.
 This setting sets the maximum amount of space in megabytes that
 PuppetDB's ActiveMQ can use for temporary message storage.
 
+### `memory-usage`
+
+This setting sets the maximum amount of memory in megabytes available for
+PuppetDB's ActiveMQ Broker.
+
+**Warning** Setting this value too high (such that memory-usage exceeds the size
+of the heap) can cause out of memory (OOM) errors. ActiveMQ does treat this as a
```
rbrw: doesn't?

rbrw:
Overall, LGTM.

PR #1951, opened 2016-05-05
(maint) install released repos even when testing nightlies

wkalt:
Upgrade tests will fail if the only installed repos are the nightly repos.

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -98,8 +98,8 @@ module PuppetDBExtensions
     puppetdb_git_ref = get_option_value(options[:puppetdb_git_ref],
       nil, "git revision of puppetdb to test against", "REF", nil)
 
-    use_nightlies = get_option_value(options[:use_nightlies], [:true, :false],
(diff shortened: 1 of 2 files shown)


acceptance/config/ec2-west-el6-64mda-ubuntu1204-64a.cfg
```diff
@@ -0,0 +1,28 @@
+HOSTS:
+  el6-64-1:
```
kbarber: as discussed, the old el5 stuff was used by the 3 node dup_collected_resources.rb, we'll need a new 3 node combination. I'd suggest having 2 el6 nodes or something. Maybe an el6 and el7 or some such, not too fussed.

acceptance/config/ec2-west-el6-64mda-ubuntu1204-64a.cfg
```diff
@@ -0,0 +1,28 @@
+HOSTS:
```
kbarber: Are we removing the old file by the way?

PR #1953, opened 2016-05-06
(maint) fix link to fact_contents page

steveax:
Prior to this release, the link to the fact_contents documentation
pointed to fact_contents.html while the file is fact-contents.html.
This commit fixes the incorrect href.


wkalt:
merged manually against stable

PR #1955, opened 2016-05-10
(maint) bump kitchensink and tk-jetty

wkalt:
Bump kitchensink to 1.3.1 and tk-jetty to 1.5.8.


ajroetker:
I think you'll need a PR on PE as well https://github.com/puppetlabs/pe-puppetdb-extensions/blob/master/project.clj#L12

PR #1957, opened 2016-05-11
(maint) .gitignore: constrain; add Emacs entries

rbrw:


.gitignore
```diff
@@ -1,21 +1,14 @@
...
+# Emacs
 *#
+*~
+.#*
 
 # Local overide for testing config
-test-resources/config/local.clj
+/test-resources/config/local.clj
 
 # Vagrant cache folder & vagrantfile if used
```
rbrw: I also wondered if we really don't want git status to tell us when there are 5 vagrant dirs in say src/puppetlabs/puppetdb...

.gitignore
```diff
@@ -1,38 +1,27 @@
...
 *#
+*~
+.#*
 
 # Local overide for testing config
-test-resources/config/local.clj
-
-# Vagrant cache folder & vagrantfile if used
-.vagrant
-Vagrantfile
```
ajroetker: Why are you removing Vagrant stuff?

rbrw: Just didn't realize we needed it (as a puppetdb-specific thing), and if we do, I wonder if perhaps it should be better constrained, i.e. maybe we'd want `git status` to tell us if there are say 5 vagrant dirs in src/puppetlabs/puppetdb.

.gitignore
```diff
@@ -1,38 +1,30 @@
...
 
 # Local overide for testing config
-test-resources/config/local.clj
+/test-resources/config/local.clj
 
-# Vagrant cache folder & vagrantfile if used
-.vagrant
-Vagrantfile
+# Vagrant
+/.vagrant/
```
rbrw: If this is too conservative, (now or later) we could add other likely places, or we could add some inverted exclusions like "not in src", or we could go back to a free-floating `.vagrant/`.

.gitignore
```diff
@@ -1,38 +1,30 @@
...
-/classes/
+
+# Leiningen
 /target/
-/docs/
-/pkg/
-/log/
-*.tar.gz
-version
-junit
```
wkalt: I think we should add this back in

PR #1958, opened 2016-05-11
(PDB-2680) Refactor PuppetDBStatus service to a plain function

senior:
These changes will allow our extensions in PE to have a separate status that includes information from PE core but can also include additional dependencies when determining whether or not PuppetDB is 'up'.

### DIFF
--- a/src/puppetlabs/puppetdb/status.clj
+++ b/src/puppetlabs/puppetdb/status.clj
@@ -4,12 +4,26 @@
             [puppetlabs.puppetdb.mq :as mq]
             [puppetlabs.puppetdb.schema :as pls]
             [schema.core :as s]
-            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]
+            [trptcolin.versioneer.core :as versioneer]))
 
 (def status-details-schema {:maintenance_mode? s/Bool
-                            :queue_depth s/Int
+                            :queue_depth (s/maybe s/Int)
                             :read_db_up? s/Bool
                             :write_db_up? s/Bool})
+
+;; This is vendored from the tk-status-service because version checking fails
+;; semver validation on PDB snapshots. When we address this upstream we can put
+;; the tk version back in.
+(defn get-artifact-version
+  [group-id artifact-id]
+  (let [version (versioneer/get-version group-id artifact-id)]
+    (when (empty? version)
+      (throw (IllegalStateException.
+               (format "Unable to find version number for '%s/%s'"
+                 group-id
+                 artifact-id))))
+    version))
 
 (pls/defn-validated status-details :- status-details-schema
   "Returns a map containing status information on the various parts of
@@ -37,3 +51,12 @@
                 :else :error)]
     {:state state
      :status status-details}))
+
+(defn register-pdb-status
+  "Registers the PuppetDB instance in TK status using `register-fn`
+  with the associated `status-callback-fn`"
+  [register-fn status-callback-fn]
+  (register-fn "puppetdb-status"
+               (get-artifact-version "puppetlabs" "puppetdb")
+               1
+               status-callback-fn))
--- a/src/puppetlabs/puppetdb/status.clj
+++ b/src/puppetlabs/puppetdb/status.clj
@@ -1,0 +1,39 @@
+(ns puppetlabs.puppetdb.status
+  (:require [puppetlabs.puppetdb.scf.storage-utils :as sutils]
+            [puppetlabs.puppetdb.utils :as utils]
+            [puppetlabs.puppetdb.mq :as mq]
+            [puppetlabs.puppetdb.schema :as pls]
+            [schema.core :as s]
+            [puppetlabs.trapperkeeper.services.status.status-core :as status-core]))
+
+(def status-details-schema {:maintenance_mode? s/Bool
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/status.clj
```diff
@@ -0,0 +1,39 @@
...
+  (let [globals (shared-globals-fn)]
+    {:maintenance_mode? (maint-mode-fn?)
+     :queue_depth (utils/nil-on-failure
+                   (->> [:command-processing :mq :endpoint]
+                        (get-in config)
+                        (mq/queue-size "localhost")))
+     :read_db_up? (sutils/db-up? (:scf-read-db globals))
+     :write_db_up? (sutils/db-up? (:scf-write-db globals))}))
+
+(pls/defn-validated create-status-map
```
ajroetker: doesn't look like this defn-validated is used

test/puppetlabs/puppetdb/status_test.clj
```diff
@@ -9,10 +9,13 @@
...
-                                            (assoc :prefix "/status/v1/services")
-                                            base-url->str-with-prefix
-                                            client/get)
+      (let [{:keys [body] :as pdb-resp} (try (-> svc-utils/*base-url*
+                                                 (assoc :prefix "/status/v1/services")
+                                                 base-url->str-with-prefix
+                                                 client/get)
+                                             (catch Exception e
+                                               (.printStackTrace e)))
+            _ (clojure.pprint/pprint (json/parse-string body true))
```
ajroetker: leftover from testing?

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -72,7 +71,6 @@
    #'dashboard-redirect-service
    #'pdb-routing-service
    #'maint-mode-service
-   #'pdb-status-service
```
ajroetker: Might wanna have the PE PR handy because this will break extensions

test/puppetlabs/puppetdb/status_test.clj
```diff
@@ -9,10 +9,13 @@
...
-      (let [{:keys [body] :as pdb-resp} (-> svc-utils/*base-url*
-                                            (assoc :prefix "/status/v1/services")
-                                            base-url->str-with-prefix
-                                            client/get)
+      (let [{:keys [body] :as pdb-resp} (try (-> svc-utils/*base-url*
+                                                 (assoc :prefix "/status/v1/services")
+                                                 base-url->str-with-prefix
+                                                 client/get)
+                                             (catch Exception e
+                                               (.printStackTrace e)))
```
ajroetker: Is this left over from testing as well?

src/puppetlabs/puppetdb/status.clj
```diff
@@ -0,0 +1,39 @@
...
+  "Returns a map containing status information on the various parts of
+  a running PuppetDB system. This data can be interpreted to determine
+  whether the system is considered up"
+  [config :- {s/Any s/Any}
+   shared-globals-fn :- pls/Function
+   maint-mode-fn? :- pls/Function]
+  (let [globals (shared-globals-fn)]
+    {:maintenance_mode? (maint-mode-fn?)
+     :queue_depth (utils/nil-on-failure
+                   (->> [:command-processing :mq :endpoint]
```
ajroetker: Pre-existing but while we're moving this could you put this with the get-in below `(get-in config [...])`?

src/puppetlabs/puppetdb/status.clj
```diff
@@ -0,0 +1,39 @@
+(ns puppetlabs.puppetdb.status
+  (:require [puppetlabs.puppetdb.scf.storage-utils :as sutils]
+            [puppetlabs.puppetdb.utils :as utils]
+            [puppetlabs.puppetdb.mq :as mq]
+            [puppetlabs.puppetdb.schema :as pls]
+            [schema.core :as s]))
+
+(def status-details-schema {:maintenance_mode? s/Bool
```
ajroetker: What do you think about calling this `foss-status-details-schema` to signal that it will be different in PE?

ajroetker: Or if our plan is to duplicate some of the stuff in this whole namespace, maybe calling it pdb_status.clj like we have with pdb_routing.clj and naming the pe namespace pe_status.clj (like pe_routing.clj).

senior: I wasn't necessarily thinking we'd have a parallel pe_status, but we could. My rationale for refactoring this into a separate namespace is to limit the touchpoints between extensions and pdb.

test/puppetlabs/puppetdb/status_test.clj
senior: Ugh, I must have lost a commit in my rebase, I'll get rid of these.

src/puppetlabs/puppetdb/status.clj
ajroetker: So we're you thinking we'd have a `(def pe-status-details-schema (assoc status-details-schema :rbac-up? <status> :activity-up? <status>)`?

senior: Yes, so in PE assoc in those two additional pieces, add an fn to take that info into account and we should be good to go

ajroetker: Sounds great!

PR #1960, opened 2016-05-12
(PE-15587) Set noop flag on a report to true only if all its resource events are noop

kbrezina:
This commit improves the algorithm which tries to recognize that puppet run was started with `--noop` param. Originally, any noop resource event caused that the whole report was marked as noop. However, a noop puppet run would have all resource events in noop state. With this commit, all resource events need to be noop to mark the whole report as noop.

The current implementation is not able to recognize an unchanged noop puppet run since it doesn't contain any events. So, this improvement is just a workaround. To be able to recognize noop puppet runs properly, puppet needs to report in which mode was the run started. This problem is logged as PUP-6294.

### DIFF
--- a/puppet/lib/puppet/reports/puppetdb.rb
+++ b/puppet/lib/puppet/reports/puppetdb.rb
@@ -38,7 +38,7 @@ Puppet::Reports.register_report(:puppetdb) do
       end
 
       resources = build_resources_list
-      is_noop = resources.any? { |rs| has_noop_event?(rs) } and resources.none? { |rs| has_failed_event?(rs) }
+      is_noop = resources.any? { |rs| has_noop_event?(rs) } && resources.none? { |rs| has_enforcement_event?(rs) }
 
 
       defaulted_catalog_uuid = defined?(catalog_uuid) ? catalog_uuid : transaction_uuid
@@ -74,8 +74,8 @@ Puppet::Reports.register_report(:puppetdb) do
 
   # @return TrueClass
   # @api private
-  def has_failed_event?(resource)
-    resource["events"].any? { |event| event["status"] == 'failed' }
+  def has_enforcement_event?(resource)
+    resource["events"].any? { |event| event["status"] != 'noop' }
   end
 
   # @return Array[Hash]


kbrezina:
ping @senior @scottyw

scottyw:
LGTM

wkalt:
I'm +1 with the understanding that this still needs an agent-side fix. Most reports don't contain many events if they contain events at all, so there are still a lot of cases where this will misrepresent reality (e.g noop resources happen to be the only ones that fire).

kbarber:
@kbrezina the commit message doesn't exactly state why this has changed, and why its better, nor does the ticket either - its just that we'll have to create release notes explaining it, also I wouldn't mind a summary of the change so I can reason about its value. Can we actually provide in summary why the new change is a good one? What was broken before, why its fixed etc.

kbarber:
@kbrezina also, double check the milestone is what you are after since right now this is targeted at 4.1.0 and above. If this is a bug fix for example on an old version, then it will need rebasing.

puppet/lib/puppet/reports/puppetdb.rb
```diff
@@ -38,7 +38,7 @@ def report_to_hash
       end
 
       resources = build_resources_list
-      is_noop = resources.any? { |rs| has_noop_event?(rs) } and resources.none? { |rs| has_failed_event?(rs) }
+      is_noop = resources.any? { |rs| has_noop_event?(rs) } && resources.none? { |rs| has_enforcement_event?(rs) }
```
kbarber: Don't we now have the possibility to just simplify this, and scan all resource events to make sure they are noop? This would make the logic a little faster as well I imagine.

kbrezina:
@kbarber I've improved the commit message.
Concerning the target version, we definitely need it to go out with PE 2016.2.0. So, 4.1.0 seems to be just fine. @scottyw Do we want to backport it even to older PE versions?

kbarber:
@kbrezina I think we should still back port this to 4.0.x if we feel its a bug, we have other users that see the field other than PE, can you rebase this?

puppet/lib/puppet/reports/puppetdb.rb
kbrezina: @kbarber When the report contains no resource events then we cannot mark it as noop. So, I'm not sure we can evaluate it just in one cycle.

kbrezina:
@kbarber PR against the stable branch - https://github.com/puppetlabs/puppetdb/pull/1962

kbrezina:
Replaced by https://github.com/puppetlabs/puppetdb/pull/1962 - closing

PR #1961, opened 2016-05-12
(maint) Only log warning for ssl-client-cn rejection when not nil

ajroetker:
This commit only logs a warning when rejecting a request for a
ssl-client-cn when the ssl-client-cn is present. Prior to this patch
when using token auth in PE PuppetDB, this warning would be printed for
perfectly valid requests.


ajroetker:
transient failure

ajroetker:

PR #1962, opened 2016-05-13
(PE-15587) Set noop flag on a report to true only if all its resource events are noop

kbrezina:
This commit improves the algorithm which tries to recognize that puppet run was started with --noop param. Originally, any noop resource event caused that the whole report was marked as noop. However, a noop puppet run would have all resource events in noop state. With this commit, all resource events need to be noop to mark the whole report as noop.

The current implementation is not able to recognize an unchanged noop puppet run since it doesn't contain any events. So, this improvement is just a workaround. To be able to recognize noop puppet runs properly, puppet needs to report in which mode was the run started. This problem is logged as PUP-6294.

### DIFF
--- a/puppet/lib/puppet/reports/puppetdb.rb
+++ b/puppet/lib/puppet/reports/puppetdb.rb
@@ -38,7 +38,7 @@ Puppet::Reports.register_report(:puppetdb) do
       end
 
       resources = build_resources_list
-      is_noop = resources.any? { |rs| has_noop_event?(rs) } and resources.none? { |rs| has_failed_event?(rs) }
+      is_noop = resources.any? { |rs| has_noop_event?(rs) } && resources.none? { |rs| has_enforcement_event?(rs) }
 
 
       defaulted_catalog_uuid = defined?(catalog_uuid) ? catalog_uuid : transaction_uuid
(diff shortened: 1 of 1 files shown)


kbrezina:
ping @kbarber @senior

puppet/lib/puppet/reports/puppetdb.rb
```diff
@@ -38,7 +38,7 @@ def report_to_hash
       end
 
       resources = build_resources_list
-      is_noop = resources.any? { |rs| has_noop_event?(rs) } and resources.none? { |rs| has_failed_event?(rs) }
+      is_noop = resources.any? { |rs| has_noop_event?(rs) } && resources.none? { |rs| has_enforcement_event?(rs) }
```
wkalt: since a resource can have multiple events, we should make sure they're all noop instead of that they each have a noop. Additionally this could be rewritten to check
```
resources.all? { |rs| all_noop_events?(rs) }
```
to avoid a second pass through the list of resources.

kbrezina: @wkalt When the report contains no resource events then we cannot mark the whole report as noop. It needs to have at least 1 noop resource event.
So, I'm not sure we can evaluate it just in one cycle.

wkalt: @kbrezina ah, I see your point. looks good to me then.

PR #1963, opened 2016-05-16
(PDB-2704) Log at-at exceptions

rbrw:
The way at-at's uses Java Executors causes any exceptions thrown by a
task to be silently swallowed, so wrap the task with a try/catch that
logs exceptions as errors.

You can demonstrate the problem by adding a throw to the
db-maintenance-tasks function and running this:

  lein test :only puppetlabs.puppetdb.acceptance.node-ttl/test-node-ttl

It should hang.  After this patch, the test should still hang, but
should also log the exception.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -289,14 +289,20 @@
             dlo-compression-interval-millis (to-millis dlo-compression-interval)
             seconds-pos? (comp pos? to-seconds)
             db-maintenance-tasks (fn []
-                                   (do
-                                     (when (seconds-pos? node-ttl) (auto-expire-nodes! node-ttl write-db))
-                                     (when (seconds-pos? node-purge-ttl) (purge-nodes! node-purge-ttl write-db))
-                                     (when (seconds-pos? report-ttl) (sweep-reports! report-ttl write-db))
+                                   (try
+                                     (when (seconds-pos? node-ttl)
+                                       (auto-expire-nodes! node-ttl write-db))
+                                     (when (seconds-pos? node-purge-ttl)
+                                       (purge-nodes! node-purge-ttl write-db))
+                                     (when (seconds-pos? report-ttl)
+                                       (sweep-reports! report-ttl write-db))
                                      ;; Order is important here to ensure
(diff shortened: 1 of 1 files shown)


rbrw:
If you merge this, leave the ticket open because we still need to fix extensions.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -288,15 +288,19 @@
...
-            db-maintenance-tasks (fn []
-                                   (do
-                                     (when (seconds-pos? node-ttl) (auto-expire-nodes! node-ttl write-db))
-                                     (when (seconds-pos? node-purge-ttl) (purge-nodes! node-purge-ttl write-db))
-                                     (when (seconds-pos? report-ttl) (sweep-reports! report-ttl write-db))
-                                     ;; Order is important here to ensure
-                                     ;; anything referencing an env or resource
-                                     ;; param is purged first
-                                     (garbage-collect! write-db)))]
+            db-maintenance-tasks #(try
```
senior: for long anonymous functions, I prefer the (fn [] ...) syntax.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -288,15 +288,19 @@
...
+                                    (when (seconds-pos? node-purge-ttl)
+                                      (purge-nodes! node-purge-ttl write-db))
+                                    (when (seconds-pos? report-ttl)
+                                      (sweep-reports! report-ttl write-db))
+                                    ;; Order is important here to ensure
+                                    ;; anything referencing an env or resource
+                                    ;; param is purged first
+                                    (garbage-collect! write-db)
+                                    (catch Exception ex
+                                      (log/error ex)))]
```
senior: Do we really want to swallow the exception here and not rethrow it? Today at-at doesn't do error handling properly, so the exception gets ignored, but do you expect it to always be that way?

rbrw: Indeed - we do not.

src/puppetlabs/puppetdb/cli/services.clj
rbrw: Right - changed.

PR #1972, opened 2016-06-03
(PDB-2737) Split message consumers from processing

senior:
This commit separates the thread listening to ActiveMQ and the threads
that are processing the message (and persisting the changes to
disk). The intent of this change is to factor the code in a way that
allows a new queuing system to be dropped in place of ActiveMQ. The code
as it is has a subtle bug relating to how JMS expects messages to be
acknowledged. When acknowledging a message, it acknowledges all
previously unacknowledged messages for that thread. This code consumes all
messages (from ActiveMQ's perspective) via a single thread. This means it's
possible for a slow message to be in-flight, and a faster/smaller
message that is processed after, to acknolwedge that slower message. If
that slower message were then to fail, we'd have problems as ActiveMQ
would already have marked the message as successfully processed.

Once we have removed ActiveMQ from this pipeline, we'll be able to
acknowledge the message(s) in a way that corresponds with our usage.

### DIFF
--- a/src/puppetlabs/puppetdb/mq_listener.clj
+++ b/src/puppetlabs/puppetdb/mq_listener.clj
@@ -1,5 +1,8 @@
 (ns puppetlabs.puppetdb.mq-listener
-  (:import [javax.jms ExceptionListener JMSException MessageListener Session])
+  (:import [javax.jms ExceptionListener JMSException MessageListener Session
+            ConnectionFactory Connection Queue Message]
+           [java.util.concurrent Semaphore ThreadPoolExecutor TimeUnit SynchronousQueue
+            RejectedExecutionException ExecutorService])
   (:require [clojure.tools.logging :as log]
             [puppetlabs.puppetdb.command.dlo :as dlo]
             [puppetlabs.puppetdb.mq :as mq]
@@ -14,7 +17,8 @@
             [puppetlabs.trapperkeeper.services :refer [defservice service-context service-id]]
             [schema.core :as s]
             [puppetlabs.puppetdb.config :as conf]
-            [puppetlabs.puppetdb.schema :as pls]))
+            [puppetlabs.puppetdb.schema :as pls]
+            [clojure.core.async :as async]))
 
 ;; ## Performance counters
 
@@ -297,13 +301,13 @@
      (time! (global-metric :retry-persistence-time)
             (publish-fn (json/generate-string msg) (mq/delay-property delay :seconds)))))
 
-(defn create-message-handler
-  [publish discarded-dir message-fn]
+(defn wrap-message-handler-middleware
+  [send-delayed-msg-fn discarded-dir message-fn]
   (let [discard        #(dlo/store-failed-message %1 %2 discarded-dir)
         on-discard     #(handle-command-discard % discard)
         on-parse-error #(handle-parse-error %1 %2 discard)
         on-fatal       #(handle-command-failure %1 %2 discard)
-        on-retry       #(handle-command-retry %1 %2 publish)]
+        on-retry       #(handle-command-retry %1 %2 send-delayed-msg-fn)]
     (-> message-fn
         (wrap-with-discard on-discard maximum-allowable-retries)
         (wrap-with-exception-handling on-retry on-fatal)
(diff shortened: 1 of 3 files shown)


senior:
Note that the target branch here is 'cmd-proc-refactor', not master

senior:
@rbrw @ajroetker I have added another commit to this that adds an uncaughtExceptionHandler and some additional logging on RejectedExcecutionExceptions for the threadpool

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -339,25 +344,142 @@
...
+      (let [q (create-queue session endpoint)]
+        (with-open [producer (.createProducer session q)]
+
+          (.send producer (mq/to-jms-message session message-to-convert message-properties)))))))
+
+(defn create-command-consumer
+  "Create and return a command handler. This function does the work of
+  consuming/storing a command. Handled commands are acknowledged here"
+  [{:keys [discard-dir] :as mq-context} command-handler]
+  (let [message-handler (wrap-message-handler-middleware (send-delayed-message mq-context)
```
rbrw: Might prefer handle-message, so we have (handle-message ...), or even just (process ...).

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -339,25 +344,142 @@
...
+                                                         discard-dir
+                                                         command-handler)]
+    (fn [^Message message]
+      ;; When the queue is shutting down, it sends nil message
+      (when message
+        (try
+          (message-handler (mq/convert-jms-message message))
+          (.acknowledge message)
+          (catch Exception ex
+            (log/error ex "Exception raised from processing message, message is not acknowledged and will be retried")))))))
```
rbrw: "Unable to process message; won't acknowledge or retry"?
I'm guessing the exception may be clear via ex?

src/puppetlabs/puppetdb/mq_listener.clj
senior: process is confusing as the real function that does the work is called process-message (and is closed over down below). `handle-message` is fine though

src/puppetlabs/puppetdb/mq_listener.clj
senior: It will retry, I'll add an explicit rollback so that it clears that up, I'll change the log message as well

rbrw: Hah, I was mostly just trying to shorten it.

ajroetker:
:+1: from me codewise, I'm good for merging this when you are @rbrw

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -339,25 +344,143 @@
...
+              consumer (.createConsumer session (create-queue session endpoint))]
+    (try
+      ;; delivering the promise here we have ensured that we can
+      ;; connect to the MQ and that the queue has been created (for
+      ;; the first startup of PuppetDB)
+      (deliver mq-connected-promise true)
+      (loop []
+        (mq-message-handler (.receive consumer))
+        (recur))
+      (catch javax.jms.IllegalStateException e
```
rbrw: Might be worth a comment describing the context of this exception handler.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -339,25 +344,143 @@
...
+        (log/info "Received IllegalStateException, shutting down"))
+      (catch Exception e
+        (log/error e)))))
```
rbrw: Not sure, but wonder if it might also be worth an initial error message here (or in the caller) making it really clear that this receiver is dying.  Perhaps the exception is sufficient, but I could imagine that it might be quicker to grep for "receiver thread shutting down" or something.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -339,25 +344,143 @@
...
+  `pool-thread-factory`. Using this thread factory ensures that we
+  have distinguishable names for threads in this pool and that all
+  uncaught exceptions are logged as errors. Without this uncaught
+  exceptions get the default which is standard error"
+  [pool-thread-factory]
+  (-> (BasicThreadFactory$Builder.)
+      (.wrappedFactory pool-thread-factory)
+      (.namingPattern "cmd-proc-thread-%d")
+      (.uncaughtExceptionHandler logging-exception-handler)
+      (.daemon true)
```
rbrw: Perhaps we already know this is appropriate/safe, but generally speaking, I'm a little wary since (I think) daemon threads can be abruptly killed at shutdown, i.e. no finally-blocks/stack-unwinding, etc.  So if we're not sure, then I'd suggest we just make sure our threads can be (safely) stopped - whether via explicit time-to-die? checking of some kind, or possibly carefully arranged interruption, and then wait for them to quit.

rbrw: ...but I'll try to refresh my memory wrt the relevant bits of JCIP this weeknd.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -339,25 +344,143 @@
...
+                                                        discard-dir
+                                                        command-handler)]
+    (fn [^Message message]
+      ;; When the queue is shutting down, it sends nil message
+      (when message
+        (try
+          (handle-message (mq/convert-jms-message message))
+          (.acknowledge message)
+          (catch Exception ex
+            (.rollback message)
```
rbrw: Will be addressed in next version.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -339,25 +344,143 @@
...
+        (log/error e "Message not submitted to command processing threadpool")
+        (.release semaphore)))))
+
+(defn run-mq-listener-thread
+  "Creates (and starts) an isolated thread to continually receive new
+  ActiveMQ messages and pass them to the `message-consumer-fn`"
+  [mq-context message-consumer-fn mq-connected-promise]
+  (doto (Thread. (fn []
+                   (create-mq-receiver mq-context message-consumer-fn mq-connected-promise)))
+    (.setDaemon false)
```
rbrw: Let's double-check this one later too...

test/puppetlabs/puppetdb/mq_listener_test.clj
```diff
@@ -0,0 +1,125 @@
...
+                handler-fn (tu/mock-fn)
+                identity-handler (command-submission-handler threadpool-ctx handler-fn)]
+            (try
+
+              (is (= 1 (.availablePermits semaphore)))
+              (is (not (test-protos/called? handler-fn)))
+
+              (identity-handler "this arg does nothing")
+
+              (is (.tryAcquire semaphore 1 TimeUnit/SECONDS)
```
rbrw: Might make this shorter in the next version - for this one test it's no big deal, but very many of these would add up?

rbrw:
Though I think there are questions (some above) we'll want to revisit, this seems just fine as a place to start.

PR #1974, opened 2016-06-08
(PDB-2495) Leverage noop flag from puppet report when the flag is available

kbrezina:
The flag is set to true if the corresponding puppet run was stared in noop mode.

### DIFF
--- a/puppet/lib/puppet/reports/puppetdb.rb
+++ b/puppet/lib/puppet/reports/puppetdb.rb
@@ -38,7 +38,7 @@ Puppet::Reports.register_report(:puppetdb) do
       end
 
       resources = build_resources_list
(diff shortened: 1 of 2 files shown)


kbrezina:
@kbarber IMHO, nothing else is needed to resolve this ticket, right?

kbarber:
@kbrezina yeah, this should close the wound.

Hrm, we have no tests around any of this I've just noticed. Can you investigate a suitable rspec test? Looks like we don't have one for the fallback logic either, hrm.

https://github.com/puppetlabs/puppetdb/blob/master/puppet/spec/unit/reports/puppetdb_spec.rb

kbrezina:
@kbarber I've added an rspec test.

PR #1975, opened 2016-06-08
(maint) don't execute unnecessary updates in maybe-activate-node

wkalt:
Without this change we unnecessary lock the certname row even when we don't
update anything.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -209,6 +209,12 @@
   (not (empty? (jdbc/query ["SELECT 1 FROM certnames WHERE certname=? LIMIT 1"
                             certname]))))
 
+(defn certname-metadata
+  [certname]
+  (first (jdbc/query ["select certname, expired, deactivated
+                       from certnames where certname=?"
+                      certname])))
+
 (defn delete-certname!
   "Delete the given host from the db"
   [certname]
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -209,6 +209,12 @@
   (not (empty? (jdbc/query ["SELECT 1 FROM certnames WHERE certname=? LIMIT 1"
                             certname]))))
 
+(defn certname-metadata
+  [certname]
+  {:pre [certname]}
```
senior: Why did you use a precondition here to check for non-nil?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1422,17 +1428,19 @@
   "Reactivate the given host, only if it was deactivated or expired before
   `time`.  Returns true if the node is activated, or if it was already active.
 
-  Adds the host to the database if it was not already present."
+   Adds the host to the database if it was not already present."
```
senior: seems like this should be reverted as it's not aligned with what's above it

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1422,17 +1428,19 @@
...
-                                {:deactivated nil, :expired nil}
-                                ["certname=? AND (deactivated<? OR expired<?)"
-                                 certname timestamp timestamp])]
-    (pos? (first replaced))))
+  (let [{:keys [deactivated expired] :as metadata} (first (certname-metadata certname))]
+    (when-not metadata
+      (add-certname! certname))
+    (let [timestamp (to-timestamp time)
+          replaced  (when (or deactivated expired)
+                      (jdbc/update! :certnames
```
senior: I guess we have a race condition here with the node expiration GC process? Does this require anything using it to be wrapped in a repeatable read TX?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -209,6 +209,12 @@
   (not (empty? (jdbc/query ["SELECT 1 FROM certnames WHERE certname=? LIMIT 1"
                             certname]))))
 
+(defn certname-metadata
+  [certname]
+  {:pre [certname]}
+  (jdbc/query ["select certname, expired, deactivated from certnames where certname=?"
+               certname]))
```
ajroetker: should we just call `first` here since this will never return more than one row?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1422,17 +1428,19 @@
...
-  (when-not (certname-exists? certname)
-    (add-certname! certname))
-  (let [timestamp (to-timestamp time)
-        replaced  (jdbc/update! :certnames
-                                {:deactivated nil, :expired nil}
-                                ["certname=? AND (deactivated<? OR expired<?)"
-                                 certname timestamp timestamp])]
-    (pos? (first replaced))))
+  (let [{:keys [deactivated expired] :as metadata} (first (certname-metadata certname))]
+    (when-not metadata
```
ajroetker: should this be an if-not? then we could move the `let` into the else clause

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: yeah, there's a race here. I'd argue that we probably want RR for store report for this reason and for https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/scf/storage.clj#L1231-1231 . I'm putting changing store-report to repeatable read in the storage changes, but I'd be hesitant to do that as part of this change because it could end up hurting performance more than this fix helps. Maybe this should all just be addressed in the storage patch to master?

senior: That seems like the more safe thing to do, especially since you're making that RR change anyway.

PR #1976, opened 2016-06-10
(PDB-2771) expose depth for fact-path querying

wkalt:
Expose depth as a queryable field for the /fact-paths endpoint. This will allow
the middleware to query fact-paths more efficiently by excluding structured
facts.


wkalt:
@dmcpl yo

senior:
:+1:

PR #1987, opened 2016-06-22
(PDB-2732) Refactor message retry logic for Stockpile

senior:
Message retries will work differently once Stockpile is introduced to the codebase. Currently retried messages are enqueued with a delay. In the future, we'll schedule a delayed task to reprocess the message. This PR refactors what before was called our message handling middleware to a single function, parameterized three functions. A function to delay a message (for retry), one function to discard a message and a function to process the message. With Stockpile, we switch the delay function and maybe the discard function.

This PR should only introduce one behavior change from the existing code. It removes the exponential backoff that was in place. Now it sleeps 1 second for each retry. This change is temporary as setting a higher value (i.e. 60 seconds) causes our tests to fail without a faster retry loop in place.

### DIFF
--- a/src/puppetlabs/puppetdb/mq_listener.clj
+++ b/src/puppetlabs/puppetdb/mq_listener.clj
@@ -19,7 +19,8 @@
             [schema.core :as s]
             [puppetlabs.puppetdb.config :as conf]
             [puppetlabs.puppetdb.schema :as pls]
-            [clojure.core.async :as async]))
+            [clojure.core.async :as async]
+            [puppetlabs.puppetdb.utils.metrics :as mutils]))
 
 ;; ## Performance counters
 
@@ -137,30 +138,6 @@
 ;; architecture.
 ;;
 
-(defn wrap-with-meter
-  "Wraps a message processor `f` and a `meter` such that `meter` will be marked
-  for each invocation of `f`."
-  [f meter]
-  (fn [msg]
-    (mark! meter)
-    (f msg)))
-
-(defn try-parse-command
-  "Tries to parse `msg`, returning the parsed message if the parse is
-  successful or a Throwable object if one is thrown."
-  [msg]
-  (try+
-   (cmd/parse-command msg)
-   (catch Throwable e
-     e)))
-
-(defn parse-command
-  [msg]
-  (try+
-   (cmd/parse-command msg)
-   (catch Throwable e
-     (throw+ {:kind ::parse-error} e "Error parsing command"))))
-
 (defn annotate-with-attempt
   "Adds an `attempt` annotation to `msg` indicating there was a failed attempt
   at handling the message, including the error and trace from `e`."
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -167,154 +158,88 @@
...
-  middleware."
-  [f on-discard max-retries]
-  (fn [{:keys [command version annotations] :as msg}]
-    (let [retries (count (:attempts annotations))]
-      (create-metrics-for-command! command version)
-      (mark! (cmd-metric command version :seen))
-      (update! (global-metric :retry-counts) retries)
-      (update! (cmd-metric command version :retry-counts) retries)
-
-      (when (>= retries max-retries)
```
senior: There was a bug in the old code. After the 15th attempt to process a message, we schedule a retry (maybe a day later) but when that message is processed again it gets discarded. As a result, we delay/enqueue the 16th attempt to later discard it without processing it.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -167,154 +158,88 @@
...
-        (mark! (cmd-metric command version :discarded))
-        (on-discard msg))
-
-      (when (< retries max-retries)
-        (let [result (time! (cmd-metric command version :processing-time)
-                            (f msg))]
-          (mark! (cmd-metric command version :processed))
-          result)))))
-
-(defn wrap-with-thread-name
```
senior: This is no longer needed as the thread name is added via the threadpool

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -359,29 +284,38 @@
...
+    (time!
+     (global-metric :retry-persistence-time)
+     (with-open [connection (create-connection conn-pool)
+                 session (create-session connection)]
+       (let [q (create-queue session endpoint)
+             message-with-error (-> message
+                                    (annotate-with-attempt exception)
+                                    json/generate-string)]
+         (with-open [producer (.createProducer session q)]
+           (->> (mq/delay-property 1 :seconds)
```
senior: I started with 60 seconds here, which seems right if we have some number of immediate retries. Not sure how exactly that would look, maybe 4 quick retries, then delay it for a minute, then 4 quick retries etc until we get 16? This value can't be 60 seconds right now without breaking tests. Specifically our full stack tests. They hit a constraint violation and need to retry. Our code that blocks waiting for results doesn't wait for 60+ seconds.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -172,45 +172,44 @@
...
-          (binding [*logger-factory* (atom-logger log-output)]
-            (mql/handle-command-retry (make-cmd 1) (RuntimeException. "foo") publish))
-          (is (= (get-in @log-output [0 1]) :debug))
-          (is (instance? Exception (get-in @log-output [0 2])))))
-
-      (testing "Exception with stacktrace, no more retries"
-        (let [log-output (atom [])]
-          (binding [*logger-factory* (atom-logger log-output)]
-            (mql/handle-command-retry (make-cmd mql/maximum-allowable-retries) (RuntimeException. "foo") publish))
+  (testing "Should log retries as debug for less than 4 attempts"
```
senior: This is an example of how we could test this stuff going forward. Below doesn't use the test message handler macros but is able to isolate the logic on when to retry, from what happens when a message is retried.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -167,154 +158,88 @@
...
-    (-> message-fn
-        (wrap-with-discard on-discard maximum-allowable-retries)
-        (wrap-with-exception-handling on-retry on-fatal)
-        (wrap-with-command-parser on-parse-error)
-        (wrap-with-meter (global-metric :seen))
-        (wrap-with-thread-name "command-proc"))))
+          (let [ex (:throwable &throw-context)]
+            (log/errorf ex "[%s] [%s] Fatal error on attempt %d" id command retries)
+            (-> parsed-result
+                (annotate-with-attempt ex)
```
ajroetker: Can annotate-with-attempt go inside `discard-message-fn`?

ajroetker: Nope it cannot.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -172,45 +172,44 @@
 (defn make-cmd
   "Create a command pre-loaded with `n` attempts"
   [n]
-  {:command nil :version nil :annotations {:attempts (repeat n {})}})
+  {:headers {:id "foo-id-1"
+             :received (tfmt/unparse (tfmt/formatters :date-time) (now))}
+   :body (json/generate-string  {:command "some catalog" :payload "foo" :version 10 :annotations {:attempts (repeat n {})}})})
```
ajroetker: Channeling my inner @rbrw, you might want to break up this line a little bit...

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -137,22 +138,12 @@
...
-(defn try-parse-command
-  "Tries to parse `msg`, returning the parsed message if the parse is
-  successful or a Throwable object if one is thrown."
+(defn parse-command
   [msg]
   (try+
    (cmd/parse-command msg)
    (catch Throwable e
-     e)))
+     (throw+ {:kind ::parse-error} e "Error parsing command"))))
```
rbrw: Assuming this exception is just for the narrow use of message-handler-with-retries, I might consider moving this function down to just before that function.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -167,154 +158,88 @@
...
+
+  (mark-both-metrics! command version :seen)
+  (update-both-metrics! command version :retry-counts retries)
+
+  (mutils/multitime!
+   [(global-metric :processing-time)
+    (cmd-metric command version :processing-time)]
+   (f))
+
+  (mark-both-metrics! command version :processed))
```
rbrw: Not sure it matters here, but any reason this might want to return the result of (f), and does it affect anything if (f) throws?

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -167,154 +158,88 @@
...
+
+(defn message-handler-with-retries
+  "This function processing of the message, retrying messages that
```
rbrw: Processes the message, retrying...?

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -167,154 +158,88 @@
...
+  fail and discarding messages that have fatal errors or have exceeded
+  their maximum allowed attempts. `delay-message-fn` and
+  `discard-message-fn` are both functions of two arguments, a
+  `message` and an `exception`. `process-message-fn` is a function
+  that accepts a message as it's argument"
```
rbrw: Alternately, wondered about something like "Calls (delay-message message exception), (discard-message msg exception), and (process-message message) when appropriate."

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -167,154 +158,88 @@
...
 
-     (mark! (global-metric :seen))
-     (time! (global-metric :processing-time)
-            (f msg))
-     (mark! (global-metric :processed))
-
-     (catch fatal? {:keys [cause]}
-       (on-fatal msg cause)
-       (mark! (global-metric :fatal)))
+     (let [{:keys [command version annotations] :as parsed-result} (parse-command msg)
```
rbrw: As we discussed, we might change the arguments to the callbacks, or we could wait until the next revision.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -167,154 +158,88 @@
...
-        (wrap-with-command-parser on-parse-error)
-        (wrap-with-meter (global-metric :seen))
-        (wrap-with-thread-name "command-proc"))))
+          (let [ex (:throwable &throw-context)]
+            (log/errorf ex "[%s] [%s] Fatal error on attempt %d" id command retries)
+            (-> parsed-result
+                (annotate-with-attempt ex)
+                (discard-message-fn ex))))
+        
+        (catch Throwable exception
```
rbrw: Perhaps change this to Exception?

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -137,22 +138,12 @@
...
-    (f msg)))
-
-(defn try-parse-command
-  "Tries to parse `msg`, returning the parsed message if the parse is
-  successful or a Throwable object if one is thrown."
+(defn parse-command
   [msg]
   (try+
    (cmd/parse-command msg)
    (catch Throwable e
```
rbrw: Suspect this should be changed to Exception too.

test/puppetlabs/puppetdb/command_test.clj
rbrw: Hah - yeah, I was going to suggest that even if you're not going to be me, there might still be a few lines worth shortening a bit.

rbrw:
OK, I think that's all the comments I had at the moment.

src/puppetlabs/puppetdb/mq_listener.clj
senior: I don't think we need it for anything, but I think you're right, it's less surprising if it returns the result

src/puppetlabs/puppetdb/mq_listener.clj
senior: Figured out why it was Throwable. There are assertions in the command parsing code. If something like payload is missing, it will throw a `AssertionError` which is not a subclass of `Exception`. I added a specific clause to the parse-command try/catch for that error so that we can continue to handle it like we have before.

senior:
@rbrw @ajroetker addressed your comments

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -167,154 +151,98 @@
...
-  [f prefix]
+   (let [command-result (f)]
+     (mark-both-metrics! command version :processed)
+     command-result)))
+
+(defn parse-command
+  [msg]
+  (try+
+   (cmd/parse-command msg)
+   (catch AssertionError e
```
rbrw: Perhaps fine for now, but I wonder if we should revisit this and adjust parse-command to at least document this as part of its API, and perhaps consider having it return/throw something other than an AssertionError.

PR #1988, opened 2016-06-22
(maint) Intended catalog failed

MichalBryxi:
- While developing, we often need to pull in various data into PuppetDB. So we use `puppetlabs.puppetdb.cli.benchmark` tool pretty extensively to achieve that.
- This PR allows us in basic setup fill in needed data to populate fields.

- I think that there should also be a case for `noop=>false, status=>failed`, but that kind of report is not included in this data set. I wanted to make this simple so not included in this PR. If this PR is ok, I would like to also include this case.

### DIFF
--- a/resources/puppetlabs/puppetdb/benchmark/samples/reports/pg1.vm-304bebf8e5fc109ced84e8b086f9bcf8324f3d78.json
+++ b/resources/puppetlabs/puppetdb/benchmark/samples/reports/pg1.vm-304bebf8e5fc109ced84e8b086f9bcf8324f3d78.json
@@ -1,4 +1,5 @@
 {
+    "cached_catalog_status": "on_failure",
     "certname": "pg1.vm",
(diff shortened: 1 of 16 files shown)


wkalt:
@MichalBryxi should this field be included for all the sample reports?

MichalBryxi:
@wkalt I cannot tell whether the field itself should be in all the reports. Not having it does not break anything (current state). 

What I do know is that we probably don't want to include this value `on_failure` in every case. That would just move all reports from one corner to other. And here we need variety.

wkalt:
@MichalBryxi seems to me like we might as well add the field to all of them, with whatever variety we want.

MichalBryxi:
@wkalt No probs. Will force-push that change. BTW: Do we care about the failed tests?

MichalBryxi:
@wkalt Added the property to all reports.

PR #1989, opened 2016-06-27
(PDB-2783) Revisit command versioning

mwaggett:
*Replace hard-coded command versions with calls to the maximum supported command versions defined in command.clj.
*Edit command versioning policy to more closely match the query versioning policy.

### DIFF
--- a/documentation/versioning_policy.markdown
+++ b/documentation/versioning_policy.markdown
@@ -79,11 +79,15 @@ Commands can be versioned on an individual command basis, so we are generally fr
 
 Commands are primarily represented by a corresponding wire format. Wire formats are versioned along with the corresponding command.
 
-The reasons to trigger a new command version are more common, and in general if we aren’t sure it's easy enough to create another version anyway.
+Changes to an existing command version can be made if the change is backward compatible. For example:
 
-Some examples of changes that will require a new command version:
+* Addition of optional parameters within a command or wire format
+* Change of a required parameter to optional
+* Addition of a new enum value for an existing field
 
-* Any change to the parameters or parameter values within a command or wire format
+Some examples of changes that *will* require a new command version:
+
+* Removal or renaming of parameters or parameter values within a command or wire format
 * Change to serialization for wire formats inside payload
 
 The [API commands][commands] documentation contains more concrete information about the existing commands, versions and statuses for this version of PuppetDB.
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -108,6 +108,10 @@
    "replace catalog" (version-range 4 9)
    "store report" (version-range 3 8)
    "deactivate node" (version-range 1 3)})
+   
+(def latest-catalog-version (apply max (get supported-command-versions "replace catalog")))
+(def latest-report-version (apply max (get supported-command-versions "store report")))
+(def latest-facts-version (apply max (get supported-command-versions "replace facts")))
 
 (defn- die-on-header-payload-mismatch
   [name in-header in-body]
--- a/test/puppetlabs/puppetdb/command_test.clj
+++ b/test/puppetlabs/puppetdb/command_test.clj
@@ -482,7 +482,7 @@
 (deftest catalog-with-updated-resource-line
   (dotestseq [version catalog-versions
               :let [command {:command (command-names :replace-catalog)
-                             :version 9
+                             :version latest-catalog-version
                              :payload basic-wire-catalog}
                     command-1 (stringify-payload command)
                     command-2 (stringify-payload
@@ -509,7 +509,7 @@
 (deftest catalog-with-updated-resource-file
   (dotestseq [version catalog-versions
               :let [command {:command (command-names :replace-catalog)
-                             :version 9
+                             :version latest-catalog-version
                              :payload basic-wire-catalog}
                     command-1 (stringify-payload command)
                     command-2 (stringify-payload
@@ -536,7 +536,7 @@
 (deftest catalog-with-updated-resource-exported
   (dotestseq [version catalog-versions
               :let [command {:command (command-names :replace-catalog)
-                             :version 9
+                             :version latest-catalog-version
                              :payload basic-wire-catalog}
                     command-1 (stringify-payload command)
                     command-2 (stringify-payload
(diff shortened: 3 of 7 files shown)


test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -866,7 +866,7 @@
 
 (deftest replace-facts-bad-payload
   (let [bad-command {:command (command-names :replace-facts)
-                     :version 5
+                     :version (apply max (get supported-command-versions "replace facts"))
```
wkalt: @mwaggett this change seems fine to me at a high level, but it'd be cleaner if these apply max things could just be swapped out for references to that command-versions map you have in export.clj .

If you go that route I'd also think about moving that map to command.clj where supported-command-versions lives, since at that point it'll be more general than just export.

mwaggett: @wkalt yeah, I agree that this looks kind of messy right now, but wasn't sure what cleanest/least redundant thing to do was.
I'll try your suggestion. :]

ajroetker:
Could you speak to what specifically you think this change buys us? I can see how it would cut down on having to update all these calls when we bump a command version, but it also means that all these tests assume a specific version of the command they're submitting as well.

mwaggett:
@ajroetker Mostly not having to update all those things is what I was thinking. I think it's fine, and even desirable, for these tests to always assume the latest command version. If we want to test other versions, then hard-coding is fine, but I feel like most cases would just want the most recent version, right? since that's what all these tests had in the first place.

Is there a drawback that you have in mind?

src/puppetlabs/puppetdb/command.clj
```diff
@@ -108,6 +108,11 @@
    "replace catalog" (version-range 4 9)
    "store report" (version-range 3 8)
    "deactivate node" (version-range 1 3)})
+   
+(def latest-command-versions
```
senior: Any reason not to have a def here?
``` clojure
(def latest-catalog-version (apply max (get supported-command-versions "replace catalog"))
(def latest-report-version (apply max (get supported-command-versions "store report"))
(def latest-facts-version (apply max (get supported-command-versions "replace facts"))})
```

mwaggett: Nope, that's a good idea.

documentation/versioning_policy.markdown
```diff
@@ -79,11 +79,11 @@ Commands can be versioned on an individual command basis, so we are generally fr
 
 Commands are primarily represented by a corresponding wire format. Wire formats are versioned along with the corresponding command.
 
-The reasons to trigger a new command version are more common, and in general if we aren’t sure it's easy enough to create another version anyway.
+Changes to an existing command version can be made if the change is largely deemed backward compatible (so long as the consumer is less strict about new unexpected parameters). For example, addition of optional parameters within a command or wire format will not require a new version.
```
ajroetker: This is a really long line :)

"Changes to an existing command version may be made when the change is backward compatible."

What's the `consumer` mentioned in "consumer is less strict about new unexpected parameters"?

"For example, the addition of new optional fields to a command do not require a new version."

documentation/versioning_policy.markdown
```diff
@@ -79,11 +79,11 @@ Commands can be versioned on an individual command basis, so we are generally fr
...
 Commands are primarily represented by a corresponding wire format. Wire formats are versioned along with the corresponding command.
 
-The reasons to trigger a new command version are more common, and in general if we aren’t sure it's easy enough to create another version anyway.
+Changes to an existing command version can be made if the change is largely deemed backward compatible (so long as the consumer is less strict about new unexpected parameters). For example, addition of optional parameters within a command or wire format will not require a new version. 
 
-Some examples of changes that will require a new command version:
+Some examples of changes that *will* require a new command version:
 
-* Any change to the parameters or parameter values within a command or wire format
+* Removal or renaming of parameters or parameter values within a command or wire format
```
ajroetker: Changing a parameter to be required from optional as well

documentation/versioning_policy.markdown
mwaggett: Most of that wording was copied from the querying part of the versioning policy, but I can edit it.

wkalt: I think "largely deemed backward incompatible" is too squishy. The parenthetical should just be the second part of the preceding sentence, assuming that's our commitment.

ajroetker: > Changes to an existing command version can be made if the change is backward compatible.
For example, optional fields in commands do not require the field to be present in the command and therefore the addition of an optional field is backwards compatible and would not require a new version.

@senior @kbarber @rbrw @wkalt  ^ any opinions?

wkalt: "The only allowable change to an existing command version is the addition of an optional field."

senior: I don't think that's true. What if we added another edge type [here](https://docs.puppet.com/puppetdb/4.1/api/wire_format/catalog_format_v8.html#data-type-relationship). I think the user would feel it's pretty reasonable that we start allowing that new edge time in v8, not requiring a version bump. I like keeping the wording as backward incompatible or something else similarly generic. I think just saying optional fields is too specific without adding much value.

documentation/versioning_policy.markdown
senior: More strict validation of existing fields could probably go in this list as well.

documentation/versioning_policy.markdown
senior: @wkalt @mwaggett @ajroetker how about:

> Changes to a existing command versions can be made if the change is optional/additive and thus is backward compatible. An examples of backward compatible change would be adding an optional field with a default value. Similarly, adding a new enum value for a given existing field field could be added to a command version without breaking existing users.

documentation/versioning_policy.markdown
```diff
@@ -79,11 +79,14 @@ Commands can be versioned on an individual command basis, so we are generally fr
 
 Commands are primarily represented by a corresponding wire format. Wire formats are versioned along with the corresponding command.
 
-The reasons to trigger a new command version are more common, and in general if we aren’t sure it's easy enough to create another version anyway.
+Changes to an existing command version can be made if the change is backward compatible. For example:
```
ajroetker: probably need a newline here to match the syntax below

ajroetker:
Lgtm but you'll need a PR on extensions as well to match the latest-command-version stuff you've done here

PR #1990, opened 2016-06-28
(PDB-2838) server side changes for remediation field

wkalt:
Add support for the new "remediation" field in the /reports and /nodes
responses, along with corresponding changes for the benchmark tool.

### DIFF
--- a/src/puppetlabs/puppetdb/reports.clj
+++ b/src/puppetlabs/puppetdb/reports.clj
@@ -13,6 +13,7 @@
 
 (def event-wireformat-schema
   {:status s/Str
+   :corrective_change (s/maybe s/Bool)
    :timestamp pls/Timestamp
    :property (s/maybe s/Str)
    :new_value (s/maybe pls/JSONable)
@@ -27,6 +28,7 @@
    :file (s/maybe s/Str)
    :line (s/maybe s/Int)
    :containment_path [s/Str]
+   :corrective_change (s/maybe s/Bool)
    :events [event-wireformat-schema]})
 
 (def metric-wireformat-schema
@@ -52,6 +54,7 @@
    :end_time pls/Timestamp
    :producer_timestamp pls/Timestamp
    :producer (s/maybe s/Str)
+   :corrective_change (s/maybe s/Bool)
    :resources [resource-wireformat-schema]
    :noop (s/maybe s/Bool)
    :noop_pending (s/maybe s/Bool)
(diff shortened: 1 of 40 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1062,6 +1062,14 @@
     "ALTER TABLE reports ADD COLUMN noop_pending boolean"
     "CREATE INDEX idx_reports_noop_pending on reports using btree (noop_pending) where (noop_pending = true)"))
 
+(defn add-remediation-columns
+  []
+  (jdbc/do-commands
+    "alter table reports add column remediation boolean"
+    "alter table resource_events add column remediation boolean"
+    "create index idx_reports_remediation on reports(remediation)"
```
senior: Are we sure this is needed?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1199,12 +1199,19 @@
       #(not (or (empty? %) (kitchensink/string-contains? "[" %)))
       (reverse containment-path)))))
 
-(def store-resources-column? (atom false))
-(defn maybe-resources
-  [row-map]
-  (if @store-resources-column?
+(def enterprise-data? (atom false))
```
senior: Why are we not storing this boolean?

wkalt: in the database? Not sure why we would. For what it's worth this is just a generalization of the atom that was already there; just wanted to change it not to be resource-specific.

wkalt:
@senior I removed the index you pointed to because I couldn't get a query to use it -- it might be dependent on the proportion of true/false values though, so we might want to keep any eye on it.

src/puppetlabs/puppetdb/reports.clj
```diff
@@ -66,17 +68,21 @@
...
  (-> report-wireformat-schema
-     (dissoc :producer :noop_pending)))
+     (dissoc :producer :noop_pending :corrective_change)))
 
 (def report-v6-wireformat-schema
   (-> report-v7-wireformat-schema
       (dissoc :catalog_uuid :cached_catalog_status :code_id)))
 
-(def resource-event-v5-wireformat-schema
+(def resource-event-v8-wireformat-schema
```
ajroetker: There are no resource-events in the v8 wireformat, only resources

src/puppetlabs/puppetdb/scf/storage.clj
ajroetker: Fwiw @mullr and I talked about this quite a bit and settled on *not* grouping all these things together.

src/puppetlabs/puppetdb/reports.clj
wkalt: those contain resource events though right?

PR #1991, opened 2016-06-28
(PDB-2661) Document the puppetdb_query function

wkalt:
Add a blurb to docs describing usage of puppetdb_query.

### DIFF
--- a/documentation/api/query/tutorial-pql.markdown
+++ b/documentation/api/query/tutorial-pql.markdown
@@ -90,6 +90,17 @@ only accept unencrypted traffic from `localhost`.
 This requires that you specify a certificate (issued by the same CA PuppetDB
 trusts), a private key, and a CA certificate.
 
+### Querying with Puppet code
+
+The PuppetDB terminus includes the `puppetdb_query` function, which can be used
+to query PuppetDB from within a Puppet manifest. For example,
+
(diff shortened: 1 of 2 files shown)


documentation/api/query/tutorial-pql.markdown
```diff
@@ -90,6 +90,17 @@ only accept unencrypted traffic from `localhost`.
 This requires that you specify a certificate (issued by the same CA PuppetDB
 trusts), a private key, and a CA certificate.
 
+### Querying with puppet code
+
+The PuppetDB terminus includes the `puppetdb_query` function, which can be used
+to query PuppetDB from within a Puppet manifest. For example,
+
+    $last_node_query = 'nodes{order by report_timestamp desc limit 1}'
```
ajroetker: might be helpful to note you can use the AST syntax too, `$last_node_query = [ "from", "nodes", <stuff> ]`

ajroetker: Also showing an example of how an exported resource query translates to a `puppetdb_query` could be useful

ajroetker: The last thing that might be interesting for users is showing how you can use this for getting a list of nodes with some property (which you can't do with exported resources).

wkalt: thanks, these are good suggestions

PR #1993, opened 2016-07-01
(PDB-2847) Add a gated threadpool, fed by channel

senior:
This commit adds code that allows data from an core.async input channel
to be processed on a bounded threadpool. When the threadpool reaches
it's max size, it blocks the calling thread on put. The threadpool is
shutdown once the input channel has been closed and in-flight work as
finished.

### DIFF
--- a/src/puppetlabs/puppetdb/threadpool.clj
+++ b/src/puppetlabs/puppetdb/threadpool.clj
@@ -5,14 +5,16 @@
             RejectedExecutionException ExecutorService]
            [org.apache.commons.lang3.concurrent BasicThreadFactory BasicThreadFactory$Builder])
   (:require [clojure.core.async :as async]
-            [clojure.tools.logging :as log]))
+            [clojure.tools.logging :as log]
+            [puppetlabs.i18n.core :as i18n]))
 
 (def logging-exception-handler
   "Exception handler that ensures any uncaught exception that occurs
   on this thread is logged"
   (reify Thread$UncaughtExceptionHandler
     (uncaughtException [_ thread throwable]
-      (log/error throwable "Error processing command"))))
+      (log/error throwable
+                 (i18n/trs "Error processing command")))))
 
 (defn thread-factory
   "Creates a command thread factory, wrapping the
@@ -30,8 +32,9 @@
 
 (defn shutdown
   "Shuts down the gated threadpool, ensuring in-flight work is
-  complete before tearing it down. Currently waits up to 5 minutes for
-  that work to complete"
+  complete before tearing it down. Waits up to `shutdown-timeout` (in
+  milliseconds) for that work to complete before forcibly shutting
+  down the threadpool that work to complete"
   [{:keys [^Semaphore semaphore ^ExecutorService threadpool shutdown-timeout]}]
 
   ;; Prevent new work from being accepted
@@ -43,12 +46,14 @@
   ;; This will block, waiting for the threadpool to shutdown
   (when-not (.awaitTermination threadpool shutdown-timeout java.util.concurrent.TimeUnit/MILLISECONDS)
 
-    (log/warn "Attempted to shutdown threadpool, not stopped after 5 minutes, forcibly shutting it down")
+    (log/warn
+     (i18n/trs "Threadpool not stopped after {0} milliseconds, forcibly shutting it down" shutdown-timeout))
 
     ;; This will force the shutdown of the threadpool and will
     ;; not allow current threads to finish
     (.shutdownNow threadpool)
-    (log/warn "Threadpool forcibly shutdown")))
+    (log/warn
+     (i18n/trs "Threadpool forcibly shutdown"))))
 
 (defrecord GatedThreadpool [^Semaphore semaphore ^ExecutorService threadpool shutdown-timeout]
   java.io.Closeable
@@ -90,7 +95,7 @@
                              (finally
                                (.release semaphore)))))
     (catch RejectedExecutionException e
-      (log/error e "Message not submitted to command processing threadpool")
+      (log/error e (i18n/trs "Message not submitted to command processing threadpool"))
       (.release semaphore))))
 
 (defn exec-from-channel
--- a/src/puppetlabs/puppetdb/threadpool.clj
+++ b/src/puppetlabs/puppetdb/threadpool.clj
@@ -1,0 +1,113 @@
+(ns puppetlabs.puppetdb.threadpool
+  (:import [javax.jms ExceptionListener JMSException MessageListener Session
+            ConnectionFactory Connection Queue Message]
+           [java.util.concurrent Semaphore ThreadPoolExecutor TimeUnit SynchronousQueue
+            RejectedExecutionException ExecutorService]
+           [org.apache.commons.lang3.concurrent BasicThreadFactory BasicThreadFactory$Builder])
+  (:require [clojure.core.async :as async]
+            [clojure.tools.logging :as log]))
+
+(def logging-exception-handler
+  "Exception handler that ensures any uncaught exception that occurs
+  on this thread is logged"
+  (reify Thread$UncaughtExceptionHandler
+    (uncaughtException [_ thread throwable]
+      (log/error throwable "Error processing command"))))
+
+(defn thread-factory
+  "Creates a command thread factory, wrapping the
+  `pool-thread-factory`. Using this thread factory ensures that we
+  have distinguishable names for threads in this pool and that all
+  uncaught exceptions are logged as errors. Without this uncaught
+  exceptions get the default which is standard error"
+  [threadpool-name-pattern pool-thread-factory]
+  (-> (BasicThreadFactory$Builder.)
+      (.wrappedFactory pool-thread-factory)
+      (.namingPattern threadpool-name-pattern)
+      (.uncaughtExceptionHandler logging-exception-handler)
+      (.daemon true)
+      .build))
+
+(defn shutdown
+  "Shuts down the gated threadpool, ensuring in-flight work is
+  complete before tearing it down. Currently waits up to 5 minutes for
+  that work to complete"
+  [{:keys [^Semaphore semaphore ^ExecutorService threadpool shutdown-timeout]}]
+
+  ;; Prevent new work from being accepted
+  (.drainPermits semaphore)
+
+  ;; Shutdown the threadpool, allowing in-flight work to finish
+  (.shutdown threadpool)
+
+  ;; This will block, waiting for the threadpool to shutdown
+  (when-not (.awaitTermination threadpool shutdown-timeout java.util.concurrent.TimeUnit/MILLISECONDS)
+
+    (log/warn "Attempted to shutdown threadpool, not stopped after 5 minutes, forcibly shutting it down")
+
+    ;; This will force the shutdown of the threadpool and will
+    ;; not allow current threads to finish
+    (.shutdownNow threadpool)
+    (log/warn "Threadpool forcibly shutdown")))
+
+(defrecord GatedThreadpool [^Semaphore semaphore ^ExecutorService threadpool shutdown-timeout]
+  java.io.Closeable
+  (close [this]
+    (shutdown this)))
+
+(defn create-threadpool
+  "Creates an unbounded threadpool with the intent that access to the
+  threadpool is bounded by the semaphore. Implicitly the threadpool is
+  bounded by `size`, but since the semaphore is handling that aspect,
+  it's more efficient to use an unbounded pool and not duplicate the
+  constraint in both the semaphore and the threadpool"
+  [size name-pattern shutdown-timeout-in-ms]
+  (let [threadpool (ThreadPoolExecutor. 1
+                                        Integer/MAX_VALUE
+                                        1
+                                        TimeUnit/MINUTES
+                                        (SynchronousQueue.))]
+    (->> threadpool
+         .getThreadFactory
+         (thread-factory name-pattern)
+         (.setThreadFactory threadpool))
+
+    (map->GatedThreadpool
+     {:semaphore (Semaphore. size)
+      :threadpool threadpool
+      :shutdown-timeout shutdown-timeout-in-ms})))
+
+(defn call-on-threadpool
+  "Executes `f` on a gated threadpool. Calls `on-complete` with the
+  results of inoking `f`"
+  [{:keys [^Semaphore semaphore ^ExecutorService threadpool] :as gated-threadpool}
+   f]
+  (.acquire semaphore)
+  (try
+    (.execute threadpool (fn []
+                           (try
+                             (f)
+                             (finally
+                               (.release semaphore)))))
+    (catch RejectedExecutionException e
+      (log/error e "Message not submitted to command processing threadpool")
+      (.release semaphore))))
+
+(defn exec-from-channel
+  "Executes `on-input` for each input found on `in-chan`, with a
+  threadpool of size `parallelism`. The results of the `on-input`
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,119 @@
...
+      (log/error throwable "Error processing command"))))
+
+(defn thread-factory
+  "Creates a command thread factory, wrapping the
+  `pool-thread-factory`. Using this thread factory ensures that we
+  have distinguishable names for threads in this pool and that all
+  uncaught exceptions are logged as errors. Without this uncaught
+  exceptions get the default which is standard error"
+  [pool-thread-factory]
+  (-> (BasicThreadFactory$Builder.)
```
mullr: `..` macro?

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,119 @@
...
+      (.wrappedFactory pool-thread-factory)
+      (.namingPattern "cmd-proc-thread-%d")
```
mullr: I think it'll be valuable to give the whole pool a name and use that here, since we'll be spinning up a special pool for initial sync.

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,119 @@
...
+      (.uncaughtExceptionHandler logging-exception-handler)
+      (.daemon true)
+      .build))
+
+(defrecord GatedThreadpool [^Semaphore semaphore ^ExecutorService threadpool])
```
mullr: What's the purpose of this being a defrecord rather than a plain map? Avoiding reflection?

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,119 @@
...
+         (.setThreadFactory threadpool))
+
+    (map->GatedThreadpool
+     {:semaphore (Semaphore. size)
+      :threadpool threadpool})))
+
+(defn shutdown
+  "Shuts down the gated threadpool, ensuring in-flight work is
+  complete before tearing it down. Currently waits up to 5 minutes for
+  that work to complete"
```
mullr: Would generally expect to pass an explicit timeout to a shutdown function.

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,119 @@
...
+    ;; This will force the shutdown of the threadpool and will
+    ;; not allow current threads to finish
+    (.shutdownNow threadpool)
+    (log/warn "Threadpool forcibly shutdown")))
+
+
+(defn call-on-threadpool
+  "Executes `f` on a gated threadpool. Calls `on-complete` with the
+  results of inoking `f`"
+  ([threadpool f]
```
mullr: param naming is a confusing here; in this case `threadpool` is your defrecord, but below it's an ExecutorService from inside the defrecord.

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,119 @@
...
+                                (.release semaphore)))))
+     (catch RejectedExecutionException e
+       (log/error e "Message not submitted to command processing threadpool")
+       (.release semaphore)))))
+
+(defn exec-from-channel
+  "Executes `on-input` for each input found on `in-chan`, with a
+  threadpool of size `parallelism`. The results of the `on-input`
+  invocations will be put to the `out-chan` in the order it has
+  completed"
```
mullr: This is the blocking version, right? Mention of that in docstring would be nice, as well as something about how it'll close out-chan for you.

mullr: ... with a blocking one, is there any point to it closing out-chan at all?

src/puppetlabs/puppetdb/threadpool.clj
senior: I'll call it gated-threadpool and include it via :as below

src/puppetlabs/puppetdb/threadpool.clj
senior: I'd say that's up to you, seems odd to close a channel I don't own anyway

src/puppetlabs/puppetdb/threadpool.clj
senior: I prefer ->, but it's not a big deal to me. What do you think @rbrw.

src/puppetlabs/puppetdb/threadpool.clj
senior: Is that something you want to provide in your call to exec-from-channel?

senior: Also reminds me, I think I have some tests around ensuring these things are named right, I'll track those down.

src/puppetlabs/puppetdb/threadpool.clj
senior: Something you'd expect to include in the exec-from-channel

mullr:
this generally looks good and should work for me. 

I know we've talked about every conceivable variation of this, but here's another: exec-from-channel takes no-channels as parameter. It returns an input-chan, which looks like a regular channel but is actually a wrapper that holds a reference to all the processing machinery. When you close input-chan, it shuts everything down asynchronously. Or, you can call a special blocking-close function that will wait for everything to drain. Any output is done using the caller-provided function. 

Mostly this is just an idea for thinking about and discussion.

src/puppetlabs/puppetdb/threadpool.clj
mullr: > Is that something you want to provide in your call to exec-from-channel?

Yeah, I think that's reasonable.

src/puppetlabs/puppetdb/threadpool.clj
mullr: Yes, I think so.

senior:
@mullr pushed up the changes we talked about, should be good to go now. I have some more tests from the stockpile branch that I will create a follow up PR for. It's mostly testing around logging and the threadpool name etc

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,113 @@
...
+  ;; Prevent new work from being accepted
+  (.drainPermits semaphore)
+
+  ;; Shutdown the threadpool, allowing in-flight work to finish
+  (.shutdown threadpool)
+
+  ;; This will block, waiting for the threadpool to shutdown
+  (when-not (.awaitTermination threadpool shutdown-timeout java.util.concurrent.TimeUnit/MILLISECONDS)
+
+    (log/warn "Attempted to shutdown threadpool, not stopped after 5 minutes, forcibly shutting it down")
```
wkalt: I can't see where 5 minutes is coming from here -- all the callsites seem to indicate 1 second, but the message should probably be dynamic.

wkalt: I guess you said this got yanked out of the stockpile branch -- maybe that explains it

senior: Good catch, yeah, it wasn't parameterized before and was hard coded to 5 minutes. I'll change that to just include the parameter

ajroetker: Also, not to be *that guy* but these should all be i18n strings as well.

senior:
@mullr @ajroetker @wkalt added i18n to the log messages and exceptions. Had some logging tests from the other branch that I moved over too. That caused me to think some more about shutdown state, I've made some changes so that it throws now, rather than log when shutting down, probably worth talking through if there are questions about it

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+          (throw+ {:kind ::rejected
+                   :message cmd
+                   :incoming-channel in-chan}
+                  e
+                  (i18n/trs "Threadpool shutting down, message rejected"))))
+      (recur (async/<!! in-chan)))))
+
+
+
+
```
ajroetker: You really like these trailing newlines don't cha?

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+  [gated-threadpool in-chan on-input]
+  (loop [cmd (async/<!! in-chan)]
+    (when cmd
+      (try
+        (call-on-threadpool
+         gated-threadpool
+         (fn []
+           (on-input cmd)))
+        (catch RejectedExecutionException e
+          (throw+ {:kind ::rejected
```
ajroetker: How come we catch this exception, rethrow it and catch it again to rethrow it?

senior: Catching it on line 101 is so that we can release the semaphore. An argument could be made for NOT releasing the sempahore here. Happy to talk through that, but it's an edge case that we shouldn't hit. We catch it here so that we can provide context, like the specific command that failed. This is something that shouldn't happen if we're using this correctly, but I figured if it did, you'd want to know which command didn't get processed.

src/puppetlabs/puppetdb/threadpool.clj
senior: I do consider myself a whitespace craftsman. I must have removed my whitespace trimming emacs library, I'll fix this up.

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+      (.release semaphore)
+      (throw e))))
+
+(defn exec-from-channel
+  "Executes `on-input` for each input found on `in-chan`, with a
+  threadpool of size `parallelism`. The results of the `on-input`
+  invocations will be put to the `out-chan` in the order it has
+  completed. Throws when the threadpool is shutting down and there are
+  still messages in the channel"
+  [gated-threadpool in-chan on-input]
```
ajroetker: This is a small, picky note but this function reminds me a lot of `map` and it might be nice to have the signature be `[gated-threadpool f chan]`. Also the docstring is wrong now I think because `out-chan` isn't a thing.

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+    ;; RejectedExecutionExceptions only occur when attempting ot
+    ;; excecute new work on a threadpool that is in the process of
+    ;; shutting down or already has been shut down
+    (catch RejectedExecutionException e
+      (.release semaphore)
+      (throw e))))
+
+(defn exec-from-channel
+  "Executes `on-input` for each input found on `in-chan`, with a
+  threadpool of size `parallelism`. The results of the `on-input`
```
ajroetker: I don't see parallelism in this function either(?)

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+  (close [this]
+    (shutdown this)))
+
+(defn create-threadpool
+  "Creates an unbounded threadpool with the intent that access to the
+  threadpool is bounded by the semaphore. Implicitly the threadpool is
+  bounded by `size`, but since the semaphore is handling that aspect,
+  it's more efficient to use an unbounded pool and not duplicate the
+  constraint in both the semaphore and the threadpool"
+  [size name-pattern shutdown-timeout-in-ms]
```
ajroetker: size is never used

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+            [puppetlabs.i18n.core :as i18n]
+            [slingshot.slingshot :refer [throw+]]))
+
+(def logging-exception-handler
+  "Exception handler that ensures any uncaught exception that occurs
+  on this thread is logged"
+  (reify Thread$UncaughtExceptionHandler
+    (uncaughtException [_ thread throwable]
+      (log/error throwable
+                 (i18n/trs "Error processing command")))))
```
ajroetker: Is there any more information that we could throw in here (Error processing command on thread) or some such?

src/puppetlabs/puppetdb/threadpool.clj
rbrw: ...it'd be nice if both gitk, and github would highlight trailing space the way that git diff/log/etc. do.

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+      (.namingPattern threadpool-name-pattern)
+      (.uncaughtExceptionHandler logging-exception-handler)
+      (.daemon true)
+      .build))
+
+(defn shutdown
+  "Shuts down the gated threadpool, ensuring in-flight work is
+  complete before tearing it down. Waits up to `shutdown-timeout` (in
+  milliseconds) for that work to complete before forcibly shutting
+  down the threadpool that work to complete"
```
rbrw: vestigial "that work to complete"?

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+    ;; excecute new work on a threadpool that is in the process of
+    ;; shutting down or already has been shut down
+    (catch RejectedExecutionException e
+      (.release semaphore)
+      (throw e))))
+
+(defn exec-from-channel
+  "Executes `on-input` for each input found on `in-chan`, with a
+  threadpool of size `parallelism`. The results of the `on-input`
+  invocations will be put to the `out-chan` in the order it has
```
rbrw: Perhaps "in the order they have completed".

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+         (.setThreadFactory threadpool))
+
+    (map->GatedThreadpool
+     {:semaphore (Semaphore. size)
+      :threadpool threadpool
+      :shutdown-timeout shutdown-timeout-in-ms})))
+
+(defn call-on-threadpool
+  "Executes `f` on a gated threadpool. Calls `on-complete` with the
+  results of inoking `f`"
```
rbrw: Perhaps "result of invoking".

src/puppetlabs/puppetdb/threadpool.clj
```diff
@@ -0,0 +1,130 @@
...
+         (thread-factory name-pattern)
+         (.setThreadFactory threadpool))
+
+    (map->GatedThreadpool
+     {:semaphore (Semaphore. size)
+      :threadpool threadpool
+      :shutdown-timeout shutdown-timeout-in-ms})))
+
+(defn call-on-threadpool
+  "Executes `f` on a gated threadpool. Calls `on-complete` with the
```
rbrw: Actually, wonder if this docstring might be stale.

src/puppetlabs/puppetdb/threadpool.clj
senior: Maybe the thread id? Not really meaningful by itself, but you could differentiate errors between threads. We should never have this code get called, as we should be handling the exceptions properly in our code. This gets called if somehow an exception was to bubble up.

src/puppetlabs/puppetdb/threadpool.clj
senior: It's included in the semaphore below. If size wasn't used, that would be really bad :-)

src/puppetlabs/puppetdb/threadpool.clj
senior: I thought I fixed this up, apparently not

PR #1995, opened 2016-07-08
(PDB-2852) export: write data via Files/copy

rbrw:
Files/copy has been observed to be notably more efficient than io/copy.


senior:
:+1:  do we have testing on the 4.0.x branch?

rbrw:
Moving to stable.

PR #1999, opened 2016-07-12
(PDB-2867) Fix navigation header link

briancain:


documentation/_puppetdb_nav.html
```diff
@@ -4,7 +4,7 @@ <h2 id="puppetdb">PuppetDB {{ page.doc.version }}</h2>
   <li><strong>General information</strong>
     <ul>
       <li><a href="{{puppetdb}}/index.html">Overview and requirements</a></li>
-      <li><a href="{{puppetdb}}/contributing.html">Contributing to PuppetDB</a></li>
+      <li><a href="{{puppetdb}}/CONTRIBUTING.html">Contributing to PuppetDB</a></li>
```
briancain: I didn't really think it would be case sensitive, but I guess it is?

PR #2000, opened 2016-07-13
(PDB-2632) add inventory endpoint and support for dotted fields

wkalt:
Add an inventory endpoint of fact name=>value pairs, along with capability to
query it and the resources parameters column with "dotted syntax" like
[= parameters.foo bar]. Also add support in PQL. This does not include
projection, regex matching of path elements, or array indexing of path
elements.

### DIFF
--- a/documentation/api/query/v4/inventory.markdown
+++ b/documentation/api/query/v4/inventory.markdown
@@ -1,0 +1,127 @@
+---
+title: "PuppetDB 4.1: Inventory endpoint"
+layout: default
+canonical: "/puppetdb/latest/api/query/v4/inventory.html"
+---
+
+[curl]: ../curl.html#using-curl-from-localhost-non-sslhttp
+[subqueries]: ./ast.html#subquery-operators
+[dotted]: ./ast.html#dot-notation
+[environments]: ./environments.html
+[factsets]: ./factsets.html
+[catalogs]: ./catalogs.html
+[facts]: ./facts.html
+[fact-contents]: ./fact_contents.html
+[events]: ./events.html
+[edges]: ./edges.html
+[resources]: ./resources.html
+[nodes]: ./nodes.html
+
+The `/inventory` endpoint enables an alternative query syntax for digging into
+structured facts, and can be used instead of the `facts`, `fact-contents`, and
+`factsets` endpoints for most fact-related queries.
+
+## `/pdb/query/v4/inventory`
+
+This will return an array of `node inventories` matching the given query.
+Inventories for deactivated nodes are not included in the response.
+
+### URL parameters
+
+* `query`: optional. A JSON array containing the query in prefix notation
+  (`["<OPERATOR>", "<FIELD>", "<VALUE>"]`). See the sections below for the
+  supported operators and fields. For general information about queries, see
+  [our guide to query structure.][query]
+
+
+### Query operators
+
+See [the AST query language page][ast] for the full list of available operators.
+
+Note: This endpoint supports [dot notation][dotted] on the `facts` and
+`trusted` response fields.
+
+### Query fields
+
+* `certname` (string): the name of the node associated with the inventory.
+
+* `timestamp` (string): the time at which PuppetDB received the facts in the inventory.
+
+* `environment` (string): the environment associated with the inventory's
+  certname.
+
+* `facts` (json): a JSON hash of fact names to fact values.
+
+* `trusted` (json): a JSON hash of trusted data for the node.
+
+### Subquery relationships
+
+The following list contains related entities that can be used to constrain the result set with implicit subqueries. For more information, consult the documentation for [subqueries][subqueries].
+
+* [`environments`][environments]: the environment where a node inventory exists.
+* [`factsets`][factsets]: the factset corresponding to a node inventory's certname.
+* [`catalogs`][catalogs]: the catalog corresponding to a node inventory's certname.
+* [`facts`][facts]: the facts corresponding to a node inventory's certname.
+* [`fact-contents`][fact-contents]: the fact contents corresponding to a node inventory's certname.
+* [`events`][events]: the events corresponding to a node inventory's certname.
+* [`edges`][edges]: the catalog edges corresponding to a node inventory's certname.
+* [`resources`][resources]: the resources corresponding to a node inventory's certname.
+* [`nodes`][nodes]: the node corresponding to an inventory.
+
+### Response format
+Successful responses will be in `application/json`.
+
+The result will be a JSON array with one entry per certname. Each entry is of
+the form:
+
+    {
+      "certname": <node certname>,
+      "timestamp": <timestamp of fact reception>,
+      "environment": <node environment>,
+      "facts": {
+                 <fact name>: <fact value>,
+                 ...
+               },
+      "trusted": {
+                   <data name>: <data value>,
+                   ...
+                 }
+    }
+
+### Examples
+
+[Using `curl` from localhost][curl]
+
+    curl -X GET http://localhost:8080/pdb/query/v4/inventory -d 'query=["=", "facts.operatingsystem", "Darwin"]'
+
+    [ {
+        "certname" : "mbp.local",
+            "timestamp" : "2016-07-11T20:02:33.190Z",
+            "environment" : "production",
+            "facts" : {
+                "kernel" : "Darwin",
+                "operatingsystem" : "Darwin",
+                "memoryfree" : "3.51 GB",
+                "macaddress_p2p0" : "0e:15:c2:d6:f8:4e",
+                "system_uptime" : {
+                    "days" : 0,
+                    "hours" : 1,
+                    "uptime" : "1:52 hours",
+                    "seconds" : 6733
+                },
+                "netmask_lo0" : "255.0.0.0",
+                "sp_physical_memory" : "16 GB",
+                "operatingsystemrelease" : "14.4.0",
+                "macosx_productname" : "Mac OS X",
+                "sp_boot_mode" : "normal_boot",
+                "macaddress_awdl0" : "6e:31:ef:e6:36:54",
+                ...
+            },
+            "trusted" : {
+                "domain" : "local",
+                "certname" : "mbp.local",
+                "hostname" : "mbp",
+                "extensions" : { },
+                "authenticated" : "remote"
+            }
+    } ]
--- a/src/puppetlabs/puppetdb/pql/transform.clj
+++ b/src/puppetlabs/puppetdb/pql/transform.clj
@@ -123,6 +123,10 @@
             (second arg)
             (vec (rest arg)))))])
 
+(defn transform-field
+  [& args]
+  (str/join "." args))
+
 (def transform-specification
   {:from               transform-from
    :subquery           transform-subquery
@@ -145,5 +149,6 @@
    :exp                transform-exp
    :groupby            transform-groupby
    :limit              transform-limit
+   :field              transform-field
    :offset             transform-offset
    :orderby            transform-orderby})
--- a/src/puppetlabs/puppetdb/query/facts.clj
+++ b/src/puppetlabs/puppetdb/query/facts.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.puppetdb.query.facts
   "Fact query generation"
   (:require [puppetlabs.puppetdb.cheshire :as json]
+            [puppetlabs.puppetdb.scf.storage-utils :as sutils]
             [puppetlabs.puppetdb.facts :as facts]
             [puppetlabs.puppetdb.query :as query]
             [puppetlabs.puppetdb.query.paging :as paging]
@@ -37,8 +38,7 @@
   (fn [rows]
     (if (empty? rows)
       []
-      (->> rows
-           (map deserialize-fact-value)))))
+      (map #(utils/update-when % [:value] sutils/parse-db-json) rows))))
 
 (defn munge-path-result-rows
   [_ _]
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.puppetdb.query-eng.engine
   (:require [clojure.core.match :as cm]
             [clojure.string :as str]
+            [puppetlabs.i18n.core :as i18n]
             [clojure.set :refer [map-invert]]
             [clojure.tools.logging :as log]
             [honeysql.core :as hcore]
@@ -23,6 +24,10 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Plan - functions/transformations of the internal query plan
+
+(defn validate-dotted-field
+  [dotted-field]
+  (and (string? dotted-field) (re-find #"(facts|trusted)\..+" dotted-field)))
 
 (def field-schema (s/cond-pre s/Keyword
                               SqlCall SqlRaw
@@ -68,6 +73,11 @@
     [column :- column-schema
      value])
 
+(s/defrecord JsonContainsExpression
+  [field :- s/Str
+   column-data :- column-schema
+   value])
+
 (s/defrecord InExpression
     [column :- [column-schema]
      ;; May not want this if it's recursive and not just "instance?"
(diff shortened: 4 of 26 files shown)


wkalt:
I need to give this another once over in the morning for code cleanup, more docs, and better testing of the PQL regex.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -350,6 +350,8 @@
                                         database-metrics-registry)
         discard-dir (io/file (conf/mq-discard-dir config))]
 
+    (def my-db write-db)
```
ajroetker: leftover from testing?

src/puppetlabs/puppetdb/query/facts.clj
```diff
@@ -37,8 +38,8 @@
   (fn [rows]
     (if (empty? rows)
       []
-      (->> rows
-           (map deserialize-fact-value)))))
+      (do (def foo rows)
```
ajroetker: leftover testing code here as well

wkalt:

wkalt:

wkalt:

wkalt:

wkalt:

wkalt:

wkalt:

wkalt:

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1068,6 +1076,68 @@
...
+         PRIMARY KEY (resource)"
+      "ALTER TABLE catalog_resources ADD CONSTRAINT catalog_resources_resource_fkey
+         FOREIGN KEY (resource) REFERENCES resource_params_cache(resource) ON DELETE CASCADE"
+      "ALTER TABLE resource_params ADD CONSTRAINT resource_params_resource_fkey
+       FOREIGN KEY (resource) REFERENCES resource_params_cache(resource) ON DELETE CASCADE"
+      "CREATE INDEX rpc_hash_expr_idx ON resource_params_cache(encode(resource, 'hex'))")))
+
+(defn fact-values-value-to-jsonb
+  []
+  (let [values (->> (sutils/sql-hash-as-str "value_hash")
```
ajroetker: might it be worth hardcoding this in case of change as well?

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -207,6 +207,10 @@
   (hcore/raw
    (format "%s = ANY(?)" (first (hfmt/format column)))))
 
+(defn json-contains
+  [field value]
```
ajroetker: Looks like value doesn't get used?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -488,14 +488,17 @@
     (insert-records*
      :resource_params_cache
      (map (fn [[resource-hash params]]
-            {:resource (sutils/munge-hash-for-storage resource-hash) :parameters (when params (sutils/db-serialize params))})
+            {:resource (sutils/munge-hash-for-storage resource-hash)
+             :parameters (when params (sutils/munge-jsonb-for-storage params))})
```
ajroetker: (some-> params sutils/...)

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -264,6 +264,49 @@
...
+  "Splits a string on . unless the . is contained within a quoted substring."
+  [string]
+  (loop [[s & splits] (string/split string #"\.")
+         temp []
+         result []
+         quoted false]
+    (if (nil? s)
+      result
+      (if quoted
+        (if (= \" (last s))
```
ajroetker: These `(last s)` should be something like

```
(.substring s (dec (.length s))
```

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -264,6 +264,49 @@
...
+(defn dotted-query->fact-path
+  "Splits a string on . unless the . is contained within a quoted substring."
+  [string]
+  (loop [[s & splits] (string/split string #"\.")
+         temp []
+         result []
+         quoted false]
+    (if (nil? s)
+      result
+      (if quoted
```
ajroetker: What do you think about splitting this into two functions, one that does some loop recur here, and another which handles the quoted case? something like

```
(defn dotted-query->fact-path
  "Splits a string on . unless the . is contained within a quoted substring."
  [string]
  (loop [[s & splits] (string/split string #"\.")
         result []]
    (if (nil? s)
      result
      (if (and (= \" (first s)) (or (= (count s) 1) (not (= \" (last s)))))
          (let [[result rest] (handle-quoted-bs [s] splits)]
             (recur rest result))
          (recur splits (conj result s)))))))
```

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1124,6 +1209,22 @@
       (instance? ArrayBinaryExpression node)
       (instance? ArrayRegexExpression node)))
 
+(defn path->nested-map
+  "Given path a.b.c and value d, produce {a {b {c d}}}"
+  [path value]
```
ajroetker: is path supposed to be a string?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1183,6 +1288,33 @@
...
+                    path (if (= head "trusted")
+                           (cons head path)
+                           path)
+                    value-column (cond
+                                   (string? value) "value_string"
+                                   (ks/boolean? value) "value_boolean"
+                                   (integer? value) "value_integer"
+                                   (float? value) "value_float"
+                                   :else (throw (IllegalArgumentException.
+                                                 (format  "Value %s of type %s unsupported." value (type value)))))]
```
ajroetker: moar i18n

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1183,6 +1288,33 @@
...
+
+                (if (or (and (number? value) (= op "~"))
+                        (and (string? value) (contains? #{"<=" "<" ">" ">="} op)))
+                  (throw (format "Operator '%s' not allowed on value '%s'" op value))
```
ajroetker: i18n

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1284,7 +1416,7 @@
                                       ["select_latest_report"]]]
 
                                     (throw (IllegalArgumentException.
-                                             (format "Field 'latest_report?' not supported on endpoint '%s'" entity))))]
+                                            (format "Field 'latest_report?' not supported on endpoint '%s'" entity))))]
```
ajroetker: i18n

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -277,7 +281,8 @@
 (defn munge-jsonb-for-storage
   "Prepare a clojure object for storage depending on db type."
   [value]
-  (let [json-str (json/generate-string value)]
+  (let [json-str (-> (json/generate-string value)
+                     (.replaceAll "\\\\u0000" "\\\ufffd"))]
```
ajroetker: Might want to check to make sure there's no reflection going on here

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -264,6 +264,49 @@
       (apply assoc m new-kvs)
       m)))
 
+(defn maybe-strip-escaped-quotes
+  [s]
+  (if (= \" (first s) (last s))
```
ajroetker: same note about `last` from above

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -264,6 +264,49 @@
...
+    (subs s 1 (dec (count s)))
```
ajroetker: Looks like this might come in handy as a utils function

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -625,3 +625,92 @@
...
+            (query-to-vec
+              (format "SELECT %s as resource, parameters FROM resource_params_cache"
+                      (sutils/sql-hash-as-str "resource")))
+            parsed-responses (for [response responses] (assoc response :parameters (sutils/parse-db-json (response :parameters))))]
+        (is (= parsed-responses
+               [{:resource "a0a0a0" :parameters {:a "apple" :b {:1 "bear" :2 "button" :3 "butts"}}}
+                {:resource "b1b1b1" :parameters {:c "camel" :d {:1 "dinosaur" :2 "donkey" :3 "daffodil"}}}]))))
+    (testing "should change only value column type"
+      (let [schema-diff (diff-schema-maps before-migration (schema-info-map *db*))]
+        (is (= #{}
```
ajroetker: why not `empty?` instead of `= #{}`?

ajroetker: Actually nvm I like this better

wkalt:
@ajroetker addressed your comments. Good call on reflection but it looks fine.

documentation/api/query/v4/inventory.markdown
```diff
@@ -0,0 +1,117 @@
...
+
+* `facts` (json): a JSON hash of fact names to fact values.
+
+* `trusted` (json): a JSON hash of trusted data for the node.
+
+### Subquery relationships
+
+The following list contains related entities that can be used to constrain the result set with implicit subqueries. For more information, consult the documentation for [subqueries][subqueries].
+
+* [`environments`][environments]: the environment where a node inventory exists.
```
senior: All of these links need to have refs at the top of the file, right?

src/puppetlabs/puppetdb/pql/transform.clj
```diff
@@ -123,6 +123,10 @@
             (second arg)
             (vec (rest arg)))))])
 
+(defn transform-field
+  [& args]
+  (str/join #"." args))
```
senior: I'm surprised this works. It looks like it does because `str` is called on the regex. I think you probably just intended to have a string there.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -182,6 +262,7 @@
                                        [:= :catalogs.id :latest_catalogs.catalog_id]
 
                                        [:factsets :fs]
+
```
senior: Maybe an accident? These clauses are all in pairs.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1183,6 +1289,33 @@
...
+                           path)
+                    value-column (cond
+                                   (string? value) "value_string"
+                                   (ks/boolean? value) "value_boolean"
+                                   (integer? value) "value_integer"
+                                   (float? value) "value_float"
+                                   :else (throw (IllegalArgumentException.
+                                                 (i18n/tru "Value {0} of type {1} unsupported." value (type value)))))]
+
+                (if (or (and (number? value) (= op "~"))
```
senior: Could value be a boolean?

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1068,6 +1080,68 @@
     "alter table reports add column corrective_change boolean"
     "alter table resource_events add column corrective_change boolean"))
 
+(defn resource-params-cache-parameters-to-jsonb
```
senior: Both of these migrations seem pretty risky. Have you done some testing on larger datasets? Seems like we could have memory issues and excessive migration times.

wkalt: They're both relatively quick with the 5000 node set we have, which has 136926 fact values but only 711 cached resource params. I don't think that's representative enough. I'll try and get some more user info today to see what's appropriate.

I think the memory issue can be reduced by streaming the select through the application into a new table before deleting the old one, but I haven't figured out a way to avoid the trip through the application since postgres doesn't seem to have the function I need to go from a json-serialized string to proper json. Will keep looking for that as well.

wkalt:
@senior addressed your comments

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1068,6 +1076,91 @@
...
+  (jdbc/do-commands
+    (sql/create-table-ddl :resource_params_cache_transform
+                          ["resource" "bytea NOT NULL"]
+                          ["parameters" "jsonb"]))
+
+  (migrate-through-app
+    :resource_params_cache
+    :resource_params_cache_transform
+    ["encode(resource::bytea, 'hex') as resource" "parameters"]
+    500
```
wkalt: @senior I just made this number up, figuring it corresponded to the size of a large factset. We could probably make it larger to reduce the number of insert statements required, at the cost of more memory consumption.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -68,6 +73,11 @@
     [column :- column-schema
      value])
 
+(s/defrecord JsonContainsExpression
```
senior: Did you intend to have schema here?

documentation/api/query/v4/inventory.markdown
wkalt: added these btw

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1124,6 +1209,22 @@
       (instance? ArrayBinaryExpression node)
       (instance? ArrayRegexExpression node)))
 
+(defn path->nested-map
+  "Given path [a b c d] and value d, produce {a {b {c d}}}"
+  [path value]
+  (reduce #(assoc {} (utils/maybe-strip-escaped-quotes %2) %1)
```
senior: I think assoc is a bit confusing here, maybe `hash-map` instead?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1124,6 +1209,22 @@
...
+          (reverse (conj (vec path) value))))
```
senior: Since we know this is a vector, we can use rseq here instead.

wkalt:

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -264,6 +264,55 @@
...
+      [(string/join "." (conj result s)) splits]
+      (recur (conj result s) splits))))
+
+(defn dotted-query->fact-path
+  [string]
+  (loop [[s & splits :as v] (string/split string #"\.")
+         result []]
+    (if (nil? s)
+      result
+      (if (and (= \" (first s)) (or (= (count s) 1) (not (= \" (str-last s)))))
```
senior: I think `clojure.string/starts-with?` and `clojure.string/ends-with?` would be easier to read here

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -264,6 +264,55 @@
...
+  [path]
+  (flatten
+    (for [s path]
+      (if (re-find #"\[\d+\]$" s)
+        (-> s
+            (string/split #"(?=\[\d+\]$)")
+            (update 1 #(Integer/parseInt (subs % 1 (dec (count %))))))
+        s))))
+
+(defn handle-quoted-thing
```
senior: `handle-quoted-path-segment` maybe?

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -264,6 +264,55 @@
...
+  [v]
+  (loop [result []
+         [s & splits] v]
+    (if (and (= \" (str-last s)) (not (= (count s) 1)))
+      [(string/join "." (conj result s)) splits]
+      (recur (conj result s) splits))))
+
+(defn dotted-query->fact-path
```
senior: I think I understand what this is doing. You have a dotted query string that may or may not have segments that are quoted. You want to split them by `.`, but if it's quoted, you want the quoting to denote the segment in case there is a `.` inside the quotes that you'd want to ignore.

I think I've got another idea for implementation, but could you put some tests together for these two functions, seems pretty easy to test and would be good to get some coverage there.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -264,6 +264,55 @@
       (apply assoc m new-kvs)
       m)))
 
+(defn str-last
```
senior: My guess is this will use reflection, but if you use `clojure.string/starts-with?` and `clojure.string/ends-with?` you probably won't need it anymore.

src/puppetlabs/puppetdb/utils.clj
wkalt: haha, sorry about that

kbarber:
@wkalt you're my hero :-).

PR #2001, opened 2016-07-15
(PDB-2865) Add quick-retries to PDB command processing

ajroetker:
Current versions of PuppetDB incur a high cost when a command is
retried. The command gets written to a scheduler queue, then re-enqueued
in the regular queue, then reprocessed. A very common case is a database
transaction issue, where two commands are attempting to update the same
data. One thread wins, one loses and has to be retried. We should
implement a quicker "inner loop" retry, where we don't incur the I/O
overhead in this common case. When the quick retries fail, we can then
go through the normal retry logic.

This commit implements this quick retry in command processing (separate
from the queue logic) and removes the exponential backoff for the
traditional retry logic in favor of a set 10 minute retry time.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -56,6 +56,7 @@
    In either case, the command itself, once string-ified, must be a
    JSON-formatted string with the aforementioned structure."
   (:require [clojure.tools.logging :as log]
+            [puppetlabs.i18n.core :as i18n]
             [puppetlabs.puppetdb.scf.storage :as scf-storage]
             [puppetlabs.puppetdb.catalogs :as cat]
             [puppetlabs.puppetdb.reports :as report]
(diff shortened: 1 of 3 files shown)


src/puppetlabs/puppetdb/command.clj
```diff
@@ -410,9 +410,24 @@
           :producer-timestamp (-> cmd :payload :producer_timestamp)}
          (when ex {:exception ex})))
 
+(defn call-with-quick-retry [num-retries f]
+  (loop [n num-retries]
+    (let [result (try+
```
senior: I think it makes sense to have a success indicator, but I think we want to return the results of invoking (f) as the return value of call-with-quick-retry

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -227,22 +221,64 @@
...
+  (with-redefs [metrics.meters/mark! (call-counter)]
+    (let [publish (call-counter)]
       (testing "Exception with stacktrace, no more retries"
         (let [log-output (atom [])]
           (binding [*logger-factory* (atom-logger log-output)]
             (mql/handle-command-retry (make-cmd mql/maximum-allowable-retries) (RuntimeException. "foo") publish))
           (is (= (get-in @log-output [0 1]) :error))
           (is (instance? Exception (get-in @log-output [0 2]))))))))
 
+
```
senior: Extra whitespace

senior:
Minor comments, but looks good!

src/puppetlabs/puppetdb/command.clj
ajroetker: fixed that up @senior

PR #2002, opened 2016-07-18
(PDB-2888) Set hisotrical-catalogs-limit default to 2

ajroetker:
We only ever "need" the latest two catalogs in the console for the
resource-graph feature. This commit doesn't store extra data for users
who don't need it.


wkalt:
@dmcpl any objection here?

dmcpl:
@ajroetker @wkalt this should be fine, I think the only reason we currently need the historical endpoint is because we want to ensure that we can fetch the catalog with uuid matching the latest report.

Hitting the historical endpoint ensured that if the catalog with the required uuid was superseded between the call to /reports then /catalogs we could still retrieve it.

PR #2004, opened 2016-07-19
(PE-16739) Update FOSS PuppetDB check in to be on interval

briancain:
Prior to this commit, the checkin for puppetdb would only be run on
start up. This commit updates that to use overtone/at-at's `every`
function to run that checkin on a defined interval. In this case, once
at start up, and then again every 24 hours after.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -51,7 +51,7 @@
             [metrics.gauges :refer [gauge-fn]]
             [metrics.timers :refer [time! timer]]
             [metrics.reporters.jmx :as jmx-reporter]
-            [overtone.at-at :refer [mk-pool interspaced stop-and-reset-pool!]]
+            [overtone.at-at :refer [mk-pool every interspaced stop-and-reset-pool!]]
             [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.command.constants :refer [command-names]]
@@ -247,11 +247,14 @@
           (.unlock lock))))))
 
 (defn maybe-check-for-updates
-  [config read-db]
+  [config read-db job-pool]
   (if (conf/foss? config)
-    (-> config
-        conf/update-server
-        (version/check-for-updates! read-db))
+    (let [checkin-interval-millis (* 1000 60 60 24)] ; once per day
+      (every checkin-interval-millis #(-> config
+                                          conf/update-server
+                                          (version/check-for-updates! read-db))
+             job-pool
+             :desc "A reoccuring job to checkin the PuppetDB version"))
     (log/debug "Skipping update check on Puppet Enterprise")))
 
 (defn shutdown-mq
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -249,9 +249,13 @@
 (defn maybe-check-for-updates
   [config read-db]
   (if (conf/foss? config)
-    (-> config
-        conf/update-server
-        (version/check-for-updates! read-db))
+    (let [job-pool (mk-pool)
```
wkalt: I think this will leak threads in HUP scenarios

briancain: Ah yeah, do you think it'll be smarter to move this check to use the same job pool, then? Basically I'm thinking move the `maybe-check-for-updates` conditional into where the `let` is just below and just pass the job-pool that was created there into the function. I could assoc this job pool onto context to be shut down later, but this job is pretty small and could probably use the same one.

wkalt: yeah, that seems fine to me.

cprice404:
FWIW it would be nice to get projects moved over to https://github.com/puppetlabs/trapperkeeper-scheduler for stuff like this, because we will probably be replacing at-at with something better in the future.

briancain:
Hey @cprice404 , that is our plan! I opted not to bring in trapperkeeper-scheduler in this case, since puppetdb is already using overtone/at-at for scheduling. Our plan for other services (puppet-server is next...) is to use tk-scheduler however.

briancain:
Rebased from feedback @wkalt

ajroetker:
@briancain I think we've been pretty strict thus far about shutting down at-at jobs. I do think we'll want to add it to the context to be shut down. @senior can verify

briancain:
@ajroetker ah, I just pushed an update to make the version check job run on the job pool that already exists...so I believe that should no longer be a problem.

ajroetker:
@briancain sorry looks like my chrome just needed a refresh

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -247,11 +247,14 @@
...
-  [config read-db]
+  [config read-db job-pool]
   (if (conf/foss? config)
-    (-> config
-        conf/update-server
-        (version/check-for-updates! read-db))
+    (let [checkin-interval-millis (* 1000 60 60 24)] ; once per day
+      (every checkin-interval-millis #(-> config
+                                          conf/update-server
+                                          (version/check-for-updates! read-db))
```
ajroetker: Is the intent for the user that they should see the "New version available" message everyday in their log?

briancain: That seems to be a side effect. The main intent is for products to check in to dujour on a more regular basis instead of only at startup (i.e. once per day).

wkalt: I don't know that I particularly mind this happening -- @ajroetker do you think it's an issue?

ajroetker: @wkalt If you're good I'm good!

PR #2005, opened 2016-07-19
(PDB-2891) Allow whitespace around expression parens in PQL

ajroetker:
Prior to this commit trailing whitespace after a left-paren in an
expression (in PQL) would raise a syntax error. This commit changes the
PQL grammar to allow whitespace around the expression parens.


senior:
Based on our new definition around what goes into a Z release and what doesn't, shouldn't this be targetted at master? Other than that, :+1:

PR #2011, opened 2016-07-22
(PDB-2528) Expose PQL Parser in HTTP Endpoint

mwaggett:


src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -9,6 +9,7 @@
             [puppetlabs.puppetdb.query.paging :as paging]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.http :as http]
+            [puppetlabs.puppetdb.pql :as pql]
```
ajroetker: this isn't used

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -168,44 +169,46 @@
...
-         (i18n/trs "Error executing query ''{0}'' with query options ''{1}''. Returning a 400 error code."
-                   query query-options))
-        (http/error-response e))
-      (catch org.postgresql.util.PSQLException e
-        (if (= (.getSQLState e) "2201B")
-          (do (log/debug e (i18n/trs "Caught PSQL processing exception"))
-              (http/error-response (.getMessage e)))
-          (throw e))))))
+    (if (:parse_pql_only query-options)
+      (http/json-response query)
```
ajroetker: I think this logic should probably go to the caller of producer-streaming-body

ajroetker: edit: nevermind I can't think of a way to split this up nicely

ajroetker:
I'd retitle your commits, the commits should describe the work being done in the commit, not necessarily the description of the ticket.

mwaggett:
@ajroetker You're saying change the 'Expose PQL Parser in HTTP Endpoint' part since that didn't end up being exactly what I did, even though that's the ticket title?

documentation/api/query/v4/index.markdown
```diff
@@ -30,6 +30,9 @@ operator or a [PQL][pql] query string.
 a query with a [`from`][from] is required to choose the [entity][entities] for which to query. For
 general info about queries, see [our guide to query structure.][query]
 
+* `parse_pql_only`: optional. A boolean value. When true, the query response
```
wkalt: what does this do for AST? Wondering if having "pql" in the name is appropriate.

wkalt: also if it does work we should document the behavior on AST queries here

test/puppetlabs/puppetdb/testutils/http.clj
```diff
@@ -48,7 +48,10 @@
      (is (= http/status-ok (:status response)))
      (-> response
          :body
-         slurp
+         ((fn [response-body]
+            (if (string? response-body)
+              response-body
+              (slurp response-body))))
          (json/parse-string true)
```
wkalt: I think this would be nicer if we split it out into it's own function -- maybe-slurp or something.

documentation/api/query/v4/index.markdown
mwaggett: @wkalt If the query is in AST, it still returns AST. I can add that to the documentation.
As for what to call it, I know what you mean, but it's really only useful for parsing PQL into AST. Like, if someone were to use it with an AST query, nothing would break, but it's not what the param is meant to be used for.
Still think I should change it?

test/puppetlabs/puppetdb/testutils/http.clj
mwaggett: haha, @mullr predicted I would get called out for that.

documentation/api/query/v4/index.markdown
wkalt: yeah, I'd say change it to something that covers the functionality. ast_only? compile_only? Up to you.

wkalt: @mwaggett ^

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -219,7 +220,11 @@
   [b]
   (if (instance? Boolean b)
     b
-    (Boolean/parseBoolean b)))
+    (case b
+      "true" true
+      "false" false
+      (throw (IllegalArgumentException.
+              (str "Unsupported parameter value '" b "'. Acceptable values are true or false."))))))
```
ajroetker: should wrap this in i18n

src/puppetlabs/puppetdb/query_eng.clj
ajroetker: This should at least go at the top level of the function @mwaggett We should avoid doing any work we don't need to if the flag is passed in

ajroetker: Also it seems weird to me now that we're supporting this for any other endpoint than the root endpoint, since that's the only endpoint that allows PQL. Preferably we'd move this logic to just the produce-streaming-body for the root endpoint in that case.

mwaggett: @ajroetker It is only supported for the root endpoint -- that's the only place where this param is allowed in the param spec. 
There doesn't seem to be a separate produce-streaming-body for the root endpoint, so I'm not sure where you're suggesting I put that logic.

mwaggett: @ajroetker And as for moving it to the top level of the function, do you mean before the 'let'? because I had thought about that and I think I talked to @wkalt about it, but I think we concluded that it is fine after?

ajroetker: The main entry point for the root endpoint is query-handler [here](https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/http/query.clj#L352-L357) which is called [here](https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/http/handlers.clj#L58) I'd prefer either a separate query-handler function or to have this logic moved to the query-handler function. I don't see any reason it should go in produce-streaming-body.

wkalt: seems reasonable to me ^

test/puppetlabs/puppetdb/http/index_test.clj
```diff
@@ -136,6 +137,12 @@
             (is (= 1 (count results)))
             (is (= "host1" (:certname (first results))))))))
 
+    (testing "ast only"
+      (doseq [query ["nodes{}"
+                     "facts { [certname,name] in fact_contents[certname,name] { limit 1 order by certname } }"]]
```
ajroetker: might want to add another test for the behavior on AST queries

ajroetker:
lgtm overall

test/puppetlabs/puppetdb/http/index_test.clj
mwaggett: @ajroetker done.

wkalt:
@mwaggett @ajroetker the only thing that sticks out at me is that this will return invalid AST if invalid AST is supplied. That's probably because we've changed it to intercept the AST before passing to produce-streaming-body, instead of pulling it out after validation. I know this was already flipped once, but what do you guys think about that?

mwaggett:
@wkalt @ajroetker oof, yeah, I don't like that.......

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -219,7 +221,11 @@
   [b]
   (if (instance? Boolean b)
     b
-    (Boolean/parseBoolean b)))
+    (case b
+      "true" true
```
senior: I don't understand why this change was needed, was this a fix for something? This changes the behavior from what was there before. parseBoolean will ignore case and accept True as an example. It will also accept any string, returning false for things that don't fit.

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -349,9 +356,17 @@
   [globals]
   (select-keys globals [:scf-read-db :warn-experimental :url-prefix :pretty-print]))
 
+(defn valid-query?
+  [version query-map]
+  (let [{:keys [remaining-query entity query-options]} (qeng/user-query->engine-query version query-map)
+        validated-query (qeng/query->sql remaining-query entity version query-options)]
```
senior: I'd push this into the body of the let. When I see something let bound, I am expecting to see it used either in the let or in the body, which `validated-query` is not.

PR #2012, opened 2016-07-22
(PDB-2843) add function support to fact-contents

wkalt:
This corrects a bug where the fact-contents and edges endpoints would not allow
usage of functions.


ajroetker:
@senior did we ever shake out which branch this kind of stuff should go on?

PR #2013, opened 2016-07-25
(PDB-2837) update anonymization to consider producer

wkalt:
Update the anonymization tool to anonymize producer like any other node name.


ajroetker:
lgtm

PR #2019, opened 2016-07-28
(PDB-2310) Add helpful error message for invalid subquery operators

mwaggett:
This commit checks for subqueries under 'extract' nodes and throws an exception
with an informative error message if the subquery operator is not among the
acceptable 'select_<entity>' operators.


senior:
:+1:

PR #2021, opened 2016-07-28
(maint) ensure nss is up to date

wkalt:
An out of date nss can cause issues with openjdk. Ensure it's up to date on both the database and the master.


senior:
:+1:  will merge once the tests pass

PR #2023, opened 2016-08-02
(PDB-2851) Reject POST query requests that are missing content-type

mwaggett:
This commit adds a check for content type in POST requests and throws
an exception if the content type is "application/x-www-form-urlencoded",
which is the default if no content type is specified, instead of returning
all results from the endpoint.

### DIFF
--- a/src/puppetlabs/puppetdb/http/server.clj
+++ b/src/puppetlabs/puppetdb/http/server.clj
@@ -9,6 +9,7 @@
                                                     wrap-with-metrics
                                                     wrap-with-illegal-argument-catch
                                                     verify-accepts-json
+                                                    verify-content-type
                                                     make-pdb-handler]]
             [ring.util.response :as rr]
             [puppetlabs.comidi :as cmdi]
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -270,12 +271,15 @@
...
-                         (parse-fn query))))))
+  (if (not= (:content-type req) "application/x-www-form-urlencoded")
+    (-> (with-open [reader (-> req :body clojure.java.io/reader)]
+          (json/parse-stream reader true))
+        (update :query (fn [query]
+                         (if (vector? query)
+                           query
+                           (parse-fn query)))))
+    (throw (IllegalArgumentException.
+            (i18n/trs "POST queries must include Content-Type header")))))
```
ajroetker: Probably want a test for this

ajroetker:
@wkalt @senior Should we only accept requests that are application/json?

mwaggett:
@ajroetker fwiw, only that default Content Type causes the issue described in the ticket. Like, I can't literally say -H 'Content-Type:butts' and it'll execute the query correctly.

PR #2025, opened 2016-08-03
(PDB-2934) Fix issue with running tests at the REPL

senior:
Running individual tests at the REPL will result in a
NullPointerException due to an issue in our with-test-db test
function. This commit fixes that so the tests can be ran from the
command line or from the REPL.


rbrw:
lgtm

PR #2029, opened 2016-08-09
(PDB-2946) single-arity usage of or/and broken in subqueries

mwaggett:
This commit adds 'or' and 'and' to the list of operators to ignore when checking
for invalid subqueries in nodes of the form ["extract" column [subquery-name _]].

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -1900,7 +1900,7 @@
                    :state (cond-> state column-validation-message (conj column-validation-message))
                    :cut true})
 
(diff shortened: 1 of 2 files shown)


wkalt:
are these the only affected operators?

mwaggett:
@wkalt I think so? 'not' and 'group_by' are already on the list of operators to ignore and @senior and I couldn't think of any others when I made this fix the first time around.

wkalt:
@mwaggett okay, seems like this will work. My only concern here is that by building our logic around that blacklist we're opening ourselves up to future bugs when operators are added, since the queries that will hit issues may tend to be pretty obscure. Not an issue with this specific patch though.

PR #2030, opened 2016-08-10
(PE-17059) Make sure that reports.noop flag is always set.

kbrezina:
Is was discovered that the latest PDB terminus with an older puppet agent can produce reports with noop set to nil. When noop flag in not defined or is nil in Puppet report the original PuppetDB heuristic needs to be used to set the flag.

### DIFF
--- a/puppet/lib/puppet/reports/puppetdb.rb
+++ b/puppet/lib/puppet/reports/puppetdb.rb
@@ -38,7 +38,9 @@ Puppet::Reports.register_report(:puppetdb) do
       end
 
       resources = build_resources_list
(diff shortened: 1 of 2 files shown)


wkalt:
@kbrezina this seems fine to me. Just to check, is there some issue with nil-valued noops in reports in the middleware or UI? They can also arise from old pre-upgrade data.

kbrezina:
Currently, the problems are only in PE Overview with the latest reports received within last hour or so (configurable). Still, it makes sense to improve handling of nils in noop.

wkalt:
@kbrezina so in the case of existing data, I imagine users upgrading from certain versions will see unknown for all their nodes. Is that what you'd expect?

wkalt:
@kbrezina also this should be on stable

wkalt:
merged to stable manually

PR #2031, opened 2016-08-10
(PDB-2939) Don't try to store same report twice

rbrw:
There's no reason to, and the attempt causes PDB to wander off into the
long (eventually fruitless) retry process.

We believe that's also what was causing one of the convergence-test
failures.  i.e. if the database connection was slow enough, two syncs
that were scheduled one right after the other could enqueue the same
command twice (in that case the addition of a report), and then the test
would fail (time out) because the second addition never showed up as a
completed command since it was repeatedly rejected as a duplicate.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -1193,45 +1193,49 @@
                 :as report} (normalize-report orig-report)
                 report-hash (shash/report-identity-hash report)]
            (jdbc/with-db-transaction []
-             (let [certname-id (certname-id certname)
-                   row-map {:hash (sutils/munge-hash-for-storage report-hash)
-                            :transaction_uuid (sutils/munge-uuid-for-storage transaction_uuid)
-                            :catalog_uuid (sutils/munge-uuid-for-storage catalog_uuid)
-                            :code_id code_id
-                            :cached_catalog_status cached_catalog_status
-                            :metrics (sutils/munge-jsonb-for-storage metrics)
-                            :logs (sutils/munge-jsonb-for-storage logs)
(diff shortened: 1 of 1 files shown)


wkalt:
looks good to me

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1193,45 +1193,49 @@
...
-                                              maybe-corrective-change)]
-               (when-not (empty? resource_events)
-                 (->> resource_events
-                      (sp/transform [sp/ALL :containment_path] #(some-> % sutils/to-jdbc-varchar-array))
-                      (map adjust-event-metadata)
-                      (apply jdbc/insert! :resource_events)))
-               (when update-latest-report?
-                 (update-latest-report! certname)))))))
+             (let [shash (sutils/munge-hash-for-storage report-hash)]
+               (when-not (-> "select 1 from reports where hash = ? limit 1"
```
senior: I can see three ways of solving this problem

1. Check if the report exists already
2. Catch uniqueness constraint violations
3. Both

It seems like 1 falls short in the case where the same report is submitted twice and might be processed at the same time. Seems unlikely in the real world, possible in our tests. 2 would waste some time processing a report to only get a failure and ignore it (discard the report) but would catch the case I mentioned earlier. 3 would be more efficient in the common ways this error can occur and also protects us on concurrent submissions.

senior: Talking with @rbrw and thinking this through again, I don't think the scenario I had above manifests as a real problem. Right now, with quick retries in place, we'll fail on a constraint violation then, immediately try again. That second time through would see the hash and resolve itself. What we have here is fine.

PR #2032, opened 2016-08-10
(PDB-2952) fix failing spec tests for ruby 2.2

wkalt:
Ruby 2.2 as changed URI validation to allow underscores, which we were using in
tests as example invalid URLs. This just changes those tests to use # instead.


mwaggett:
Looks good to me.. ¯\_(ツ)_/¯

PR #2036, opened 2016-08-15
(ORCH-1491) Better PQL examples and docs structure

kbarber:
This patch adds a new PQL examples page with a list examples containing:

* a title
* description
* pql
* output

It also makes the PQL documentation a top-level element in navigation so it
and other related articles can be found easier.

The tutorial and references pages also have links to each other now so they
are easier to find.

### DIFF
--- a/documentation/api/query/examples-pql.markdown
+++ b/documentation/api/query/examples-pql.markdown
@@ -1,0 +1,447 @@
+---
+title: "PuppetDB 4.2 » Puppet query language (PQL) » Examples"
+layout: default
+---
+
+[tutorial]: ./tutorial-pql.html
+[reference]: ./v4/pql.html
+
+## Example Queries
+
+Below is a list of example PQL queries that you may find useful.
+
+Other resources you may also find useful include:
+
+* [PQL tutorial][tutorial]
+* [PQL reference guide][reference]
+
+***
+
+### Filtering on node names
+
+Query nodes with `green` in their name.
+
+``` ruby
+nodes { certname ~ 'green' }
+```
+
+*Output:*
+
+``` json
+[
+    {
+        "cached_catalog_status": "not_used",
+        "catalog_environment": "production",
+        "catalog_timestamp": "2016-08-15T11:06:26.275Z",
+        "certname": "greenserver.vm",
+        "deactivated": null,
+        "expired": null,
+        "facts_environment": "production",
+        "facts_timestamp": "2016-08-15T11:06:26.140Z",
+        "latest_report_corrective_change": null,
+        "latest_report_hash": "4a956674b016d95a7b77c99513ba26e4a744f8d1",
+        "latest_report_noop": false,
+        "latest_report_noop_pending": null,
+        "latest_report_status": "changed",
+        "report_environment": "production",
+        "report_timestamp": "2016-08-15T11:06:18.393Z"
+    }
+]
+```
+
+***
+
+### Basic fact filtering
+
+Nodes with operating system name `CentOS`.
+
+``` ruby
+inventory { facts.os.name = "CentOS" }
+```
+
+*Output (abbreviated):*
+
+``` json
+[
+    {
+        "certname": "centos70.vm",
+        "environment": "production",
+        "facts": {
+            ...
+            "operatingsystem": "CentOS",
+            "operatingsystemmajrelease": "7",
+            "operatingsystemrelease": "7.0.1406",
+            "os": {
+                "architecture": "x86_64",
+                "family": "RedHat",
+                "hardware": "x86_64",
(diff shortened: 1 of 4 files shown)


kbarber:
I would have preferred the implicit syntax for inventory querying, but this does not work yet. See this PR for more details: https://github.com/puppetlabs/puppetdb/pull/2035

senior:
:+1:

documentation/api/query/examples-pql.markdown
```diff
@@ -0,0 +1,447 @@
...
+
+*Output (abbreviated):*
+
+``` json
+[
+    {
+        "certname": "centos70.vm",
+        "environment": "production",
+        "facts": {
+            ...snip...
```
wkalt: spoke with the docs team about how this would look -- recommendation was to use ... or just leave the snip out, since we don't to that anywhere else.

documentation/api/query/examples-pql.markdown
```diff
@@ -0,0 +1,447 @@
...
+```
+
+***
+
+### Mcollective sub-collective filtering
+
+Show nodes that belong in the sub-collective `dc1`.
+
+``` ruby
+inventory[certname] { facts.mcollective.server.collectives[*] = "dc1" } }
```
wkalt: This was specced but never made it in. The best you can do is
```
inventory[certname] {facts.mcollective.server.collectives.match("\d+") = "dc1"}
```
I created PDB-2989 to actually get the wildcard working -- may or may not do the range stuff in the same ticket.

documentation/api/query/examples-pql.markdown
```diff
@@ -0,0 +1,447 @@
...
+
+Show nodes that have the profile class `Profile::Remote_mgmt` applied to it.
+
+``` ruby
+nodes { resources { type = "Class" and title = "Profile::Remote_mgmt" } }
+```
+
+*Output:*
+
+``` json
```
wkalt: have we always been able to do this?

documentation/api/query/examples-pql.markdown
```diff
@@ -0,0 +1,447 @@
...
+
+### List of nodes with report status failed
+
+List all nodes that have had a failure on their last run.
+
+``` ruby
+nodes { latest_report_status = 'failed' }
+```
+
+*Output:*
```
wkalt: The *Output* thing is not how we do these examples in other places -- any reason to change?

wkalt:
Just a couple minor comments. The new examples should help a lot.

documentation/api/query/examples-pql.markdown
kbarber: @melinda-campbell and I went through many iterations really based on how it looked, not comparing to status quo. Have you got anything in mind?

documentation/api/query/examples-pql.markdown
kbarber: Cool cool, thanks. Umm, I think this parsed though, let me double check.

documentation/api/query/examples-pql.markdown
kbarber: No worries, will do.

documentation/api/query/examples-pql.markdown
kbarber: Umm, probably, but @nfagerlund only let me know about it within the last few months I think. Nice huh? We need a pql variant, I'm cheating and using ruby because its close enough.

kbarber:
@wkalt yeah I hope so. And I think people want a place to publish their examples too, so we might promote a few contributions, and learn something in the process too (I keep finding new ways are using our querying now).

The repositioning in the LHS nav should help too, I found users didn't know there was more help elsewhere (I've been watching some user testing, that involved the docs), have the PQL stuff in its own section so people can see they can change might make a difference here. I think some people are example oriented, others reference and others prefer tutorials perhaps. Its like a triple-pronged attack :-).

We'll probably find so many examples we'll need further sectioning also, but I wasn't sure that was premature yet.

This does feel like a reasonable baseline for future improvement, let me get your comments fixed tomorrow mate. Thanks for the review.

documentation/api/query/examples-pql.markdown
wkalt: What we do everywhere else is just put the query and the response in the same code block, separated by a line https://docs.puppet.com/puppetdb/4.2/api/query/v4/facts.html#examples. If we like this Output style better we should probably change the rest of the docs to match, but I'd rather not do that unless someone has an issue with what we do today.

kbarber: But those examples you have listed are in the same box because they are a CLI command and a resulting response. An example query in this page isn't the same. The examples aren't CLI commands basically.

PR #2048, opened 2016-08-25
(PDB-2905) Migrate dlo to stockpile

rbrw:


src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,174 @@
...
+      (parse-metadata qmeta))))
+
+(defn- update-cmd-stats
+  "Updates stats to reflect the receipt of the named command."
+  [stats command size]
+  (let [addnp (fn [v n] (+ (or v 0) n))]
+    (-> stats
+        (update-in ["all" :count] inc)
+        (update-in ["all" :size] addnp size)
+        (update-in [command :count] addnp 1)
```
wkalt: why the different handling between global and per-command counts?

wkalt: I guess you're creating the per-command counts on first receipt. Makes sense.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,174 @@
...
+    {:count (gauge-fn registry ["puppetlabs.puppetdb.dlo" category "count"]
+                      #(stat [category :count]))
+     :size (gauge-fn registry ["puppetlabs.puppetdb.dlo" category "size"]
+                     #(stat [category :size]))}))
+
+(defn- note-new-discard! [dlo command size]
+  ;; Swap twice so that :stats will already have the values that the
+  ;; gauges are going to refer to.
+  (swap! dlo (fn [cur]
+               (assoc cur :stats (update-cmd-stats (:stats cur) command size))))
```
wkalt: is this just an update?

rbrw: Umm... yep.

senior:
The dashboard needs updated to remove the DLO metrics that are no longer valid

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,173 @@
...
+        stat (fn [what] (get-in (:stats @dlo) what))]
+    {:count (gauge-fn registry ["puppetlabs.puppetdb.dlo" category "count"]
+                      #(stat [category :count]))
+     :size (gauge-fn registry ["puppetlabs.puppetdb.dlo" category "size"]
+                     #(stat [category :size]))}))
+
+(defn- note-new-discard! [dlo command size]
+  ;; Swap twice so that :stats will already have the values that the
+  ;; gauges are going to refer to.
+  (swap! dlo update :stats update-cmd-stats command size)
```
senior: Why not just combine the two, let-bind the results of the first?

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,173 @@
...
+  (swap! dlo assoc-in [:metrics command] (cmd-gauges dlo command)))
+
+(defn- write-event-info
+  "Writes a command attempt summary to out, including timestamp,
+  error information, and stacktrace."
+  [out what timestamp error trace]
+  (.write out (format "%s @ %s\n" what timestamp))
+  (.write out error)
+  (.newLine out)
+  (doseq [elt trace]
```
senior: If you switch to a PrintWriter, you could instead use 
[this method](https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#printStackTrace(java.io.PrintWriter)

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,173 @@
...
+                                          (str (when-not (zero? i) \newline)
+                                               "Attempt " (inc i))
+                                          timestamp error trace))
+                      attempts))
+  (.newLine out)
+  (write-event-info out
+                    (str (when-not (zero? (count attempts)) \newline)
+                         "Attempt " (inc (count attempts)))
+                    (timestamp)
+                    (str exception)
```
senior: I don't think this is what we want here. This will output the class of the exception in addition to the error message

```
(str (Exception. "A nice descriptive error message"))
;=> "java.lang.Exception: A nice descriptive error message"
```
What I think we want is the attempt and time, then the "user friendly" description of the message (via getMessage) which is "a nice descriptive error message" above, then below that the full stacktrace (which will include the class and description).

```
java.lang.Exception: A nice descriptive error message
	at puppetlabs.puppetdb.command$eval37920.invokeStatic(form-init5913706765094242906.clj:50)
	at puppetlabs.puppetdb.command$eval37920.invoke(form-init5913706765094242906.clj:50)
...
```

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,173 @@
...
+(defn initialize
+  [path registry]
+  "Initializes the dead letter office (DLO), at the given path (a Path),
+   creating the directory if it doesn't exist, adds related metrics to
+   the registry, and then returns a representation of the DLO."
+  (try
+    (Files/createDirectory path (make-array FileAttribute 0))
+    (catch FileAlreadyExistsException ex
+      (when-not (Files/isDirectory path (make-array LinkOption 0))
+        (throw (Exception. (trs "DLO path {0} is not a directory"))))))
```
senior: Looks like it's missing the path argument to correlate with {0}

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,173 @@
...
+  (let [stats (gather-stats path)
+        dlo (atom {:path path :stats stats :registry registry})]
+    (swap! dlo
```
senior: Why swap here?

src/puppetlabs/puppetdb/nio.clj
```diff
@@ -0,0 +1,14 @@
+(ns puppetlabs.puppetdb.nio
+  (:import
+   [java.nio.file Path Files Paths StandardCopyOption]))
+
+(def copt-atomic StandardCopyOption/ATOMIC_MOVE)
```
senior: What's copt?

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: Ahh, ok, I think I may have just matched what we had; I'll adjust it.

src/puppetlabs/puppetdb/nio.clj
senior: oh, maybe a copy option?

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: Not sure I follow yet -- since the gauges refer to the stat from the "top", I wanted the stats to be fully in place before we create the gauges that refer to them, and I wanted the gauges to refer to the stats by "get-in" lookup, not by pointer reference.

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: So that we can pass the reference to the "final" dlo object to cmd-gauges.

src/puppetlabs/puppetdb/nio.clj
rbrw: Right, stolen from stockpile, but I'm fine with something better; I'd just tried to pick something simple/short and opted (at least atm) for direct bindings so we don't have a hash table look up for a simple #define -- suppose that could be argued overoptimization.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,173 @@
...
+              ;; entry-cmd-err-filename below.
+              (if-let [cmeta (and (.endsWith p ".json")
+                                  (parse-cmd-filename (-> p .getFileName str)))]
+                (update-cmd-stats (:command cmeta) (Files/size p))
+                stats))
+            {"all" {:count 0 :size 0}}
+            path-stream)))
+
+(defn- cmd-gauges [dlo category]
+  "Adds gauges to the dlo for the given category (e.g. \"all\"
```
senior: Docstring is in the wrong place here

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,173 @@
...
+  [stats command size]
+  (let [addnp (fn [v n] (+ (or v 0) n))]
+    (-> stats
+        (update-in ["all" :count] inc)
+        (update-in ["all" :size] addnp size)
+        (update-in [command :count] addnp 1)
+        (update-in [command :size] addnp size))))
+
+(defn- gather-stats [dlo]
+  "Gathers statistics about the filenames directly inside the dlo
```
senior: Docstring is in the wrong place here

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: Hmm, interesting - though (the horrors) looks like it formats with tabs.

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: Umm - oops; apparently missed that when I switched to trs.

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: Oh, looks like I actually dropped the stacktrace for the one exception we had; though in truth, I wasn't worrying to much about it since I knew the next PR (where we have all the exceptions) would/does revamp the whole thing.

rbrw: Wait, so I'm confused -- while this won't indent the stacktrace, I think it may already print more or less exactly what you have above, and if we think it's better to trade RAM for a larger write (and simpler code), I we can just `(.write out (str/join "\n  " trace))`, or use printStackTrace for the output?

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,173 @@
...
+  "Updates stats to reflect the receipt of the named command."
+  [stats command size]
+  (let [addnp (fn [v n] (+ (or v 0) n))]
+    (-> stats
+        (update-in ["all" :count] inc)
+        (update-in ["all" :size] addnp size)
+        (update-in [command :count] addnp 1)
+        (update-in [command :size] addnp size))))
+
+(defn- gather-stats [dlo]
```
senior: Elsewhere in the code, `dlo` is the atom containing the metrics, here `dlo` means the dlo directory. That tripped me up a bit.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,173 @@
...
+            {"all" {:count 0 :size 0}}
+            path-stream)))
+
+(defn- cmd-gauges [dlo category]
+  "Adds gauges to the dlo for the given category (e.g. \"all\"
+  \"replace command\").  The gauges will pull their values from the
+  dlo atom's :stats."
+  (let [registry (:registry @dlo)
+        stat (fn [what] (get-in (:stats @dlo) what))]
+    {:count (gauge-fn registry ["puppetlabs.puppetdb.dlo" category "count"]
```
senior: Category here is what you're calling command elsewhere in this file, right? Any reason not to preserve that?

src/puppetlabs/puppetdb/command/dlo.clj
senior: I've looked into the code a little deeper and I think there are a few ways that we can reduce the complexity here. I think the easiest way is to switch to the counter gauge. It supports increment and increment by a specified number. If that was in place, you wouldn't need to keep track of the counter and the gauge separately, which would also remove the self referencing atom update.

rbrw: Sounds right, thanks; I hadn't realized that the counter handled that, and the gauges were just carried over from the more complex thing we had before we dropped the filesystem monitoring.

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: As we discussed, I'll switch to printStackTrace -- particularly since it handles chained exceptions in a well known format.

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: Hmm, well I think the distinction I was trying to make is that here the category can be a command name or "all", and elsewhere, just a command name.

senior: Got it. FWIW, in the mq_listener area of the code we call it global rather than all https://github.com/puppetlabs/puppetdb/blob/cmd-proc-refactor/src/puppetlabs/puppetdb/mq_listener.clj#L162

resources/public/dashboard/index.html
```diff
@@ -257,19 +257,14 @@
...
        format: d3.format(",.3s"),
        snag: getPercentileSnag},
-      {description: "DLO Compression",
-       addendum: "round trip time, seconds",
-       url: "/metrics/v1/mbeans/puppetlabs.puppetdb.dlo:name=global.compression",
-       format: d3.format(",.3s"),
-       snag: getPercentileSnag},
       {description: "DLO Size on Disk",
        addendum: "bytes",
-       url: "/metrics/v1/mbeans/puppetlabs.puppetdb.dlo:name=global.filesize",
```
senior: why change the metric name here?

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -1,210 +1,144 @@
...
+  writes a summary of the failure to out."
+  [out attempts exception]
+  (let [write-header (fn [out i timestamp]
+                       (.format out "%sAttempt %d @ %s\n"
+                                (into-array Object [(if (zero? i) "" "\n")
+                                                    (inc i)
+                                                    timestamp])))]
+    (dorun (map-indexed (fn [i {:keys [timestamp error trace]}]
+                          (write-header out i timestamp)
+                          (.print out trace))
```
senior: What's the difference between this stacktrace and the one below on line 67?

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -150,7 +150,8 @@
   (let [attempts (get annotations :attempts [])
         attempt  {:timestamp (kitchensink/timestamp)
                   :error     (str e)
-                  :trace     (map str (.getStackTrace e))}]
+                  :trace (with-out-str
+                           (.printStackTrace e (java.io.PrintWriter. *out*)))}]
```
senior: Probably should with-open the PrintWriter here

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -22,7 +26,24 @@
       (throw+ {:kind ::parse-error} e "Error parsing command"))))
 
 (defn metadata-str [received command version certname]
-  (format "%s_%s_%s_%s" (tcoerce/to-long received) command version certname))
+  (format "%s_%s_%s_%s.json" (tcoerce/to-long received) command version certname))
```
wkalt: I think this is causing the .json to get doubled up when there are preexisting messages in the queue -- starting with a nonempty queue will spew a bunch of file not found errors.

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: Attempts aren't exceptions, they're just serialized backtrace (preexisting behavior), though I've tweaked them in this pr to just be the one big printStackTrace string rather than a list of stacktrace "elements" -- an intermediate step; the forthcoming pr (already up for review) will revamp this to make everything an exception.

src/puppetlabs/puppetdb/mq_listener.clj
rbrw: Definitely.

resources/public/dashboard/index.html
rbrw: This just changes it to match the new dlo names; there we now only have "all" or a command name, each with a total size and count, but happy to change that if we prefer something else.

wkalt:

PR #2052, opened 2016-08-30
(maint) Add note about configuring token auth for Couch users

ajroetker:
This commit adds a note about configuring token auth for Couch users,
because we forgot to manage the client-auth setting for Jetty in the
Couch PE module.


documentation/api/query/curl.markdown
```diff
@@ -67,6 +67,10 @@ viewing or editting node data depending on the operation.
       -H "X-Authentication: <token contents>"
       --tlsv1 \
       --cacert /etc/puppetlabs/puppet/ssl/certs/ca.pem \
+      
+**Note:** PE 2016.2 users will need to set `client-auth = want` under the
+`[jetty]` header of their jetty.ini confiuration, later versions of PE have this
```
wkalt: configuration, and also I'd change the comma after it to a period.

PR #2053, opened 2016-08-31
(PDB-3031) Separate stockpile open and load

senior:
This commit opens the stockpile queue, then loads the command channel
with existing entries in a background thread. This allows the startup
process to continue while the commands are being re-enqueued in memory.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -78,16 +78,12 @@
             [robert.hooke :as rh]
             [schema.core :as s]
             [slingshot.slingshot :refer [throw+ try+]]
-            [stockpile :as stock]
             [clojure.core.async :as async]
             [puppetlabs.puppetdb.mq-listener :as mql]
             [puppetlabs.puppetdb.queue :as queue])
   (:import [javax.jms ExceptionListener]
-           [java.nio.file Files LinkOption]
            [java.util.concurrent.locks ReentrantLock]
-           [org.joda.time Period]
-           [java.nio.file Files]
-           [java.nio.file.attribute FileAttribute]))
+           [org.joda.time Period]))
 
 (def cli-description "Main PuppetDB daemon")
 (def database-metrics-registry (get-in metrics/metrics-registries [:database :registry]))
@@ -267,6 +263,8 @@
     (.close ds))
   (when-let [pool (:job-pool context)]
     (stop-and-reset-pool! pool))
+  (when-let [command-loader (:command-loader context)]
+    (future-cancel command-loader))
   context)
 
 (defn initialize-schema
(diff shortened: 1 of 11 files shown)


senior:
This PR is waiting on a Stockpile release. I'll change the dependency and rebase once that has been done.

senior:
Pushed up the stockpile 0.0.2 release bump along with the namespace changes

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -266,6 +262,9 @@
     (.close ds))
   (when-let [pool (:job-pool context)]
     (stop-and-reset-pool! pool))
+  (when-let [command-load-future (:command-load-future context)]
+    (when-not (.isDone command-load-future)
```
rbrw: We can probably omit this guard since I think future-cancel should just return false if it's already finished.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -353,15 +338,18 @@
                         (if (:catalog-facts-bash command-processing)
                           (queue/sorted-command-buffer max-enqueued)
                           max-enqueued))
+          q (queue/create-or-open-stockpile (conf/stockpile-dir config))
```
rbrw: I'd suggest grabbing the next-likely-id here, and passing it to the loader below so that we'll be insensitive to the startup time of any other producers/consumers.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -146,3 +149,32 @@
...
+        (Files/createDirectories (get-path stockpile-root)
+                                 (make-array FileAttribute 0))
+        (stock/create queue-path)))))
+
+(defn load-persisted-messages
+  "Enqueue existing stockpile messages to `command-chan`. Note that
+  there is no guarantee on the enqueuing order of commands read from
+  stockpile's reduce function"
+  [q command-chan]
+  (let [max-id (stock/next-likely-id q)]
```
rbrw: Wondered about ceiling or new-id-threshold or something instead of max-id.

test/puppetlabs/puppetdb/queue_test.clj
```diff
@@ -124,3 +127,42 @@
...
+         (tqueue/store-command q "replace catalog" 1 "foo1" {:message "payload 1"})
+         (tqueue/store-command q "replace catalog" 1 "foo2" {:message "payload 2"})
+         (tqueue/store-command q "replace catalog" 1 "foo3" {:message "payload 3"})
+         (tqueue/store-command q "replace catalog" 1 "foo4" {:message "payload 4"})
+
+         (let [existing-q (create-or-open-stockpile temp-path)
+               command-chan (async/chan 40)]
+           (load-persisted-messages existing-q command-chan)
+           (is (= #{"foo1" "foo2" "foo3" "foo4"}
+                  (set (map :certname (repeatedly 4 #(async/poll! command-chan)))))))))))
```
rbrw: If it's easy, might want to check here that after we read the 4 messages, there's nothing else in the channel.

senior:
@rbrw pushed up fixes based on your comments

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -353,15 +337,19 @@
                         (if (:catalog-facts-bash command-processing)
                           (queue/sorted-command-buffer max-enqueued)
                           max-enqueued))
+          [q message-loader] (queue/create-or-open-stockpile (conf/stockpile-dir config))
```
rbrw: Not important, and feel free to leave it alone, but I'd be tempted to call this load-messages so that it'll be `(load-messages ...)` in the invocation below.

rbrw:
Looks good.

PR #2054, opened 2016-09-02
(PDB-2734) Add stockpile upgrade code

senior:
This commit adds the ability to upgrade from an existing ActiveMQ queue to Stockpile. This happens automatically on startup. If an ActiveMQ queue is present, it moves all of the messages out of the queue and into Stockpile. It leaves a "lock" file in place so that there's no need to initialize the ActiveMQ broker and connections upon restart.

@ajroetker did most of this work, I took what he had and adapted it to how stockpile was integrated into PDB

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -22,12 +22,8 @@
 
    * Message queue
 
-     We use an embedded instance of AciveMQ to handle queueing duties
-     for the command processing subsystem. The message queue is
-     persistent, and it only allows connections from within the same
-     VM.
-
-     Refer to `puppetlabs.puppetdb.mq` for more details.
+     We use stockpile to durable store commands. The \"in memory\"
+     representation of that queue is a core.async channel.
 
    * REST interface
 
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -22,7 +22,7 @@
 
    * Message queue
 
-     We use stockpile to durable store commands. The \"in memory\"
+     We use stockpile to durably store commands. The \"in memory\"
      representation of that queue is a core.async channel.
 
    * REST interface
--- a/src/puppetlabs/puppetdb/cli/shovel.clj
+++ b/src/puppetlabs/puppetdb/cli/shovel.clj
@@ -1,0 +1,378 @@
+(ns puppetlabs.puppetdb.cli.shovel
+  (:import [javax.jms
+            ExceptionListener
+            JMSException
+            MessageListener
+            Session
+            TextMessage
+            BytesMessage
+            ConnectionFactory
+            Connection
+            MessageConsumer
+            Queue
+            Message]
+           [java.io File ByteArrayInputStream]
+           [org.apache.activemq.broker BrokerService]
+           [org.apache.activemq.usage SystemUsage MemoryUsage]
+           [org.apache.activemq.pool PooledConnectionFactory]
+           [org.apache.activemq ActiveMQConnectionFactory ScheduledMessage])
+  (:require [clojure.tools.logging :as log]
+            [clojure.java.io :as io]
+            [me.raynes.fs :as fs]
+            [stockpile :as stockpile]
+            [puppetlabs.puppetdb.cheshire :as json]
+            [puppetlabs.i18n.core :as i18n]
+            [puppetlabs.trapperkeeper.config :as config]
+            [puppetlabs.trapperkeeper.logging :as logutils]
+            [puppetlabs.kitchensink.core :as kitchensink]
+            [puppetlabs.puppetdb.config :as conf]
+            [puppetlabs.puppetdb.utils :as utils]
+            [puppetlabs.puppetdb.command.dlo :as dlo]
+            [puppetlabs.kitchensink.core :as ks]
+            [schema.core :as s]))
+
+(def cli-description "Transfer messages from ActiveMQ to Stockpile")
+
+(defn mq-dir [config]
+  (str (io/file (get-in config [:global :vardir]) "mq")))
+
+(defn ^Session create-session [^Connection connection]
+  (.createSession connection false Session/CLIENT_ACKNOWLEDGE))
+
+(defn ^Queue create-queue [^Session session endpoint]
+  (.createQueue session endpoint))
+
+(defn ^Connection create-connection [^ConnectionFactory factory]
+  (doto (.createConnection factory)
+    (.setExceptionListener
+     (reify ExceptionListener
+       (onException [this ex]
+         (log/error ex (i18n/trs "receiver queue connection error")))))
+    (.start)))
+
+(defn extract-headers
+  "Creates a map of custom headers included in `message`, currently only
+  supports String headers."
+  [^Message msg]
+  (into {}
+        (for [k (enumeration-seq (.getPropertyNames msg))]
+          [(keyword k) (.getStringProperty msg k)])))
+
+(defprotocol ConvertMessageBody
+  (convert-message-body ^String [msg]))
+
+(extend-protocol ConvertMessageBody
+  TextMessage
+  (convert-message-body [msg]
+    (-> msg
+        .getText
+        (.getBytes "UTF-8")))
+  BytesMessage
+  (convert-message-body [msg]
+    (let [len (.getBodyLength msg)
+          buf (byte-array len)
+          n (.readBytes msg buf)]
+      (when (not= len n)
+        (throw (Exception. (i18n/trs "Only read {0}/{1} bytes from incoming message" n len))))
+      buf)))
+
+(defn coerce-clj->json-byte-stream
+  "Converts clojure data to JSON serialized bytes on a
+  ByteArrayInputStream. Cheshire will only output to writers and
+  stockpile will only accept input streams, so some conversion needs
+  to be done in this fn"
+  [payload]
+  (let [baos (java.io.ByteArrayOutputStream.)]
+    (with-open [osw (java.io.OutputStreamWriter. baos java.nio.charset.StandardCharsets/UTF_8)]
+      (json/generate-stream payload osw))
+    (.close baos)
+    (-> baos
+        .toByteArray
+        java.io.ByteArrayInputStream.)))
+
+(defn convert-old-command-format
+  "Converts from the command format that didn't include the certname
+  in the toplevel key list and had all the command data dested under
+  payload"
+  [headers message-bytes]
+  (let [{:keys [command version] :as old-command} (json/parse-strict message-bytes true)]
+    {:command command
+     :version (Long/parseLong version)
+     :certname (or (get-in old-command [:payload :certname])
+                   (get-in old-command [:payload :name]))
+     :payload (coerce-clj->json-byte-stream (:payload old-command))}))
+
+(def upgrade-message-schema
+  {:command s/Str
+   :version s/Int
+   :certname s/Str
+   :payload java.io.ByteArrayInputStream})
+
+(s/defn coerce-to-new-command :-  upgrade-message-schema
+  "Commands coming out of this function are suitable for enqueuing
+  into stockpile. They must include command, version, certname and a
+  payload"
+  [{:keys [command version certname] :as headers} message-bytes]
+  (if (and command version certname)
+    {:command command
+     :version (Long/parseLong version)
+     :certname certname
+     :payload (java.io.ByteArrayInputStream. message-bytes)}
+    (convert-old-command-format headers message-bytes)))
+
+(defn discard-amq-message [message-bytes id headers exception dlo]
+  (let [now (ks/timestamp)]
+    (dlo/discard-bytes message-bytes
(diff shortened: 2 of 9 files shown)


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -22,12 +22,8 @@
 
    * Message queue
 
-     We use an embedded instance of AciveMQ to handle queueing duties
-     for the command processing subsystem. The message queue is
-     persistent, and it only allows connections from within the same
-     VM.
-
-     Refer to `puppetlabs.puppetdb.mq` for more details.
+     We use stockpile to durable store commands. The \"in memory\"
```
rbrw: "durably"?

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+(defn coerce-clj->json-byte-stream
+  "Converts clojure data to JSON serialized bytes on a
+  ByteArrayInputStream. Cheshire will only output to writers and
+  stockpile will only accept input streams, so some conversion needs
+  to be done in this fn"
+  [payload]
+  (let [baos (java.io.ByteArrayOutputStream.)]
+    (with-open [osw (java.io.OutputStreamWriter. baos java.nio.charset.StandardCharsets/UTF_8)]
+      (json/generate-stream payload osw))
+    (.close baos)
```
rbrw: Wonder if this close is redundant, and if not, if baos could be moved to another binding in the with-open.

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+    (-> baos
+        .toByteArray
+        java.io.ByteArrayInputStream.)))
+
+(defn convert-old-command-format
```
rbrw: or `old-command->new`

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+  "Converts from the command format that didn't include the certname
+  in the toplevel key list and had all the command data dested under
```
rbrw: dested?

src/puppetlabs/puppetdb/cli/shovel.clj
rbrw: hmm, hadn't seen the coerce function yet, so dunno...

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+
+(defn consume-everything
+  "Attempts to read all available messages from ActiveMQ, giving up
+  after 5 seconds if one is not available"
+  [^MessageConsumer consumer process-message]
+  (try
+    (loop [id 0]
+      ;; We'll wait 5 seconds for a message otherwise we'll assume the Queue is
+      ;; drained
+      (when-let [msg (.receive consumer 5000)]
```
rbrw: As discussed earlier, we may want to double-check that we can tell when we're completely finished with the existing amqp messages since, given the final guard file, we'll never check again.

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+                                             (let [msg-id (.getStringProperty msg
+                                                                              ScheduledMessage/AMQ_SCHEDULED_ID)
+                                                   remove (doto (.createMessage session)
+                                                            (.setStringProperty ScheduledMessage/AMQ_SCHEDULER_ACTION
+                                                                                ScheduledMessage/AMQ_SCHEDULER_ACTION_REMOVE)
+                                                            (.setStringProperty ScheduledMessage/AMQ_SCHEDULED_ID msg-id))]
+                                               (.send producer remove))
+                                             (catch Exception e nil)))]
+          (.send producer request)
+          (Thread/sleep 2000)
```
rbrw: Wondered what this does.

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+                                           (try
+                                             (process-message msg)
+                                             (let [msg-id (.getStringProperty msg
+                                                                              ScheduledMessage/AMQ_SCHEDULED_ID)
+                                                   remove (doto (.createMessage session)
+                                                            (.setStringProperty ScheduledMessage/AMQ_SCHEDULER_ACTION
+                                                                                ScheduledMessage/AMQ_SCHEDULER_ACTION_REMOVE)
+                                                            (.setStringProperty ScheduledMessage/AMQ_SCHEDULED_ID msg-id))]
+                                               (.send producer remove))
+                                             (catch Exception e nil)))]
```
rbrw: Is this just best-effort, or wondered if it might retry, or maybe this could/should discard?

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+  {:pre [((some-fn nil? integer?) megabytes)
+         (fn? usage-fn)]}
+  (when-let [limit (some-> megabytes (* 1024 1024))]
+    (log/info (i18n/trs "Setting ActiveMQ {0} limit to {1} MB" desc megabytes))
+    (-> (.getSystemUsage broker)
+        usage-fn
+        (.setLimit limit)))
+  broker)
+
+(defn ^:dynamic enable-jmx
```
rbrw: As discussed, might want to check to make sure we're not actually firing up amqp jmx if we didn't mean to,

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+
+(defn activemq-connection-factory [spec]
+  (proxy [PooledConnectionFactory java.lang.AutoCloseable]
+      [(ActiveMQConnectionFactory. spec)]
+    (close [] (.stop this))))
+
+;; Questionable code
+
+(defn upgrade-lockfile-path [vardir]
+  (str (io/file vardir "upgraded_mq.lock")))
```
rbrw: Offhand, I'm tempted not to call this a .lock since we're not actually handling it as a real lockfile (wrt the establishment algorithm), maybe just something like "mq-migrated" or whatever.

rbrw: Also might be tempted to leave it a file, or use (get-path ...) if paths work, unless we only need the string later, and pondered passing in the config rather than the vardir.

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+
+(defn needs-upgrade?
+  "Returns true if the user has an unmigrated ActiveMQ director in
+  their vardir"
+  [config]
+  (let [vardir (get-in config [:global :vardir])]
+    (and (fs/exists? (io/file vardir "mq"))
+         (not (fs/exists? (upgrade-lockfile-path vardir))))))
```
rbrw: Check the lock file first?

rbrw: And while I'm fine keeping fs/exists?, I'm half tempted to just say (.exists file) or (Paths/exists path) and avoid the middleman.

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+
+(defn lock-upgrade
+  "Puts a lockfile in vardir to indicate the user has already migrated
+  to Stockpile"
+  [config]
+  (fs/touch (upgrade-lockfile-path (get-in config [:global :vardir]))))
```
rbrw: (If we happen to end up with paths due to any of the above, note that there's a (.toFile path), since iirc reynes/fs doesn't know how to handle paths.)

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,378 @@
...
+                   (build-and-start-broker! "localhost" mq-dir cmd-proc-config))
+        process-message (create-message-processor enqueue-fn dlo)]
+
+    ;; Drain the scheduler so we don't get any extra messages on the Queue when
+    ;; we're processing
+    (create-scheduler-receiver conn-pool mq-endpoint process-message)
+    (create-mq-receiver conn-pool mq-endpoint process-message)
+    (create-mq-receiver conn-pool retired-mq-endpoint process-message)
+
+    ;; Stopping
```
rbrw: Could/should these be protected by some finally blocks?

src/puppetlabs/puppetdb/cli/shovel.clj
senior: I don't think it's redundant, but it looks like it's not needed https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayOutputStream.html#close(). I didn't have it in a with-open because I was using the close to ensure all the data was written to the byte array (flushed etc), then I needed to call the toByteArray below. It seems odd not to close a stream, but I can remove that since it's a no-op for a ByteArrayInputStream

src/puppetlabs/puppetdb/cli/shovel.clj
senior: Yeah, good question, I didn't write that, maybe @ajroetker knows why?

src/puppetlabs/puppetdb/cli/shovel.clj
senior: I think we're good here. I see no other way to determine this. We've closed off all enqueues so all messages will be there on startup, they should all be ready. I am switching to receiveNoWait to remove the 5 second timeout

src/puppetlabs/puppetdb/cli/shovel.clj
senior: Discard seems right, I'll add that

src/puppetlabs/puppetdb/cli/shovel.clj
senior: receiveNoWait must not do what we think it does, switching to that causes my tests to fail, switching back to receive with a 5 second timeout

src/puppetlabs/puppetdb/cli/shovel.clj
rbrw: Hmm, I'd guess that the byte array doesn't do buffering, but good point (and I was more thinking about the fact that the wrapper closes it?); so I suppose I might just first, i.e. (-> (with-open ... (.flush b) (.toByteArray b)) ...) or similar.

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,375 @@
...
+
+      (finally
+        ;; Stopping
+
+        (log/info (i18n/trs "Stopping ActiveMQConnectionFactory"))
+        (when conn-pool
+          (.stop conn-pool))
+
+        (log/info (i18n/trs "Stopping broker"))
+        (when broker
```
rbrw: Wondered if these whens were necessary, or maybe it's handling the case where the constructors fail, and the attempt to use them in the try detonates; in any case, wondered whether nesting finally blocks like this might could make sense (assumes we do need th whens):

    (let [mq-dir (str (io/file (:vardir global-config) "mq"))
          mq-broker-url ...
          mq-discard-dir ...
          mq-endpoint ...
          process-message ...
          id (atom 0)]
      (when-let [conn-pool (activemq-connection-factory mq-broker-url)]
        (try
          (when-let [broker (do ...)]
            (try
              (create-scheduler-receiver conn-pool mq-endpoint process-message dlo id)
              ...
              (finally
                (stop-broker! broker))))
          (finally
            (log/info (i18n/trs "Stopping ActiveMQConnectionFactory"))
            (.stop conn-pool)))))

src/puppetlabs/puppetdb/cli/shovel.clj
rbrw: Reading the toByteArrayDocs, I don't think we need to close or flush, so this is fine.

src/puppetlabs/puppetdb/cli/shovel.clj
senior: I think the real issue here is the broker and the connection pool are constructed outside the try. The whens and nesting I think fall out from that. I'll fix that up.

src/puppetlabs/puppetdb/cli/shovel.clj
```diff
@@ -0,0 +1,375 @@
...
+            Message]
+           [java.io File ByteArrayInputStream]
+           [org.apache.activemq.broker BrokerService]
+           [org.apache.activemq.usage SystemUsage MemoryUsage]
+           [org.apache.activemq.pool PooledConnectionFactory]
+           [org.apache.activemq ActiveMQConnectionFactory ScheduledMessage])
+  (:require [clojure.tools.logging :as log]
+            [clojure.java.io :as io]
+            [me.raynes.fs :as fs]
+            [stockpile :as stockpile]
```
rbrw: This is the stockpile ref I meant earlier.

PR #2055, opened 2016-09-02
(PDB-3029) Test dlo/initialize with existing items

rbrw:
...and fix the related bug.

### DIFF
--- a/src/puppetlabs/puppetdb/command/dlo.clj
+++ b/src/puppetlabs/puppetdb/command/dlo.clj
@@ -17,35 +17,46 @@
 (def command-names (cons "unknown" queue/metadata-command-names))
 
 (defn- cmd-counters
-  "Adds gauges to the dlo for the given category (e.g. \"all\"
-  \"replace command\").  The gauges will pull their values from the
-  dlo atom's :stats."
+  "Adds gauges to the dlo for the given category (e.g. \"global\"
+  \"replace command\")."
   [registry category]
-  (let [dlo-ns "puppetlabs.puppetdb.dlo"]
-    {:count (counters/counter registry [dlo-ns category "messages"])
-     :size (counters/counter registry [dlo-ns category "filesize"])}))
+  (let [dns "puppetlabs.puppetdb.dlo"]
+    {:messages (counters/counter registry [dns category "messages"])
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -49,10 +49,10 @@
   (with-open [path-stream (Files/newDirectoryStream path)]
     (doseq [path path-stream]
       ;; Assume the trailing .json here and in entry-cmd-err-filename below.
-      (when-let [cmd (:command
-                      (and (.endsWith path ".json")
-                           (parse-cmd-filename (-> path .getFileName str))))]
-        (update-metrics metrics cmd (Files/size path))))))
+      (let [name (-> path .getFileName str)]
+        (when-let [cmd (:command (and (.endsWith name ".json")
+                                      (parse-cmd-filename name)))]
```
wkalt: parse-cmd-filename has a regex bug that will prevent this from working -- [0-9+] needs to be swapped with [0-9]+.

test/puppetlabs/puppetdb/command/dlo_test.clj
```diff
@@ -54,11 +93,23 @@
                #(err-attempt-line? 1 %)
                #(= "java.lang.Exception: thud-1" %)]
               (-> (:info discards) .toFile io/reader line-seq)))
-         (is (= content (-> (:command discards) .toFile io/reader slurp))))
-       (is (= 1 (cval ["global" "messages"])))
-       (is (= cmd-size (cval ["global" "filesize"])))
-       (is (= 1 (cval ["replace catalog" "messages"])))
-       (is (= cmd-size (cval ["replace catalog" "filesize"])))))))
+         (is (Arrays/equals content
```
senior: Is there a reason not to compare the content as strings? We have functions that make this easy and if they are different, seems like it'd be easier to find the issue in a string vs. byte array.

senior:
:+1: I like it compared to the atom stuff you had before, leaving it as immutable until everything is read in is nice, then tucking it inside the atom. Just a minor question on comparing the content of two dlo commands in one of the tests, also needs a rebase.

PR #2056, opened 2016-09-02
(PDB-2932) queue metrics

wkalt:
Add stockpile metrics for queue depth and persistence time.


src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -109,6 +122,17 @@
       (swap! metrics assoc-in storage-path
              (create-metrics [(keyword command) (keyword (str version))])))))
 
+(defn initialize-depth!
+  [q]
+  (let [queue-dir (str (.toString (:directory q)) "/q")]
```
senior: I'm not convinced that this is necessary. We already get this information from stockpile on startup, why not hook in there so we don't have to reread all of the files on disk?

resources/public/dashboard/index.html
```diff
@@ -204,9 +204,9 @@
        snag: getValueSnag},
       {description: "Command Queue",
        addendum: "depth",
-       url: "/metrics/v1/mbeans/org.apache.activemq:type=Broker,brokerName=localhost,destinationType=Queue,destinationName=puppetlabs.puppetdb.commands",
+       url: "/metrics/v1/mbeans/puppetlabs.puppetdb.mq%3Aname%3Dglobal.depth",
        format: d3.format(",s"),
-       snag: function(res) { return res["QueueSize"]; }},
+       snag: function(res) { return res["Count"]; }},
```
senior: Why change the name here?

src/puppetlabs/puppetdb/mq.clj
```diff
@@ -17,7 +17,8 @@
 (def mq-metrics-registry (get-in metrics/metrics-registries [:mq :registry]))
 
 (def metrics (atom {:message-persistence-time (timer mq-metrics-registry
-                                                     (metrics/keyword->metric-name [:global] :message-persistence-time))}))
+                                                     (metrics/keyword->metric-name
```
senior: I think we've got some general purging of MQ related dead code. My guess is this will go when that does.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -76,16 +78,22 @@
 (defn create-metrics [prefix]
   (let [to-metric-name-fn #(metrics/keyword->metric-name prefix %)]
     {:processing-time (timer mq-metrics-registry (to-metric-name-fn :processing-time))
-     :retry-persistence-time (timer mq-metrics-registry (to-metric-name-fn :retry-persistence-time))
-     :generate-retry-message-time (timer mq-metrics-registry (to-metric-name-fn :generate-retry-message-time))
+     :retry-persistence-time (timer mq-metrics-registry (to-metric-name-fn
```
senior: I think this goes away

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -76,16 +78,22 @@
...
+                                                          :retry-persistence-time))
+     :generate-retry-message-time (timer mq-metrics-registry (to-metric-name-fn
```
senior: I think this can be removed

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -76,16 +78,22 @@
...
+                                                               :generate-retry-message-time))
+     :message-persistence-time (timer mq-metrics-registry
+                                      (to-metric-name-fn :message-persistence-time))
      :retry-counts (histogram mq-metrics-registry (to-metric-name-fn :retry-counts))
+     :depth (counter mq-metrics-registry (to-metric-name-fn :depth))
      :seen (meter mq-metrics-registry (to-metric-name-fn :seen))
+     :invalidated (counter mq-metrics-registry (to-metric-name-fn :invalidated))
```
senior: :+1:

senior: Wait, is this being used anywhere?

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -201,9 +216,11 @@
              (queue/cmdref->cmd q cmdref)
              retries (count (:attempts cmdref))]
          (try+
-          (call-with-command-metrics command version retries
-                                     #(process-message cmd))
+           (call-with-command-metrics command version retries
+                                      #(process-message cmd))
           (queue/ack-command q cmd)
+          (update-depth command version dec!)
```
senior: Need to update the queue depth when commands are deleted as well

resources/public/dashboard/index.html
senior: Talked with Wyatt about this, this is changing anyway since the prefix isn't including the acivemq broker information, I'm good to go with this.

src/puppetlabs/puppetdb/mq_listener.clj
```diff
@@ -76,16 +79,24 @@
...
      :seen (meter mq-metrics-registry (to-metric-name-fn :seen))
      :processed (meter mq-metrics-registry (to-metric-name-fn :processed))
      :fatal (meter mq-metrics-registry (to-metric-name-fn :fatal))
      :retried (meter mq-metrics-registry (to-metric-name-fn :retried))
      :discarded (meter mq-metrics-registry (to-metric-name-fn :discarded))}))
 
-(def metrics (atom {:global (create-metrics [:global])}))
+(def metrics (atom {}))
+
+(defn clear-counters!
```
senior: Should this be in test.services?

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -144,5 +153,11 @@
...
-(defn sorted-command-buffer [^long n]
-  (SortedCommandBuffer. (TreeMap.) (HashMap.) n))
+(defn sorted-command-buffer
+  ([^long n]
+   (sorted-command-buffer n (constantly nil)))
+  ([^long n ^clojure.lang.IFn delete-update-fn]
+   ;; accepting a function here is a hack to get around a cyclic dependency
+   ;; between this ns, mq-listener.clj, and dlo.clj. My hope is we'll be able
+   ;; to get rid of it somehow when we refactor mq-listener and command.clj.
+   (SortedCommandBuffer. (TreeMap.) (HashMap.) n delete-update-fn)))
```
senior: Bummer. We'll fix this up soon. Another thing to think about here, is much of the information that we're looking for here is already known in the buffer. We don't currently have a threadsafe way to do this, but, if we did, you could just ask for the size of the buffer and you'd know the number of enqueue (as an example). Similarly, walking the datastructure could get the number of invalidated.

test/puppetlabs/puppetdb/testutils/queue.clj
```diff
@@ -3,6 +3,7 @@
             [clojure.java.shell :as shell]
             [me.raynes.fs :refer [delete-dir]]
             [stockpile :as stock]
+            [metrics.counters :refer [clear!]]
```
senior: Is this still needed?

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -7,6 +7,7 @@
             [puppetlabs.puppetdb.testutils.log
              :refer [notable-pdb-event? with-log-suppressed-unless-notable]]
             [puppetlabs.puppetdb.scf.storage-utils :as sutils]
+            [metrics.counters :refer [clear!]]
```
senior: Once clear-counters! is moved here, this will need to stay?

src/puppetlabs/puppetdb/mq_listener.clj
wkalt: yeah, maybe. We might end up wanting it for HUP stuff depending on what we do there, but happy to move it.

src/puppetlabs/puppetdb/queue.clj
wkalt: Agreed -- it's a more attractive approach than the counters.

test/puppetlabs/puppetdb/testutils/services.clj
wkalt: I was just anticipating the comment :-)

PR #2057, opened 2016-09-07
(PDB-3043) Fix admin-cleanup-test race condition

senior:
Travis seems to be particularly good at exposing this test race
condition. The symptom is a noisy but passing
`puppetlabs.puppetdb.admin-clean-test/admin-clean-competition`. The test
finishes, but the database connection pool is shutdown before
the cleanup finishes. This then causes the next test
`puppetlabs.puppetdb.admin-clean-test/admin-clean-status` to fail.


rbrw:
Looks good.  Wonder if this should go on stable/master?

PR #2060, opened 2016-09-09
(maint) Update PDB to use new signing key

melissa:


melissa:
This won't work until we rebuilt the debian repos with the new signing key. But I'm not sure this ever worked, because apt.puppetlabs.com/debian does not exist

melissa:
cc/ @wkalt are you releasing any pdb versions with pre-ezbake automation? If so, we'll likely need to update that branch to use the new key

melissa:

melissa:
@wkalt I have a feeling this change will not impact tests, as the url to pubkey.gpg was wrong anyway

melissa:
That, and the key provided by pubkey.gpg will always be the key in use for the debian repos. If this runs before we've updated to the new key, it may just reimport the old key, which should be fine

wkalt:
This was just an update to some dead test data. I've removed the file now, closing the PR.

PR #2062, opened 2016-09-12
(PDB-3047) Corrective change counts

kbrezina:
This PR subdivides `successes` and `noops` counts into intentional and corrective parts in `event-counts` and `aggregate-event-counts` endpoints for PE. Events are mapped to the corresponding counts based on value of `corrective_change` flag.

The response formats for FOSS are not changed.

### DIFF
--- a/documentation/api/query/v4/aggregate-event-counts.markdown
+++ b/documentation/api/query/v4/aggregate-event-counts.markdown
@@ -71,6 +71,25 @@ event-count results that were aggregated.
         "total": 3
     } ]
 
+#### Puppet Enterprise
+
+In PE, the `successes` and `noops` counts are subdivided into intentional and corrective parts.
+Events are mapped to the corresponding counts based on the value of `corrective_change` flag.
+
+    [ {
+        "summarize_by": "containing_class",
+        "intentional_successes": 2,
+        "corrective_successes": 0,
+        "failures": 0,
+        "intentional_noops": 0,
+        "corrective_noops": 0,
+        "skips": 1,
+        "total": 3
+    } ]
+
+`intentional_successes`, `corrective_successes`, `intentional_noops`, and `corrective_noops` fields
+can be used in `counts_filter` too.
+
 ### Examples
 
 You can use [`curl`][curl] to query information about aggregated resource event counts:
--- a/documentation/api/query/v4/event-counts.markdown
+++ b/documentation/api/query/v4/event-counts.markdown
@@ -130,6 +130,37 @@ When summarizing by `containing_class`, the `subject` will contain a `title` key
       }
     ]
 
+#### Puppet Enterprise
+
+In PE, the `successes` and `noops` counts are subdivided into intentional and corrective parts.
+Events are mapped to the corresponding counts based on the value of `corrective_change` flag.
+
+    [
+      {
+        "subject_type": "certname",
+        "subject": { "title": "foo.local" },
+        "failures": 0,
+        "intentional_successes": 2,
+        "corrective_successes": 0,
+        "intentional_noops": 0,
+        "corrective_noops": 0,
+        "skips": 1
+      },
+      {
+        "subject_type": "certname",
+        "subject": { "title": "bar.local" },
+        "failures": 1,
+        "intentional_successes": 0,
+        "corrective_successes": 0,
+        "intentional_noops": 0,
+        "corrective_noops": 0,
+        "skips": 1
+      }
+    ]
+
+`intentional_successes`, `corrective_successes`, `intentional_noops`, and `corrective_noops` fields
+can be used in `counts_filter` too.
+
 ### Examples
 
 You can use [`curl`][curl] to query information about resource event counts:
--- a/test/puppetlabs/puppetdb/http/aggregate_event_counts_test.clj
+++ b/test/puppetlabs/puppetdb/http/aggregate_event_counts_test.clj
@@ -10,66 +10,71 @@
                                                         query-response
                                                         query-result
                                                         vector-param]]
-            [puppetlabs.puppetdb.testutils.reports :refer [store-example-report!]]))
+            [puppetlabs.puppetdb.testutils.reports :refer [with-corrective-change
+                                                           without-corrective-change
+                                                           store-example-report!]]))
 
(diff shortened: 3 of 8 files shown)


kbrezina:
@wkalt @senior Could you please take a look?

documentation/api/query/v4/event-counts.markdown
```diff
@@ -60,7 +61,13 @@ This endpoint builds on top of the [`events`][events] endpoint, and supports all
 ### Response format
 
 The response is a JSON array of maps. Each map contains the counts of events that matched the input
-parameters. The events are counted based on their statuses: `failures`, `successes`, `noops`, `skips`.
+parameters. The events are counted based on their statuses and corrective_change flags (PE only):
+`failures` (status = 'failure'),
+`successes` (status = 'success' and corrective_change = false),
+`noops` (status = 'noop' and corrective_change = false),
```
wkalt: FOSS users will never have corrective_change = false, so we might want to make this a little bit more clear. I also think it would be preferable to return null for the corrective_* fields in FOSS rather than 0, since 0 isn't really the truth (they may well have corrective noops, but they won't be reflected in the count). I understand that you can't get nulls out of postgres's count function -- might be worth some kind of special flag for this.

wkalt:
To the questions -- this doesn't require an API version bump because the affected endpoints are marked experimental. The store-corrective-change atom gets flipped when the PE service starts, and is fine to use in tests, but it's probably best to have a test case when it's on and another when it's off.

wkalt:

documentation/api/query/v4/event-counts.markdown
kbrezina: @wkalt What kind of flag do you mean? A special URL parameter? That has been discussed earlier. I had the impression we wanted to avoid that.

Maybe, you mean something else here.

wkalt: No, I mean a flag within the PDB service that dispatches on the atom you asked about and either replaces the 0 counts with null in FOSS, or just removes the keys entirely. I don't think there's any point in exposing the flag to users.

kbrezina: @wkalt I seem to be still missing something. Is there an existing atom that would hold a is-pe? value in PDB? Or, do you think I should create such an atom? How can I recognize PE environment?

wkalt: @kbrezina That atom here is scf-store/store-corrective-change. In FOSS it's false, in PE it's switched to true on initialization of the service here: https://github.com/puppetlabs/pe-puppetdb-extensions/blob/master/src/puppetlabs/pe_puppetdb_extensions/reports.clj#L61-63

kbrezina:
@wkalt I've changed the PR based on the discussion we had yesterday. The response formats for FOSS are not changed now. Could you take a look?

Tests for `@store-corrective-change? = false` still need to be added.

test/puppetlabs/puppetdb/http/aggregate_event_counts_test.clj
```diff
@@ -10,215 +10,250 @@
...
 
 (deftest-http-app query-with-environment
   [[version endpoint] endpoints
    method [:get :post]]
-
-  (store-example-report! (:basic reports) (now))
-  (store-example-report! (assoc (:basic2 reports)
+  (with-corrective-change
+    (store-example-report! (:basic reports) (now))
+    (store-example-report! (assoc (:basic2 reports)
```
wkalt: indentation is off here

test/puppetlabs/puppetdb/http/event_counts_test.clj
```diff
@@ -256,354 +290,450 @@
...
+              :subject {:type "Notify" :title "hi"}}}))))
+
+(deftest-http-app query-with-environment
+  [[version endpoint] endpoints
+   method [:get :post]]
+
+  (with-corrective-change
+    (store-example-report! (:basic reports) (now))
+    (store-example-report! (assoc (:basic2 reports)
+                           :certname "bar.local"
```
wkalt: indentation

test/puppetlabs/puppetdb/testutils/reports.clj
```diff
@@ -51,6 +51,26 @@
      (scf-store/add-report!* example-report timestamp update-latest-report?)
      (report-for-hash :v4 report-hash))))
 
+(defmacro with-or-without-corrective-change
+  "Runs body in a context where scf-store/store-corrective-change? is set to true or false based on corrective-change? parameter"
```
wkalt: break this line

documentation/api/query/v4/aggregate-event-counts.markdown
```diff
@@ -71,6 +71,24 @@ event-count results that were aggregated.
         "total": 3
     } ]
 
+#### Puppet Enterprise
+
+In PE, `successes` and `noops` counts are further subdivided into intentional and corrective parts. Events are mapped to the corresponding
+counts based on value of `corrective_change` flag.
```
wkalt: "the `successes` and `noops` counts" ... "based on the value of `corrective_change`"

wkalt: I'd get rid of "further" as well.

documentation/api/query/v4/event-counts.markdown
```diff
@@ -130,6 +130,36 @@ When summarizing by `containing_class`, the `subject` will contain a `title` key
       }
     ]
 
+#### Puppet Enterprise
+
+In PE, `successes` and `noops` counts are further subdivided into intentional and corrective parts. Events are mapped to the corresponding
+counts based on value of `corrective_change` flag.
```
wkalt: same as comment above

documentation/api/query/v4/aggregate-event-counts.markdown
wkalt: also I'd be inclined to break the first line -- looks a little long

wkalt:
@kbrezina left some super minor stylistic notes. This looks good to me otherwise and behaves as expected. Once we have some tests with the atom set to false I'm good to go.

kbrezina:
@wkalt I've fixed the typos. Tests are written with and without support for corrective changes now.

Thanks a lot for your review and valuable feedback. 👍

PR #2063, opened 2016-09-12
(maint) Update PDB to use new signing key

melissa:
We are in the process of introducing a new signing key. This commit
updates PDB to switch from the old key to the new one

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -6,7 +6,7 @@ cows: 'base-precise-i386.cow base-squeeze-i386.cow base-trusty-i386.cow base-whe
 pbuild_conf: '/etc/pbuilderrc'
 packager: 'puppetlabs'
 gpg_name: 'info@puppetlabs.com'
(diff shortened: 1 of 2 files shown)


melissa:
Travis failures seem to be unrelated to this change

melissa:

melissa:
Wait, the only reason we'd ship a new PDB 2.3 is for PE 3.8, right?

If we're shipping to the foss repos, we need to use the new key, but if we're shipping to PE 3.8.x, we need the old key. So, the pe-puppetdb repo for 3.x shouldn't inherit this change

wkalt:
@melissa ok cool, sounds good

melissa:
So, for clarification

1) we need this change in this repo if we ever ship PDB 2.3.x to the foss repos (to use the new key)

2) this change cannot be merged up into pe-puppetdb for PE 3.8.x (to use the old key)

If we're never going to ship another PDB package in the 2.3.x series to the open source repos, this PR can be closed

PR #2065, opened 2016-09-14
(PDB-3039) Safely embed certnames in queue & DLO filenames

aperiodic:
Change `puppetdb.queue/metadata-str` to replace illegal filename
characters and underscores in the certname it's given, truncate the
certname if its length would cause the metadata string to exceed 255
characters, and add the SHA1 hash of the original certname if the
certname in the metadata string differs in any way (via character
replacement or truncation) from the original certname.

Change the *-filename functions in `puppetlabs.command.dlo` to truncate
just the metadata portions of the filenames if they would cause the
filenames to exceed 255 bytes in length. Truncating the metadata portion
is acceptable since all these filenames contain stockpile IDs, which
ensures they'll be unique.

See the individual commits & messages for more details.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -69,7 +69,8 @@
             [puppetlabs.puppetdb.utils :as utils]
             [slingshot.slingshot :refer [try+ throw+]]
             [puppetlabs.puppetdb.mq-listener :as mql]
-            [puppetlabs.puppetdb.command.constants :refer [command-names]]
+            [puppetlabs.puppetdb.command.constants
+             :refer [command-names supported-command-versions]]
             [puppetlabs.trapperkeeper.services
              :refer [defservice service-context]]
             [schema.core :as s]
@@ -109,19 +110,6 @@
     ~@body
     (catch Throwable e#
       (throw+ (fatality e#)))))
-
-(defn version-range [min-version max-version]
-  (set (range min-version (inc max-version))))
-
-(def supported-command-versions
-  {"replace facts" (version-range 2 5)
-   "replace catalog" (version-range 4 9)
-   "store report" (version-range 3 8)
-   "deactivate node" (version-range 1 3)})
-
-(def latest-catalog-version (apply max (get supported-command-versions "replace catalog")))
-(def latest-report-version (apply max (get supported-command-versions "store report")))
-(def latest-facts-version (apply max (get supported-command-versions "replace facts")))
 
 (defn- die-on-header-payload-mismatch
   [name in-header in-body]
--- a/src/puppetlabs/puppetdb/command/dlo.clj
+++ b/src/puppetlabs/puppetdb/command/dlo.clj
@@ -6,8 +6,7 @@
    [puppetlabs.i18n.core :refer [trs]]
    [puppetlabs.kitchensink.core :refer [timestamp]]
    [puppetlabs.puppetdb.nio :refer [copts copt-atomic copt-replace oopts]]
-   [puppetlabs.puppetdb.queue :as queue
-    :refer [cmdref->entry metadata-str parse-cmd-filename]]
+   [puppetlabs.puppetdb.queue :as q]
    [puppetlabs.stockpile.queue :as stock])
   (:import
    [java.nio.file AtomicMoveNotSupportedException
@@ -48,7 +47,7 @@
               ;; entry-cmd-err-filename below.
               (let [name (-> p .getFileName str)]
                 (if-let [cmd (:command (and (.endsWith name ".json")
-                                            (parse-cmd-filename name)))]
+                                            (q/parse-cmd-filename name)))]
                   (update-metrics (ensure-cmd-metrics metrics registry cmd)
                                   cmd
                                   (Files/size p))
@@ -140,7 +139,7 @@
   [cmdref stockpile dlo]
   (let [{:keys [path registry metrics]} dlo
         {:keys [command received attempts]} cmdref
-        entry (cmdref->entry cmdref)
+        entry (q/cmdref->entry cmdref)
         cmd-dest (.resolve path (entry-cmd-data-filename entry))]
     ;; We're going to assume that our moves will be atomic, and if
     ;; they're not, that we don't care about the possibility of
@@ -173,7 +172,7 @@
   ;; indicator that the unknown message may be complete.
   (let [{:keys [path registry metrics]} dlo
         digest (digest/sha1 [bytes])
-        metadata (metadata-str received "unknown" 0 digest)
+        metadata (q/metadata-str received "unknown" 0 digest)
         cmd-dest (.resolve path (str id \- metadata))]
     (Files/write cmd-dest bytes (oopts []))
     (let [info-dest (store-failed-command-info id metadata "unknown"
--- a/src/puppetlabs/puppetdb/command/dlo.clj
+++ b/src/puppetlabs/puppetdb/command/dlo.clj
@@ -14,8 +14,6 @@
     FileAlreadyExistsException Files LinkOption]
    [java.nio.file.attribute FileAttribute]))
 
-(def command-names (cons "unknown" queue/metadata-command-names))
-
 (defn- cmd-counters
   "Adds gauges to the dlo for the given category (e.g. \"global\"
   \"replace command\")."
--- a/src/puppetlabs/puppetdb/command/dlo.clj
+++ b/src/puppetlabs/puppetdb/command/dlo.clj
@@ -7,6 +7,7 @@
    [puppetlabs.kitchensink.core :refer [timestamp]]
    [puppetlabs.puppetdb.nio :refer [copts copt-atomic copt-replace oopts]]
    [puppetlabs.puppetdb.queue :as q]
+   [puppetlabs.puppetdb.utils :refer [utf8-length utf8-truncate]]
    [puppetlabs.stockpile.queue :as stock])
   (:import
    [java.nio.file AtomicMoveNotSupportedException
@@ -37,6 +38,13 @@
     metrics
     (assoc metrics cmd (cmd-counters registry cmd))))
 
+(defn- parse-cmd-filename
+  [filename]
+  (let [parse-metadata (q/metadata-parser (cons "unknown" q/metadata-command-names))
+        id-meta-split-rx #"([0-9]+)-(.*)"]
+    (when-let [[_ id qmeta] (re-matches id-meta-split-rx filename)]
+      (parse-metadata qmeta))))
+
 (defn- metrics-for-dir
   "Updates metrics to reflect the discarded commands directly inside
   the path; does not recurse."
@@ -44,10 +52,10 @@
   (with-open [path-stream (Files/newDirectoryStream path)]
     (reduce (fn [metrics p]
               ;; Assume the trailing .json here and in
-              ;; entry-cmd-err-filename below.
+              ;; entry-cmd-data-filename below.
               (let [name (-> p .getFileName str)]
                 (if-let [cmd (:command (and (.endsWith name ".json")
-                                            (q/parse-cmd-filename name)))]
+                                            (parse-cmd-filename name)))]
                   (update-metrics (ensure-cmd-metrics metrics registry cmd)
                                   cmd
                                   (Files/size p))
@@ -68,23 +76,36 @@
                           (.printStackTrace exception out))
                         attempts))))
 
+(defn- strip-metadata-suffix
+  [queue-metadata-str]
+  (str/replace queue-metadata-str #"\.json$" ""))
+
+(defn- dlo-filename
+  [id metadata suffix]
+  (str id \-
+       (utf8-truncate metadata (- 255
+                                  (count (str id))
+                                  1 ; dash after id
+                                  (utf8-length suffix)))
+       suffix))
+
 (defn- entry-cmd-data-filename
-  "Returns a string representing the filename for the given stockpile
-  `entry`. The filename is similar to what stockpile would store the
-  message as '10291-1469469689-cat-4-foo.org.json'"
+  "Returns a string representing the filename for the given stockpile `entry`.
+  The filename is similar to what stockpile would store the message as
+  '10291-1469469689_cat_4_foo.org.json', but in contrast 'unknown' is an allowed
+  command name."
   [entry]
   (let [id (stock/entry-id entry)
-        meta (stock/entry-meta entry)]
-    (str id \- meta)))
+        metadata (stock/entry-meta entry)]
+    (dlo-filename id metadata "")))
 
-(defn- err-filename
-  "Creates a filename string for error metadata associated the
+(defn- entry-cmd-err-filename
+  "Creates a filename string for error metadata associated with the
   stockpile message at `id`. This filename is similar to how stockpile
   would store it but includes a error stuffix to differentiate it,
-  similar tot he following '10291-1469469689-cat-4-foo.org-err.txt'"
+  similar to the following '10291-1469469689_cat_4_foo.org_err.txt'"
   [id metadata]
-  (assert (str/ends-with? metadata ".json"))
-  (str id \- (subs metadata 0 (- (count metadata) 5)) "-err.txt"))
+  (dlo-filename id (strip-metadata-suffix metadata) "_err.txt"))
 
 (defn- store-failed-command-info
   [id metadata command attempts dir]
@@ -92,12 +113,13 @@
   ;; we want the command because id isn't unique by itself (given
   ;; unknown commands).
   (let [tmp (Files/createTempFile dir
-                                  (str "tmp-err-" id \- command) ""
+                                  ;; no need to truncate, implementation will
+                                  (str "tmperr-" id \- command) ""
                                   (make-array FileAttribute 0))]
     ;; Leave the temp file if something goes wrong.
     (with-open [out (java.io.PrintWriter. (.toFile tmp))]
       (write-failure-metadata out attempts))
-    (let [dest (.resolve dir (err-filename id metadata))
+    (let [dest (.resolve dir (entry-cmd-err-filename id metadata))
           moved? (try
                    (Files/move tmp dest (copts [copt-atomic]))
                    (catch FileAlreadyExistsException ex
@@ -130,8 +152,8 @@
   "Stores information about the failed `stockpile` `cmdref` to the
   `dlo` (a Path) for later inspection.  Saves two files named like
   this:
-    10291-1469469689-cat-4-foo.org.json
-    10291-1469469689-cat-4-foo.org-err.txt
+    10291-1469469689_cat_4_foo.org.json
+    10291-1469469689_cat_4_foo.org_err.txt
   The first contains the failed command itself, and the second details
   the cause of the failure.  The command will be moved from the
   `stockpile` queue to the `dlo` directory (a Path) via
@@ -161,8 +183,8 @@
   later inspection.  `attempts` must be a list of {:time t :exception
   ex} maps in reverse chronological order.  Saves two files named like
   this:
-    10291-1469469689-unknown-0-BYTESHASH
-    10291-1469469689-unknown-0-BYTESHASH.txt
+    10291-1469469689_unknown_0_BYTESHASH
+    10291-1469469689_unknown_0_BYTESHASH.txt
   The first contains the bytes provided, and the second details the
   cause of the failure.  Returns {:info Path :command Path}."
   [bytes id received attempts dlo]
--- a/src/puppetlabs/puppetdb/constants.clj
+++ b/src/puppetlabs/puppetdb/constants.clj
@@ -1,0 +1,8 @@
+(ns puppetlabs.puppetdb.constants)
+
+(def filename-forbidden-characters
+  [\/ \u0000 \\ \:])
+
+(def sha1-hex-length
+  "The length (in ASCII/UTF-8 bytes) of a SHA1 sum when encoded in hexadecimal."
+  40)
--- a/src/puppetlabs/puppetdb/queue.clj
+++ b/src/puppetlabs/puppetdb/queue.clj
@@ -8,7 +8,8 @@
             [puppetlabs.stockpile.queue :as stock]
             [clj-time.coerce :as tcoerce]
             [puppetlabs.puppetdb.cheshire :as json]
-            [puppetlabs.puppetdb.command.constants :as constants]
+            [puppetlabs.puppetdb.command.constants :as command-constants]
+            [puppetlabs.puppetdb.constants :as constants]
             [metrics.timers :refer [timer time!]]
             [metrics.counters :refer [inc!]]
             [clj-time.core :as time]
@@ -17,10 +18,11 @@
             [clojure.core.async :as async]
             [clojure.core.async.impl.protocols :as async-protos]
             [puppetlabs.puppetdb.nio :refer [get-path]]
-            [puppetlabs.puppetdb.utils :refer [match-any-of]]))
+            [puppetlabs.puppetdb.utils :refer [match-any-of utf8-length
+                                               utf8-truncate]]))
 
 (def metadata-command-names
-  (vals constants/command-names))
+  (vals command-constants/command-names))
 
 (defn stream->json [^InputStream stream]
   (try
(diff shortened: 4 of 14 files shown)


src/puppetlabs/puppetdb/command.clj
```diff
@@ -69,7 +69,7 @@
             [puppetlabs.puppetdb.utils :as utils]
             [slingshot.slingshot :refer [try+ throw+]]
             [puppetlabs.puppetdb.mq-listener :as mql]
-            [puppetlabs.puppetdb.command.constants :refer [command-names]]
+            [puppetlabs.puppetdb.command.constants :as cmd-consts]
```
rbrw: I suspect this was mine originally -- for what it's worth, both here, and with respect to the q/foo changes, I sometimes prefer the shorter names because I find the prefixes "noisy" unless they're part of the meaning (or disambiguating) (say for stock/open stock/reduce).  For me, something like `cmdref->entry` is clear enough, and past a point, prefixes can make the code somewhat harder to read.

rbrw: (Though I realize the refer lists might be tedious to maintain, so it's not remotely categorical.)

aperiodic: I'm more attached to the `q/foo` changes, since I found it hard to remember what's in the DLO vs. what's in the queue namespace, and the `q/` alias is very short. I'd be fine with dropping the new alias in here, though, if that's the consensus.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -6,16 +6,14 @@
...
-    :refer [cmdref->entry metadata-str parse-cmd-filename]]
+   [puppetlabs.puppetdb.queue :as q]
+   [puppetlabs.puppetdb.utils :refer [utf8-length utf8-truncate]]
    [puppetlabs.stockpile.queue :as stock])
   (:import
    [java.nio.file AtomicMoveNotSupportedException
     FileAlreadyExistsException Files LinkOption]
    [java.nio.file.attribute FileAttribute]))
 
-(def command-names (cons "unknown" queue/metadata-command-names))
```
rbrw: This might not be critical, but I'd originally wanted to limit the knowledge about "unknown" to the dlo.  i.e. anywhere else it wouldn't be correct.

src/puppetlabs/puppetdb/command.clj
rbrw: I think I'd probably also lean against some of the changes just wrt the general issue of introducing potential merge conflicts (for lines that wouldn't otherwise be changing), but if there ends up being a consensus in favor, no worries.

rbrw: Actually I might take that back a bit - looks like some of the changes I was thinking about perhaps aren't "additional" because the namespace is changing (because they weren't refers to begin with); so dunno.

src/puppetlabs/puppetdb/command/dlo.clj
aperiodic: Yeah. This PR maintains that.

aperiodic: Sorry, this one doesn't exactly, but the next one (#2067) adds stricter behavior around acceptable command names such that "unknown" will throw an error with the default queue metadata serializer.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -30,15 +32,78 @@
...
+      (let [name-hash (kitchensink/utf8-string->sha1 certname)]
+        (format "%d_%s_%d_%s_%s.json"
+                recvd-long command version safe-certname name-hash)))))
 
 (defn- metadata-rx [valid-commands]
   (re-pattern (str
                "([0-9]+)_("
-               (str/join "|" (map #(format "(?:%s)" (re-quote-replacement %))
-                                  valid-commands))
+               (str/join "|" (map #(format "\\Q%s\\E" %) valid-commands))
```
rbrw: Nice catch wrt re-quote-replacement, and it may not matter, but I wondered if this still might not be the "comprehensive quoting" I'd originally wanted, i.e. say one of the command names contains an "\E".

aperiodic: Come on, that's perverse. Do you really think we're going to have a PuppetDB command with that in the name? I don't think that's worth guarding against.

src/puppetlabs/puppetdb/command.clj
aperiodic: Yeah, I think it's reasonable to keep the :refer in here (esp. since command & command-constants are so closely related so the alias doesn't have much benefit), but the other ones that were already aliases I think should stay aliases.

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: Ahh, ok having a guard elsewhere could be fine too.

src/puppetlabs/puppetdb/queue.clj
rbrw: Oh, it may not matter if we know that the input here is trusted, and it may well be (or we can guard elsewhere), but just happened to notice it while thinking through the behavior.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -60,13 +125,6 @@
 
 (def parse-metadata (metadata-parser))
 
-(def parse-cmd-filename
-  (let [parse-metadata (metadata-parser (cons "unknown" metadata-command-names))]
-    (fn [s]
-      (let [rx #"([0-9]+)-(.*)"]
-        (when-let [[_ id qmeta] (re-matches rx s)]
-          (parse-metadata qmeta))))))
```
aperiodic: @rbrw FWIW, keeping "unknown" a DLO thing is part of why I moved this function over to the DLO namespace.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -343,6 +345,29 @@
...
      :certname certname :payload (json/->RawJsonString payload)}))
 
+(defn utf8-bytes [s] (.getBytes ^String s "UTF-8"))
+
+(defn utf8-length
+  "Return the length in bytes of the given string when encoded in UTF-8"
+  [s]
+  (count (.getBytes ^String s "UTF-8")))
+
+(def ^Charset utf8-charset (Charset/forName "UTF-8"))
```
rbrw: Might just grab it from here instead: https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -343,6 +345,29 @@
...
+
+(defn utf8-truncate
+  "Truncate the given string such that its UTF-8 representation is at most
+  `max-bytes` long. Note that the returned string may be empty."
+  [s max-bytes]
+  (let [utf8-bytebuff (ByteBuffer/allocate max-bytes)
+        chars-in (CharBuffer/wrap (chars (.toCharArray ^String s)))
```
rbrw: Wondered if we might not need the chars call here.

src/puppetlabs/puppetdb/utils.clj
rbrw: Also, might move it up, and use it in the length and bytes calls.

src/puppetlabs/puppetdb/utils.clj
aperiodic: Yup, it's superfluous. I'll remove it.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -343,6 +345,29 @@
...
+        encoder (.newEncoder utf8-charset)]
+    (.encode encoder chars-in utf8-bytebuff true)
```
rbrw: Haven't thought hard about it yet, but wondered if we might want to sanity check the CoderResult.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -343,6 +345,29 @@
...
+    (.flush encoder utf8-bytebuff)
+    (let [utf8-bytes (->> (.array utf8-bytebuff)
+                       (remove zero?)
+                       byte-array)]
```
rbrw: I'm not sure, but I think we might be able to replace this with something like (String. bytebuf 0 (.position bytebuf) utf8-charset), assuming the encoder puts position right at the end of the valid bytes.

aperiodic: Oh, interesting. I'll try that out.

src/puppetlabs/puppetdb/constants.clj
```diff
@@ -0,0 +1,4 @@
+(ns puppetlabs.puppetdb.constants)
+
+(def filename-forbidden-characters
```
rbrw: OK, so at first I was a little wary of this, since it's not the unicode characters we're worried about, it's the possibility that the corresponding "ascii" bytes might show up in the utf-8 encoding of some unicode string, but maybe it's actually fine given the way utf-8 works (thinking out loud...), i.e. because the 7-bit ascii range "passes-through", covering "/" and ":", and the only null byte in standard utf-8 is for unicode null itself?  If that's right, then this approach is perhaps OK for anything in the 7-bit range, including null, but might or might not be for any other chars.

(Also had no idea bout MUTF-8 https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8)

aperiodic: We don't have to worry about UTF-8 encoding accidentally producing ASCII bytes. It's one of the design goals of UTF-8 that it doesn't: all multi-byte sequences that encode bytes > 127 have the high bit set in every byte, so they'll never contain ASCII bytes.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -30,15 +32,78 @@
...
+(defn sanitize-certname
+  "Replace any underscores and filename forbidden characters found in `certname`
+  with dashes."
+  [certname]
+  (let [forbidden-chars (conj constants/filename-forbidden-characters
+                              "_")]
+    (str/replace
+      certname
+      (re-pattern (str "("
+                       (->> forbidden-chars
```
rbrw: Since at the moment, we only use forbidden-chars here, I wondered if we might even consider just putting the right regex here directly, i.e. #"[/\_:]" or similar.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -30,15 +32,78 @@
...
+                           (map #(format "\\Q%s\\E" %)) ; escape regex chars
+                           (str/join "|"))
+                       ")"))
+      "-")))
+
+(defn max-certname-length
```
rbrw: Wondered if we could rearrange this so that we just go ahead and format the string prefix that we know we have to have, and then convert that to utf-8, count it, to provide our lower limit, then generate the remainder to fit the available space and concatenate the two.  Seems like that might save some allocation/computation.

src/puppetlabs/puppetdb/queue.clj
aperiodic: Eh, I think that diminishes intent for no real gain.

rbrw: Well, the gain would be trivial code here, but agree that there are competing considerations.

src/puppetlabs/puppetdb/queue.clj
rbrw: i.e. (ish)
```
(let [prefix (format-... )
      prefix-u8 (.getBytes prefix "UTF-8")
      remaining (- max-len (count prefix-u8))
      suffix-u8 (render-suffix ... remaining)]
  (concat-byte-arrays prefix-u8 suffix-u8))
```

src/puppetlabs/puppetdb/queue.clj
aperiodic: Since we do it in a couple places in this namespace alone, should we have a function for going from a collection of strings/characters to a regex (or perhaps regex string fragment to pass to re-pattern) that matches those strings/characters literally?

rbrw: Sounds plausible to me, and then it might even be justifiable to try to get it "exactly right", i.e. more like what I was (not) thinking when I used re-quote-replacement in the first place, and that might be as trivial as figuring out if there's an escape for \E within \Q \E (or just reject \E).

rbrw: Oh, and I take it back (my C-f failed me), looks like the var is at least used in the tests too.

src/puppetlabs/puppetdb/queue.clj
aperiodic: I highly doubt that operating on small strings that can fit entirely in the L1 cache is ever going to be a CPU bottleneck when there's disk & network IO going on, nor do I see how they're going to exert significant memory pressure when the average command body is way larger. I'd prefer to keep this as clear as possible until I see some data or a compelling argument that is ever going to be more than a blip on CPU or memory usage.

src/puppetlabs/puppetdb/constants.clj
rbrw: Agreed, that's what I was just thinking through -- we're fine as long as we don't make any non-ascii-7-bit chars "special".

src/puppetlabs/puppetdb/queue.clj
rbrw: Got it - I see that now.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -71,31 +76,43 @@
                           (.printStackTrace exception out))
                         attempts))))
 
+(defn- strip-metadata-suffix
+  [queue-metadata-str]
+  (str/replace queue-metadata-str #"\.json$" ""))
```
rbrw: Perhaps (subs s 0 (- (count s) 5)?

rbrw: or (subs s 0 (- (count s) (count extension))

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -71,31 +76,43 @@
...
+
 (defn- entry-cmd-data-filename
   [entry]
-  ;; 10291-1469469689-cat-4-foo.org.json
+  ;; 10291_1469469689_cat_4_foo.org.json
```
rbrw: *almost* - stockpile uses (and has always used) dashes as its separator, so it'll be "10291-1469469689_cat_4_foo.org.json"

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -71,31 +76,43 @@
...
   (let [id (stock/entry-id entry)
-        meta (stock/entry-meta entry)]
-    (str id \- meta)))
+        metadata (stock/entry-meta entry)]
+    (str id \_
```
rbrw: Needs to remain a dash to be compatible with stockpile.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -71,31 +76,43 @@
...
+         (utf8-truncate metadata (- 255
```
rbrw: Not sure how yet, but feel like we might want to abstract this a bit more, i.e. so that everyone just calls (constrain-metadata ...) or similar, and it worries about the details.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -71,31 +76,43 @@
...
 
 (defn- store-failed-command-info
   [id metadata command attempts dir]
   ;; We want the metdata and command so we don't have to reparse, and
   ;; we want the command because id isn't unique by itself (given
   ;; unknown commands).
   (let [tmp (Files/createTempFile dir
-                                  (str "tmp-err-" id \- command) ""
+                                  ;; no need to truncate, implementation will
+                                  (str "tmp-err_" id \_ command) ""
```
rbrw: Must be dash here too.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -71,31 +76,43 @@
...
+                                    (count (str id)))))))
 
-(defn- err-filename
+(defn- entry-cmd-err-filename
   [id metadata]
-  ;; 10291-1469469689-cat-4-foo.org-err.txt
-  (assert (str/ends-with? metadata ".json"))
-  (str id \- (subs metadata 0 (- (count metadata) 5)) "-err.txt"))
+  ;; 10291_1469469689_cat_4_foo.org_err.txt
+  (str id \_
```
rbrw: Dash too.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -171,8 +188,8 @@
   ;; indicator that the unknown message may be complete.
   (let [{:keys [path registry metrics]} dlo
         digest (digest/sha1 [bytes])
-        metadata (metadata-str received "unknown" 0 digest)
-        cmd-dest (.resolve path (str id \- metadata))]
+        metadata (q/metadata-str received "unknown" 0 digest)
+        cmd-dest (.resolve path (str id \_ metadata))]
```
rbrw: Dash, and in the examples above.

src/puppetlabs/puppetdb/command/dlo.clj
aperiodic: Why? That's both harder to understand and more likely to do the wrong thing (if the metadata-str doesn't have the correct suffix).

rbrw: Hmm, well if we need to be worried about it not having the correct suffx, then I'd wonder if we might want to verify that in either approach, but I was originally just thinking about it from the perspective that subs is probably a lot simpler.

I suppose if you wanted to use subs, and if a missing extension was ok, you could guard it with an "if endswith?", but in any case subs vs re isn't a big deal to me either way, since the strings should be fairly short, I just mentioned it offhand.  (With subs, I suspected that the implementation might be more likely to be able to "work from the right" (plus or minus some utf-16 sync issues).  If re can't too, then it might be O(n) vs O(1).)

src/puppetlabs/puppetdb/command/dlo.clj
aperiodic: Ok. Somewhat regrettable for us since it means DLO filenames will have two different kinds of separators, but probably too much work to change it at this point.

src/puppetlabs/puppetdb/command/dlo.clj
aperiodic: Either way it's O(1) because there's an upper limit on the length of a queue metadata string.

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: I'm a little ambivalent - I suppose it does make a clear (and pointless?) distinction between which bits are stockpile and which pdb, but yeah, assuming Ryan didn't pick _ to intentionally differ, it's mostly a "historical" accident; I originally just chose dash as a personal preference.

aperiodic:
@rbrw updated to address comments. I think I addressed pretty much everything besides changing `strip-metadata-suffix`.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -343,6 +360,29 @@
...
+(defn utf8-length
+  "Return the length in bytes of the given string when encoded in UTF-8"
+  [s]
+  (count (.getBytes ^String s utf8)))
+
+(defn utf8-truncate
+  "Truncate the given string such that its UTF-8 representation is at most
+  `max-bytes` long. Note that the returned string may be empty."
+  [s max-bytes]
+  (let [utf8-bytebuff (ByteBuffer/allocate max-bytes)
```
senior: TIL about CharBuffer/wrap and the UTF8 encoder class. :+1: 

One question I have is, should be be truncating the string to n chars _before_ we truncate? I'm just thinking of how this encoder/flush stuff is working. If we have a string that is too long, it could hit that `max-bytes` limit mid char, or even if not mid-char, the byte buffer would be full and flush-result would have no effect, is that right?

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -30,16 +33,79 @@
...
 (defn metadata-str [received command version certname]
-  (format "%d_%s_%d_%s.json"
-          (tcoerce/to-long received) command version certname))
+  (let [certname (or certname "unknown-host")
+        recvd-long (tcoerce/to-long received)
+        safe-certname (embeddable-certname received command version certname)]
+    (if (= safe-certname certname)
+      (format "%d_%s_%d_%s.json" recvd-long command version safe-certname)
+      (let [name-hash (kitchensink/utf8-string->sha1 certname)]
+        (format "%d_%s_%d_%s_%s.json"
```
senior: One issue that I think we should discuss. This PR basically changes a certname if that certname is too long before storing the command in stockpile. That metadata string is not what we're storing in memory, we're storing the pieces that form it. Let's say we have a catalog with a certname that is too long. We would encode it as described, however the "too long" certname would be in memory as the full string. PDB is stopped and started, we receive a new catalog from that same certname, we wouldn't recognize that those two nodes are the same. The one we read from disk would be represented in memory as "<way too...>-<some hash>" and the new one would be represented as "<way too long>".

Although I don't think the above is a big deal, more of a performance optimization related to our "bash in place" code, but would be curious how that complicates things. What got me thinking about this is interning certnames so save memory on enqueued commands.

src/puppetlabs/puppetdb/queue.clj
rbrw: Umm, let's pretend this never happened?  Agreed that it might not matter performance-wise, but regardless I was just wrongheaded here - thinking about a situation that doesn't match what we're *actually* doing (we don't keep the *utf-8* prefix bytes, they're regenerated elsewhere later), and so was handwaving about something that would just be more expensive than what you have.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -30,16 +33,79 @@
...
+(defn truncated-certname-length
+  "Given the received-at time, command, and command version that will be in
+  a metadata string, returns the length (in bytes) to truncate certnames to when
+  they are longer than the `max-certname-length` for the string or they contain
+  characters that must be sanitized. This is less than the string's
+  `max-certname-length` to leave space for the SHA1 hash."
+  [received command version]
+  (- (max-certname-length received command version)
+     1 ; for the additional field separator between certname and hash
+     (utf8-length (kitchensink/utf8-string->sha1 ""))))
```
aperiodic: Sure, sounds reasonable.

src/puppetlabs/puppetdb/utils.clj
aperiodic: The encoder is smart enough to stop if there's not enough room left in the buffer to encode a multi-byte character, as shown [in the tests](https://github.com/puppetlabs/puppetdb/pull/2065/files#diff-17ba5b2b53251cf7d606c36ba5254e2dR115).

src/puppetlabs/puppetdb/command/dlo.clj
rbrw: Fair enough - suppose I was mostly thinking of the constant factors i.e. length of string vs ~4, but in any case, it's not critical, and agreed that the inline subs isn't as clear; perhaps eventually something like this (with a better name?) in utils might be best of both worlds:
```
(defn chomp-string-suffix [s suffix]
  (if-let [n (str/last-index-of s suffix)]
    (subs s 0 n)
    s))
```

src/puppetlabs/puppetdb/queue.clj
aperiodic: I think this is rare enough that we shouldn't worry about optimizing for it. You need to have a certname around 180 characters long before there's a chance that it could get truncated, and even our longest FQDNs around here are only about 60 characters long.

aperiodic:
@senior updated to define the SHA1 hex sum length statically.

src/puppetlabs/puppetdb/utils.clj
rbrw: @senior hmm, i see what you mean, and so far it's not completely clear to me from the docs that the bytebuffer from 0 to position is *guaranteed* to be a valid encoding when flush returns OVERFLOW, but I suspect that for at least an encoding like UTF-8, it will be.

src/puppetlabs/puppetdb/queue.clj
rbrw: Don't know if it matters on this front, but if we end up with any Kanji certnames, the limit may be closer to 30-60.

src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -40,14 +38,21 @@
     metrics
     (assoc metrics cmd (cmd-counters registry cmd))))
 
+(defn- parse-cmd-filename
+  [filename]
+  (let [parse-metadata (q/metadata-parser (cons "unknown" q/metadata-command-names))
```
rbrw: Assuming the compiler' or clojure's not doing anything fancy here, I'd lean toward moving this parser to a top-level binding or similar, so that we don't (re)create it once per filename.

aperiodic: That happens in #2067

PR #2066, opened 2016-09-14
(PDB-3053) Create unique index on certnames.latest_report_id

npwalker:


wkalt:
I'm +1 pending validation in the scale testing environment. This index was previously removed as part of a showstopper bug preventing fast access to the latest report to a node, but makes good sense to me intuitively and should make aggregate-event-counts queries faster.

wkalt:
mentioned in channel, but this should go against stable to get into a davis Z release

wkalt:
failure is transient

senior:
@wkalt @npwalker Where are we at with this? Has it been tested in the scale testing environment? Why are you thinking Davis Z release?

wkalt:
@senior to my knowledge it hasn't been tested yet (aside from the last time the scale env tested it). If it makes a big difference I think it's worth a Z -- I think Nick was seeing an improvement of 10-100x for a customer.

npwalker:
@senior yea we're talking about 5+ minutes to 5 seconds for the query that backs event inspector.  

Any place that joins on latest_report_id to get the latest report is likely missing out on some performance without this index.

kbarber:
@npwalker looks like this needs a rebase.

npwalker:
@kbarber done.

PR #2067, opened 2016-09-15
(PDB-2982) Shorter Command Names in Queue Metadata

aperiodic:
(**NB**: this commit depends on #2065)

Replace `puppetdb.queue/metadata-command-names` with a map called
`metadata-command->puppetdb-comand` that defines the mapping between
the command names that can appear in the queue metadata that forms the
message filenames and the command names used by the rest of PuppetDB.
The inverse mapping is dynamically constructed under the var
`puppetdb-command->metadata-command`.

Replace `puppetdb.queue/metadata-str` with `metadata-serializer`, which
takes a map from PuppetDB command names to the metadata command names
and returns a function with the same signature as `metadata-str`; viz.
taking all the queue metadata components and returning a metadata
string. The only differences are that the returned function looks up the
command name it's given in the translation map and puts the metadata
name it finds there in the metadata string, and that if this lookup
fails then an IllegalArgumentException is thrown.

Bind a metadata serializer with the default translation map of
`queue/puppetdb-command->metadata-command` to `queue/serialize-metadata`

Change `metadata-parser` to take a translation map rather than
a sequential collection comand names, have the returned function perform
the inverse mapping from the metadata command name back to the PuppetDB
command name, and to add a docstring to `metadata-parser`.

Add `parse-metadata` and `serialize-metadata` vars to the
`puppetdb.command.dlo` namespace that add an "unknown" -> "unknown"
entry to both translation maps, since the DLO uses that command name as
a fallback when the message is so garbled that the command name cannot
be recovered. Consequently, change all invocations of
`queue/metadata-str` in the DLO to call `dlo/serialize-metadata`, and
likewise `queue/metadata-parser` to `dlo/parse-metadata`.

Change `puppetdb.command.dlo/discard-cmdref` to create its own queue
metadat string and use `stock/entry` directly in order to create
a stockpile entry for the discarded cmdref, rather than using
`queue/cmdref->entry`, as the latter creates a metadata string using
only the default command names & metadata abbreviations, so it will now
throw an error if the discarded cmdref's command is "unknown".

Update tests to align with the new function names and arguments, as well
as the abbreviated command names in queue metadata strings.

Clarify that `puppetdb.queue/embeddable-certname` and the two
functions it calls to determine the certname length all take the
metadata command name, rather than the command name that the rest of
PuppetDB uses.

Remove a redundant check for a ".json" suffix on the filename from
`puppetdb.command.dlo/metrics-for-dir`, since the metadata parser also
performs this check, and can handle random strings that aren't the form
it expects, so there's no need to do this pre-screening.

### DIFF
--- a/src/puppetlabs/puppetdb/command/dlo.clj
+++ b/src/puppetlabs/puppetdb/command/dlo.clj
@@ -61,8 +61,7 @@
               ;; Assume the trailing .json here and in
               ;; entry-cmd-data-filename below.
               (let [name (-> p .getFileName str)]
-                (if-let [cmd (:command (and (.endsWith name ".json")
-                                            (parse-cmd-filename name)))]
+                (if-let [cmd (:command (parse-cmd-filename name))]
                   (update-metrics (ensure-cmd-metrics metrics registry cmd)
                                   cmd
                                   (Files/size p))
--- a/src/puppetlabs/puppetdb/command/dlo.clj
+++ b/src/puppetlabs/puppetdb/command/dlo.clj
@@ -38,10 +38,17 @@
     metrics
     (assoc metrics cmd (cmd-counters registry cmd))))
 
+(def ^:private parse-metadata
+  (q/metadata-parser (assoc q/metadata-command->puppetdb-command
+                            "unknown" "unknown")))
+
+(def ^:private serialize-metadata
+  (q/metadata-serializer (assoc q/puppetdb-command->metadata-command
+                                "unknown" "unknown")))
+
 (defn- parse-cmd-filename
   [filename]
-  (let [parse-metadata (q/metadata-parser (cons "unknown" q/metadata-command-names))
-        id-meta-split-rx #"([0-9]+)-(.*)"]
+  (let [id-meta-split-rx #"([0-9]+)-(.*)"]
     (when-let [[_ id qmeta] (re-matches id-meta-split-rx filename)]
       (parse-metadata qmeta))))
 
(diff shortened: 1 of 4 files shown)


src/puppetlabs/puppetdb/command/dlo.clj
```diff
@@ -38,10 +38,17 @@
     metrics
     (assoc metrics cmd (cmd-counters registry cmd))))
 
+(def ^:private parse-metadata
```
senior: Finally @rbrw has some company

aperiodic:
2/3 of the EC2 instances in the jenkins FOSS acceptance test passed; the other errored during provisioning.

senior:
@aperiodic I think you might need to rebase to get rid of those commits there were already merged in from your previous PR, I'm still seeing them here

aperiodic:
@senior rebased

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -21,8 +21,17 @@
...
+  ;; note that if there are multiple metadata names for the same command then
+  ;; puppetdb-command->metadata-command (defined below) will have to be modified
+  ;; in order to establish a "preferred" metadata name for each command
+  {"catalog" "replace catalog"
+   "facts" "replace facts"
+   "rm-node" "deactivate node"
+   "report" "store report"})
+
+(def puppetdb-command->metadata-command
+  (into {} (for [[md pdb] metadata-command->puppetdb-command] [pdb md])))
```
senior: If only we had a bidimap...

aperiodic: what's that?

aperiodic: "BIDIrectional map"?

senior: https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/BidiMap.html

PR #2071, opened 2016-09-21
(doc) Fix typo

rlinehan:


rlinehan:
Should this be targeted at stable instead and merged up?

wkalt:
@rlinehan yeah, should be stable. For little stuff like this feel free to just commit directly as well.

PR #2074, opened 2016-09-28
(maint) .travis.yml: require sudo explicitly

rbrw:
We already require it, but haven't had to specify it because it's the
default for puppetlabs/puppetdb on travis ("default for repositories
enabled before 2015").

Without this explicit setting, anyone who tries to set up their own
travis build will end up defaulting to "sudo: false", which won't work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: clojure
 lein: lein2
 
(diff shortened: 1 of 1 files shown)


rbrw:
For reference: https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

PR #2075, opened 2016-09-28
(PDB-3059) Add memoize-resource-identity-hashes opt

rbrw:


kbarber:
Seems to be related to: https://github.com/puppetlabs/pe-puppetdb-extensions/pull/226

PR #2078, opened 2016-09-30
(PDB-3073) Add support for querying resource events by corrective_change flag

kbrezina:
There is a need to query `events` by `corrective_change` flag in PE. This PR enables that.

Currently, there is no INDEX in `resource_events` for the `corrective_change` flag.  Since we always query `events` by `status` and `corrective_change` fields at the same time `resource_events_status_idx` seems to be sufficient. However, a combined index `CREATE INDEX xyz_idx ON resource_events (status, corrective_change)` would be ideal. @senior , @wkalt What is your policy concerning new INDEXes?

### DIFF
--- a/src/puppetlabs/puppetdb/query.clj
+++ b/src/puppetlabs/puppetdb/query.clj
@@ -682,7 +682,7 @@
              ;; these fields allow NULL, which causes a change in semantics when
              ;; wrapped in a NOT(...) clause, so we have to be very explicit
              ;; about the NULL case.
-             [(field :guard #{"property" "message" "file" "line" "containing_class"})]
+             [(field :guard #{"property" "message" "file" "line" "containing_class" "corrective_change"})]
              (if-not (nil? value)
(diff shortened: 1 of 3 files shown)


kbrezina:
@senior @wkalt Could you review this PR?

wkalt:
@kbrezina I'm +1. We don't have a specific policy for adding an index -- it would just be a matter of adding a migration.

As for the question of which index to create, I'm not convinced the multicolumn index is what we want, since it won't get used by queries for corrective_change in isolation. Since corrective_change is expected to be mostly false, a partial index on corrective_change=true might be the most space-efficient option here and would work for isolated queries on corrective_change. It may be worth testing this against the multicolumn index to understand what the tradeoffs are.

Do you want to deal with the index as a separate ticket or should I merge this now?

kbrezina:
@wkalt I agree. This can be merged and the index can be handled separately.

PR #2077, opened 2016-09-29
(PDB-3069) Remove command annotations

aperiodic:
With stockpile, the `:id` and `:received` annotations are now top-level
fields of commands. As these were the only annotations, remove the
annotations field from commands entirely.

PR #2081, opened 2016-09-30
(PDB-3067) Improve command logging

aperiodic:
(**NB:** these commits depend on #2077 and should be merged after it.)

Change the message identifier in the command namespace's log/info
invocations to be a concatenation of the stockpile ID and the timestamp
the message was received at, in order to guarantee uniqueness across
restarts (as the stockpile ID may reset).

Change the overall structure of the logged message in the same form so
it's clear *what* is being logged.

When a command is placed on the queue, log its ID, command name, and
certname at the debug level, similarly to the logging after the command is
processed.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -122,9 +122,12 @@
   (try
     (.acquire write-semaphore)
     (time! (get @metrics :message-persistence-time)
-           (async/>!! command-chan
-                      (queue/store-command
-                        q command version certname command-stream command-callback)))
+           (let [cmd (queue/store-command
+                       q command version certname command-stream command-callback)
+                 {:keys [id received]} cmd]
+             (async/>!! command-chan cmd)
+             (log/debugf (i18n/trs "[%s-%d] '%s' command enqueued for %s")
+                         id (tcoerce/to-long received) command certname)))
     (finally
       (.release write-semaphore)
       (when command-stream
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -138,7 +138,8 @@
     (jdbc/with-transacted-connection' db :repeatable-read
       (scf-storage/maybe-activate-node! certname producer-timestamp)
       (scf-storage/replace-catalog! catalog received))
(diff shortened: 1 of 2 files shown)


senior:
@aperiodic I think this needs a rebase, I'm seeing your commits from the other PR

aperiodic:
@senior rebased

src/puppetlabs/puppetdb/command.clj
```diff
@@ -210,8 +216,8 @@
     (jdbc/with-transacted-connection db
       (scf-storage/maybe-activate-node! certname producer-timestamp)
       (scf-storage/add-report! report received))
-    (log/infof "[%s] [%s] puppet v%s - %s"
-               id
+    (log/infof "[%s-%d] '%s' puppet v%s command processed for %s"
+               id, (tcoerce/to-long received)
```
wkalt: I'm +1 generally, but I don't like this comma :)

senior: Agreed on the comma. Also, these strings need to be wrapped in a trs call.

aperiodic: I typically use a comma when some lines have only one argument and other lines have multiple, but I'm not too attached to it. I'll add the `trs`

aperiodic:
Two out of the three jenkins targets passed; in the third, the EC2 instance failed to provision correctly (SSH attempts all timed out). So let's just call it good.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -122,9 +122,12 @@
...
     (.acquire write-semaphore)
     (time! (get @metrics :message-persistence-time)
-           (async/>!! command-chan
-                      (queue/store-command
-                        q command version certname command-stream command-callback)))
+           (let [cmd (queue/store-command
+                       q command version certname command-stream command-callback)
+                 {:keys [id received]} cmd]
+             (async/>!! command-chan cmd)
+             (log/debugf "[%s-%d] '%s' command enqueued for %s"
```
senior: Missed one trs call here

senior: Missed one trs call here

aperiodic:
@senior added it

PR #2085, opened 2016-10-03
(PDB-3066) Detect queue upgrade end via sentinel

rbrw:
Before beginning to transfer messages from activemq to stockpile, send
an "end-of-line" tracer command that when received later should indicate
the end of the queue.

In the unlikely (and unsupported) downgrade case, older versions of
puppetdb should just reject the sentinel as garbage.

### DIFF
--- a/src/puppetlabs/puppetdb/amq_migration.clj
+++ b/src/puppetlabs/puppetdb/amq_migration.clj
@@ -24,6 +24,7 @@
             [puppetlabs.trapperkeeper.config :as config]
             [puppetlabs.kitchensink.core :as kitchensink]
             [puppetlabs.puppetdb.command.dlo :as dlo]
+            [puppetlabs.puppetdb.mq :as mq]
             [puppetlabs.kitchensink.core :as ks]
             [schema.core :as s]))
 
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/amq_migration.clj
```diff
@@ -329,28 +346,35 @@
         mq-discard-dir (str (io/file mq-dir "discard"))
         mq-endpoint (:endpoint mq-config default-mq-endpoint)
         process-message (create-message-processor enqueue-fn dlo)
-        id (atom 0)]
+        id (atom 0)
+        fail-msg "Unable to receive ActiveMQ messages. Migration of existing messages failed."]
```
wkalt: Would it be better to say "Some ActiveMQ messages failed to transfer" here? The current error sort of gives the impression that nothing happened, which isn't necessarily the case.

wkalt: also we should probably use i18n

src/puppetlabs/puppetdb/amq_migration.clj
```diff
@@ -136,31 +137,36 @@
...
               consumer (.createConsumer session (create-queue session endpoint))]
     (log/info (i18n/trs "Processing messages from the Queue: {0}" endpoint))
-    (consume-everything consumer process-message id)))
+    (mq/send-message! connection endpoint
+                      "\"ignore me\""
+                      {"command" "end-of-line"})
+    (try
+      (loop []
+        (when-let [msg (.receive consumer 120000)]
+          (or (= "end-of-line" (.getStringProperty msg "command"))
```
senior: Seems like this would be a useful thing to log.

src/puppetlabs/puppetdb/amq_migration.clj
```diff
@@ -136,31 +137,36 @@
...
+              producer (.createProducer session (create-queue session endpoint))
               consumer (.createConsumer session (create-queue session endpoint))]
     (log/info (i18n/trs "Processing messages from the Queue: {0}" endpoint))
-    (consume-everything consumer process-message id)))
+    (mq/send-message! connection endpoint
+                      "\"ignore me\""
+                      {"command" "end-of-line"})
+    (try
+      (loop []
+        (when-let [msg (.receive consumer 120000)]
```
senior: Would it be useful here to log something after a period of time? This will cause startup to hang for 2 minutes. I can't think of a reason WHY it would ever take that long, but if it did. There's no indication here what's going on. Any opinions here @wkalt ?

rbrw: I don't know that it's sufficient, but the wait will be preceded by the line:
```
Processing messages from the Queue: puppetlabs.puppetdb.commands
```
But I'm happy to add some kind of timeout-based output if we like.  Were you suggesting it should be per-receive if a timer expires, or periodic for the whole loop, or...?

senior: I think so, not sure what the best way to go about it is. We wait for 2 minutes, but 2 minutes would be very abnormal. I would expect normal situations to get a message immediately, certainly within a few seconds. Maybe we can make one receive call with something like a 10 second wait. If we get nothing, we log that we're going to wait for 2 more minutes (maybe warn?) and invoke receive again? What do you think @wkalt ?

rbrw: @senior ok, sounds fine to me.

rbrw:
Attempted to address the current outstanding concerns.

senior:
:+1: will merge when the pipeline is clear

PR #2087, opened 2016-10-06
(PDB-3077) Add service reload and logback size-based rotation

camlow325:
This PR bumps the Trapperkeeper and lein-ezbake dependencies up to 1.5.1 and 1.0.0, respectively, in order for PuppetDB packages to be built with support for service reload functionality (built on top of SIGHUP).  This PR also includes some logback configuration changes to support moving to logback file-size based log rotation and disabling the logic in ezbake that previously enabled log rotation via logrotate.

### DIFF
--- a/resources/ext/config/logback.xml
+++ b/resources/ext/config/logback.xml
@@ -10,8 +10,8 @@
         <append>true</append>
         <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
             <fileNamePattern>/var/log/puppetlabs/puppetdb/puppetdb-%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
-            <!-- each file should be at most 10MB, keep 90 days worth of history, but at most 1GB total-->
-            <maxFileSize>10MB</maxFileSize>
+            <!-- each file should be at most 200MB, keep 90 days worth of history, but at most 1GB total-->
+            <maxFileSize>200MB</maxFileSize>
             <maxHistory>90</maxHistory>
             <totalSizeCap>1GB</totalSizeCap>
         </rollingPolicy>
@@ -36,8 +36,8 @@
         <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
             <!-- rollover daily -->
             <fileNamePattern>/var/log/puppetlabs/puppetdb/puppetdb-status-%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
-            <!-- each file should be at most 10MB, keep 90 days worth of history, but at most 1GB total-->
-            <maxFileSize>10MB</maxFileSize>
+            <!-- each file should be at most 200MB, keep 90 days worth of history, but at most 1GB total-->
+            <maxFileSize>200MB</maxFileSize>
             <maxHistory>90</maxHistory>
             <totalSizeCap>1GB</totalSizeCap>
         </rollingPolicy>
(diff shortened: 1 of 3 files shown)


camlow325:
I noticed that there are some helpers in the beaker tests that do full stops / starts of the puppetdb service that could be moved over to using `service puppetdb reload` instead, such as restart_puppetdb.  I didn't bother trying to change those in this PR as I wasn't sure if the PuppetDB team thought we might be losing valuable existing test coverage in the process.  Moving some of these to `reload`, though, could provide some performance improvements so I figured I'd point that out as potential future work.

camlow325:
Ping @cprice404 and @jpinsonault for additional review.

camlow325:
In addition to running the Clojure unit tests locally to check for any regressions related to the TK dependency bump, I installed a PuppetDB package built with these changes and manually smoke tested the use of the service `reload` and logback rotation on the puppetdb.log and puppetdb-access.log files.  All seemed to work as I'd expected it to.

resources/ext/config/logback.xml
```diff
@@ -7,10 +7,13 @@
...
-
-        <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
-            <fileNamePattern>/var/log/puppetlabs/puppetdb/puppetdb-%d{yyyy-MM-dd}.log.gz</fileNamePattern>
+        <append>true</append>
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+            <fileNamePattern>/var/log/puppetlabs/puppetdb/puppetdb-%d{yyyy-MM-dd}.%i.log.gz</fileNamePattern>
+            <!-- each file should be at most 10MB, keep 90 days worth of history, but at most 1GB total-->
+            <maxFileSize>10MB</maxFileSize>
             <maxHistory>90</maxHistory>
+            <totalSizeCap>1GB</totalSizeCap>
```
cprice404: Were we going to go ahead and add the snippet for the status debug logging in here while we were messing with this file?  TK-400?

cprice404:
LGTM

camlow325:
@cprice404 I added the status log debugging snippet in 2701971.  Hope you and @jpinsonault might take a peek to see that it looks like what you were expecting.  I'll at least do some smoke testing with this to see that it doesn't break existing logging.

jpinsonault:
A puppetdb-status.log file is created even if nothing has tried to log to it?

camlow325:
> A puppetdb-status.log file is created even if nothing has tried to log to it?

Yeah, apparently.  That's what I saw with the logback config on this PR anyway.

jpinsonault:
Oh that's unfortunate, but not terrible I guess

jpinsonault:
I'm 👍

camlow325:
Rebased this at 26c2a5d.  In the process, I fixed a dumb mistake I'd made where I used .gz in one of the logback xml files as the rotated log file instead of .zip.  .zip is the format that logback uses for its log rotation.

camlow325:
After talking about this further with @cprice404 and @jpinsonault, we decided it would be better to use .gz instead of .zip for the compressed log archives.  That's what PuppetDB was already using anyway - which I'd initially overlooked.  I amended the comments that added the logging segments to switch out the extension.  I'll do a sanity check with a local ezbake build to see that this all seems to be working properly.

camlow325:
In 4a95495, I changed the `maxFileSize` for the different log files from 10 MB to 200 MB, based on an offline conversation with @wkalt and @cprice404.

camlow325:
@wkalt - are you +1 to merge this PR at this point?  Let me know if you think anything else needs to change first.  Thanks!

PR #2088, opened 2016-10-10
(PDB-3071) Log a warning if retired server/port config items are in use.

wkalt:
It's relatively common for users to miss the retirement of server/port
when upgrading from versions of PDB that supported it. This has the
effect of defaulting their server/port to puppetdb:8081, producing agent
run failures that log a puppetdb hostname different from what they
believe they've configured.

This patch logs warnings when server and port are detected in the config
file. I think it would be preferable to remove the defaulting behavior
and throw an error when server_urls is not present, but I don't have a
sense of whether anyone relies on the defaulting today.


wkalt:
these failures are transient

PR #2089, opened 2016-10-10
(PDB-3060) `concurrent-writes` setting

aperiodic:
Change the default value for the concurrent writes configuration setting
from 100 to the minimum of half the CPU coresb (rounded up) and 4.

Add documentation for the command processing `concurrent-writes`
configuration setting to the configuration documentation file, and add
some commented-out lines to the default `config.ini` file that set its
value and explain what it's for.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -620,6 +620,17 @@ tuned based on what you see in the performance dashboard.][dashboard]
 
 This setting defaults to half the number of cores in your system.
 
+### `concurrent-writes`
+
+This sets a limit on the number of threads that can write to the disk at
+any one time. The default value is the smaller number of half the number
+of CPU cores and 4.
+
+If your load is low, your disk is fast (i.e. an SSD), and commands
+aren't being processed quickly enough, then you could increasing this
+value in order to alleviate that, but this is unlikely to be the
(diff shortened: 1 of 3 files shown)


aperiodic:
Rebasing

documentation/configure.markdown
```diff
@@ -620,6 +620,18 @@ tuned based on what you see in the performance dashboard.][dashboard]
...
+
+This sets a limit on the number of threads that can write to the disk at
+any one time. The default value is the smaller number of half the number
+of CPU cores and 4.
+
+If your load is low, your disk is fast (i.e. an SSD), and commands
+aren't being processed quickly enough, then you could increasing this
+value in order to alleviate that, but this is unlikely to be the
+bottleneck for command processing. Note that any value higher than that
+for the command processing `threads` parameter has no effect.
```
wkalt: I don't think this is true.

wkalt: based on the document that went out, on a slow disk having 32 write threads with 8 command processing threads had a significantly different profile/enqueue rate than 8 writer threads.

aperiodic: My misunderstanding. Updated to remove this line and tweak the config.ini comment to not imply this.

PR #2090, opened 2016-10-10
(PDB-3058) version 6 reports incorrectly rejected

wkalt:
This was a schema error introduced with the corrective_change feature.
The main operational impact was that users upgrading from 3.2.x to 4.2
with enqueued data would have their commands fail on restart.

This commit corrects the schema to accept v6 reports, and also does a
bit of test cleanup so we won't have this problem again. Specifically it
consolidates synthesis of old command versions under the example reports
namespace and causes test compilation to fail if the synthetic reports
are invalid.

### DIFF
--- a/test/puppetlabs/puppetdb/command_test.clj
+++ b/test/puppetlabs/puppetdb/command_test.clj
@@ -10,9 +10,14 @@
             [puppetlabs.puppetdb.scf.storage :as scf-store]
             [puppetlabs.puppetdb.scf.storage-utils :as sutils]
             [puppetlabs.puppetdb.catalogs :as catalog]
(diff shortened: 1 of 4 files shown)


test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -10,9 +10,10 @@
             [puppetlabs.puppetdb.scf.storage :as scf-store]
             [puppetlabs.puppetdb.scf.storage-utils :as sutils]
             [puppetlabs.puppetdb.catalogs :as catalog]
-            [puppetlabs.puppetdb.examples.reports :as report-examples]
+            [puppetlabs.puppetdb.examples.reports :refer :all]
```
senior: Why the refer all here?

PR #2091, opened 2016-10-11
(PDB-3077) Update doc reference from logrotate to Logback for rotation

camlow325:
This commit changes a paragraph in the documentation which had referred
to the configuration of logrotate for logging to instead reference the
use of Logback's own file rotation.


camlow325:
@wkalt, @cprice404, and @jpinsonault - This is just a small doc update that I missed in my first PR - #2087.

PR #2092, opened 2016-10-11
(PDB-3092) Create a partial index in resource_events for querying by corrective events

kbrezina:
The main purpose of this index is to help `resource_events` queries with `corrective_change = true` - that is the case of `PE Events` queries.

It's expected that minority of events are corrective, so the partial index wouldn't consume a lot of disk space and improve performance.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1163,6 +1163,11 @@
     "create index fact_values_value_float_idx on fact_values(value_float)"
     "create index fact_values_value_integer_idx on fact_values(value_integer)"))
 
+(defn add-corrective-change-index
+  []
+  (jdbc/do-commands
+    "CREATE INDEX resource_events_status_for_corrective_change_idx ON resource_events (status) WHERE corrective_change"))
+
 (def migrations
   "The available migrations, as a map from migration version to migration function."
   {28 init-through-2-3-8
@@ -1192,7 +1197,8 @@
    49 add-corrective-change-columns
    50 remove-historical-catalogs
    51 fact-values-value-to-jsonb
-   52 resource-params-cache-parameters-to-jsonb})
+   52 resource-params-cache-parameters-to-jsonb
+   53 add-corrective-change-index})
 
 
 (def desired-schema-version (apply max (keys migrations)))


kbrezina:
@wkalt This index has already been discussed in #2078

wkalt:
@kbrezina has the index been tested?

kbrezina:
@wkalt I've tested it locally with `events` queries from `PE Events`. The response times were several times faster when the query had `corrective_change = true` condition and this index was in place.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1163,6 +1163,11 @@
     "create index fact_values_value_float_idx on fact_values(value_float)"
     "create index fact_values_value_integer_idx on fact_values(value_integer)"))
 
+(defn add-corrective-change-index
+  []
+  (jdbc/do-commands
+    "CREATE INDEX resource_events_status_for_corrective_change_idx ON resource_events (status) WHERE corrective_change;"))
```
kbarber: _Nitpick_: is that semi-colon superfluous at the end here?

kbarber: So the goal is to create a new index for 'status', that only maps corrective_change = true cases? So that lookup is faster? Wasn't sure if the index on status was intentional or not, or if you mean corrective_change.

Its no issue if this is true, but you should probably explain the 'why' or specifics like this that might confuse the reviewer in your commit message a little better, so I don't need to ask the question :-).

kbrezina: The purpose of this index is to speed up queries with `corrective_change = true`. Just a small percentage of events should be corrective so the partial index seems to be more appropriate here. When PE queries `resource_events` by `corrective_change` it always specifies `status` too (e.g. `[:and [:= :status "success"] [:= :corrective_change true]]` that is why the index is for `status`. IMHO, a multicolumn index for `status` and `corrective_change` is not needed since `corrective_change` is always true in case of this partial index.

kbrezina: Good catch. The semi-colon is removed now.

wkalt: In https://github.com/puppetlabs/puppetdb/pull/2078 I was envisioning this as an index on corrective_change where corrective_change=true. Part of the idea there was to benefit consumers other than the console. Have we checked to see if this index gives any benefit over that one?

wkalt:
I left a comment on Ken's outdated diff. To summarize, I was wondering if we were missing an opportunity to benefit queries other than the console's by making this index on status instead of corrective_change. I did some testing on a table of 23,942,016 events with a random 5% set to corrective_change=true. Results are here: https://gist.github.com/b671019d324afd5556f1c33a7c59138b

Takeaways are:
* The planner is smart enough to use this index in queries against corrective_change alone (status uninvolved)
* Queries against corrective_change and corrective_change + status both execute faster with Karel's index than with a partial index on corrective_change alone. The second result is expected, the first doesn't make much sense to me but is enough to convince me we're not hurting queries against corrective_change only by making the index on status.

With that in mind I'm +1 on this change.

kbrezina:
@wkalt @kbarber Thanks for your review. 👍 
I'll try to provide more details next time.

kbarber:
@kbrezina no worries mate, thanks for the work, and thanks @wkalt for digging in for proof.

PR #2093, opened 2016-10-12
(PDB-3089) log stockpile enqueues at info level

rbrw:


wkalt:
I'm not completely convinced that this is needed -- seems like it generates a fair amount of noise. If we want it though I think we should swap out the ids being logged in processing with the <id>-<date> string logged in enqueuing in order to make the connection clear.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -190,13 +191,20 @@
    (let [current-time (time/now)
          entry (stock/store q
                             command-stream
-                            (serialize-metadata current-time command version certname))]
+                            (serialize-metadata current-time command version certname))
+         received (kitchensink/timestamp current-time)]
+     (log/infof "enqueued command: %s v%d for %s as %d-%d"
```
ajroetker: `(log/info (i18n/trs`

rbrw:
Think we've handled this issue another way -- closing.

PR #2094, opened 2016-10-12
(PDB-3038) Merge command.clj and mq_listener.clj

senior:
This command takes the puppetlabs.puppetdb.command and
puppetlabs.puppetdb.mq-listener namespaces and puts them together. It
also collapses the two related TK services into a single TK service. As
part of this merge the API for several of the message handler functions
changed which caused test failures (most notably those in
puppetlabs.puppetdb.command-test). This commit also refactors those
tests to use the new function signatures and simplify how those
functions are used.

This commit also fixes an issue with command metrics that were causing
test failures when attempting to run only puppetlabs.puppetdb.command-test


test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -957,17 +993,19 @@
...
-                    attempts (:attempts msg)]
-                (is (= 1 (count calls)))
-                (is (= 1 (count call)))
-                (is (= 1 (count attempts)))
-                (is (re-matches
-                     #"(?sm).*ERROR: could not serialize access due to concurrent update.*"
-                     (-> attempts first :exception full-sql-exception-msg)))))
+            (handle-message (store-command' q new-facts-cmd))
+            (reset! second-message? true)
+            (let [failed-cmdref (async/<!! command-chan)]
```
kbarber: So as discussed, potentially having a timeout here will alleviate any blocking issues in tests, even if it is stupidly big. A stack trace early, will show us the line of code thats failing, but a permanent blocking call, will not show us a trace, so solving a lost command scenario will be harder. Its probably not going to happen, just thinking about how one can defend from this kind of thing.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -317,18 +433,124 @@
...
+
+    (mutils/multitime!
+     [(global-metric :processing-time)
+      (cmd-metric command version :processing-time)]
+
+     (process-command-and-respond! cmd scf-write-db response-chan stats)
+     (mark-both-metrics! command version :processed))
+
+    (queue/ack-command q cmd)
+    (update-counter! :depth command version dec!)))
```
rbrw: Not much pdb can do about it, so mostly just an observation, but I suppose if stock/discard managed to fail in a way that leaves the command in place, depth will be wrong until the next restart.  But stock/discard's pretty unlikely to fail as long as there's no "destination".

src/puppetlabs/puppetdb/command.clj
```diff
@@ -317,18 +433,124 @@
...
+  unhandled and expected to be caught by the caller."
+  [cmdref q scf-write-db response-chan stats]
+  (let [{:keys [command version] :as cmd} (queue/cmdref->cmd q cmdref)
+        retries (count (:attempts cmdref))]
+
+    (create-metrics-for-command! command version)
+
+    (mark-both-metrics! command version :seen)
+    (update! (global-metric :retry-counts) retries)
+    (update! (cmd-metric command version :retry-counts) retries)
```
rbrw: Might be exactly what we want, but just noticed that we bump the retry counts before processing (suppose if we didn't want that, the alternative would be in a finally block or similar).

src/puppetlabs/puppetdb/command.clj
```diff
@@ -317,18 +433,124 @@
...
+     [(global-metric :processing-time)
+      (cmd-metric command version :processing-time)]
+
+     (process-command-and-respond! cmd scf-write-db response-chan stats)
+     (mark-both-metrics! command version :processed))
+
+    (queue/ack-command q cmd)
+    (update-counter! :depth command version dec!)))
+
+(def command-delay-in-ms (* 1000 60 60))
```
rbrw: Not important, but thought about `command-delay-ms`.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -57,42 +57,38 @@
...
+  (close [_]
+    (async/close! command-chan)
+    (async/close! response-chan)
+    (fs/delete-dir (:path dlo))
+    (#'overtone.at-at/shutdown-pool-now! @(:pool-atom delay-pool))))
+
+(defn call-with-message-handler [q]
+  (let [delay-pool (mk-pool)
+        command-chan (async/chan 10)
+        response-chan (async/chan 10)
```
rbrw: I imagine 10 here just because we know we never need more than that in the existing tests?

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -57,42 +57,38 @@
...
-  `(test-msg-handler* ~command ~delay-var ~discard-var *db* ~@body))
+(defrecord CommandHandlerContext [message-handler command-chan dlo delay-pool response-chan q]
+  java.io.Closeable
+  (close [_]
+    (async/close! command-chan)
+    (async/close! response-chan)
+    (fs/delete-dir (:path dlo))
+    (#'overtone.at-at/shutdown-pool-now! @(:pool-atom delay-pool))))
+
+(defn call-with-message-handler [q]
```
rbrw: Maybe we talked about this name?  Wondered about something other than call-with since it's not really calling anything.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -102,92 +98,119 @@
...
+(def default-timeout-in-ms
+  (* 1000 60 5))
+
+(defn take-with-timeout!!
+  "Takes from `port` via <!!, but will throw an exception if
+  `timeout-in-ms` expires"
+  [port timeout-in-ms]
+  (async/alt!!
+    (async/timeout timeout-in-ms)
+    (throw (Exception. (format "Take from channel failed after timeout of '%s' ms was exceeded" timeout-in-ms)))
```
rbrw: Code line's a bit long, maybe wrap it a bit, and could perhaps get away with just "Channel take timed out after '%s' ms" or something.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -75,21 +74,130 @@
...
+  "Create a subtree of metrics for the given command and version (if
+  present).  If a subtree of metrics already exists, this function is
+  a no-op."
+  [command version]
+  (let [storage-path [(keyword (str command version))]]
+    (when (= ::not-found (get-in @metrics storage-path ::not-found))
+      (swap! metrics assoc-in storage-path
+             (create-metrics [(keyword command) (keyword (str version))])))))
+
+(defn inc-cmd-metrics
```
wkalt: preexisting but maybe we should change this to inc-cmd-depth

src/puppetlabs/puppetdb/command.clj
```diff
@@ -75,21 +74,130 @@
...
+;; * `:retried`: the number of commands re-queued for retry
+;;
+;; * `:discarded`: the number of commands discarded for exceeding the
+;;   maximum allowable retry count
+;;
+;; * `:processing-time`: how long it takes to process a command
+;;
+;; * `:retry-counts`: histogram containing the number of times
+;;   messages have been retried prior to suceeding
+;;
```
senior: I didn't change any of this, just moved it over, but I'll add it, maybe reformat some of those comments to not take up so much space

src/puppetlabs/puppetdb/command.clj
```diff
@@ -317,18 +433,124 @@
...
+       (process-delete-cmdref cmdref q scf-write-db response-chan stats)
+       (let [retries (count (:attempts cmdref))]
+         (try+
+          (process-cmdref cmdref q scf-write-db response-chan stats)
+          (catch fatal? obj
+            (mark! (global-metric :fatal))
+            (let [ex (:cause obj)]
+              (log/error
+               (:wrapper &throw-context)
+               (i18n/trs "[{0}] [{1}] Fatal error on L2 attempt {2} for {3}" id command retries certname))
```
senior: I agree. I think the log messages look good just by removing L2, and keeping the rest, so I did that for all of them.

src/puppetlabs/puppetdb/command.clj
rbrw: Whatever we might eventually want, I see that this matches existing behavior.

PR #2095, opened 2016-10-13
(PDB-3083) Handle stockpile/store failure cases

rbrw:


rbrw:
Not sure how/if we want to test this; as yet, I don't see an easy (portable) way to "really" test it.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -183,14 +186,36 @@
...
 
+(defn store-in-stockpile [q command-stream received
+                          command version certname]
+  (try+
+   (stock/store q command-stream
+                (serialize-metadata received command version certname))
+
+   ;; Q: just delete the data for now?
+   (catch [:kind ::unable-to-commit] {:keys [stream-data]}
+     (-> "Unable to commit {0} v{1} for {2} received {3}; cleaning up"
```
senior: This will be double logged if you log it here and rethrow. Would probably be better to rethrow with this better error message as the exception message

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -183,14 +186,36 @@
...
+         (i18n/trs command version (pr-str certname) received)
+         (log/error))
+     (Files/delete stream-data)
+     (throw (:cause &throw-context)))
+
+   ;; Q: just delete the data for now?
+   (catch [:kind ::path-cleanup-failure-after-error] {:keys [path exception]}
+     (->> (i18n/trs "Failed to store {0} v{1} for {2} received {3}; cleaning up"
```
senior: Same comment as above

PR #2097, opened 2016-10-13
(PDB-3077) Upgrade to lein-ezbake 1.1.2, fix Debian startup problems

camlow325:
This commit upgrades the lein-ezbake dependency from 1.0.0 to 1.1.2.
This is needed to fix problems with the PuppetDB service not starting
properly on Debian-based platforms due to a permissions issue around the
use of the service rundir.  The rundir problem was previously introduced
by the bump to lein-ezbake 1.0.0.

In the newer ezbake release, subcommands run with a umask of 027.  This
commit includes a couple of explicit chmod calls in the migration and
ssl-setup subcommands which ensure that any modified config files are
still world-readable.  World-readability is required for these files in
order for the puppetdb service, running as user puppetdb, to be able to
still read the config files, which are owned by root:root.

### DIFF
--- a/resources/ext/cli/config-migration
+++ b/resources/ext/cli/config-migration
@@ -99,6 +99,7 @@ function replaceline {
   tmp="$3.tmp.$(date +%s)"
   sed "s/$1/$(echo "$2" | sed -e 's/[\/&]/\\&/g')/g" "$3" > "$tmp"
   mv "$tmp" "$3"
+  chmod 644 "$3"
 }
 
 # This funtion gives a generic error message if a given path or file isn't found
--- a/project.clj
+++ b/project.clj
@@ -143,7 +143,7 @@
              :ezbake {:dependencies ^:replace [[puppetlabs/puppetdb ~pdb-version]
                                                [org.clojure/tools.nrepl "0.2.3"]]
                       :name "puppetdb"
-                      :plugins [[puppetlabs/lein-ezbake "1.0.0"
+                      :plugins [[puppetlabs/lein-ezbake "1.1.2"
                                  :exclusions [org.clojure/clojure]]]}
              :testutils {:source-paths ^:replace ["test"]}
              :ci {:plugins [[lein-pprint "1.1.1"]]}}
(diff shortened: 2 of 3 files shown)


resources/ext/cli/config-migration
```diff
@@ -99,6 +99,7 @@ function replaceline {
   tmp="$3.tmp.$(date +%s)"
   sed "s/$1/$(echo "$2" | sed -e 's/[\/&]/\\&/g')/g" "$3" > "$tmp"
   mv "$tmp" "$3"
+  chmod 644 "$3"
```
wkalt: @senior @rbrw this seems fine to me -- you see any problem?

rbrw: Hmm, perhaps depends on the broader context, i.e. if replaceline is used on a sensitive file, or one with special permissions, this won't preserve them, and it could also leave the content more vulnerable than 664 until the chmod.

For a moment I also wondered about the embedded sed expansion, but perhaps that's fine.  Though since this is  bash, suppose it might be possible to use a `${parameter//pattern/string}` expansion instead.

Also, don't know if it matters in the current context, but on principle, I'd use mktemp (or something as safe) to create the temp file, i.e. `tmpfile="$(mktemp "foo-XXXXXXX")"`.  mktemp is fairly portable, though if say Linux vs BSD matters, we'll need to make sure we use common args.

camlow325: All good points, thanks.  I'd only been trying to preserve backward compatibility with how this was working before just after package install, allowing the service to still be able to boot properly.  At initial package install, the config files have 644 permissions and are owned by root:root.

Assuming these config files may have sensitive data, it might be worth a broader investigation into ensuring that all config files are installed with service owner permissions and a 0640 mode - along with your additional recommendations about how to ensure that world-accessibility isn't, even briefly, re-granted as these data migrations are occurring.  I'm not planning on following up on this piece further at this point but might be worth pursuing with the Release Engineering team further.  I take it that they may own (or at least had initially written) some of these scripts?

PR #2098, opened 2016-10-17
(MAINT) Bump beaker to 2.50.0

camlow325:
Previously, PuppetDB's Gemfile was picking up beaker version 2.43.0.
This was transitively picking up hocon version 0.9.5.  puppetserver
recently added some variable interpolation to the puppetserver.conf
file.  Some fixes in hocon 1.0 and later are needed in order for the
interpolation to be handled correctly.

This commit bumps the beaker dependency in the Gemfile to ~> 2.50.0,
which transitively will cause test jobs to pick up a newer hocon gem
dependency, >= 1.0.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,7 @@ group :acceptance do
     gem 'beaker', *location_for(beaker_version)
   else
     # use the pinned version
-    gem 'beaker', '~> 2.43.0'
+    gem 'beaker', '~> 2.50.0'
   end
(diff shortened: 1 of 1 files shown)


camlow325:
I ran `bundle install --path ./vendor/bundler` locally with Ruby 1.9.3 and saw that the gems all installed okay.  I haven't run any beaker tests with this locally, though.

PR #2099, opened 2016-10-17
(PDB-2810) Update benchmark sample data

senior:
The sample data included with PuppetDB that is used whenever you invoke
the benchmark tool without a dataset had become stale. Unchanged
resources for reports, producer, catalog_uuid and code_id are recent
additions that were missing from the sample data (and from the
randomization that benchmark does). This commit adds this new data.

### DIFF
--- a/resources/puppetlabs/puppetdb/benchmark/samples/catalogs/host-0.json
+++ b/resources/puppetlabs/puppetdb/benchmark/samples/catalogs/host-0.json
@@ -7943,5 +7943,8 @@
   "certname" : "host-0",
   "hash" : "d3299d0dfa96afc41ae74c430ab1f42df75ebd07",
   "environment" : "production",
-  "version" : "1425333104"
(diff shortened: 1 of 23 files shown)


kbarber:
I've looked through the data, the only comments were:

* the containment paths for each resource are inaccurate (they don't match the resource, they are all just copied from a single example), the discussion we had around this was that it wasn't important 'yet' at least not for load testing, but might affect modelling in a UI if people want to use this data for this purpose.
* to avoid some of that ^^ we probably want to instead persist the method we use to generate data of this nature, ie. orchestrate actual puppet generating the data somehow, so things like containment_path are always correct ... this is for later though :-).

Tested this on puppetdb, and pe-puppetdb - seems to work okay. I presume this is going to create additional load if used now in a PE environment, but this reflects reality, so its 'as it should be'.

Looks good mate, I'm +1 once the build is green.

PR #2104, opened 2016-10-20
(PDB-3082) Add new queue metrics

aperiodic:
Add three new metrics to the mq metrics registry:
  * `size`: a histogram of the size of received commands as reported by
    the HTTP request's Content-Length header.
  * `awaiting-retry`: a count of the number of commands that are
    currently waiting to be re-inserted into the queue.
  * `queue-time`: a histogram of the time in seconds that messages
    waited in the queue before being processed.


senior:
@aperiodic I'm :+1: on this, just needs a rebase

PR #2107, opened 2016-10-21
(PDB-2986) Update non-internationalized logging messages to use i18n

senior:
This PR adds a new logging utilities namespace that makes this conversion process easier. Then the next commit switches everything over to that new namespace.

Marking this as "Don't Merge Yet" as it should go in after our releases.

### DIFF
--- a/src/puppetlabs/puppetdb/admin.clj
+++ b/src/puppetlabs/puppetdb/admin.clj
@@ -47,7 +47,7 @@
       :else
       (http/json-response
        {:kind "conflict"
-        :msg (i18n/trs "Another cleanup is already in progress")
+        :msg (i18n/tru "Another cleanup is already in progress")
         :details nil}
        http/status-conflict))))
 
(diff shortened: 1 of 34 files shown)


src/puppetlabs/puppetdb/utils/logging.clj
```diff
@@ -0,0 +1,57 @@
+(ns puppetlabs.puppetdb.utils.logging
```
kbarber: Any discussions with people, where this might land if it was an okay pattern to repeat for other projects? I was actually considering this kind of thing for hafjell the other day (I swear) and was surprised we didn't already have logging wrappers like this :-).

kbarber:
@senior looks like you need a rebase also mate ^^

wkalt:
looks like this missed migrate.clj

wkalt:
@senior migrate.clj is the only gap I'm seeing. Other than that, looks good to me.

senior:
@wkalt fixed up migrate.clj

rbrw:

rbrw:
Looks good.  I'll merge it after the retest, if not before.

PR #2113, opened 2016-10-31
(PDB-3099) implicit facts->inventory relationship

wkalt:
We had the other half of this covered already.


senior:
:+1: will merge once the release goes out

PR #2114, opened 2016-11-01
(maint) Update beaker to ~3.4, unpin from 2.30.1

kbarber:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,9 @@ else
 end
 
 group :test do
+  # Add test-unit for ruby 2.2+ support (has been removed from stdlib)
+  gem 'test-unit'
+
   # Pinning for Ruby 1.9.3 support
   gem 'json_pure', '~> 1.8'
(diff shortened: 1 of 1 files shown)


senior:
@kbarber looks like the version bump causes gem conflicts

kbarber:
@senior actually, looks like we need to use a newer ruby on the slaves, let me dig in when I have some time.

rbrw:
May not be quite the same thing, but Wyatt just pinned a gem to fix a similar error in stable and master 39f423d3ffad0ba61169e3411cab65587f8c6f81

kbarber:
@rbrw well, for this PR I'll have to remove the pinning (when we get ruby 2) I guess now so thanks.

kbarber:
I've raised a PR for CI here: https://github.com/puppetlabs/ci-job-configs/pull/2039 with the intent to have it switch to ruby 2.3.1.

kbarber:

PR #2117, opened 2016-11-01
(PDB-3154) pdb_support_guide: add stockpile info

rbrw:


documentation/pdb_support_guide.markdown
```diff
@@ -31,15 +31,78 @@ The terminus resides on the Puppet master and redirects agent data to
...
-be submitted to the database.  Commands may not always be processed in
-the order received, but will ususally be handled in roughly the order
-submitted.  Although "stale" commands might be ignored.  For example,
-older fact sets or catalogs that are still in the queue might be
-dropped when newer ones are received for the same certname.
+The commands sent from the terminus to PDB are likely to be (currently
+always) deferred to a filesystem queue for later processing, at which
+point they will be submitted to the database.  Commands may not always
+be processed in the order received, but will ususally be handled in
+roughly the order submitted, though "stale" commands might be ignored.
```
senior: There's a lot of grey area here, and I'm not sure why. These docs are tied to the version of PDB that we are about to release. In that version, they will always go to disk, and we'll not ignore stale commands. The not processed in order stuff, though technically correct, really it can only happen if we are starting up with a backed up queue. Although it might be worth mentioning at some point, the first few sentences doesn't seem like the right place to me.

On the things that we are planning to change in the near future, it might be worth noting at the end or just leaving out entirely (what if we never get around to it?)

documentation/pdb_support_guide.markdown
```diff
@@ -31,15 +31,78 @@ The terminus resides on the Puppet master and redirects agent data to
...
+For example, older fact sets or catalogs that are still in the queue
+might be dropped when newer ones are received for the same certname.
+
+Previous versions of PDB stored deferred messages in ActiveMQ, which
+conglomerated them in opaque binary files inside `vardir`/mq.  Now PDB
+stores the messages via [stockpile], which records each message as a
+normal file inside `vardir`/stockpile/cmd/q.  Whenever PDB starts up
+it will check to see if it has already migrated older messages from
+ActiveMQ to stockpile, and if not, it will migrate them, and make a
+note not to attempt the migration again.
```
senior: Seems like it's worth referencing the specific file we're dropping. Also a note around removing/renaming the AMQ dir if the users run into issues and just want to discard what they had.

documentation/pdb_support_guide.markdown
```diff
@@ -31,15 +31,78 @@ The terminus resides on the Puppet master and redirects agent data to
...
+rough examinations of the queue.  For example, something like this
+
+    find cmd/q | grep -cE '^cmd/q/[0-9]+-[0-9]+_facts_'
+
+should provide a count of "replace facts" commands in the queue, and
+something like this:
+
+    find cmd/q -printf "%s %p\n" | sort -n | tail
+
+should list the largest commands in the queue.
```
senior: Might be worth talking about how they're plain text and can be examined directly

documentation/pdb_support_guide.markdown
```diff
@@ -31,15 +31,78 @@ The terminus resides on the Puppet master and redirects agent data to
...
+
+You can manually remove command files from the queue, but doing so is
+only recommended when PDB is not running, and while you can add files
+to the queue at any time, they won't be noticed until PDB is restarted
+or reloaded (i.e. receives a HUP signal).
```
senior: Might be good to caveat this as not a typical thing to do, be aware of collisions etc

documentation/pdb_support_guide.markdown
```diff
@@ -31,15 +31,78 @@ The terminus resides on the Puppet master and redirects agent data to
...
+
+As a result of this, PDB expects the `vardir` filesystem to be able to
+handle all of your certnames when encoded as UTF-8 as filenames,
+excepting of course, any characters that are translated to dashes as
+mentioned above.
+
+PDB also expects the `vardir` filesystem to be able to handle
+filenames up to a maximum of 255 bytes, although the actual queue
+message filename lengths will depend on your certnames.  Filesystems
+like ext4 and xfs should work fine.
```
senior: Might not be a bad idea to talk about our tmp-* files as well

documentation/pdb_support_guide.markdown
rbrw: Handled in the previous paragraph.

documentation/pdb_support_guide.markdown
rbrw: Added below.

PR #2119, opened 2016-11-07
(PE-18318) Update docs for command_broadcast flag

mullr:
This is being set to true by PE.

### DIFF
--- a/documentation/puppetdb_connection.markdown
+++ b/documentation/puppetdb_connection.markdown
@@ -99,9 +99,10 @@ The default value is false.
 
 **Note: For use with Puppet Enterprise only.**
 
-When using multiple `server_urls`, this flag can be set to `true` to cause commands to be sent to all configured PuppetDB instances.
+When set to `true` in installations using multiple `server_urls`, commands are sent to all configured PuppetDB instances. 
 
-The default value is false.
+In open source Puppet and PE versions earlier than 2016.5, the default setting is `false`. In PE 2016.5 and later, the default setting is `true`. 
+
 
 #### `min_successful_submissions`
 


mullr:
@melamos here's the puppetdb docs change. look ok?

documentation/puppetdb_connection.markdown
```diff
@@ -101,7 +101,7 @@ The default value is false.
 
 When using multiple `server_urls`, this flag can be set to `true` to cause commands to be sent to all configured PuppetDB instances.
 
-The default value is false.
+The default value is `false` when not specified, but Puppet Enterprise 2016.5 and later explicitly configures `command_broadcast` to `true`.
```
wkalt: Would it be clearer to say something like "Defaults to 'false' in FOSS and 'true' in Puppet Enterprise > 2016.5". To me the current wording implies that what PE is doing is something other than defaulting, which is a little confusing.

melamos: How about:

When set to `true` in installations using multiple `server_urls`, commands are sent to all configured PuppetDB instances. 

In open source Puppet and PE versions earlier than 2016.5, the default setting is `false`. In PE 2016.5 and later, the default setting is `true`.

wkalt:
This should ultimately go against stable, but we might want to hold off on merging until the 2016.5 release due to the reference.

mullr:
ok @melamos , updated to use that wording. Any thought about when to push this? doesn't seem like a huge deal for this to go live a little early, but I could be wrong.

melamos:
I don't think we should release information about code changes before the actual product release. Can this PR hang out for a bit and be merged when Everett releases?

mullr:
@melamos sure, that sounds good

senior:
Marking this as "Don't Merge Yet", assuming we still want to wait for mid-December

senior:
@mullr when can this be merged?

PR #2122, opened 2016-11-11
Nov2016 link fixes

nfagerlund:
I went through our most recent broken link report and fixed a bunch of stuff from the PuppetDB docs. I also switched to "magic URLs" for a few cross-project links. (So once a version of PE includes this PuppetDB version, links from here will go to THAT specific PE version, not just /latest/.)


wkalt:
+1, this fixes a lot of stuff

PR #2123, opened 2016-11-11
(maint) update support guide and faq

wkalt:
Update the FAQ to remove dead links to the old catalog duplication page.
Fold some of the information from that duplication page into the support
guide, and link to that instead.


ajroetker:
lgtm

PR #2125, opened 2016-11-12
(MAINT) Remove classname/subprotocol from tests

senior:
This specifically removes the fields from our "call-with-pdb-instance"
tests which leads test output like:

  The [database] classname setting has been retired and will be ignored.
  The [database] subprotocol setting has been retired and will be ignored.


rbrw:
Now why didn't I think of that.

PR #2131, opened 2016-11-18
(PDB-3191) Attempt to flush output before exit

rbrw:
While we may still want to pursue a more coherent exit strategy
eventually (PDB-2053), this hopes to make it less likely that we'll lose
output (diagnostic or otherwise) when we quit via System/exit.

The underlying issue is that the JVM does not necessarily flush stdout
or stderr during a System/exit.

### DIFF
--- a/src/puppetlabs/puppetdb/utils.clj
+++ b/src/puppetlabs/puppetdb/utils.clj
@@ -18,6 +18,20 @@
            [java.nio.charset Charset CharsetEncoder CoderResult StandardCharsets]
            [org.postgresql.util PGobject]))
 
+(defn flush-and-exit [status]
+  "Attempts to flush *out* and *err*, reporting any failures to *err*,
+  if possible, and then invokes (System/exit status)."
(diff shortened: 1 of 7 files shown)


rbrw:
Not sure we want this, but thought I'd propose it for discussion.

wkalt:

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -18,6 +18,20 @@
            [java.nio.charset Charset CharsetEncoder CoderResult StandardCharsets]
            [org.postgresql.util PGobject]))
 
+(defn flush-and-exit [status]
+  "Attempts to flush *out* and *err*, reporting any failures to *err*,
+  if possible, and then invokes (System/exit status)."
+  (let [out-ex (try (flush) nil (catch Exception ex ex))]
```
rbrw: Since I didn't see that documented, I thought it'd be safer to avoid depending on it.

wkalt:
this looks good to me

PR #2132, opened 2016-11-18
(maint) add test-unit gem to gemfile

wkalt:
Add test-unit gem to gemfile. This was apparently removed from ruby
stdlib in 2.2.0, and has become necessary since we've bumped our CI
tests to use ruby 2.3.0.


wkalt:
I haven't seen a failure this week -- seems like @kbarber must have fixed this last weekend.

PR #2142, opened 2016-12-07
(maint) Update beaker to ~3.4, unpin from 2.43.0

camlow325:


camlow325:
This is just a cherry-pick of the commit from #2114 that went into other branches already.

PR #2143, opened 2016-12-07
(PE-17301) Add coarse record count to tk-status response 

mullr:


src/puppetlabs/puppetdb/query/summary_stats.clj
```diff
@@ -5,6 +5,19 @@
...
+   "select count(*) from certnames"
+
+   :catalog_count
+   "select count(*) from catalogs"
+
+   :factset_count
+   "select count(*) from factsets"
+
+   :report_count
+   "select count(*) from reports"})
```
wkalt: this PR is on hold related to perf concerns around this

senior: Marking as do not merge

mullr:
the only thing I can think of here is to build on top of the bucketed summary query and the associated caching mechanism. I tried adding counts to the query, and it seemed to perform the same as before (as expected), but it would mean that this part of the patch needs to be moved over to the PE side so it can use that part of the code.

mullr:
Closing this for now, may revisit it in the future.

PR #2144, opened 2016-12-08
(PDB-3220) Drop resource_events_resource_type_idx

npwalker:
This commit drops the resource_events_resource_type_idx index
because the primary column duplicates the
resource_events_resource_timestamp index and as such we see that
the resource_type_idx is never used in practice.


wkalt:
I've never seen this used either. +1.

senior:
I'm ok with the change, but it seems like it should be targeted at master

npwalker:
@senior okay targeted at master... I could really use a briefing on how I should decide where to target changes.  I seems I always pick the wrong one.

PR #2145, opened 2016-12-08
(maint) Externalize strings in config.clj

ajroetker:


src/puppetlabs/puppetdb/config.clj
```diff
@@ -289,23 +288,27 @@
     (cond
      (nil? vardir)
      (throw (IllegalArgumentException.
-             "Required setting 'vardir' is not specified. Please set it to a writable directory."))
+             (format "%s %s"
+                     (trs "Required setting ''vardir'' is not specified.")
+                     (trs "Please set it to a writable directory."))))
```
senior: I just used (str ...) instead of (format "" ..).

rbrw: I might prefer str or str join if we need to break things up like this, but it's not critical, so I'm fine with a merge either way.

PR #2147, opened 2016-12-15
(PDB-3037) Remove unused activeMQ code

briancain:
This commit makes another attempt to go through the puppetdb codebase
and remove any additional code that is no longer used related to
ActiveMQ. It goes through the `mq` namespace and removes several
duplicated functions that now exist within the `mq-migration` namespace.
It also moves tests over tests for any function that now lives in that
namespace but had tests in the `mq` namespace. Any other code that used
these functions from the mq namesce now use the functions that exist
within the `mq-migration` namespace.

FYI: I still need to double check if I left any unused `:require`'s around in the mq namespace after deleting the code.


briancain:
Tested this by doing:

- `lein test`
- Running the benchmark tool.

If there are other good ways (acceptance tests maybe?) to test this that would be good to know! Just to be sure nothing broke :)

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -428,7 +429,7 @@
   (close-buf! [this]
     (println "Shutting down AMQ...")
     (reset! queue-count 0)
-    (mq/stop-broker! broker)
+    (amq/stop-broker! broker)
```
senior: This is good, @rbrw is working on a PR now to remove this and move it to stockpile.

test/puppetlabs/puppetdb/amq_migration_test.clj
```diff
@@ -183,3 +186,68 @@
...
+      (let [dir (kitchensink/absolute-path (fs/temp-dir "ignore-kahadb-corruption"))
+            broker-name "test"]
+        (try
+          ;; Start and stop a broker, then corrupt the journal
+          (let [broker (build-embedded-broker broker-name dir {})]
+            (start-broker! broker)
+            (stop-broker! broker)
+            (spit (fs/file dir "test" "KahaDB" "db-1.log") "asdf"))
+          ;; Now lets use the more resilient build-and-start-broker!
+          (let [broker (build-and-start-broker! broker-name dir {})]
```
senior: @rbrw any idea why we don't use build-and-start-broker! in the amq-migration code? Reading this test it seems like we should.

briancain:
Rebased.

PR #2148, opened 2016-12-21
(PDB-3228) Allow configuration of the ActiveMQ Broker's MemoryLimit

kilianw:
Same changes as https://github.com/puppetlabs/puppetdb/pull/1949/files (v 3.2) but to 2.3

https://tickets.puppetlabs.com/browse/PDB-3228?filter=-2

the default 64MB.


senior:
Support for this was added in PuppetDB 4.1.0. We don't currently have any plans to release another 2.3.x version of PuppetDB, as that line of releases has been EOL'd.

@kilianw Thanks for the contribution, but I'm going to close this given where we are with the 2.3.x branch. Is there something that is preventing your upgrade to a more recent version of PuppetDB?

PR #2149, opened 2016-12-27
(PDB-3056) pre_suite: fix gem install i18n

senior:
On the Debian 8 hosts, "gem install i18n" is failing like this:

  # gem install i18n -v 0.6.11
  ERROR:  Loading command: install (LoadError)
          /usr/lib/x86_64-linux-gnu/ruby/2.1.0/openssl.so: symbol SSLv2_method, version OPENSSL_1.0.0 not defined in file libssl.so.1.0.0 with link time reference - /usr/lib/x86_64-linux-gnu/ruby/2.1.0/openssl.so
  ERROR:  While executing gem ... (NoMethodError)
      undefined method `invoke_with_build_args' for nil:NilClass

Fix it by upgrading to the latest available openssl.


senior:
@rbrw I found our 3.2.x pipelines failing with the same error that you fixed for PDB-3056. This PR is a cherry-pick of that commit on 3.2.x. There was conflicts so I had to fix it up some.

PR #2152, opened 2016-12-30
(PDB-3240) Convert log event to map for logging predicates

senior:
The 'with-log-suppressed-unless-notable' macro passes it's predicates
log event objects, NOT the Clojure map versions created by the
event->map function. The predicates were written to expect the map
version. This resulted in ALL log messages being filtered out. This
commit makes that conversion in the predicates to ensure we don't
lose related log messages in tests.

### DIFF
--- a/test/puppetlabs/puppetdb/testutils/log.clj
+++ b/test/puppetlabs/puppetdb/testutils/log.clj
@@ -10,7 +10,8 @@
    [ch.qos.logback.core.spi LifeCycle]
    [ch.qos.logback.classic Level Logger]
    [ch.qos.logback.classic.encoder PatternLayoutEncoder]
-   [org.slf4j LoggerFactory]))
+   [org.slf4j LoggerFactory]
+   [ch.qos.logback.classic.spi ILoggingEvent]))
 
 (defmacro with-started
   "Ensures that if a given name's init form executes without throwing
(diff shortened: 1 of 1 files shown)


senior:
@rbrw  I lost a day to trying to track down some issues in the amq migration tests due to it filtering out all of the log events :-(. This is just a quick fix so that we're not ignoring those log entries anymore. We should chat about better fixes for this, probably upstream in TK. Maybe TK should just always convert the Java object to a map as it's easier to work with in Clojure. Another way to attack this would be to add some schemas to event->map and validate the predicates.

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -4,13 +4,15 @@
    [environ.core :refer [env]]
    [puppetlabs.kitchensink.core :as kitchensink]
    [puppetlabs.puppetdb.testutils :refer [temp-file]]
-   [me.raynes.fs :as fs])
+   [me.raynes.fs :as fs]
+   [puppetlabs.trapperkeeper.testutils.logging :as tklog])
```
rbrw: Not used?

rbrw:
lgtm

test/puppetlabs/puppetdb/testutils/log.clj
senior: Good catch, I removed it

PR #2155, opened 2017-01-10
(PDB-3051) benchmark: rm activemq; fix shutdown 

rbrw:


rbrw:
It's possible you might want to look at the individual commits here since they're arranged with some straightforward refactorings first, followed by the more complicated changes -- might be easier to follow.

rbrw:
The individual commit messages might also be helpful.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -472,27 +429,148 @@
...
+                             ;; Do we need to make this longer b/c CI mess?
+                             (.join fill-queue 10000))
+                           false
+                           (catch Exception ex
+                             (println-err ex)
+                             true))]
+                     (fs/delete-dir tmpdir)
+                     (and (not close-error?)
+                          (not join-error?)
+                          (not (.isAlive fill-queue)))))))]
```
briancain: This all seems kind of complicated to place in a let block. I wonder if it would be cleaner to move it outside the let?

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -375,89 +380,41 @@
...
                   messages-per-second (float (/ events-since-last-report time-diff-seconds))]
-              (println "Sending" messages-per-second "messages/s"
-                       "(load equivalent to" (int (/ messages-per-second expected-node-message-rate)) "nodes)")
-              (flush)
+              (println-err
+               (trs
+                "Sending {0} messages/s (load equivalent to {1} nodes)"
+                messages-per-second
+                (int (/ messages-per-second expected-node-message-rate))))
+              (binding [*out* *err*] (flush))
```
senior: `println-err` doesn't this already via println

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -472,27 +429,148 @@
...
+  (println-err (trs "Populating queue in the background"))
+  (doseq [host (random-hosts numhosts pdb-host catalogs reports facts)
+          :while (>!! mq (assoc host :lastrun (rand-lastrun run-interval)))]
+    true)
+  (println-err (trs "Finished populating queue")))
+
+(defn cycle-commands [numhosts run-interval rand-perc simulation-threads
+                      rate-monitor-ch command-send-ch
+                      pdb-host catalogs reports facts]
+  "Populates the message queue with messages, and then cycles through
```
senior: comment needs to be before the arglist

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -472,27 +429,148 @@
...
+  reappears at the head of the queue.  Returns a stop function of no
+  arguments that can be used to effect an orderly shutdown."
+  ;; The stop function is also registered as a shutdown hook (thread),
+  ;; and is set up as a oneshot, protected by an atomic boolean.  All
+  ;; of the work that needs to be completed on a SIGINT, etc. is
+  ;; coordinated from that hook via the stop function because after a
+  ;; signal, only daemon threads, shutdown hooks, and finalizers will
+  ;; still run, and daemon threads will only run until all of the
+  ;; hooks and finalizers are finished.
+  (let [preftmp (get-path (or (System/getenv "TMPDIR")
```
senior: This needs some docs

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -472,27 +429,148 @@
...
+          hosts-per-second (/ numhosts (* run-interval-minutes 60))]
+      (relay-messages-at-rate mq rate-limited-mq-out hosts-per-second)
+      (pipeline simulation-threads
+                mq
+                (map #(update-host % rand-perc (time/now) command-send-ch))
+                rate-limited-mq-out)
+      stop)))
+
+(defn benchmark
+  [args]
```
senior: Docstring needs pulled up

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -472,27 +429,148 @@
...
+        mq (chan (message-buffer message-dir numhosts))
+        rate-limited-mq-out (chan)
+        channels [rate-limited-mq-out mq rate-monitor-ch command-send-ch]
+        fill-queue (doto (Thread. #(populate-queue mq numhosts run-interval
+                                                   pdb-host
+                                                   catalogs reports facts))
+                     (.setDaemon true))
+        shutdown-thread (promise)  ;; Because no letrec...
+        stopping? (AtomicBoolean.)
+        stop (fn []
```
senior: I'll chew on this for a bit, see if I can find a way to shrink it some.

senior: I've spent some time understanding this and I think I see a way to simplify it. I'll chat with you about it when you're around.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -472,27 +429,148 @@
...
+                (map #(update-host % rand-perc (time/now) command-send-ch))
+                rate-limited-mq-out)
+      stop)))
+
+(defn benchmark
+  [args]
+  "Feeds commands to PDB as requested by args.  If --runinterval has
+  been specified, returns a function that can be called to stop
+  processing and clean up.  That function returns true or false to
+  indicate whether the shutdown encountered any problems."
```
ajroetker: I think you want this docstring before the [args] vector

src/puppetlabs/puppetdb/cli/benchmark.clj
rbrw: Good question - just kept the flush from before; perhaps so, assuming it respects *flush-on-newline*.

src/puppetlabs/puppetdb/cli/benchmark.clj
rbrw: I'd mentioned it in load_testing_tool.markdown; perhaps I should also add it to the --help.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -472,27 +431,156 @@
...
+        channels [rate-limited-mq-out mq rate-monitor-ch command-send-ch]
+        stop (delay
+              ;; Currently assumes that it's ok to leave already
+              ;; in-flight "channel takes" running.  i.e. stop
+              ;; precipitates a stop, but doesn't hang around for
+              ;; anything but the queue deletion.  Also, we don't
+              ;; bother to try to close any of the channels during
+              ;; shutdown shutdown because all those (non-daemon)
+              ;; threads should be dead.
+              (let [close-err? (not-every? identity
```
senior: What's the impact of this, seems like nothing to me, since the message is removed from disk before it's dequeued. It either gets dequeued and shutdown before it can submit it, or it gets shutdown after it gets submitted. Either way I think there's no impact right?

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -503,57 +591,48 @@
...
+                                            rate-monitor-ch command-send-ch
+                                            pdb-host catalogs reports facts)])
+                  (catch Exception ex
+                    (deliver result ex))))
+          (.setDaemon true)
+          .start)
+        (let [x @result]
+          (if (vector? x)
+            (first x)
+            (throw x)))))))
```
senior: What's the scenario that x is not a vector?

senior:
@puppetlabs/puppetdb-maintainers need another reviewer for this one

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -472,27 +431,156 @@
...
+                                               catalogs reports facts)))
+                       (.setDaemon true))
+          ;; Use delay as one-shot
+          stop (delay
+                @fill-started ;; So we know it's OK to join
+                (let [result (do (.join fill-queue 500)
+                                 (or (not (.isAlive fill-queue))
+                                     (do
+                                       (println-err (trs "Waiting up to 10s for shutdown"))
+                                       ;; Do we need to make this longer b/c CI mess?
```
rbrw: Hmm, I'm actually leaning the other way (shorter) for the default since (currently) benchmark will be hard to kill during this period (i.e. kill -9) -- I'm also now wondering if we might not need to worry about this one as much; i'll ping you in a bit.

src/puppetlabs/puppetdb/cli/benchmark.clj
rbrw: If you're talking about the question of leaving "in-flight" takes alone, then yeah, I think it should be OK with the current code, but I wanted to mention it so that reviewers would ponder it for a moment too -- in the non-testing case, my spot-check of core.async isDaemon indicated that its threads aren't daemon threads, and so whatever they're doing, they'll be done.  Worst case I might expect some noise from postgresql about a broken transaction/connection.  During testing, they'll keep running (i.e. (stop) doesn't stop them), but that should be a similar story  i.e. worst case they'll end up talking to a db that's been torn down?

src/puppetlabs/puppetdb/cli/benchmark.clj
rbrw: See the two delivers above -- if it's a vector, it's the normal result from cycle-commands, otherwise it's an exception it threw.

rbrw: ...and it's really just a redirection of the daemon-thread's result (exception or not) back onto the current thread, since the only reason for the temporary switch to a daemon thread is so that we won't be unceremoniously plonked by shutdown.

rbrw:
nb. Ryan's investigating among other things, the possibility that using our new queue threadpool infrastructure might simplify the consumer end.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -233,29 +240,46 @@
                        (update-report uuid stamp))
         factset (some-> factset
                         (update-factset rand-percentage stamp))]
-    (when catalog (>!! command-send-ch [:catalog host 9 catalog]))
-    (when report (>!! command-send-ch [:report host 8 report]))
-    (when factset (>!! command-send-ch [:factset host 5 factset]))
+
+    (when catalog
+      (when-not (>!! command-send-ch [:catalog host 9 catalog])
+        (throw+ channel-closed-map)))
```
senior: :+1:

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -503,57 +560,81 @@
...
+
+          create-commands-result (promise)
+          create-command-thread (Thread. (fn []
+                                           (try+
+                                            (create-messages-thunk)
+                                            (deliver create-commands-result [true])
+                                            (catch [:kind ::command-send-channel-closed] _
+                                              (when-not @shutting-down?
+                                                (throw+)
+                                                (deliver create-commands-result [true])))
```
senior: This line is unreachable, I think it just needs to be moved up a line.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -503,57 +560,81 @@
...
+                                     rate-monitor-ch command-send-ch
+                                     pdb-host catalogs reports facts))
+
+          {submit-commands-thunk :run-task
+           stop-command-submission :stop-task} (submit-commands threadpool base-url command-send-ch rate-monitor-ch shutting-down?)
+
+          create-commands-result (promise)
+          create-command-thread (Thread. (fn []
+                                           (try+
+                                            (create-messages-thunk)
```
senior: I think this indentation is off.

senior:
@rbrw :+1: from me, I like the way this came together

@mullr could you take a look at this? @rbrw and I have gone back and forth on it and I think we've put together something that will work.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -56,10 +56,15 @@
...
             [puppetlabs.puppetdb.amq-migration :as amq]
             [taoensso.nippy :as nippy]
-            [puppetlabs.i18n.core :refer [trs]])
-  (:import [org.apache.activemq.broker BrokerService]
-           [javax.jms MessageConsumer MessageProducer Session]
-           [clojure.core.async.impl.protocols Buffer]))
+            [puppetlabs.i18n.core :refer [trs]]
+            [puppetlabs.puppetdb.nio :refer [get-path]]
+            [puppetlabs.puppetdb.threadpool :as tp])
+  (:import
```
mullr: these two lines should be joined

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -221,6 +226,8 @@
              "producer" (random-producer))
       (update "values" (partial randomize-map-leaves rand-percentage))))
 
+(def channel-closed-map {:kind ::command-send-channel-closed})
```
mullr: channel-closed-error?

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -233,29 +240,46 @@
...
+               (loop [mutated-hosts hosts
+                      msgs-to-send num-msgs
+                      stamp (time/minus (time/now) (time/minutes (* 30 num-msgs)))]
+                 (if (zero? msgs-to-send)
+                   (close! command-send-ch)
+                   (let [mutated-hosts (mapv #(update-host % rand-percentage stamp command-send-ch) mutated-hosts)]
+                     (recur mutated-hosts
+                            (dec msgs-to-send)
+                            (time/plus stamp (time/minutes 30)))))))
+   :stop-task (fn [] (close! command-send-ch))
```
mullr: hm. The normal way to stop a loop that's putting stuff onto a channel is to close the channel. Then the loop should check if >! (>!!) returns false and break out.

test/puppetlabs/puppetdb/cli/benchmark_test.clj
```diff
@@ -86,38 +107,38 @@
...
-                                 "--config" "anything.ini"
-                                 "--numhosts" "1"
-                                 "--nummsgs" "10")
-        reports (->> submitted
-                     (filter #(= :report (:entity %)))
-                     (map :payload))]
-    (is (= 10 (->> reports (map :configuration_version) distinct count)))
-    (is (= 10 (->> reports (map :start_time) distinct count)))
-    (is (= 10 (->> reports (map :end_time) distinct count)))))
+(comment (deftest consecutive-reports-are-distinct
```
mullr: don't want these tests anymore?

test/puppetlabs/puppetdb/cli/benchmark_test.clj
```diff
@@ -25,56 +25,77 @@
...
-                                 "--nummsgs" "3")]
+#_(deftest multiple-messages-and-hosts
+  (let [submitted (benchmark-nummsgs {}
+                                     "--config" "anything.ini"
+                                     "--numhosts" "2"
+                                     "--nummsgs" "3")]
     (is (= 18 (count submitted)))))
 
-(deftest archive-flag-works
+#_(deftest archive-flag-works
```
mullr: ?

test/puppetlabs/puppetdb/cli/benchmark_test.clj
```diff
@@ -25,56 +25,77 @@
...
+         (remove-watch submitted watch-key)
+         (is (>= (count @submitted) enough-records))
+         (force-shutdown))))))
 
-(deftest multiple-messages-and-hosts
-  (let [submitted (run-benchmark {}
-                                 "--config" "anything.ini"
-                                 "--numhosts" "2"
-                                 "--nummsgs" "3")]
+#_(deftest multiple-messages-and-hosts
```
mullr: ?

mullr:
@senior @rbrw ok, here's a slightly refined version of the thing Rob and I hacked up. I think it's pretty good. 

https://github.com/mullr/puppetdb/tree/ticket/master/pdb-3051-benchmark-rm-amq-coreasync

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -385,175 +376,186 @@
...
+                                                   catalogs reports facts)
+        _ (start-simulation-loop numhosts run-interval nummsgs rand-perc
+                                 simulation-threads command-send-ch mq-ch)
+        join-fn (fn join-benchmark[]
+                  (async/<!! populate-finished-ch)
+                  (async/<!! command-sender-finished-ch)
+                  (async/<!! rate-monitor-finished-ch))
+        stop-fn (fn stop-benchmark []
+                  (async/close! mq-ch)
+                  ;; todo: timeout here
```
senior: todo ^^^^

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -331,34 +313,43 @@
...
+     rate-monitor-ch
+     (map (fn [[command host version payload]]
+            (let [submit-fn (case command
+                              :catalog client/submit-catalog
+                              :report client/submit-report
+                              :factset client/submit-facts)]
+
+              (try
+                (submit-fn base-url host version payload)
+                ::submitted
```
senior: is this and ::error below used anymore?

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -385,175 +376,186 @@
...
+                                                         (if nummsgs
+                                                           (async/take (* numhosts nummsgs) command-send-ch)
+                                                           command-send-ch)
+                                                         rate-monitor-ch
+                                                         threads)
+        populate-finished-ch (start-populate-queue mq-ch numhosts run-interval pdb-host
+                                                   catalogs reports facts)
+        _ (start-simulation-loop numhosts run-interval nummsgs rand-perc
+                                 simulation-threads command-send-ch mq-ch)
+        join-fn (fn join-benchmark[]
```
senior: missing a space after benchmark

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -385,175 +376,186 @@
...
+        rate-monitor-ch (chan)
+
+        ;; processes
+        rate-monitor-finished-ch (start-rate-monitor rate-monitor-ch run-interval commands-per-puppet-run)
+        command-sender-finished-ch (start-command-sender base-url
+                                                         (if nummsgs
+                                                           (async/take (* numhosts nummsgs) command-send-ch)
+                                                           command-send-ch)
+                                                         rate-monitor-ch
+                                                         threads)
```
senior: Why have a static simulation-threads and a user specified threads?

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -385,175 +376,186 @@
...
+        temp-dir (get-path (or (System/getenv "TMPDIR")
+                               (System/getProperty "java.io.tmpdir")))
+
+        ;; channels
+        mq-ch (chan (message-buffer temp-dir numhosts))
+        _ (.addShutdownHook (Runtime/getRuntime) (Thread. #(async/close! mq-ch)))
+
+        command-send-ch (chan)
+        rate-monitor-ch (chan)
+
```
senior: Since these are unbuffered, could this potentially slow things? Seems like a small buffer of maybe threads might improve things?

senior:
Looks good 👍 , just some minor comments

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -385,175 +376,186 @@
...
+      (.add q path)))
+
   (close-buf! [this]
-    (println "Shutting down AMQ...")
-    (reset! queue-count 0)
-    (amq/stop-broker! broker)
-    (fs/delete-dir mq-dir))
+    (.clear q)
+    (println-err "Cleaning up temp files from" (.toString storage-dir))
+    ;; todo: timeout here
```
mullr: @rbrw thoughts about what this timeout should be, or how it should be implemented?

src/puppetlabs/puppetdb/cli/benchmark.clj
mullr: @rbrw the mechanism here is easy, I can implement that; ideas on the value?

src/puppetlabs/puppetdb/cli/benchmark.clj
mullr: It's not. I'm not sure it ever was, actually. I'm kind of loathe to remove it, as the rate monitor needs /something/ to count.

src/puppetlabs/puppetdb/cli/benchmark.clj
mullr: That's actually why the rate monitor is there; the fundamental limiting factor is the speed at which we can send commands. It'll provide backpressure upstream. The rate monitor should tell us how fast that actually happened.

src/puppetlabs/puppetdb/cli/benchmark.clj
mullr: I think simulation-threads was added later. You want me to make that an arg too?

test/puppetlabs/puppetdb/cli/benchmark_test.clj
```diff
@@ -86,18 +105,18 @@
...
-                                   "--numhosts" "2"
-                                   "--nummsgs" "3"
-                                   "--archive" export-out-file)]
-      (is (= 18 (count submitted))))))
+    (let [submitted (benchmark-nummsgs {}
+                                       "--config" "anything.ini"
+                                       "--numhosts" "2"
+                                       "--nummsgs" "3"
+                                       "--archive" export-out-file)]
+      (is (= (* 2 3 3) (count submitted))))))
```
rbrw: @mullr Worth one step further, i.e. let bind numhosts and nummsgs and use them here and in the args above?  Not sure.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -43,23 +43,27 @@
...
+            [puppetlabs.puppetdb.utils :as utils :refer [println-err]]
             [puppetlabs.kitchensink.core :as kitchensink]
             [clj-time.core :as time]
             [puppetlabs.puppetdb.client :as client]
             [puppetlabs.puppetdb.reports :as reports]
             [puppetlabs.puppetdb.random :refer [random-string random-bool]]
             [puppetlabs.puppetdb.archive :as archive]
             [slingshot.slingshot :refer [try+ throw+]]
-            [clojure.core.async :refer [go go-loop >! <! >!! <!! timeout chan close! dropping-buffer pipeline]]
+            [clojure.core.async :refer [go go-loop >! <! >!! <!! timeout chan close!] :as async]
```
mullr: I kinda hate wrapping require forms

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -375,185 +366,200 @@
...
-(defn chan? [x]
-  (satisfies? clojure.core.async.impl.protocols/Channel x))
-
+                                   (if facts 1 0))
+        temp-dir (get-path (or (System/getenv "TMPDIR")
+                               (System/getProperty "java.io.tmpdir")))
+
+        ;; channels
+        mq-ch (chan (message-buffer temp-dir numhosts))
+        _ (.addShutdownHook (Runtime/getRuntime) (Thread. #(async/close! mq-ch)))
```
mullr: I think I prefer leaving the full thing inline here; we only do it twice, and it makes it clear what's going on.

rbrw: Sounds good - I think the only question would be whether the other code should change anytime this does; if so, then linking them might make more sense, so we don't forget, but not a big deal either way.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -375,185 +366,200 @@
...
+        join-fn (fn join-benchmark
+                  ([] (join-benchmark nil))
+                  ([timeout-ms]
+                   (let [t-ch (if timeout-ms (async/timeout timeout-ms) (chan))]
+                     (async/alt!! t-ch false populate-finished-ch true)
+                     (async/alt!! t-ch false command-sender-finished-ch true)
+                     (async/alt!! t-ch false rate-monitor-ch true))))
+        stop-fn (fn stop-benchmark []
+                  (async/close! mq-ch)
+                  (join-fn 5000))
```
mullr: join (and stop) actually do that :) they return true on successful join, or false on timeout.

PR #2158, opened 2017-01-13
(PDB-3108) Add producer-timestamp to the command header, enable "bash in place"

senior:
This PR adds producer-timestamp to the command POST headers and threads it through the stack. The parameter is used in the "bash in place" code to determine which command is older. This also required modifying the terminus to include that parameter in the POST.

While working on this patch, I found it difficult to thread the new producer-timestamp value through the stack for enqueuing. There's also a commit in this set that refactors the enqueuing functions to pass a "command request" through the stack. That map gets created (and validated) by the service, which is then passed through the stack.

Note to reviewers, it will probably be easier to look at the commits individually.

~~This PR depends on https://github.com/puppetlabs/puppetdb/pull/2152, which needs to go in first.~~

### DIFF
--- a/puppet/lib/puppet/indirector/facts/puppetdb.rb
+++ b/puppet/lib/puppet/indirector/facts/puppetdb.rb
@@ -18,6 +18,9 @@ class Puppet::Node::Facts::Puppetdb < Puppet::Indirector::REST
 
   def save(request)
     profile("facts#save", [:puppetdb, :facts, :save, request.key]) do
+
+      current_time = Time.now
+
       payload = profile("Encode facts command submission payload",
                         [:puppetdb, :facts, :encode]) do
         facts = request.instance.dup
@@ -30,12 +33,12 @@ class Puppet::Node::Facts::Puppetdb < Puppet::Indirector::REST
           # when we attempt to use ActiveSupport 2.3.16 on RHEL 5 with
           # legacy storeconfigs.
           "environment" => request.options[:environment] || request.environment.to_s,
-          "producer_timestamp" => Puppet::Util::Puppetdb.to_wire_time(Time.now),
+          "producer_timestamp" => Puppet::Util::Puppetdb.to_wire_time(current_time),
           "producer" => Puppet[:node_name_value]
         }
       end
 
-      submit_command(request.key, payload, CommandReplaceFacts, 5)
+      submit_command(request.key, payload, CommandReplaceFacts, 5, current_time.clone.utc)
     end
   end
 
--- a/puppet/lib/puppet/indirector/facts/puppetdb.rb
+++ b/puppet/lib/puppet/indirector/facts/puppetdb.rb
@@ -30,7 +30,7 @@ class Puppet::Node::Facts::Puppetdb < Puppet::Indirector::REST
           # when we attempt to use ActiveSupport 2.3.16 on RHEL 5 with
           # legacy storeconfigs.
           "environment" => request.options[:environment] || request.environment.to_s,
-          "producer_timestamp" => request.options[:producer_timestamp] || Time.now.iso8601(5),
+          "producer_timestamp" => Puppet::Util::Puppetdb.to_wire_time(Time.now),
           "producer" => Puppet[:node_name_value]
         }
       end
--- a/puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
+++ b/puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
@@ -8,6 +8,8 @@ require 'puppet/util/puppetdb/command_names'
 require 'json'
 require 'puppet/resource/catalog'
 require 'puppet/version'
+require 'date'
+require 'time'
 
 describe Puppet::Resource::Catalog::Puppetdb do
   before :each do
@@ -25,8 +27,7 @@ describe Puppet::Resource::Catalog::Puppetdb do
     end
     let(:options) {{
       :transaction_uuid => 'abcdefg',
-      :environment => 'my_environment',
-      :producer_timestamp => "a test",
+      :environment => 'my_environment'
     }}
 
     before :each do
@@ -39,10 +40,15 @@ describe Puppet::Resource::Catalog::Puppetdb do
     end
 
     it "should POST the catalog command as a JSON string" do
-      command_payload = subject.munge_catalog(catalog, options).to_json
-
+      before_test_time = Time.now
+      command_payload = subject.munge_catalog(catalog, options)
       http.expects(:post).with do |uri, body, headers|
-        expect(body).to eq(command_payload)
+        req = JSON.parse(body)
+        actual_producer_timestamp = extract_producer_timestamp(req)
+        req.delete("producer_timestamp")
+        command_payload.delete("producer_timestamp")
+        req == command_payload &&
+          actual_producer_timestamp <= Time.now.to_i
       end.returns response
 
       save
--- a/puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
+++ b/puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
@@ -40,15 +40,10 @@ describe Puppet::Resource::Catalog::Puppetdb do
     end
 
     it "should POST the catalog command as a JSON string" do
-      before_test_time = Time.now
-      command_payload = subject.munge_catalog(catalog, options)
+      command_payload = subject.munge_catalog(catalog, Time.now.utc, options)
       http.expects(:post).with do |uri, body, headers|
-        req = JSON.parse(body)
-        actual_producer_timestamp = extract_producer_timestamp(req)
-        req.delete("producer_timestamp")
         command_payload.delete("producer_timestamp")
-        req == command_payload &&
-          actual_producer_timestamp <= Time.now.to_i
+        assert_command_req(command_payload, body)
       end.returns response
 
       save
@@ -367,7 +362,7 @@ describe Puppet::Resource::Catalog::Puppetdb do
         exec{'/bin/true': subscribe => Exec["foo\nbar"]}
         MANIFEST
 
-        expect { subject.munge_catalog(catalog) }.not_to raise_error
+        expect { subject.munge_catalog(catalog, Time.now.utc) }.not_to raise_error
       end
 
       describe "exported resources" do
@@ -396,7 +391,7 @@ describe Puppet::Resource::Catalog::Puppetdb do
           Notify <<| |>>
           MANIFEST
 
-          result = subject.munge_catalog(catalog)
+          result = subject.munge_catalog(catalog, Time.now.utc)
 
           result['edges'].should include(edge)
         end
@@ -412,7 +407,7 @@ describe Puppet::Resource::Catalog::Puppetdb do
           Notify <<| |>>
           MANIFEST
 
-          result = subject.munge_catalog(catalog)
+          result = subject.munge_catalog(catalog, Time.now.utc)
 
           result['edges'].should include(edge)
         end
@@ -427,7 +422,7 @@ describe Puppet::Resource::Catalog::Puppetdb do
           MANIFEST
 
           expect do
-            subject.munge_catalog(catalog)
+            subject.munge_catalog(catalog, Time.now.utc)
           end.to raise_error(Puppet::Error, "Invalid relationship: Notify[source] { before => Notify[target] }, because Notify[target] is exported but not collected")
         end
 
@@ -440,7 +435,7 @@ describe Puppet::Resource::Catalog::Puppetdb do
           notify { target: }
           MANIFEST
 
-          result = subject.munge_catalog(catalog)
+          result = subject.munge_catalog(catalog, Time.now.utc)
 
           result['edges'].should_not include(edge)
         end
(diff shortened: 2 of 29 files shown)


senior:

mullr:
In the commit message for "(PDB-3108) Add producer-timestamp as a command header": typo at "timestmap"

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -103,6 +106,16 @@
       (utf8-truncate sanitized-certname trunc-length)
       sanitized-certname)))
 
+(defn received+producer-ts [received-ts producer-ts]
```
ajroetker: Would you mind adding a docstring to this function? I found this function a little mysterious the first time I saw it :)

puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
```diff
@@ -39,10 +40,14 @@ def save
...
-
+      command_payload = subject.munge_catalog(catalog, Time.now.utc, options)
       http.expects(:post).with do |uri, body, headers|
-        expect(body).to eq(command_payload)
+        req = JSON.parse(body)
+        actual_producer_timestamp = extract_producer_timestamp(req)
+        req.delete("producer_timestamp")
+        command_payload.delete("producer_timestamp")
+        req == command_payload &&
+          actual_producer_timestamp <= Time.now.to_i
```
ajroetker: Looks like this is used a few places, is it worth/possible putting this in a util function or some such?

src/puppetlabs/puppetdb/queue.clj
mullr: Naming gripe: this looks to me like a function that takes two timestamps and encodes them into a string in a way that preserves characters. Can we call it something about 'encoding' in the name, and have a docstring that describes the encoding?

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -112,24 +125,24 @@
...
+         (format "%s_%s_%d_%s.json" received+producer-time short-command version safe-certname)
          (let [name-hash (kitchensink/utf8-string->sha1 certname)]
-           (format "%d_%s_%d_%s_%s.json"
-                   recvd-long short-command version safe-certname name-hash)))))))
+           (format "%s_%s_%d_%s_%s.json"
+                   received+producer-time short-command version safe-certname name-hash)))))))
 
 (def serialize-metadata (metadata-serializer))
 
 (defn metadata-rx [valid-commands]
```
mullr: rx is a prescription, or maybe a reactive extension. regex?

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -112,24 +125,24 @@
...
   (re-pattern (str
-               "([0-9]+)_"
+               "([0-9]+)([+|-][0-9]+)?_"
```
mullr: I guess this is the producer timestamp offset; is it optional because it may be unspecified? If it's nil, is that part of the filename just omitted?

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -147,32 +160,62 @@
...
-                             Long/parseLong
-                             tcoerce/from-long
-                             kitchensink/timestamp)
-               :version (Long/parseLong version)
-               :command (get md-cmd->pdb-cmd md-command "unknown")
-               :certname certname}))))))
+       (when-let [[_ received-stamp producer-offset md-command version certname] (re-matches rx s)]
+         (let [received-time-long (Long/parseLong received-stamp)
+               producer-offset (and producer-offset (Long/parseLong producer-offset))]
+           (and certname
```
mullr: What are the situations where certname could be nil?

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -147,32 +160,62 @@
...
+                {:received (-> received-time-long
+                               tcoerce/from-long
+                               kitchensink/timestamp)
+                 :producer-ts (some-> producer-offset
+                                      (+ received-time-long)
+                                      tcoerce/from-long)
```
mullr: Why isn't this one run through `ks/timestamp` as well?

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -147,32 +160,62 @@
...
+                 :version (Long/parseLong version)
+                 :command (get md-cmd->pdb-cmd md-command "unknown")
```
mullr: An unknown command seems pretty bad; Maybe that should just throw? I can imagine it borking some code downstream, and having an exception closer to the source of the problem could make it easier to diagnose.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -257,18 +301,39 @@
...
-      (when (or (= command-type "replace catalog")
-                (= command-type "replace facts"))
-        (when-let [^CommandRef old-command (.get certnames-map [command-type certname])]
-          (.put fifo-queue
-                (:id old-command)
-                (assoc old-command :delete? true))
-          (delete-update-fn (:command old-command) (:version old-command)))
-        (.put certnames-map [command-type certname] cmdref))
-      (.put fifo-queue (:id cmdref) cmdref))
+      (if (and
```
mullr: The next line wants to be combined with this one. To me, anyway.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -257,18 +301,39 @@
...
+                    (:id maybe-old-command)
+                    (assoc maybe-old-command :delete? true))
+              (.put fifo-queue (:id cmdref) cmdref)
+              (delete-update-fn (:command maybe-old-command) (:version maybe-old-command)))
+
+            :else
+            (do
+              (.put fifo-queue
+                    (:id cmdref)
+                    (assoc cmdref :delete? true))
```
mullr: Why have `delete? true` in this branch? That seems wrong.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -257,18 +301,39 @@
...
+              (delete-update-fn (:command cmdref) (:version cmdref)))))
```
mullr: I can't find a place where delete-update-fn is actually provided; can in be deleted? Or did I (more likely) just miss something?

test/puppetlabs/puppetdb/amq_migration_test.clj
```diff
@@ -18,7 +18,7 @@
             [puppetlabs.trapperkeeper.services :refer [service-context]]
             [puppetlabs.kitchensink.core :as ks]
             [metrics.counters :as mc]
-            [puppetlabs.trapperkeeper.testutils.logging :refer [with-log-suppressed-unless-notable]]
+            [puppetlabs.trapperkeeper.testutils.logging :refer [with-log-suppressed-unless-notable event->map]]
```
mullr: unused import

puppet/lib/puppet/indirector/facts/puppetdb.rb
```diff
@@ -30,12 +33,12 @@ def save(request)
...
           # legacy storeconfigs.
           "environment" => request.options[:environment] || request.environment.to_s,
-          "producer_timestamp" => request.options[:producer_timestamp] || Time.now.iso8601(5),
+          "producer_timestamp" => Puppet::Util::Puppetdb.to_wire_time(current_time),
           "producer" => Puppet[:node_name_value]
         }
       end
 
-      submit_command(request.key, payload, CommandReplaceFacts, 5)
+      submit_command(request.key, payload, CommandReplaceFacts, 5, current_time.utc)
```
mullr: The distinction between `.utc` here, and without that above, is interesting. Should submit command always be dealing with utc times? If so, should it invoke `.utc` itself?

mullr: oh holy shit I just found the thing below this. Ummm. Clone the time?

src/puppetlabs/puppetdb/queue.clj
senior: It's [here](https://github.com/puppetlabs/puppetdb/pull/2158/commits/50ae08fc47a34931c261d86e88365f973dac0f26#diff-66cce979c7939fbe2d11a146eae980ccR337)

puppet/lib/puppet/indirector/facts/puppetdb.rb
senior: It seems strange to me that we are sending non-UTC times to PDB. We just convert it to UTC over there anyway. I didn't want to change that in this PR, but it seems like we should make that change.

Are you suggesting I clone the time instance so that line 36 and 41 are not order dependent?

src/puppetlabs/puppetdb/queue.clj
senior: good idea

puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
senior: Will do, I guess that goes in the helpers?

src/puppetlabs/puppetdb/amq_migration.clj
```diff
@@ -124,7 +124,8 @@
         (try
           (let [{:keys [command version certname payload] :as msg} (coerce-to-new-command headers message-bytes)]
             (try
-              (enqueue-fn command version certname payload)
+              (enqueue-fn command version certname nil payload)
```
mullr: Yes, that would be helpful.

src/puppetlabs/puppetdb/queue.clj
mullr: oh, my grep-fu is not up to snuff.

puppet/lib/puppet/indirector/facts/puppetdb.rb
mullr: Yes, that's right.

mullr: Or have submit_command clone the time and then call .utc itself

src/puppetlabs/puppetdb/queue.clj
senior: Also yes, it will be omitted from the filename. We won't have producer timestamps for old messages in the queue (or AMQ), imports etc.

src/puppetlabs/puppetdb/queue.clj
senior: I don't remember the scenarios, @rbrw might have more info. Garbage in the directory maybe? I think this is used by both the dead letter code and the queue code

src/puppetlabs/puppetdb/queue.clj
senior: Passing it through ks/timestmap means I lose the ability to compare it to other timestamps, as ks/timestamp returns a string. This was ok for received as I never needed to use it, just drag it around. I would like received to be an object as well. I wanted to make that switch, but figured it'd be best left to another ticket. I wrote up that refactor as PDB-3278

src/puppetlabs/puppetdb/queue.clj
senior: Does that seem ok for the DLO @rbrw?

src/puppetlabs/puppetdb/queue.clj
senior: I thought about putting a condition here instead of an else, something like
```
(time/before? producer-ts
             (:producer-ts maybe-old-command))
```
Would that have been helpful?

senior:
@mullr @ajroetker Fixed it up based on your feedback. I left the changes as separate commits that I'll squash later

src/puppetlabs/puppetdb/queue.clj
mullr: If we have a command that's already out of date, why bother putting it on the queue?

senior: I don't remember why I did this now, probably because it just falls in line with the other behavior. I guess what we could do instead is spin through the stream, discarding all of the bytes, then invoke the delete-update-fn?

senior: I remembered why, just as I was going in to change this. This part of the code is putting a Command Ref in the channel, which means the command is already on disk. To avoid storing this would be pretty difficult with the current code. We'd need to expose the data outside of the mutex in a way that also locks it, consult the map and then decide to discard.

mullr:
@senior One of the commits has the wrong issue tag "￼	(PDB-3180) Add validation of HTTP request params in specs ￼…" (should be 3108) and it's freaking jira out

src/puppetlabs/puppetdb/queue.clj
mullr: Ah, I see. Well, explaining the reasons in comments would be very, very helpful.

mullr:
@senior Just a bit of explanation on that cond, and I'll be +1 for this

ajroetker:
I tested this a few days ago and it looked good to me, nothing changed since then so I'm assuming it's still good!

senior:
👍 I'll rebase and fix up some of the commit messages on Monday to get it ready for merge

PR #2159, opened 2017-01-17
(PDB-3275) Add generative command tests

mullr:


mullr:
This is going to need some squashing at the end of the data, but I think I've ferreted out most of the failures. Fixed two real bugs in the process, too.

test/puppetlabs/puppetdb/generative/submit_command.clj
```diff
@@ -0,0 +1,151 @@
...
+                      "2016-12-19T23:33:00.000Z"
+                      "2016-12-19T23:34:00.000Z"
+                      "2016-12-19T23:35:00.000Z"
+                      "2016-12-19T23:36:00.000Z"
+                      "2016-12-19T23:37:00.000Z"
+                      "2016-12-19T23:38:00.000Z"
+                      "2016-12-19T23:39:00.000Z"})
+
+(defmacro with-catalog-overrides [body]
+  `(with-redefs [pdb-gen/certname #(gen/elements ["a.com" "b.com" "c.com"])
```
senior: Talked with @mullr some about this. I think it's really important that we have this kind of override capability. I'd like to see it plumbed through as first class. Sounds like having that gets us a little closer to what the spec version will look like as well which is a nice bonus

mullr:
@senior So... 

I... um... did a thing. 

Please understand, I've had no outlet for this stuff for so long....

test/puppetlabs/puppetdb/generative/generators.clj
```diff
@@ -1,109 +1,180 @@
 (ns puppetlabs.puppetdb.generative.generators
   (:require [clojure.test.check.generators :as gen]
             [com.gfredericks.test.chuck.generators :as chuck-gen]
-            [clj-time.core :as time :refer [now]]))
+            [clj-time.core :as time :refer [now]]
+            [clojure.tools.logging :as log]
```
mullr: cider added this for me, shouldn't be here.

mullr:
@senior Alternatively, I could do it with a dynamic var :wink:

test/puppetlabs/puppetdb/generative/generators.clj
```diff
@@ -0,0 +1,297 @@
...
+(defn datetime []
+  (gen/fmap clj-time.coerce/from-long
+            (gen/choose 50000000000 2000000000000)))
+
+
+
+;;; catalog generator
+
+(defn certname [] (string+))
+(defn environment [] (string+))
```
mullr: I think the main generator definition here should be a the definition of what the environment /can/, be, i.e. keep it close to what a spec would be. But I can see doing that in the overrides. Not sure if test.check knows how to do that kind of thing, but I'm sure we could teach it.

test/puppetlabs/puppetdb/generative/generators.clj
```diff
@@ -0,0 +1,297 @@
...
+(defn producer_timestamp [] (string+))
+
+(defn producer [] (string+))
+(defn code_id [] (nilable (string+)))
+
+(defn version [] (string+))
+
+(defn resource-type [] (string+))
+(defn resource-title [] (string+))
+(defn resource_spec []
```
mullr: oh, that shouldn't be an underscore.

test/puppetlabs/puppetdb/generative/generators.clj
```diff
@@ -0,0 +1,297 @@
...
+(defn relationship []
+  (gen/one-of (map gen/return
+                   ["contains"
+                    "required-by"
+                    "notifies"
+                    "before"
+                    "subscription-of"])))
+
+(defn edge []
+  (gen/hash-map :source (source)
```
mullr: I think this one is actually stale; see the resource-tree generator and resource-tree-to-edge-list function for the real thing. It only makes trees, but at least it's sort of real.

mullr:
@senior I think this is pretty baked now. I cleaned up the patchset and improved the docs on the overridable generator code, and added a few meta-tests for good measure. Still not 100% convinced it's what we should do, but I now at least think the idea is sound.

test/puppetlabs/puppetdb/generative/generators.clj
```diff
@@ -0,0 +1,285 @@
...
+(defgen :edge/source :edge/resource-spec)
+(defgen :edge/target :edge/resource-spec)
+(defgen :edge/relationship
+  (gen/elements ["contains"
+                 "required-by"
+                 "notifies"
+                 "before"
+                 "subscription-of"]))
+
+
```
wkalt: extra spaces

mullr: :chec

senior:
@mullr Looks like the zip-test stuff is still failing, have you been able to look into that?

test/puppetlabs/puppetdb/generative/spec.clj
```diff
@@ -0,0 +1,119 @@
+(comment
```
senior: It seems like we should remove this as there's no way for us to make sure the code still works, maybe it can be squirreled away in a branch?

mullr: I'll put it in a gist

mullr: https://gist.github.com/mullr/e596ec9efdd7c89aba65026aea705140

mullr:
@senior fixed the zip-test issues

senior:
@mullr you've got some failing tests on travis, out of memory. Maybe it's as easy as dialing down some of the generators?

mullr:
@senior ah forgot about that. yes, it should be. I think the generators get a little bit rowdy up towards the high iterations.

test/puppetlabs/puppetdb/generative/overrideable_generators_test.clj
```diff
@@ -0,0 +1,29 @@
+(ns puppetlabs.puppetdb.generative.overrideable-generators-test
+  (:require [clojure.test :as t]
```
senior: Unless there's a solid reason for this, I'd prefer to keeps this consistent with all of our other tests and just :refer :all

test/puppetlabs/puppetdb/generative/submit_command.clj
```diff
@@ -0,0 +1,150 @@
...
+                      "2016-12-19T23:38:00.000Z"
+                      "2016-12-19T23:39:00.000Z"})
+
+(def timestamp-gen
+  (gen/fmap clj-time.coerce/from-string
+            (gen/elements ten-timestamps)))
+
+(def overrides {:puppet/certname (gen/elements ["a.com" "b.com" "c.com"])
+                :catalog/version (gen/return "9")
+                :puppet/environment (gen/return "production")
```
senior: Seems like we should have a few of these

test/puppetlabs/puppetdb/generative/submit_command.clj
```diff
@@ -0,0 +1,150 @@
...
+
+;;; Generative tests
+
+(tc/defspec submit-commands 25
+  (prop/for-all
+   [cmd (->> :puppetdb/command
+             (gen/override overrides)
+             gen/no-shrink
+             gen/convert)]
+   (print "-") (flush)
```
senior: I think this might be the first time we're intentionally printing things in a test. It seems useful when working on the tests locally, but it seems like it would just be noise in CI. Not sure what the best way to accommodate both is. Maybe log/trace or something? Thoughts @rbrw?

test/puppetlabs/puppetdb/generative/generators.clj
```diff
@@ -0,0 +1,283 @@
...
+                      :puppet/producer_timestamp
+                      :puppet/producer
+                      :catalog/version
+                      :catalog/code_id
+                      :catalog/resource-tree)))
+
+;;; factset generator
+
+(defgen :fact/name (string+))
+(defgen :fact/value json-value)
```
senior: Will this only generate vectors and maps? Or will it generate strings/integers as well (not inside a tree structure)

test/puppetlabs/puppetdb/generative/submit_command.clj
mullr: I put these in initially to be able to tell the difference between test.check going into shrink mode (adding thousands of iterations) and a generator being completely out to lunch. Since everything is behind gen/no-shrink now, I think I can just get rid of it.

test/puppetlabs/puppetdb/generative/overridable_generators.clj
```diff
@@ -0,0 +1,135 @@
...
+  one, return a test.check generator. "
+  ([gen] (lookup gen {}))
+  ([gen overrides]
+   (let [gens (merge @generators overrides)]
+     (cond
+       (keyword? gen) (if-let [gen-val (get gens gen)]
+                        (recur gen-val overrides)
+                        (throw (ex-info (str "Generator " gen " not registered") {:generator-key gen})))
+       (gen/generator? gen) gen
+       :default (gen overrides)))))
```
mullr: You mean `(gen gens)`? No, I don't think so. I mean, I guess we could, but there's no point. In this case, 'gen' is an overridable genearator, i.e. a function of overrides -> generator. Each overridable generator calls `lookup` on its arguments, passing down overrides. So eventually it's going to bottom out in this `lookup` function again, having either a regular generator or a keyword. 

Which is to say, if we passed down `gens`, we'd end up merging in `@generators` twice. 

I do think moving the `let` into the `(keyword?)` block would be better though.

test/puppetlabs/puppetdb/generative/generators.clj
mullr: I *think* it'll give you plain values sometimes. Checking...

mullr: wow, it acutally doesn't. I didn't expect that.

mullr:
rebased on top of latest master

test/puppetlabs/puppetdb/generative/generators.clj
senior: 👍

PR #2161, opened 2017-01-18
(PDB-3256) drop root endpoint experimental status

wkalt:
Drop the experimental status of PQL, the root endpoint, implicit subqueries, and the from operator.


wkalt:
@senior @kbarber I've left these split out for now since there are a few things getting dropped. IMO if we're going to do any of them we may as well do them all, but we can start the discussion here and squash them together if we like. If you know of any teams that might have an interest in this please ping them. I'll let Beth and Ryan C. know about it.

ajroetker:

mullr:

mullr:
@wkalt can we get the ticket number in the commit summary like usual?

PR #2163, opened 2017-01-20
(PDB-2524) Data-driven dashboard

mullr:
Rework the dashboard to get the list of meters to display in a request to the
server, rather than hardcoding them in javascript. This has several benefits:

- All metrics can be returned in a single response, instead of having to poll
  each metric individually

- The meter descriptions can be localized

- The data can be overriden in PE, to allow new meters to be added there for
  PE-only functionality.

### DIFF
--- a/resources/public/dashboard/charts.js
+++ b/resources/public/dashboard/charts.js
@@ -1,18 +1,23 @@
+// Formatting middleware that collapses small values to 0
+function clampToZero(f, window) {
+    return function(n) {
+        return f(Math.abs(n) < window ? 0 : n);
+    };
+};
+
 /*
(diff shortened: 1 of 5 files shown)


mullr:
If anybody reviews this: the sketchiest part, to me, is the format of the meter defs; it has some data for sever-side processing, and some stuff meant for the client.

mullr:

wkalt:
One thing I'm noticing about this is the batched request seems to make the initial page load much slower. I'm not sure if it's worth addressing though.

mullr:
@wkalt yeah, it probably will. I guess I could change it to read the meters directly rather than going through jmx, but I doubt that's worth the effort.

wkalt:
+1 will merge after travis passes

PR #2164, opened 2017-01-25
(PDB-3250) (PDB-3252) Reduplicate fact values

rbrw:


rbrw:
This is just the first step.  The migration-test and storage-test should pass, but that may be about it.

wkalt:

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -102,7 +102,7 @@
   "Calls clojure.jdbc/db-query-with-resultset after adding (jdbc/db)
    as the first argument. Note that this will hold the whole resultset in memory
    due to the default jdbc fetchsize of 0. If streaming is required, use
-   with-query-results-cursor."
```
mullr: In the commit message, when you say column vectors, do you mean row vectors?

rbrw: Well, "vectors of column" (values), but yeah row vector might be more sensible -- I'll fix it in the next rebase.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -127,17 +127,43 @@
...
+  produced as they do for that function.  So for example, when
+  as-arrays? is logically true, the result rows will be vectors, not
+  maps, and the first result row will be a vector of column names."
+  ([query f] (call-with-query-rows query {} f))
+  ([[sql & params]
+    {:keys [as-arrays? identifiers qualifier read-columns] :as opts}
+    f]
+   (with-db-transaction []
+     (with-open [stmt (.prepareStatement (:connection *db*) sql)]
+       (doseq [[index value] (map-indexed vector params)]
```
rbrw: Hmm, I'll take a look -- what's there now is just what was already in with-query-results-cursor, which this is replacing.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -127,17 +127,43 @@
...
-       (map #(kitchensink/mapvals convert %) result-set))))
+(defn convert-if-array
+  ([x] (convert-if-array vec x))
+  ([convert x]
+   (cond
+     (kitchensink/array? x) (convert x)
+     (isa? (class x) java.sql.Array) (convert (.getArray x))
+     :else x)))
+
+(defn call-with-query-rows
```
rbrw: The other one streams too, I just thought this name was clearer, and it allows you to ask for arrays instead of maps if that's all you want/need.  And it's somewhat following the Scheme convention to scope operations on a resource with a lifetime (cf. `call-with-output-file`).  I figured the docstring might be sufficient, but if we wanted to change the name, I wondered about `call-with-lazy-query-rows`.

rbrw:
I've added commits to cover the rest of the fact_value removal related changes, and as before, you might want to look at the individual commits to start, and "ignore whitespace" might help.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1027,76 +943,116 @@
...
+     jdbc/*db* true
+     (cons
+      (str "update facts set"
+           "    value_type_id = ?,"
+           "    value_boolean = ?,"
+           "    value_integer = ?,"
+           "    value_float = ?,"
+           "    value_string = ?,"
+           "    value = ?,"
+           "    large_value_hash = ?"
```
wkalt: this fails because it's using the wrong type for large_value_hash.

```
2017-02-03 10:21:44 PST ERROR:  column "large_value_hash" is of type bytea but expression is of type character varying
at character 167
```

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1027,76 +943,116 @@
...
+           "  where factset_id = ? and fact_path_id = ?")
+      (for [[_ path path-id new-vm] change-info]
+        [(:value_type_id new-vm)
+         (:value_boolean new-vm)
+         (:value_integer new-vm)
+         (:value_float new-vm)
+         (:value_string new-vm)
+         (:value new-vm)
+         (:large_value_hash new-vm)
```
wkalt:
```
(some-> (:large_value_hash new-vm) sutils/munge-hash-for-storage)
```
or something

wkalt:
I can get a significant performance boost out of this with an index on facts(factset_id). With that change it seems to surpass previous performance on fact storage/processing.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1027,76 +943,116 @@
...
+                  (if (fact-value-changed? row new-vm)
+                    [:changed path id new-vm]
+                    [:unchanged path id])
+                  [:gone path id])))))
+
+          rm-pids (seq (map (fn [[_ _ pid]] pid)
+                            (filter #(= :gone (first %))
+                                    existing-status)))
+          new-pathstrs (let [have? (set (for [[st path] existing-status
+                                              :when #({:changed :unchanged} st)]
```
wkalt: is this supposed to be
```
:when #(#{:changed :unchanged} st)]
```
?

wkalt: looking a little closer it feels like this could actually be intended behavior, since we don't need to realize paths for unchanged facts? if that's the case it would be clearer to make it
```
:when #(= :changed st)
```

rbrw: No, that's just a mess I made; I think what I intended was just (#{:changed :unchanged} st), but now that you have me looking at it, I think i might be able to make it even simpler.

rbrw: Rather, what you said: #(#{:changed :unchanged} st)

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: Glad you noticed - thanks.

rbrw: Though we should have some tests coming up soon: https://tickets.puppetlabs.com/browse/PDB-3269

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: Hopefully have this a bit clearer in the next version.

wkalt:
I did a little bit of testing with dynamic facts submission against this patch.

To process 50 consecutive full overwrites of a large "partitions-shaped" fact, with 10k devices, too 160 seconds on this patch and 379 seconds on master.

To process 50 consecutive prepends to an otherwise static array of 10k map-valued elements took 124 seconds on this patch. On master I stopped the test because the commands were being processed more slowly than once per minute, so this is a pretty huge improvement in that case.

The first of these two tests is most similar to the problems we have with the native facter facts.

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -186,11 +186,12 @@
...
-                           (jdbc/with-query-results-cursor
-                             results-query (comp #(http/stream-json % buffer pretty-print)
-                                                 #(do (when-not (instance? PGobject %)
-                                                        (first %)) (deliver query-error nil) %)
-                                                 munge-fn)))
+                           (jdbc/call-with-query-rows
+                            results-query
+                            (comp #(http/stream-json % buffer pretty-print)
+                                  #(do (when-not (instance? PGobject %)
+                                         (first %)) (deliver query-error nil) %)
```
wkalt: this would be easier to read with deliver on a new line

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1172,6 +1173,54 @@
...
+   "drop table facts"
+   "drop table fact_values"
+   "drop sequence fact_values_id_seq"
+
+    "alter table facts_transform rename to facts"
+
+    "create index facts_factset_id_idx on facts using btree (factset_id)"
+    "create index facts_fact_path_id_idx on facts using btree (fact_path_id)"
+    "create index facts_value_integer_idx on facts(value_integer)"
+    "create index facts_value_float_idx on facts(value_float)"
```
wkalt: I think we will want indexes for all the columns that previously had them -- this is missing value_string and value_hash

wkalt: with the trgm index on value_string

wkalt: reading closer you might not need the one on the hash

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -967,10 +1000,12 @@
...
             :tags #{"file" "class" "foobar"}
             :line 20}}
-         (-> (query-to-vec "SELECT title, tags, line FROM catalog_resources WHERE certname_id = (select id from certnames where certname = ?)" certname)
-             jdbc/convert-result-arrays
-             tags->set
-             set)))
+         (call-with-query-rows
+          [(str "select title, tags, line from catalog_resources"
+                 "  where certname_id = (select id from certnames"
+                 "                         where certname = ?)")
```
wkalt: even though it's test code, we should change the queries in this file to use joins instead of correlated subqueries

src/puppetlabs/puppetdb/facts.clj
```diff
@@ -130,26 +130,37 @@
...
-      (let [value-keyword (case type-id
-                            0 :value_string
-                            1 :value_integer
-                            2 :value_float
-                            3 :value_boolean
-                            5 :value)]
-        (assoc initial-map value-keyword value
-          :value (sutils/munge-jsonb-for-storage value))))))
+      result
+      (let [value-key (case type-id
```
rbrw: Don't disagree -- I'd originally held off changing it more substantially, but happy to give it a try.

rbrw: And regarding value-key, I think it's likely awkward/ugly/inefficient, but correct -- in any case, I'll fix it.

rbrw: value_string is in the case because structured values and strings are the two cases where we might have a "large value".

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Right now we only have the (optional) large_value_hash (replacing the universal hash), which I don't think we're using for anything yet, but I suppose an index might eventually be useful.

senior: Mentioned this to @rbrw, we should really have a test here to ensure the before/after state of the migration.

rbrw: @wkalt I think the trgm index is created opportunistically (as before) below?

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -645,3 +646,198 @@
...
+                               :character_maximum_length nil
+                               :nullable? "NO"
+                               :datetime_precision nil
+                               :character_octet_length nil
+                               :column_default nil
+                               :numeric_scale 0}}}
+          expected {:index-diff exp-idx-diffs
+                    :table-diff exp-table-diffs}]
+      ;; Handy when trying to see what's wrong.
+      (when-not (= expected diff)
```
mullr: https://github.com/pjstadig/humane-test-output

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -645,3 +646,198 @@
...
+
+(deftest fact-values-reduplication-schema-diff
+  ;; This does not check the value_string trgm index since it is
+  ;; created opportunistically later in trgm-indexes!
+  (clear-db-for-testing!)
+  (fast-forward-to-migration! 55)
+  (let [initial (schema-info-map *db*)]
+    (apply-migration-for-testing! 56)
+    (let [migrated (schema-info-map *db*)
+          rename-idx (fn [smap idx-id new-table new-name]
```
mullr: https://en.wikipedia.org/wiki/SMAP

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: I've verified that by hand, and also created a new test for it that I'll get in (likely as a follow-up pr) later.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -41,7 +48,7 @@
   (sql/db-do-prepared *db* true (into [sql] params) {:multi? true}))
 
 (defn do-commands-outside-txn [& commands]
-  (let [conn (:connection *db*)
+  (let [conn ^Connection (:connection *db*)
```
senior: Seems a little odd to type-hint the right hand side of a let expression but it looks like it works.

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -645,3 +646,227 @@
...
+                               :character_maximum_length nil
+                               :numeric_precision nil
+                               :numeric_precision_radix nil
+                               :data_type "bytea"
+                               :column_name "large_value_hash"
+                               :table_name "facts"}
+                              :same nil}
+                             ;; shrank value_type_id from bigint to integer
+                             {:left-only {:data_type "bigint" :numeric_precision 64}
+                              :right-only {:data_type "integer" :numeric_precision 32}
```
senior: I'm not opposed to this change, but this is now the only id column that is integer, right?

senior: It looks like value_types.id is still a bigint, why an integer here then/

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1172,6 +1173,54 @@
...
+    "   value_boolean)"
+    "  select factset_id, fact_path_id,"
+    "         value_integer, value_float, value_type_id, value_string, value,"
+    "         case when pg_column_size(value) >= 50"
+    "              then value_hash"
+    "         end,"
+    "         value_boolean"
+    "    from facts inner join fact_values on fact_value_id = fact_values.id"]
+
+   "drop index facts_fact_value_id_idx"
```
senior: indentation is off here

test/puppetlabs/puppetdb/scf/migrate_test.clj
rbrw: Good point(s), though originally, I wasn't thinking about the id here as a pk, I was thinking of it (via domain specific knowledge) as a type tag that we know has specific, unchangable, integer values (https://github.com/puppetlabs/puppetdb/blob/57196c992dc68563532d16ae7c82f1081077b071/src/puppetlabs/puppetdb/scf/migrate.clj#L247) that we've been hard-coding on the clojure side (https://github.com/puppetlabs/puppetdb/blob/57196c992dc68563532d16ae7c82f1081077b071/src/puppetlabs/puppetdb/facts.clj#L146), and figured we could just use the smaller postgresql "best balance" type here -- but happy to leave that off if we'd rather, and wasn't thinking at the time about the mismatch created with the other table.

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Test is in.

src/puppetlabs/puppetdb/jdbc.clj
rbrw: What'd be the more natural approach?

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Ahh, I think that's just an artifact of the vector for the multi-line command above (newly allowed by https://github.com/puppetlabs/puppetdb/pull/2164/commits/14f6c75dc9900160e3f4f3680e1048ada86b5b3a in this series), which might also be a thing to discuss.

test/puppetlabs/puppetdb/scf/migrate_test.clj
rbrw: After a bit of discussion with @senior, I think we should back this out and handle it later, separately, if it still seems worth pursuing, and we should probably fix the value_types.id mismatch then too.

PR #2165, opened 2017-01-25
(PDB-3169) Increase timeout in command test

mullr:
This sometimes times out when running on shared test infrastructure; attempt to
work around this by increasing the timeout.


rbrw:
Wondered if maybe this should be default-timeout-ms from testutils.

mullr:
Will reopen on stable

PR #2166, opened 2017-01-25
(PDB-3271) Mark 2.3.x end of life

wkalt:
Add a note to the docs marking 2.3.x end of life and encouraging users
to upgrade.


mullr:
:+1:  from me, should probably get it from one other person too

PR #2170, opened 2017-01-27
(MAINT) Close datasource in max connection jdbc pool tests

camlow325:
Previously, the max connections jdbc pool-construction tests did not
close the connection pool data source before falling out of scope.
Depending upon timing, this could cause the tests to fail on
attempting to drop temp databases which are still busy being used
by the connection pool.

This commit wraps usage of the data source in each of the max
connections tests with `with-open` calls, allowing the connection pool
to be closed off before any calls to drop the temporary databases occur
(and, therefore, for the tests to complete successfully).

### DIFF
--- a/test/puppetlabs/puppetdb/jdbc_test.clj
+++ b/test/puppetlabs/puppetdb/jdbc_test.clj
@@ -54,19 +54,21 @@
   (testing "max connections setting defaults to 25"
     (call-with-antonym-test-database
       (fn []
-        (let [pool (-> *db*
-                       defaulted-read-db-config
-                       subject/pooled-datasource)]
-          (is (= 25 (.getMaximumPoolSize (:datasource pool))))))))
+        (with-open [data-source (-> *db*
+                                    defaulted-read-db-config
+                                    subject/pooled-datasource
+                                    :datasource)]
+          (is (= 25 (.getMaximumPoolSize data-source)))))))
 
   (testing "max connections setting takes effect"
     (call-with-antonym-test-database
       (fn []
-        (let [pool (-> *db*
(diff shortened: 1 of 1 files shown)


camlow325:
@wkalt - here's the fix for those tests that I mentioned to you today.

test/puppetlabs/puppetdb/jdbc_test.clj
```diff
@@ -57,7 +57,8 @@
         (let [pool (-> *db*
                        defaulted-read-db-config
                        subject/pooled-datasource)]
-          (is (= 25 (.getMaximumPoolSize (:datasource pool))))))))
```
senior: Any reason not to use the with-open instead of the let above?

test/puppetlabs/puppetdb/jdbc_test.clj
```diff
@@ -66,7 +67,8 @@
                        defaulted-read-db-config
                        (assoc :maximum-pool-size 5)
                        subject/pooled-datasource)]
-          (is (= 5 (.getMaximumPoolSize (:datasource pool)))))))))
+          (with-open [data-source (:datasource pool)]
```
senior: Looks like the same thing is possible here too

test/puppetlabs/puppetdb/jdbc_test.clj
camlow325: Yeah, seems like that would work.  I can change this and the other one as well.

camlow325: @senior I collapsed the `let`s and `with-open`s in 3832c43.  Hope that's what you were looking for.

PR #2171, opened 2017-01-27
(PDB-3053) Create unique index on certnames.latest_report_id

npwalker:
Prior to this commit, there was no index on latest_report_id.
In PDB-2789, it was thought that this index caused performance
issues but this was potentially a red-herring.

After this commit, we create a unique index on latest_report_id. 
The index is absolutely necessary for deleting reports as the
reports table has a constraint on latest_report_id and without this
index it scans the whole certnames table once for every row it
deletes from reports.  For a 100,000 rows this index causes
deleting from reports to go from 3-4 mins to 3-4 seconds.


mullr:
@npwalker This is currently targeted at master; is that where you think it should go? (i.e. PE Glisan at earliest)

PR #2173, opened 2017-01-31
(PDB-2640) Add support for gzip-compressing commands via terminus

camlow325:
This commit adds support for gzip-compressing commands which the
PuppetDB terminus sends out.  When stockpile persists the commands to
disk, it stores the gzipped json bytes with a .json.gz file extension.
Command payloads are decompressed and parsed from JSON when processed
back out of the queue.

The terminus does not gzip-compress the command content itself.
Instead, to minimize memory usage somewhat and realize a bit better
performance, it passes through its intent for the implementation
upstream from the Puppet::Network::HttpPool.http_instance to perform
the gzip compression.  Corresponding work in Puppet Server and its
upstream clj-http-client library enables the payload to be compressed.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -246,7 +246,7 @@
   [q
    command-chan
    ^Semaphore write-semaphore
-   {:keys [command certname command-stream] :as command-req} :- queue/command-req-schema]
+   {:keys [command certname command-stream compression] :as command-req} :- queue/command-req-schema]
   (try
     (.acquire write-semaphore)
     (time! (get @metrics :message-persistence-time)
@@ -407,8 +407,8 @@
 
 (defprotocol PuppetDBCommandDispatcher
   (enqueue-command
-    [this command version certname producer-ts payload]
-    [this command version certname producer-ts payload command-callback]
+    [this command version certname producer-ts payload compression]
+    [this command version certname producer-ts payload compression command-callback]
     "Submits the command for processing, and then returns its unique id.")
 
   (stats [this]
@@ -660,16 +660,16 @@
   (stats [this]
     @(:stats (service-context this)))
 
-  (enqueue-command [this command version certname producer-ts command-stream]
-                   (enqueue-command this command version certname producer-ts command-stream identity))
+  (enqueue-command [this command version certname producer-ts command-stream compression]
+                   (enqueue-command this command version certname producer-ts command-stream compression identity))
 
-  (enqueue-command [this command version certname producer-ts command-stream command-callback]
+  (enqueue-command [this command version certname producer-ts command-stream compression command-callback]
     (let [config (get-config)
           q (:q (shared-globals))
           command-chan (:command-chan (shared-globals))
           write-semaphore (:write-semaphore (service-context this))
           command (if (string? command) command (command-names command))
-          command-req (queue/create-command-req command version certname producer-ts command-callback command-stream)
+          command-req (queue/create-command-req command version certname producer-ts compression command-callback command-stream)
           result (do-enqueue-command q command-chan write-semaphore command-req)]
       ;; Obviously assumes that if do-* doesn't throw, msg is in
       (inc-cmd-depth command version)
--- a/src/puppetlabs/puppetdb/middleware.clj
+++ b/src/puppetlabs/puppetdb/middleware.clj
@@ -115,6 +115,21 @@
       (http/error-response (tru "must accept {0}" content-type)
                            http/status-not-acceptable))))
 
+(defn verify-content-encoding
+  "Verification for the specified list of content-encodings."
+  [app content-encodings]
+  {:pre [(coll? content-encodings)
+         (every? string? content-encodings)]}
+  (fn [{:keys [headers request-method] :as req}]
+    (let [content-encoding (headers "content-encoding")]
+      (if (or (not= request-method :post)
+              (empty? content-encoding)
+              (some #{content-encoding} content-encodings))
+        (app req)
+        (http/error-response (tru "content encoding {0} not supported"
+                                  content-encoding)
+                             http/status-unsupported-type)))))
+
 (defn verify-content-type
   "Verification for the specified list of content-types."
   [app content-types]
--- a/src/puppetlabs/puppetdb/middleware.clj
+++ b/src/puppetlabs/puppetdb/middleware.clj
@@ -117,14 +117,14 @@
 
 (defn verify-content-encoding
   "Verification for the specified list of content-encodings."
-  [app content-encodings]
-  {:pre [(coll? content-encodings)
-         (every? string? content-encodings)]}
+  [app allowed-encodings]
+  {:pre [(coll? allowed-encodings)
+         (every? string? allowed-encodings)]}
   (fn [{:keys [headers request-method] :as req}]
     (let [content-encoding (headers "content-encoding")]
       (if (or (not= request-method :post)
               (empty? content-encoding)
-              (some #{content-encoding} content-encodings))
+              (some #{content-encoding} allowed-encodings))
         (app req)
         (http/error-response (tru "content encoding {0} not supported"
                                   content-encoding)
--- a/src/puppetlabs/puppetdb/queue.clj
+++ b/src/puppetlabs/puppetdb/queue.clj
@@ -4,7 +4,7 @@
            [java.util TreeMap HashMap]
            [java.nio.file Files LinkOption]
            [java.nio.file.attribute FileAttribute]
-           (org.apache.commons.compress.compressors.gzip GzipCompressorInputStream))
+           [org.apache.commons.compress.compressors.gzip GzipCompressorInputStream])
   (:require [clojure.string :as str :refer [re-quote-replacement]]
             [puppetlabs.stockpile.queue :as stock]
             [clj-time.coerce :as tcoerce]
@@ -21,7 +21,8 @@
             [clojure.core.async :as async]
             [clojure.core.async.impl.protocols :as async-protos]
             [puppetlabs.puppetdb.nio :refer [get-path]]
-            [puppetlabs.puppetdb.utils :refer [match-any-of utf8-length
+            [puppetlabs.puppetdb.utils :refer [compression-file-extension-schema
+                                               match-any-of utf8-length
                                                utf8-truncate]]
             [slingshot.slingshot :refer [try+]]
             [schema.core :as s]
@@ -199,7 +200,7 @@
    :producer-ts (s/maybe pls/Timestamp)
    :callback (s/=> s/Any s/Any)
    :command-stream java.io.InputStream
-   :compression s/Str})
+   :compression compression-file-extension-schema})
 
 (s/defn create-command-req :- command-req-schema
   "Validating constructor function for command requests"
@@ -207,7 +208,7 @@
    version :- s/Int
    certname :- s/Str
    producer-ts :- (s/maybe s/Str)
-   compression :- s/Str
+   compression :- compression-file-extension-schema
    callback :- (s/=> s/Any s/Any)
    command-stream :- java.io.InputStream]
   {:command command
@@ -239,14 +240,14 @@
                       :compression compression})))
 
 (defn wrap-decompression-stream
-  [compression command-stream]
-  (condp = compression
+  [file-extension command-stream]
+  (case file-extension
     nil command-stream
     "" command-stream
     "gz" (GzipCompressorInputStream. command-stream)
     (throw+ {:kind ::parse-error} nil
-            (trs "Unsupported compression format for command: {0}"
-                 compression))))
+            (trs "Unsupported compression file extension for command: {0}"
+                 file-extension))))
 
 (defn cmdref->cmd [q cmdref]
   (let [compression (:compression cmdref)
--- a/src/puppetlabs/puppetdb/queue.clj
+++ b/src/puppetlabs/puppetdb/queue.clj
@@ -3,7 +3,8 @@
            [java.io InputStreamReader BufferedReader InputStream]
            [java.util TreeMap HashMap]
            [java.nio.file Files LinkOption]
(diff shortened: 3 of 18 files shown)


camlow325:
This is definitely going to have conflicts with #2158 since that PR also threads through another piece of metadata through several layers of the stack, a `producer-ts`, like this PR does.  Since that PR has been in the pipeline much longer than this one, probably would be best to let that land first and rework this one on top of that.

camlow325:
I suspect that there would be some additional tests which would be useful to add for this one.  Would appreciate some guidance from others more familiar with the current unit/acceptance test structure as to where additional tests could best be placed.

Where this PR relies upon Puppet Server and its upstream clj-http-client library to do the gzip-compression and the underlying functionality in MRI Ruby Puppet wouldn't inherently be capable of doing gzip-compression, I tested this one manually for now with a custom build of Puppet Server.  I did the following:

1) Built a clj-http-client library from the work in https://github.com/puppetlabs/clj-http-client/pull/67.

2) Built puppetserver from the work in https://github.com/puppetlabs/puppetserver/pull/1325 and a custom build of the clj-http-client library.

3) Stood up the custom puppetserver and puppetdb following the instructions from the [running from source docs](https://github.com/puppetlabs/puppetserver/blob/a9d9f6b7bca3900a9de6f49454734323a1fc44bc/documentation/dev_running_from_source.markdown) in the Puppet Server repo.

4) Did a few agent runs against Puppet Server and corresponding report queries to the PuppetDB server to see that reports were stored correctly.

5) Stopped the postgres server before one of the runs for long enough to see that the command payload stored in the stockpile directory had the .json.gz extension and was gzipped.

6) Restarted puppetdb with some .json.gz files in the stockpile q directory and confirmed that puppetdb picked up those files and delivered them to postgres (via report queries).

7) Did a few agent runs against a Ruby WEBrick master, version 4.8.1, and corresponding report queries to the PuppetDB server to see that the reports were stored correctly.  This validates that PuppetDB is still capable of handling uncompressed json in commands when the infrastructure behind the terminus is not capable of gzip-compressing the payload.

camlow325:
I did some testing with our automated gatling infrastructure to compare performance when the Puppet DB terminus was gzip-compressing vs. not compressing commands sent through Puppet Server to PuppetDB.

These runs simulate up to 1250 agents running with ~ 30 minute delay between runs over a 2 hour period, with catalogs shaped from those which are seen by the agent on a compile master in our SysOps environment.  See the definition of the [medium role](https://github.com/puppetlabs/puppetlabs-puppetserver_perf_control/blob/production/site/role/manifests/by_size/medium.pp) in the puppetserver_perf_control repo for more info, if interested.

The Puppet Server and Puppet DB versions used for each of the runs had identical code other than the specific code changes related to this PR which enable the use of gzip-compression for terminus commands.

No agent run errors occurred in any of the gzip or no compression runs.  The overall mean agent run time appeared to be roughly the same for gzip vs. no compression.  Runs were slightly faster with gzip-compression enabled but within one standard deviation of their no-compression counterparts:

| Compression | Mean Agent Run Time | Standard Deviation | Run Link
|--------------| ---------------------- | ------------------- | ------------
| none              |  6.142 seconds            | 745 ms                    | http://puppetserver-perf-driver68-dev:8080/job/oss-puppetserver-latest/18/
| none              |  5.992 seconds            | 596 ms                    | http://puppetserver-perf-driver68-dev:8080/job/oss-puppetserver-latest/19/
| gzip               |  5.654 seconds            | 657 ms                    | http://puppetserver-perf-driver68-dev:8080/job/oss-puppetserver-latest/25/
| gzip               |  5.556 seconds            | 580 ms                    | http://puppetserver-perf-driver68-dev:8080/job/oss-puppetserver-latest/26/

Some additional performance tuning, e.g., with respect to buffer sizes used for compression / decompression, may improve performance further.

camlow325:
I rebased this on top of the latest code in master, at 33c63ff.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -407,8 +407,8 @@
 
 (defprotocol PuppetDBCommandDispatcher
   (enqueue-command
-    [this command version certname producer-ts payload]
-    [this command version certname producer-ts payload command-callback]
+    [this command version certname producer-ts payload compression]
+    [this command version certname producer-ts payload compression command-callback]
```
mullr: we'll need to update pe-puppetdb-extensions for this as well, since sync uses this api

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -179,7 +184,10 @@
                                       tcoerce/from-long)
                  :version (Long/parseLong version)
                  :command (get md-cmd->pdb-cmd md-command "unknown")
-                 :certname certname})))))))
+                 :certname certname
+                 :compression (if (nil? compression)
+                                ""
+                                (subs compression 1))})))))))
```
mullr: (or could it use  `content-encoding->file-extension` from above?)

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -190,47 +198,65 @@
...
                       :version version
                       :certname certname
                       :received received
                       :producer-ts producer-ts
-                      :callback identity})))
+                      :callback identity
+                      :compression compression})))
+
+(defn wrap-decompression-stream
+  [compression command-stream]
```
mullr: we use 'compression' elsewhere to talk about the header, which is (I think) either "" or "gzip". This is the file-extension form, so can we call it that?

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -115,6 +115,21 @@
...
+(defn verify-content-encoding
+  "Verification for the specified list of content-encodings."
+  [app content-encodings]
+  {:pre [(coll? content-encodings)
+         (every? string? content-encodings)]}
+  (fn [{:keys [headers request-method] :as req}]
+    (let [content-encoding (headers "content-encoding")]
+      (if (or (not= request-method :post)
+              (empty? content-encoding)
+              (some #{content-encoding} content-encodings))
```
camlow325: I went with the `some` approach here for consistency with the pre-existing code in the similar [verify-content-type](https://github.com/puppetlabs/puppetdb/blob/1c98f73ee23624049ae9dcf2294bfe9469fc7c61/src/puppetlabs/puppetdb/middleware.clj#L128), but I'm good with a `set` approach you suggested instead if you think that reads better.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -3,7 +3,8 @@
            [java.io InputStreamReader BufferedReader InputStream]
            [java.util TreeMap HashMap]
            [java.nio.file Files LinkOption]
-           [java.nio.file.attribute FileAttribute])
+           [java.nio.file.attribute FileAttribute]
+           (org.apache.commons.compress.compressors.gzip GzipCompressorInputStream))
```
camlow325: Yeah, totally good with that.  Intelli-J did that for me.  😄

src/puppetlabs/puppetdb/middleware.clj
mullr: oh ha, didn't see that. That's a good enough reason to leave it this way, so I'll leave it up to you.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -132,26 +133,30 @@
...
-         (let [name-hash (kitchensink/utf8-string->sha1 certname)]
-           (format "%s_%s_%d_%s_%s.json"
-                   received+producer-time short-command version safe-certname name-hash)))))))
+           safe-certname (embeddable-certname received short-command version certname)
+           name (if (= safe-certname certname)
+                  safe-certname
+                  (format "%s_%s"
+                          safe-certname
+                          (kitchensink/utf8-string->sha1 certname)))
+           extension (str "json" (if (not-empty compression)
```
camlow325: Not sure.  I had thought @senior and I had discussed a `.json.gz` extension but I might be remembering that incorrectly.  @senior - WDYT?

src/puppetlabs/puppetdb/queue.clj
camlow325: This code just parses back the file extension as stockpile would store it to disk.  Since it's stored as `.json.gz`, this strips it to `gz` in the conversion back to a command.  The PR currently only converts from a `content-encoding` (which would be "gzip") to a file extension (which would be "gz") when the command is first dumped into the queue from the HTTP route handling, in `enqueue-command-handler`.  I could store the payload to disk as `.json.gzip` to avoid the conversion, but then we'd be using even more characters in memory.  This might also look weird from the perspective of someone inspecting files on the file system, too, since `...gz` seems more standard.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -190,47 +198,65 @@
...
+   :command-stream java.io.InputStream
+   :compression s/Str})
 
 (s/defn create-command-req :- command-req-schema
   "Validating constructor function for command requests"
   [command :- s/Str
    version :- s/Int
    certname :- s/Str
    producer-ts :- (s/maybe s/Str)
+   compression :- s/Str
```
camlow325: I was just thinking through trying to minimize the number of places where code has to reference specific compression format values - in the event that more than one may exist in the future.  Could repurpose a enumerated type elsewhere too, though, so I'm good with going that way.

src/puppetlabs/puppetdb/queue.clj
camlow325: Sure, so just rename `compression` to `file-extension` in this one function, then - or were you thinking I should do that in some other functions as well?

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -190,47 +198,65 @@
...
                       :certname certname
                       :received received
                       :producer-ts producer-ts
-                      :callback identity})))
+                      :callback identity
+                      :compression compression})))
+
+(defn wrap-decompression-stream
+  [compression command-stream]
+  (condp = compression
```
camlow325: Good call, thanks.

camlow325:
@mullr I pushed a new commit, 3a82a93, which attempts to address much of the feedback you've provided so far.  I didn't make any changes to the file naming yet - e.g., maybe `.jgz` instead of `.json.gz` to save a few characters in memory.  I'd like to get @senior's opinion on that part first.

src/puppetlabs/puppetdb/queue.clj
senior: I'm in favor of ".json.gz" just because it's clear from the user perspective what's going on there. We have code to protect ourselves if the name is too long (we'll hash it), and adding 3 chars seems ok to me.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -190,47 +199,65 @@
...
 
 (defn cmdref->cmd [q cmdref]
-  (let [entry (cmdref->entry cmdref)]
+  (let [compression (:compression cmdref)
+        entry (cmdref->entry cmdref)]
     (with-open [command-stream (stock/stream q entry)]
       (assoc cmdref
-             :payload (stream->json command-stream)
-             :entry entry))))
+        :payload (stream->json (wrap-decompression-stream
```
senior: 👍 

I was trying to refresh my memory on the order of compression with java streams. I knew there was a big different depending on how it's arranged so I ran some local tests. This is the fastest way to do it. I'm not sure why, but if instead you have the decompression on the result of the buffered stream, it's super slow (i.e. like 400x slower). Might be good to have a comment on the function above to that affect to let people in the future know that order is important here.

senior:
Nice work here @camlow325 . One thing I'm a little unclear on is how users can ensure that they get this perf improvement. Does the user need to be running a minimum (yet to be released) version of Puppet Server that supports this compression along with the new version of PuppetDB?

src/puppetlabs/puppetdb/queue.clj
camlow325: @senior That's great to know - thanks for doing that testing.  Happy to add a comment here.

TBH, I'd gone with just wrapping the `GzipCompressorInputStream` around the original input stream - without an extra buffered input stream in between - because I'd noticed that the `GzipCompressorInputStream` already managed its own internal buffer, which it defaults to 8192 bytes.  Double buffering the data before decompressing it intuitively didn't seem like it would bode well for performance.

I think I could have gone with the JDK's `GZIPInputStream` instead of the apache commons one here, for which the internal buffer size is configurable.  I naively went with `GzipCompressorInputStream` instead because we were already using it in PuppetDB - when handling archive files.  Where we probably won't have the requirement of decompressing concatenated `.gz` files in the context of handling PuppetDB commands, though, it may be better to go with `GZIPInputStream` later and do some more targeted buffer size tweaking.  Maybe that would be worth a follow-on effort if it seems worthwhile?

camlow325:
> Does the user need to be running a minimum (yet to be released) version of Puppet Server that supports this compression along with the new version of PuppetDB?

@senior Yeah, we would need to be running the PuppetDB terminus in a new-enough, to-be-released Puppet Server in order to have it start gzip-compressing request payloads.  The Puppet Server enablement piece would come from https://github.com/puppetlabs/puppetserver/pull/1325.  We'll probably need to sync up later on versioning mechanics to ensure this is documented when appropriate.

camlow325:
@senior Added a couple of more commits, one with a comment about perf related to your suggestion and one that consolidates explicit references in the code to "gzip" a bit more.

camlow325:
@senior, @mullr - The [puppetserver](https://github.com/puppetlabs/puppetserver/commit/78dd43a4623491e4bbe0575100a66db499afa642) and upstream work on the gzip compression side has been merged now.  We don't have a target date in mind for the next Puppet Server release which would have this work yet but it would be in any of the latest nightlies being published to http://nightlies.puppetlabs.com.  Anything else you think we should address before merging this PR?

mullr:
I'm +1, good to merge anytime. @senior?

PR #2176, opened 2017-02-10
Ruby acceptance tests -> Clojure integration tests

mullr:
before running the tests, you need to:

```
bundle install
lein install-gems
```


senior:
@mullr @rbrw The three of us should review this. As a reminder we're planning to do this in 3 steps/phases. The code here isn't an attempt to combine/collapse tests that are obviously covering similar things. Instead we converted tests from ruby/beaker straight into our full stack clojure equivalents. With this we should be able to compare the ruby version and the clojure version side by side. The new tests mirror the same directory structure and naming so it should be easy to connect two tests. We want to make sure we're testing the same things and haven't lost anything in the translation.

Once we have confidence that our tests are covering the same ground, we can delete the ruby based tests. We can then start to collapse and combine tests. These tests will get MUCH faster and smaller after that. Next we will probably need to create some basic/minimal ruby based smoke tests that will be our new acceptance tests. Things that can make sure our scripts work, upgrades are good to go etc.

test/puppetlabs/puppetdb/integration/storeconfigs/file_with_binary_template.clj
```diff
@@ -0,0 +1,27 @@
...
+                           (str "file { '" file-resource-path "':"
+                                "  content => template('" binary-template-path "'),"
+                                "  tag => 'binary_file',"
+                                "}")))
+
+      (testing "PDB should have stored the resource"
+        (is (= {:certname "binary-file-agent"
+                :type "File"
+                :title file-resource-path}
+               (first (int/pql-query pdb "resources [certname, type, title] { tag = 'binary_file' }"))))))))
```
senior: We should discuss how we want to go about querying in these tests. Here you're using PQL, previously we were hitting the `/resources` endpoint with an AST tag criteria.

test/puppetlabs/puppetdb/integration/storeconfigs/deactivated_nodes.clj
```diff
@@ -0,0 +1,27 @@
...
+                            (let [{:keys [out]} (int/run-puppet-as "collector" ps pdb "Notify <<| |>>")
+                                  check-result (if is-positive identity not)]
+                              (is (check-result (re-find #"Hello from exporter" out)))))]
+
+      (testing "Resources should be collected before deactivation"
+        (int/run-puppet-as "exporter" ps pdb exporter-manifest)
+        (test-collection true))
+
+      (testing "Resources from deactivated nodes should be ignored"
+        (int/run-puppet-node-deactivate pdb "exporter")
```
senior: In the ruby version it's checking standard out to ensure the node was deactivated. It doesn't seem all that useful as a test, but could it help in troubleshooting if something went wrong?

test/puppetlabs/puppetdb/integration/storeconfigs/puppetdb_query.clj
```diff
@@ -0,0 +1,30 @@
...
+            [puppetlabs.puppetdb.integration.fixtures :as int]))
+
+(deftest ^:integration puppetdb-query-function
+  (with-open [pg (int/setup-postgres)
+              pdb (int/run-puppetdb pg {})
+              ps (int/run-puppet-server [pdb] {})]
+    (testing "Initial agent run, to populate puppetdb with data to query"
+      (int/run-puppet ps pdb "notify { 'hello, world!': }"))
+
+    (let [test-out-file "/tmp/test_puppetdb_query.txt"]
```
senior: We should switch this to fs/temp-file

test/puppetlabs/puppetdb/testutils/db.clj
```diff
@@ -163,7 +163,7 @@
     db
     user]))
 
-(defn- drop-test-db [db-config]
+(defn drop-test-db [db-config]
```
senior: +1

test/puppetlabs/puppetdb/testutils/services.clj
```diff
@@ -88,6 +88,32 @@
 (defn strip-retired-config [config]
   (update config :database dissoc :classname :subprotocol))
 
+;; TODO: rework call-with-puppetdb-instance to use this
```
senior: TODO

test/puppetlabs/puppetdb/integration/soft_fail/soft_write_fail.clj
```diff
@@ -0,0 +1,26 @@
...
+(deftest ^:integration soft-write-fail
+  (with-open [pg (int/setup-postgres)
+              pdb (int/run-puppetdb pg {})
+              ps (int/run-puppet-server [pdb] {:terminus {:main {:soft_write_failure true}}})]
+    (tk-app/stop (-> pdb int/info-map :app))
+
+    (testing "Agent run should fail for manifest which collects resources"
+      (try
+        (int/run-puppet ps pdb "Notify <<| |>>")
+        (is false "The puppet run should have thrown an exception")
```
senior: Same `thrown+-with-msg?` comment as below

test/puppetlabs/puppetdb/integration/soft_fail/soft_write_fail.clj
```diff
@@ -0,0 +1,26 @@
...
+        (catch clojure.lang.ExceptionInfo e
+          (is (re-find #"Could not retrieve resources from the PuppetDB"
+                       (-> e .getData :err))))))
+
+    (testing "Agent run should succeed for manifest which exports resources"
+      (try
+        (int/run-puppet ps pdb "@@notify { 'exported notify': }"
```
senior: I don't understand this part. It looks like the agent run didn't succeed because it threw an exception?

test/puppetlabs/puppetdb/integration/apply/fact_storage.clj
```diff
@@ -0,0 +1,21 @@
...
+                                                {:env {"FACTER_foo" "testfoo"}})]
+        (is (re-find #"testfoo" out))))
+
+    (testing "Run again to ensure we aren't using puppetdb for the answer"
+      (let [{:keys [out]} (int/run-puppet-apply pdb "notice($foo)"
+                                                {:env {"FACTER_foo" "second test"}})]
+        (is (re-find #"second test" out))))
+
+    (testing "Check that the fact was put in puppetdb"
+      (= [:value "second test"] (int/pql-query pdb "facts [value] { name='foo' }")))))
```
senior: Missing `is` here, I think it will fail too.

test/puppetlabs/puppetdb/integration/db_garbage_collection/report_ttl.clj
```diff
@@ -0,0 +1,30 @@
...
+    (testing "Sleep for one second to make sure we have a ttl to exceed"
+      (Thread/sleep 1000))
+
+    (with-open [pdb (int/run-puppetdb pg {:database {:report-ttl "1s"}})
+                ps (int/run-puppet-server [pdb] {})]
+
+      ;; TODO: this is really fragile, it would be better if there were some way
+      ;; to tell that the GC had finished.  We could maybe scrape the logs if this
+      ;; proves too fragile.
+      (testing "sleep 5 seconds to allow GC to complete"
```
senior: Maybe we need to add something to [this](https://docs.puppet.com/puppetdb/4.3/api/admin/v1/cmd.html#post-pdbadminv1cmd). What do you think @rbrw?

test/puppetlabs/puppetdb/integration/db_resilience/db_fallback.clj
```diff
@@ -0,0 +1,22 @@
+(ns puppetlabs.puppetdb.integration.db-resilience.db-fallback
```
senior: Just verifying here @mullr, I didn't see much value in the postgres_restart test, is that why you didn't port it over?

test/puppetlabs/puppetdb/integration/db_resilience/db_fallback.clj
```diff
@@ -0,0 +1,22 @@
...
+    (testing "Agent run against pdb1"
+      (int/run-puppet ps pdb1 "notify { 'hello, world!': }")
+      (is (= 1 (count (int/pql-query pdb1 "nodes {}")))))
+
+    (testing "Fallback to pdb2"
+      (tk-app/stop (:app pdb1))
+      (is (= 0 (count (int/pql-query pdb2 "nodes {}"))))
+
+      (int/run-puppet ps pdb2 "notify { 'hello, world!': }")
+      (is (= 1 (count (int/pql-query pdb2 "nodes {}")))))))
```
senior: Seems like we're missing half of this test, was there a reason it was excluded? The ruby one stands back up pdb1 and ensures it falls back to the original.

mullr:
- Print out versions of all the gems on test startup

test/puppetlabs/puppetdb/integration/reports/producer.clj
```diff
@@ -0,0 +1,20 @@
...
+(deftest ^:integration producer-populated
+  (with-open [pg (int/setup-postgres)
+              pdb (int/run-puppetdb pg {})
+              ps (int/run-puppet-server-as "my_puppetserver" [pdb] {})]
+    (testing "Initial agent run, to populate puppetdb with data to query"
+      (int/run-puppet-as "my_agent" ps pdb
+                         (str "notify { \"hi\":"
+                              "  message => \"Hi my_agent\" "
+                              "}")))
+    (let [result (int/entity-query pdb "/reports"
```
mullr: On the subject of pql-query vs. entity-query: I converted things to pql-query because that's what I had at hand, and because I didn't see any real difference to the test between the two. And also it makes the code shorter.

test/puppetlabs/puppetdb/integration/reports/report_query_by_timestamp.clj
```diff
@@ -0,0 +1,35 @@
...
+      ;; Sleep 1 ms so that we are sure to get a start-time/end-time
+      ;; before/after the puppet run
+      (Thread/sleep 1)
+      (testing "Initial agent run, to populate puppetdb with data to query"
+        (int/run-puppet-as "my_agent" ps pdb
+                           (str "notify { \"hi\":"
+                                "  message => \"Hi my_agent\" "
+                                "}")))
+      (Thread/sleep 1)
+      (let [end-time (now)
```
mullr: Ruby side has a test that no events are returned if the timestamp range doesn't match.

test/puppetlabs/puppetdb/integration/reports/report_storage.clj
```diff
@@ -0,0 +1,75 @@
...
+                         ["--noop"])
+      (let [result (int/entity-query pdb "/reports"
+                                     ["=" "certname" "my_agent"]
+                                     {"order_by" (json/generate-string [{"field" "receive_time"
+                                                                         "order" "desc"}])})
+            [event :as events] (remove #(= (:resource_type %) "Schedule")
+                                       (int/entity-query pdb "/events"
+                                                         ["=" "report" (-> result first :hash)]))]
+
+        (are [x y] (= x y)
```
mullr: :heart: are

test/puppetlabs/puppetdb/integration/reports/report_storage.clj
```diff
@@ -0,0 +1,75 @@
...
+      (int/run-puppet-as "my_agent" ps pdb
+                         (str "notify { \"hi\":"
+                              "  message => \"Hi my_agent\" "
+                              "}")
+                         ["--noop"])
+      (let [result (int/entity-query pdb "/reports"
+                                     ["=" "certname" "my_agent"]
+                                     {"order_by" (json/generate-string [{"field" "receive_time"
+                                                                         "order" "desc"}])})
+            [event :as events] (remove #(= (:resource_type %) "Schedule")
```
mullr: Can we replicate the 'this is weird' comment from the ruby test? 'cause it's weird.

test/puppetlabs/puppetdb/integration/reports/report_storage.clj
```diff
@@ -0,0 +1,75 @@
...
+              ps (int/run-puppet-server-as "something_silly" [pdb] {})]
+    (testing "Initial agent run, to populate puppetdb with data to query"
+      (int/run-puppet-as "my_agent" ps pdb
+                         (str "notify { \"hi\":"
+                              "  message => \"Hi my_agent\" "
+                              "}")
+                         ["--noop"])
+      (let [result (int/entity-query pdb "/reports"
+                                     ["=" "certname" "my_agent"]
+                                     {"order_by" (json/generate-string [{"field" "receive_time"
```
mullr: There's only one report, order_by doesn't actually do anything.

test/puppetlabs/puppetdb/integration/reports/report_storage.clj
```diff
@@ -0,0 +1,75 @@
...
+          "message" (:property event)
+          "Hi my_agent" (:new_value event))))
+
+    (testing "agent run without noop"
+      (int/run-puppet-as "my_agent" ps pdb
+                         (str "notify { \"hi\":"
+                              "  message => \"Hi my_agent\" "
+                              "}"))
+
+      (let [[report] (int/entity-query pdb "/reports"
```
mullr: I should do this too, instead of calling 'first' on everything.

test/puppetlabs/puppetdb/integration/reports/transaction_uuid.clj
```diff
@@ -0,0 +1,26 @@
...
+              pdb (int/run-puppetdb pg {})
+              ps (int/run-puppet-server-as "something_silly" [pdb] {})]
+    (testing "Initial agent run, to populate puppetdb with data to query"
+      (int/run-puppet-as "my_agent" ps pdb
+                         (str "notify { \"hi\":"
+                              "  message => \"Hi foo\" "
+                              "}")))
+    (let [[report] (int/entity-query pdb "/reports"
+                                     ["=" "certname" "my_agent"]
+                                     {"order_by" (json/generate-string [{"field" "receive_time"
```
mullr: Just one report, so order_by doesn't do anything

test/puppetlabs/puppetdb/integration/storeconfigs/dup_collected_resources.clj
```diff
@@ -0,0 +1,22 @@
...
+              ps (int/run-puppet-server [pdb] {})]
+
+    (testing "Run puppet on exporters to create duplicate exported resources"
+      (doseq [certname ["exporter1" "exporter2"]]
+        (int/run-puppet-as certname ps pdb "@@notify { 'DUPE NOTIFY': }")))
+
+    (testing "Run puppet on collector and expect failure"
+      (try
+        (int/run-puppet-as "collector" ps pdb "Notify <<| title == 'DUPE NOTIFY' |>>")
+        (is false "The collector puppet run should have thrown an exception")
```
mullr: Fixed and pushed

test/puppetlabs/puppetdb/integration/soft_fail/soft_write_fail.clj
mullr: That's right. It's not supposed to succeed, since exported resource collection requires puppetdb, but we stopped it above.

mullr: oh wait, I'm sorry, I was confusing which is which.

mullr: This try/catch is in place because run-puppetdb always waits for commands to be processed. With puppetdb stopped, they will of course never be processed.

mullr: I'll add a comment to that effect.

test/puppetlabs/puppetdb/integration/db_resilience/db_fallback.clj
mullr: I think it's valuable, but we can't actually test that here. It's checking that we keep working when postgres itself goes away and comes back. i.e. we're using the connection pool appropriately. I don't know how to replicate that in this context. If we moved to self-hosting postgres out of the tests, we could do it.

test/puppetlabs/puppetdb/integration/storeconfigs/deactivated_nodes.clj
mullr: oh, the 'puppet node status' call? that was an oversight.

test/puppetlabs/puppetdb/integration/db_resilience/db_fallback.clj
mullr: At the time the second half didn't seem too important, but here in the light of day it kinda does. I'll add it.

test/puppetlabs/puppetdb/integration/fixtures.clj
```diff
@@ -0,0 +1,375 @@
...
+  (close [_] (tk-app/stop app)))
+
+(defn stringify-keys-with-ns [m]
+  (let [f (fn [[k v]] (if (keyword? k) [(subs (str k) 1) v] [k v]))]
+    ;; only apply to maps
+    (walk/postwalk (fn [x] (if (map? x) (into {} (map f x)) x)) m)))
+
+;; TODO: Get this fixed version into the clj-typesafe-config library and use it
+;; from there
+(defn map->hocon-string
```
senior: is it just the namespaced parameters that caused us issues?

mullr: yeah

test/puppetlabs/puppetdb/integration/fixtures.clj
```diff
@@ -0,0 +1,375 @@
...
+(defn run-puppetdb [postgres-server config-overrides]
+  (let [{:keys [app base-url]} (start-test-puppetdb "test-resources/integration-bootstrap.cfg"
+                                                    "test-resources/integration-puppetdb.conf"
+                                                    (ks/deep-merge config-overrides
+                                                                   {:database (info-map postgres-server)})
+                                                    10)
+        {:keys [host port]} base-url]
+    (PuppetDBTestServer. {:hostname host
+                          :port port
+                          :query-base-url (assoc base-url :prefix "/pdb/query" :version :v4)
```
senior: Any objections to switching this to use functions in services.clj? Or adapting those to allow a base-url?

mullr: :+1: but I'm not entirely sure which ones you mean

test/puppetlabs/puppetdb/integration/reports/basic_event_query.clj
```diff
@@ -0,0 +1,27 @@
...
+                              "  message => \"Hi my_agent\" "
+                              "}")))
+    (let [result (int/entity-query pdb "/events"
+                                   ["extract" ["old_value" "new_value"]
+                                    ["and"
+                                     ["=" "certname" "my_agent"]
+                                     ["=" "resource_type" "Notify"]
+                                     ["=" "status" "success"]
+                                     ["~" "property" "^[Mm]essage$"]
+                                     ["~" "message" "Hi my_agent"]]])]
```
senior: Doesn't really serve a purpose, but I skipped over it initially because I didn't have the right "unparse" function for the query. I needed it for another test anyway so I'll add it back in.

test/puppetlabs/puppetdb/integration/reports/producer.clj
senior: I guess as long as we feel there is adequate coverage of the individual endpoints, we can move to PQL. I left the queries as is more for testing the endpoints than the AST queries.

test/puppetlabs/puppetdb/integration/reports/report_storage.clj
senior: correct, not sure why it was in the original tests, I'll remove it though

test/puppetlabs/puppetdb/integration/reports/report_query_by_timestamp.clj
senior: Where are you seeing that? Seems like a good assertion to have though.

test/puppetlabs/puppetdb/integration/install_gems.clj
```diff
@@ -0,0 +1,30 @@
...
+  ;; the puppet gem brings hiera and facter with it
+  ;; TODO: it would be better to just install the puppet-agent package in CI
+  (let [puppet-version  (get env :puppet-version "latest")]
+    (if (not= puppet-version "latest")
+      (gem-run! config "install" "puppet" "--version" puppet-version)
+      (gem-run! config "install" "puppet")))
+
+  ;; Install the puppetserver vendored gems listed inside its jar; this is where
+  ;; ezbake gets them
+  (let [gem-list (string/split (slurp (io/resource "ext/build-scripts/gem-list.txt")) #"\n")]
```
rbrw: Wondered if we might need to worry about filtering blank or comment lines or...

test/puppetlabs/puppetdb/integration/install_gems.clj
```diff
@@ -0,0 +1,30 @@
+(ns puppetlabs.puppetdb.integration.install-gems
+  (:require [clojure.java.io :as io]
+            [puppetlabs.puppetserver.cli.subcommand :as cli]
+            [puppetlabs.services.jruby.jruby-puppet-core :as jruby-puppet-core]
+            [puppetlabs.services.jruby-pool-manager.jruby-core :as jruby-core]
+            [clojure.string :as string]
+            [environ.core :refer [env]]))
```
rbrw: Presumably all of these sub-execs are bottoming out in direct (non-shell) execs -- if we're not sure, might want to verify.

ext/travisci/install_puppetserver.sh
```diff
@@ -0,0 +1,37 @@
...
+    SED_COMMAND="s|$SED_REGEX|$SED_REPLACEMENT|"
+
+    set -e
+    set -x
+
+    sed -i -e "/$SED_ADDRESS/ $SED_COMMAND" $file
+}
+
+lein-pprint() {
+    export PUPPETSERVER_HEAP_SIZE=1G
```
mullr: iirc, doing it as a prefix didn't work for reasons I don't understand at all. This was the only way I got it to work

test-resources/puppetserver/puppetserver.conf
```diff
@@ -0,0 +1,136 @@
+global: {
+    logging-config: ./test-resources/puppetserver/logback-dev.xml
+}
+
+product: {
+    update-server-url: "http://some.update.server.that.does.not.exist/"
```
mullr: hah, indeed. This is copied directly from the puppetserver codebase, fwiw.

test/puppetlabs/puppetdb/integration/db_garbage_collection/report_ttl.clj
```diff
@@ -0,0 +1,30 @@
...
+        (Thread/sleep 5000))
+
+      (testing "Verify that the report has been deleted"
+        (is (= 0 (count (int/pql-query pdb "reports { certname = 'ttl-agent' }"))))))))
```
mullr: `=` is actually special inside off clojure.test assertions, gives you the 'expected' vs. 'actual' business when it fails.

test/puppetlabs/puppetdb/integration/fixtures.clj
```diff
@@ -0,0 +1,382 @@
...
+            [puppetlabs.trapperkeeper.config :as tk-config]
+            [puppetlabs.trapperkeeper.testutils.bootstrap :as tkbs]
+            [puppetlabs.trapperkeeper.core :as tk]
+            [yaml.core :as yaml]
+            [slingshot.slingshot :refer [throw+]]
+            [clj-time.format :as tfmt])
+  (:import [com.typesafe.config ConfigValueFactory]))
+
+(defprotocol TestServer
+  (info-map [this]))
```
mullr: hm. `server-info`?

mullr: Are you saying you don't approve of revere-wordy-hungarian-notation? ;)

test/puppetlabs/puppetdb/integration/fixtures.clj
```diff
@@ -0,0 +1,382 @@
...
+
+(defn mri-agent-dir []
+  (-> (sh "bundle" "show" "puppet") :out clojure.string/trim))
+
+(defn jruby-agent-dir []
+  (->> (fs/list-dir "vendor/puppetserver-gems/gems/")
+       sort
+       (filter #(re-matches #"puppet-\d+\.\d+\.\d+" (fs/base-name %)))
+       (map fs/absolute)
+       last))
```
mullr: argh you're right, I'll redo it

test/puppetlabs/puppetdb/integration/fixtures.clj
```diff
@@ -0,0 +1,382 @@
...
+
+(defn install-terminus-into [puppet-dir]
+  ;; various pieces of the terminus have to actually be inside of puppet itself to work
+  (let [base-dir "puppet"
+        base-uri (.toURI (io/file base-dir))
+        terminus-files (->> (fs/find-files base-dir #".*")
```
mullr: I think it's fine here. It may even be what the packages do.

test/puppetlabs/puppetdb/integration/install_gems.clj
mullr: don't think so. This file is read by a number of other tools, and they're all fairly primitive.

test/puppetlabs/puppetdb/integration/storeconfigs/deactivated_nodes.clj
```diff
@@ -0,0 +1,33 @@
...
+
+(deftest ^:integration deactivated-nodes
+  (with-open [pg (int/setup-postgres)
+              pdb (int/run-puppetdb pg {})
+              ps (int/run-puppet-server [pdb] {})]
+
+    (let [exporter-manifest  "@@notify { 'Hello from exporter': }"
+          test-collection (fn test-collection [should-be-active]
+                            (let [{agent-stdout :out} (int/run-puppet-as "collector" ps pdb "Notify <<| |>>")
+                                  {status-stdout :out} (int/run-puppet-node-status pdb "exporter") ]
```
rbrw: Lines getting a bit long.

test/puppetlabs/puppetdb/integration/storeconfigs/deactivated_nodes.clj
```diff
@@ -0,0 +1,33 @@
...
+            [puppetlabs.puppetdb.integration.fixtures :as int]
+            [me.raynes.fs :as fs]))
+
+(deftest ^:integration deactivated-nodes
+  (with-open [pg (int/setup-postgres)
+              pdb (int/run-puppetdb pg {})
+              ps (int/run-puppet-server [pdb] {})]
+
+    (let [exporter-manifest  "@@notify { 'Hello from exporter': }"
+          test-collection (fn test-collection [should-be-active]
```
rbrw: `should-be-active?` or even thought about refactoring, to eliminate the boolean, i.e. vaguely:
```clojurescript
(let [...
      info #(mapv :out [(int/run-puppet-as "collector" ps pdb "Notify <<| |>>")
                        (int/run-puppet-node-status pdb "exporter")])
      verify-active #(let [[agent status] (info)]
                       (is (re-find #"Hello from exporter" agent))
                       (is (re-find #"Currently active" status)))
      verify-inactive #(let [[agent status] (info)]
                         (is (not (re-find #"Hello from exporter" agent-stdout)))
                         (is (re-find #"Deactivated at" status-stdout)))]
  ...)
```
or whatever

rbrw:
OK, finished current review.

ext/travisci/install_puppetserver.sh
rbrw: You mean
```sh
  PUPPETSERVER_HEAP_SIZE=1G "$lein" with-profile dev,ci pprint "$@" | sed -e 's/^"//' -e 's/"$//'
```
Didn't work?  And do we really need this just for pprint?  Hmm, Actually, there must be something else going on since this var is already exported in the envt via line 31 below.

test-resources/puppetserver/puppetserver.conf
rbrw: Yeah, I imagined it might have been.

test/puppetlabs/puppetdb/integration/db_garbage_collection/report_ttl.clj
rbrw: Ahh, right - I forgot.

test/puppetlabs/puppetdb/integration/fixtures.clj
rbrw: Hah - yeah, I think I like `info` or `server-info` or similar better here.

test/puppetlabs/puppetdb/integration/install_gems.clj
rbrw: OK, well note that depending on how the shell script does it, it might at least be implicitly quashing "blank" lines -- though still may not be important for us, just wondered.

.travis.yml
```diff
@@ -5,23 +5,27 @@ lein: 2.7.1
...
+      jdk: openjdk7
+    - env: PDB_TEST_LANG=clojure PDB_TEST_SELECTOR=:unit
+      jdk: oraclejdk7
+    - env: PDB_TEST_LANG=clojure PDB_TEST_SELECTOR=:unit
+      jdk: oraclejdk8
+
+    # pdb + terminus + puppetserver + puppet integration tests
+    - env: PDB_TEST_LANG=clojure PDB_TEST_SELECTOR=:integration PUPPET_VERSION=4.6.2 PUPPETSERVER_VERSION=2.7.2
+      jdk: oraclejdk8
+    - env: PDB_TEST_LANG=clojure PDB_TEST_SELECTOR=:integration PUPPET_VERSION=latest PUPPETSERVER_VERSION=master
```
mullr: aw man, I really don't want to go through another travis test/fix cycle...

ext/travisci/install_puppetserver.sh
mullr: It didn't work. And yes we need it for pprint, since that loads the puppetserver jar, which will complain straight to stdout if this isn't set, making the version number non-parsable.

test/puppetlabs/puppetdb/integration/db_garbage_collection/report_ttl.clj
mullr: for the sleep above, we actually track that in metrics already. I may be able to pull it via tk.

test/puppetlabs/puppetdb/integration/fixtures.clj
```diff
@@ -0,0 +1,382 @@
...
+
+(defn start-test-puppetdb [bootstrap-config-file config-file config-overrides bind-attempts]
+  (when (zero? bind-attempts)
+    (throw (RuntimeException. "Repeated attempts to bind port failed, giving up")))
+  (let [port (svc-utils/open-port-num)
+        tmp-puppetdb-conf (str (tu/temp-file "puppetdb" ".conf"))
+        _ (adjust-config-file config-file tmp-puppetdb-conf
+                              (-> config-overrides
+                                  (assoc-in [:jetty :ssl-port] port)
+                                  (assoc-in [:global :vardir] (str (tu/temp-dir)))))
```
mullr: The temp file handling in here is sloppy. I think it's ok for now, but in fairly short order we should do a 1-dir-per-test kind of thing.

test/puppetlabs/puppetdb/integration/install_gems.clj
mullr: It's Runtime.exec, which shows no sign of being a shell exec

test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -0,0 +1,29 @@
+(ns puppetlabs.puppetdb.integration.smoke
+  (:require [clojure.test :refer :all]
+            [puppetlabs.puppetdb.integration.fixtures :as int]))
+
+(deftest ^:integration simple-agent-run
+  (with-open [pg (int/setup-postgres)
+              pdb (int/run-puppetdb pg {})
+              ps (int/run-puppet-server [pdb] {})]
+    (testing "Agent run succeeds"
+      (let [{:keys [out]} (int/run-puppet-as "my-agent" ps pdb "notify { 'hello, world!': }")]
```
mullr: `run-puppet` (well `run-bundle-exec` under it) will print all of that if the command fails

test/puppetlabs/puppetdb/integration/storeconfigs/deactivated_nodes.clj
mullr: I kinda like the boolean and laying out the assertions in this case, makes it very easy to follow what's going on

mullr:
@rbrw ok, all cleaned up

ext/travisci/install_puppetserver.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
```
mullr: added set -e and pipefail

.travis.yml
rbrw: Not critical - just thought it might be nice if we happened to know they supported that.

ext/travisci/install_puppetserver.sh
rbrw: Sounds good (and if you're not famiiar with it, note that pipefail can behave in ways you might not expect wrt, for example, grep, i.e. cause the rest of the pipeline to fail when a negative result was not an error).

ext/travisci/install_puppetserver.sh
rbrw: OK, well there's still likely something wrong here because aside from the fact that having it on the same line should work, the variable has already been exported by the time this code runs (see line 31) -- it's very suspicious if it fails when the setting in this function is removed.

test/puppetlabs/puppetdb/integration/install_gems.clj
rbrw: Yep, looks like that should be fine (random aside: had no idea the (exec string) method (as compared to the string[] varieties) behaves the way it does (tokenizing on whitespace)).

mullr:
@rlb @senior tests are green now, I think we're good to merge if y'all are ok with the codee

rbrw:
Not critical, but I might quash those last JVM arg-related commits into a single one with the change we want.

PR #2178, opened 2017-02-24
(maint) Update bidi to 2.0.12

aperiodic:


mullr:
@aperiodic generally lgtm, anything we should be watching out for as part of this update?

aperiodic:
@mullr I'm honestly not sure what all the changes are, since the [bidi changelog contains only one entry](https://github.com/juxt/bidi/blob/master/CHANGELOG.md#change-log). There shouldn't be that many changes since the 2.0.z series is the successor to 1.25.z. The documented change doesn't seem to affect PuppetDB, since the old key doesn't appear anywhere in PuppetDB's source.

PR #2179, opened 2017-02-28
(PE-18347) Add metric_id option to http requests in termini

rlinehan:
Add the `metric_id` option to the http requests in the facts terminus, query
function, and for command submission. These `metric_id`s will be used by the
http client Puppet Server uses to create metrics in its registry, which can
then be queried to determine how long requests Puppet Server makes to PuppetDB
are taking.

For the facts `find` action, the `metric_id` is `[:puppetdb :facts :find
<node-name>]`.

For the facts `search` action, the `metric_id` is `[:puppetdb :facts
:search]`. The query is not included as part of the `metric_id` because it
would not create useful metrics to capture - each search query can be pretty
much anything, and tracking a single search query over time probably won't be
very useful.

For the http call made by the puppetdb query function, the `metric_id` is
`[:puppetdb :query]`. This metric-id does not include the actual query, since
there are so many possible queries that tracking metrics on them does not seem
particularly useful.

For the http requests for submitting commands, `metric_id`s will be
`[:puppetdb :command <command name> <certname>]`.


mullr:
@rlinehan For the command metrics, why do you want certname in the metric id? That's the certname of the node being managed, so there will be lots of them.

rlinehan:
@mullr same is true for certname in the `facts.find`. We track certname also on compilation metrics and some others. Some users may find it useful if there is a specific node that is having problems. I guess it is less likely that the problem will be on the puppetdb side of things, but still seems like it could be potentially useful.

The way the metric ids work (which I realize I forgot to put in the commit message... I can amend it), is that a metric gets created for each element in the array - so for `[:puppetdb :command <command name> <node name>]` you would get `puppetdb`, `puppetdb.command`, `puppetdb.command.<command name>` and `puppetdb.command.<command name>.<node name>`.

mullr:
@rlinehan ah, I see. That sounds quite useful.

rlinehan:
Jenkins, retest this please.

rlinehan:
Realized I had missed the `resource` terminus, so amended my commit to include that and also updated the commit message to describe how metric-ids work.

rlinehan:
Hmmm looks like actually I broke things....

rlinehan:
Well, I figured out why I broke things. Locally I'm running this against Puppet Server at HEAD of master, which has https://github.com/puppetlabs/puppetserver/pull/1331/files#diff-5805ba6eb4857e236651f3c3fcdffeb0R75, which changes `.get` to allow for `options` to be passed in as a third argument.

So, the question is, do we ever make any requirements about PuppetDB and Puppet Server compatibility? Can we say that if you run PuppetDB 5 you must run Puppet Server 5?

rlinehan:
Heyyyy tests are passing! @camlow325 or @jpinsonault could you take a look at this when you get a chance and make sure this seems reasonable to you?

puppet/lib/puppet/util/puppetdb/http.rb
```diff
@@ -222,6 +222,29 @@ def self.reset_query_failover()
...
+    #
+    # Since we don't have any guarantees over what version of Puppet/Puppet
+    # Server this terminus code will run under, this method inspects the `get`
+    # method to check its arity, and then calls it with the correct number of
+    # arguments.
+    def self.multi_arity_get(http_instance, path, headers, options={})
+      # `parameters()` returns the parameters of a method as an array of
+      # tuples, with a tuple for each parameter, e.g. `[[:req :a], [:opt
+      # :b]]`. Counting this array gives the arity.
+      arity = http_instance.class.instance_method(:get).parameters.count
```
camlow325: Do you know how fast this call is?  If this is measurably slow, I wonder if it would be better to memoize the parameter count.

puppet/lib/puppet/util/puppetdb/http.rb
```diff
@@ -222,6 +222,29 @@ def self.reset_query_failover()
...
+      if arity == 2
```
camlow325: Could do a case/when here instead, I suppose.

senior:
@rlinehan We've not had a requirement for a specific version of Puppet Server to run PuppetDB. Did that part get fixed in this PR or is merging it mean we must have Puppet Server 5? Also has Puppet Server 5 been released yet? I just see 2.7.2-SNAPSHOT in the repo now.

puppet/lib/puppet/util/puppetdb/http.rb
rlinehan: It seems pretty fast - benchmarking with calling it 5000 times memoized and not resulted in 

```
                 user     system      total        real
not:         0.000000   0.000000   0.000000 (  0.004518)
memoized:    0.000000   0.000000   0.000000 (  0.000412)
```

so... a small difference, and memoizing it is very easy, so might as well.

rlinehan:
Pushed up another commit memoizing the call to get the parameter count.

@jb I have done a small amount of functional testing by running PuppetDB with this change locally along with Puppet Server at both an old commit and a newer-commit - including with the status endpoint changes (that I haven't pushed up anywhere). It works at a commit on stable (Puppet Server 2.7.2) and also on my newer changes (will be Puppet Server 5), and reports metrics to the status endpoint.

@senior Yes, with the second commit (adding the `multi_arity_get` function) this should work with any version of Puppet Server or with just Ruby Puppet. The acceptance tests were failing before because y'all are running tests against Puppet Server 2.7.2. The fact that these tests are now passing (with the new commit) means that the compatibility will not be broken.

camlow325:
This looks okay to me to merge.  @senior, you good with this?

PR #2180, opened 2017-03-01
(maint) update the docs on dashboard access

wkalt:
The docs previously gave the impression that changing the cleartext
hosts setting was the preferred mechanism of access. This updates them
to recommend an ssh tunnel instead.


mullr:
@wkalt Were you thinking this should go into the next PE LTS z release?

wkalt:
@mullr if you're suggesting we send it to an earlier branch I'm cool with that. Otherwise it's just a docs change -- it's not tied to any release and should go in as soon as it's reviewed. It's been warranted for years.

senior:
@mullr I'm good to go for merging this in, were you wanting the PR on a different branch?

PR #2182, opened 2017-03-02
(PDB-3107) Allow extra headers in commands

mullr:
New Relic's java agent likes to inject extra JMS headers; update our message
schema to allow for unexpected keys so this doesn't bring the world down.


mullr:
@mullr check this works with newrelic's agent

rbrw:
lgtm

PR #2183, opened 2017-03-02
(PDB-3298) Don't carry fact values as valuemaps

rbrw:


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -943,46 +993,63 @@
...
-      (not= (:value row) (:value new-valuemap))))
+(defn fact-value-row-changes
+  "Returns logically false if new-val does not differ from the row data,
+  logically true otherwise, and the true value will be a hash of
+  new-val if one was computed during the evaluation."
+  [new-val type-id hash v vstr vint vfloat vbool]
+  (or (not= type-id (fact-value-type-id new-val))
+      (cond
+        ;; faster direct compare
+        (or (= type-id db-fv-int)
```
senior: It might make this block more readable to name these predicates. Maybe this is `primitive-fact-value?`, below is `nil-fact-value?` and below that is `big-fact-value?`. That might help take some of the noise away here and improve clarity

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -907,22 +907,72 @@
...
+   (let [jsonb (when-not (nil? value)
+                 (sutils/munge-jsonb-for-storage value))
+         large? (and jsonb (>= (count (.getValue jsonb))
+                               facts/large-value-threshold))
+         hash (when large?
+                (or hash (storage-hash value)))]
+     (value->value-row value hash jsonb)))
+  ([value hash jsonb]
+   (let [id (fact-value-type-id value)]
+     (cond
```
senior: I was thinking about how to improve this. You could let bind `(let [v [(fact-value-type-id value) hash jsonb]]`. Below then becomes `(into v [value nil nil nil])` etc.

I also thought about case here, I thought it might work since the vars are const, but it won't condp would reduce the repeating yourself there, but there's a hit to clarity too.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -992,38 +1059,50 @@
...
-              (for [{:keys [path id] :as row} rows]
-                (if-let [new-vm (pathstr->valuemap path)]
-                  (if (fact-value-changed? row new-vm)
-                    [:changed path id new-vm]
-                    [:unchanged path id])
-                  [:gone path id])))))
-
+              (for [[path id type-id hash v vstr vint vfloat vbool :as row] rows]
+                ;; Classify row as :changed :unchanged: :gone
+                (let [new-val (pathstr->value path missing)]
```
senior: I find missing here non-idiomatic. I'd also pretty use get if I'm using a default, so I think typically you'd see
`(get pathstr->value path ::not-found)` or something similar.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -992,38 +1059,50 @@
...
+           (fn [[col-names & rows]]
              (doall
-              (for [{:keys [path id] :as row} rows]
-                (if-let [new-vm (pathstr->valuemap path)]
-                  (if (fact-value-changed? row new-vm)
-                    [:changed path id new-vm]
-                    [:unchanged path id])
-                  [:gone path id])))))
-
+              (for [[path id type-id hash v vstr vint vfloat vbool :as row] rows]
```
senior: This seems like a `map` as well

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -56,23 +56,32 @@
...
+         "       f.value_float,"
+         "       f.value_boolean"
+         "  from factsets fs"
+         "  inner join facts as f on fs.id = f.factset_id"
+         "  inner join fact_paths as fp on f.fact_path_id = fp.id"
+         "  where fp.depth = 0 and fs.certname = ?")
+    certname]
+   {:as-arrays? true}
+   (fn [[col-names & rows]]
+     (doall
```
senior: The `doall` is not necessary here.

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -135,14 +144,72 @@
...
 
+(deftest fact-value-row-changes-basic-behavior
+  (let [changes fact-value-row-changes
+        value-row (fn [x hash]
+                    (let [type-id (fact-value-type-id x)]
+                      [type-id
+                       hash
+                       (sutils/munge-jsonb-for-storage x)
+                       (when (zero? type-id) x)      ; value_string
+                       (when (= 1 type-id) x)        ; value_integer
```
senior: Why use literals here?

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -135,14 +144,72 @@
...
+                       (when (= 2 type-id) x)        ; value_float
+                       (when (= 3 type-id) x)        ; value_boolean
+                       ]))]
+    (is (not (apply changes "x" (value-row "x" nil))))
```
senior: I'd rub a little `(are..)` on this block and below

src/puppetlabs/puppetdb/scf/storage.clj
senior: Another idea on the predicates is
```
(def primitive-fact-value #{db-fv-int ...})
```

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: I suppose I felt like the alignment made it fairly clear the columns are duplicated, and with the current formulation, it's also clear there's no "extra work" per result vector, i.e. we build a single vector and return it.

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: Yeah, I obviously haven't been entirely convinced by the idiom -- I've tended to use (Object.) because then there's no question of aliasing, and so it's not possible to interfere with the function from the outside (either accidentally or intentionally), and there's no question that you have to think about, now or in the future; that said, in this case, it probably doesn't matter.

test/puppetlabs/puppetdb/scf/storage_test.clj
rbrw: OK - I'd been a little inclined against `(are ...)` because I thought it presented the wrong line numbers on failures, and if so, in simple cases, didn't mind the copy/pasting, but I'm happy to change it if y'all prefer.

test/puppetlabs/puppetdb/scf/storage_test.clj
rbrw: Forgot to convert.

test/puppetlabs/puppetdb/scf/storage_test.clj
rbrw: Ryan convinced me - although it does break line numbers, the output's not to difficult to decipher.

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: After some discussion, think we'll keep the vectors, but switch to condp to make it less chatty.

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: After a bit of discussion, think we might leave this -- trying to separate the predicate for the "primitive" case, for example, makes the linkage between the type selection and the `(or vint vfloat ...)` code a bit less clear.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -943,46 +993,64 @@
...
+        ;; things that may have a hash
+        (or (= type-id db-fv-string)
+            (= type-id db-fv-struct))
+        (if hash
+          (let [new-hash (storage-hash new-val)]
+            (when-not (= hash new-hash)
+              new-hash))
+          (cond
+            (= type-id db-fv-string)
+            (not= new-val vstr)
```
mullr: newline here, please.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -943,46 +993,64 @@
...
+            (= type-id db-fv-struct)
+            (not= new-val (sutils/parse-db-json v))
```
mullr: here too

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -992,38 +1060,50 @@
...
-
+              (map (fn [[path id type-id hash v vstr vint vfloat vbool]]
+                     ;; Classify row as :changed :unchanged: :gone
+                     (let [new-val (pathstr->value path ::not-found)]
+                       (if (= new-val ::not-found)
+                         [:gone path id]
+                         (if-let [change (fact-value-row-changes new-val
+                                                                 type-id hash
+                                                                 v vstr vint vfloat vbool)]
+                           (if (instance? PGobject change)  ; If so, it's a hash
```
mullr: This is icky, but I think it's a symptom of my comments above about fact-value-row-changes just being odd.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -907,22 +907,72 @@
...
+(defn fact-value-type-id [x]
+  (cond
+    (keyword? x) db-fv-string
+    (string? x) db-fv-string
+    (integer? x) db-fv-int
+    (float? x) db-fv-float
+    (kitchensink/boolean? x) db-fv-bool
+    (nil? x) db-fv-nil
+    (coll? x) db-fv-struct
+    :else (throw (Exception. (trs "Fact value {0} has unexpected type"
```
rbrw: I think so?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -943,46 +993,64 @@
...
-(defn fact-value-changed?
-  [row new-valuemap]
-  (or (not= (:value_type_id row) (:value_type_id new-valuemap))
-      (when-let [row-hash (:large_value_hash row)]
-        (not= row-hash (:large_value_hash new-valuemap)))
-      (not= (:value row) (:value new-valuemap))))
+(defn fact-value-row-changes
+  "Returns logically false if new-val does not differ from the row data,
+  logically true otherwise, and the true value will be a hash of
+  new-val if one was computed during the evaluation."
```
rbrw: Changing it to return `{:changed? ... :hash ...}` as discussed.

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: Yep, gone.

mullr:
travis is effectively green, merging.

PR #2184, opened 2017-03-03
Missing comma from example

Nekototori:
When trying to use the example on line 58, puppet parser errors due to the lack of comma after nodes, and is unable to do a successful query.  Adding the comma makes puppet happy.


wkalt:
@Nekototori thanks -- would you mind re-raising this on the stable branch? That way it will go live on the main docs as soon as it's merged.

Nekototori:
Re-raising on the stable.

PR #2186, opened 2017-03-07
(PDB-3297) Don't act as if facts might be arrays

rbrw:
In fact, there only a small number of array items in the current
database, yet we check nearly every value that comes back to see if it
might be an array that needs conversion.  Start stopping that.

Accordingly, don't attempt array conversion in call-with-query-rows, and
add a new call-with-array-converted-query-rows that preserves the old
behavior.  Continue to call the former in the fact related storage code,
but adjust everything else to use the latter to preserve existing
behavior.

As the original motivation, Wyatt noticed non-trivial cost for the
"isa?" test via coerce-to-array when processing facts.

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -134,24 +134,26 @@
    #(and (map? %)
          (valid-jdbc-query? (:results-query %)))))
 
-(defn coerce-to-array
-  ([x] (coerce-to-array vec x))
-  ([convert x]
-   (cond
-     (kitchensink/array? x) (convert x)
-     (isa? (class x) java.sql.Array) (convert (.getArray ^java.sql.Array x))
-     :else x)))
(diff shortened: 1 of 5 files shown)


rbrw:
This should go in after (and be rebased on) the resolution of the pending valuemaps PR (#2183).

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -134,16 +134,51 @@
...
+       (.setFetchSize stmt 500)
+       (with-open [rset (.executeQuery stmt)]
+         (try
+           (f (sql/result-set-seq rset opts))
+           (catch Exception e
+             ;; Cancel the current query
+             (.cancel stmt)
+             (throw e))))))))
+
+;; Since we only have a couple of array columns, we may want to
```
senior: I'm ok deprecating this if you think it's best, but I'd prefer we deprecated rather than leave a note asking that question.

PR #2187, opened 2017-03-07
(PDB-3323) Make report exists query use index

mullr:
While this query was correct, it didn't use the index we have in place on this
column, making it very slow.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -1194,8 +1194,8 @@
                 report-hash (shash/report-identity-hash report)]
            (jdbc/with-db-transaction []
              (let [shash (sutils/munge-hash-for-storage report-hash)]
-               (when-not (-> "select 1 from reports where hash = ? limit 1"
-                             (query-to-vec shash)
+               (when-not (-> "select 1 from reports where encode(hash, 'hex'::text) = ? limit 1"
+                             (query-to-vec report-hash)
                              seq)
                  (let [certname-id (certname-id certname)
                        row-map {:hash shash


npwalker:
Looks legit to me.  My only concern being that everyone touching PuppetDB needs to remember that when you query on the report hash you have to encode it first or you end up scanning the whole reports table.   

This is a very reasonably short term fix, I just wonder how we prevent it from happening again.  

```
pe-puppetdb=# explain analyze select 1 from reports where hash = '\x7ad27615d5c7de439855b245e5eb6d7245e812f6';
                                            QUERY PLAN                                            
--------------------------------------------------------------------------------------------------
 Seq Scan on reports  (cost=0.00..40.41 rows=1 width=0) (actual time=0.028..0.398 rows=1 loops=1)
   Filter: (hash = '\x7ad27615d5c7de439855b245e5eb6d7245e812f6'::bytea)
   Rows Removed by Filter: 192
 Planning time: 1.080 ms
 Execution time: 0.423 ms
(5 rows)

pe-puppetdb=# explain analyze select 1 from reports where encode(hash, 'hex'::text) = '\x7ad27615d5c7de439855b245e5eb6d7245e812f6';
                                                          QUERY PLAN                                                           
-------------------------------------------------------------------------------------------------------------------------------
 Index Scan using reports_hash_expr_idx on reports  (cost=0.27..8.29 rows=1 width=0) (actual time=0.044..0.044 rows=0 loops=1)
   Index Cond: (encode(hash, 'hex'::text) = '\x7ad27615d5c7de439855b245e5eb6d7245e812f6'::text)
 Planning time: 0.230 ms
 Execution time: 0.068 ms
(4 rows)

PR #2188, opened 2017-03-10
(PDB-3324) Remove acceptance tests which were ported to Clojure

mullr:


senior:
@mullr do we have an accounting for what new tests we need to write as smoke tests to replace these? or are there enough left behind to cover our scripts starting/stopping etc?

mullr:

PR #2189, opened 2017-03-14
(PDB-3307) Ensure queue depth reflects removals

rbrw:
Make sure that when files are randomly deleted from the queue, the queue
depth statistics are updated properly.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -486,22 +486,29 @@
   [cmdref q scf-write-db response-chan stats]
   (let [{:keys [command version] :as cmd} (queue/cmdref->cmd q cmdref)
         retries (count (:attempts cmdref))]
+    (if-not cmd
+      ;; Queue file is missing
+      (let [{:keys [command version]} cmdref]
+        (create-metrics-for-command! command version)
+        (mark-both-metrics! command version :seen)
+        (update-counter! :depth command version dec!)
+        (mark! (global-metric :fatal)))
+      (do
+        (create-metrics-for-command! command version)
 
-    (create-metrics-for-command! command version)
+        (mark-both-metrics! command version :seen)
+        (update! (global-metric :retry-counts) retries)
+        (update! (cmd-metric command version :retry-counts) retries)
 
-    (mark-both-metrics! command version :seen)
-    (update! (global-metric :retry-counts) retries)
-    (update! (cmd-metric command version :retry-counts) retries)
+        (mutils/multitime!
+         [(global-metric :processing-time)
+          (cmd-metric command version :processing-time)]
 
-    (mutils/multitime!
-     [(global-metric :processing-time)
-      (cmd-metric command version :processing-time)]
+         (process-command-and-respond! cmd scf-write-db response-chan stats)
+         (mark-both-metrics! command version :processed))
 
-     (process-command-and-respond! cmd scf-write-db response-chan stats)
-     (mark-both-metrics! command version :processed))
-
-    (queue/ack-command q cmd)
-    (update-counter! :depth command version dec!)))
+        (queue/ack-command q cmd)
+        (update-counter! :depth command version dec!)))))
 
 (def command-delay-ms (* 1000 60 60))
 
--- a/src/puppetlabs/puppetdb/queue.clj
+++ b/src/puppetlabs/puppetdb/queue.clj
@@ -250,21 +250,31 @@
                  file-extension))))
 
 (defn cmdref->cmd [q cmdref]
+  "Returns the command associated with cmdref, or nil if the file is
+  missing (i.e. it's been deleted)."
   (let [compression (:compression cmdref)
-        entry (cmdref->entry cmdref)]
-    (with-open [command-stream (stock/stream q entry)]
-      (assoc cmdref
-        ;; For performance reasons, it's best to wrap any decompression stream
-        ;; around the original command stream before calling into stream->json.
-        ;; stream->json wraps it's own BufferedReader around the stream before
-        ;; parsing bytes from it as json.  From testing, we've seen that if we
-        ;; wrap an extra buffered input stream between the original command
-        ;; stream and the decompressing stream, Gzip in particular, that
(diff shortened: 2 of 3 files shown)


src/puppetlabs/puppetdb/command.clj
```diff
@@ -495,22 +495,29 @@
...
-
-     (process-command-and-respond! cmd scf-write-db response-chan stats)
-     (mark-both-metrics! command version :processed))
-
-    (queue/ack-command q cmd)
-    (update-counter! :depth command version dec!)))
+    (if-not cmd
+      ;; Queue file is missing
+      (let [{:keys [command version]} cmdref]
+        ;; Does a message that's missing still qualify as :seen?
```
mullr: I think so. We observed its existence.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -250,21 +250,32 @@
                  file-extension))))
 
 (defn cmdref->cmd [q cmdref]
+  "Returns the command associated with cmdref, or logically false if
+  the file is missing (i.e. someone has deleted it)."
```
mullr: is it actually nil if the command is missing? I think being more specific is better.

src/puppetlabs/puppetdb/queue.clj
```diff
@@ -250,21 +250,32 @@
...
-        ;; parsing bytes from it as json.  From testing, we've seen that if we
-        ;; wrap an extra buffered input stream between the original command
-        ;; stream and the decompressing stream, Gzip in particular, that
-        ;; throughput would be much worse - could be up to 400x times slower.
-        :payload (stream->json (wrap-decompression-stream
-                                compression
-                                command-stream))
-        :entry entry))))
+        entry (cmdref->entry cmdref)
+        stream (do
```
mullr: superfluous `do`?

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1471,3 +1474,51 @@
                "Waited up to 5 seconds for 3 acknowledgement results")
 
            (is (= [nil nil nil] @ack-results))))))))
+
+(deftest missing-queue-files-accounted-for-in-stats
+  ;; Use a read-write lock to hold up command processing while we
```
mullr: stale comment? don't think it's a read-write lock now

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1471,3 +1474,51 @@
...
+                                             {:certname name
+                                              :producer_timestamp (Date.)})
+                                            ""))
+              depths #(mapv counters/value
+                            [(global-metric :depth)
+                             (cmd-metric "deactivate node" 3 :depth)])
+              qdir (get-path (conf/stockpile-dir (get-config)) "cmd" "q")
+              start-stats (stats)]
+          ;; Enqueue a few commands
+          (locking cmd-gate
```
mullr: It appears that the whole body of the test is under this lock; does it actually work?

src/puppetlabs/puppetdb/queue.clj
rbrw: For what it's worth, by default I tend to leave that unspecified in documentation so that the result is free to change later if that better suits a future implementation, and it also reminds not to use `(nil? result)` -- though arguably such code changes may be more likely in languages with additional false values (i.e. `return x[i] - y[i]`, etc.).

src/puppetlabs/puppetdb/queue.clj
rbrw: Ah, right, thanks - left over from debugging iirc.

test/puppetlabs/puppetdb/command_test.clj
rbrw: Hmm, it should only be locked until after we're finished trashing the queue, i.e. until ";; Lock has been released, wait for queue to drain" below.

src/puppetlabs/puppetdb/queue.clj
mullr: Postel's law?

rbrw: Hmm, well the way I tend to think if it, at least for an api spec is that there's no way to put the paste back in the tube once you've been more specific (callers would be well within their rights to use `nil?` if the function says it returns `nil`), but it doesn't matter here - happy to change it if you prefer.

rbrw: OK, so @mullr pointed out that the norm in the clojure.core docs is nil or non-nil (I'd remembered something else), and so I agree we should parallel that.

mullr:
+1; will merge once travis is green-ish.

rbrw:
Rebased this, and it's no longer passing tests here, so don't merge for now.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1471,3 +1474,63 @@
...
+      (svc-utils/with-puppetdb-instance
+        (let [dispatcher (get-service *server* :PuppetDBCommandDispatcher)
+              enqueue-command (partial enqueue-command dispatcher)
+              stats #(stats dispatcher)
+              deactivate (fn [name]
+                           (enqueue-command (command-names :deactivate-node)
+                                            3 name nil
+                                            (tqueue/coerce-to-stream
+                                             {:certname name
+                                              :producer_timestamp (Date.)})
```
senior: Swap the `(Date.)` for `(now)`

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1471,3 +1474,63 @@
...
+            (deactivate "y.local")
+            (deactivate "z.local")
+            (doseq [i (range (/ default-timeout-ms 20))
+                    :while (not= 3 (- (:received-commands (stats))
+                                      (:received-commands start-stats)))]
+              (Thread/sleep 20))
+            (is (= (map #(+ (or % 0) 3) start-depths)
+                   (depths)))
+            ;; Delete their files
+            (is (= 3 (count (doall
```
senior: you don't need the `doall`

PR #2190, opened 2017-03-15
(PDB-3353) Prototype perf regression tests

mullr:
This is decidedly not yet baked (I think the junit report part is actually broken), but it's useful for discussion. I'd like to hear thoughts about the approach.

One change I know I want to make is separating the allowed range from the test itself; that'll allow us to specify different ranges for different test runners. 

This whole thing, by the way, is a blatent ripoff of https://www.chromium.org/developers/testing/chromium-build-infrastructure/performance-test-plots

### DIFF
--- a/test/puppetlabs/puppetdb/performance/cli.clj
+++ b/test/puppetlabs/puppetdb/performance/cli.clj
@@ -1,0 +1,148 @@
+(ns puppetlabs.puppetdb.performance.cli
+  (:require [me.raynes.fs :as fs]
+            [clojure.java.shell :refer [sh]]
+            [clojure.string :as s]
+            [clojure.edn :as edn]
+            [clojure.set :as set]
+            [clojure.data.xml :as xml]
+            [clojure.tools.namespace.find :as nsf]
+            [clojure.java.classpath :as cp]
+            [clojure.string :as s]
+            [puppetlabs.puppetdb.utils :refer [println-err]]))
+
+;; thanks, StackOverflow
+(defn- find-by-var-meta [metadata-flag]
+  (->> (all-ns)
+       (mapcat ns-publics)
+       (keep
+        (fn [[_ v]]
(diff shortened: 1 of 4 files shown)


mullr:
@camlow325 you may find this interesting

test/puppetlabs/puppetdb/performance/cli.clj
```diff
@@ -0,0 +1,134 @@
...
+        variance (first (:variance new))]
+    (assoc test :errors
+           (remove nil?
+                   [(when (>= dt regress) (str "Performance regression of " dt
+                                               " is past threshold of " regress
+                                               " (" old-time "s -> " new-time "s)"))
+                    (when (<= (- improve) dt) (str "Suspicious performance improvement: "
+                                                   dt " is past threshold of " regress
+                                                   " (" old-time "s -> " new-time "s)"))
+                    (when (> variance max-dev) (str "Variance of " variance " is outside specified bound ("
```
wkalt: needs revisiting -- std and variance aren't comparable like this, but we don't know if variance is really variance here.

mullr: This is why I thought the `variance` field was actually standard deviation: https://github.com/hugoduncan/criterium/blob/develop/src/criterium/core.clj#L976

I haven't actually looked at how it's computed though.

mullr: OH are they taking the square root of that? I missed the line above it, where it calls 'report-point-estimate-sqrt', that would make more sense

wkalt: it looks like variance is variance and it's taking the root to report std deviation. I'm not completely sure what we're trying to do here -- where's max-dev coming from?

mullr:
Build didn't fail, I cancelled it because it's irrelevant for this

PR #2191, opened 2017-03-15
(PDB-3328) Munge package inventory facts in terminus

mullr:
This adds terminus support for that package inventory feature. When the
_puppet_inventory_1 fact is seen it is removed, reformatted to be in the tabular
format expected by puppetdb, and placed in a top-level 'package_inventory' key
in the facts command payload.

### DIFF
--- a/puppet/lib/puppet/indirector/facts/puppetdb.rb
+++ b/puppet/lib/puppet/indirector/facts/puppetdb.rb
@@ -26,7 +26,12 @@ class Puppet::Node::Facts::Puppetdb < Puppet::Indirector::REST
         facts = request.instance.dup
         facts.values = facts.values.dup
         facts.values[:trusted] = get_trusted_info(request.node)
-        {
+
+        inventory = facts.values['_puppet_inventory_1']
+        package_inventory = inventory['packages'] if inventory.respond_to?(:keys)
+        facts.values.delete('_puppet_inventory_1')
+
+        payload_value = {
           "certname" => facts.name,
           "values" => facts.values,
           # PDB-453: we call to_s to avoid a 'stack level too deep' error
@@ -36,6 +41,12 @@ class Puppet::Node::Facts::Puppetdb < Puppet::Indirector::REST
           "producer_timestamp" => Puppet::Util::Puppetdb.to_wire_time(current_time),
           "producer" => Puppet[:node_name_value]
         }
+
+        if inventory
+          payload_value['package_inventory'] = package_inventory
+        end
+
+        payload_value
       end
 
       submit_command(request.key, payload, CommandReplaceFacts, 5, current_time.clone.utc)
--- a/test/puppetlabs/puppetdb/integration/inventory/package_inventory.clj
+++ b/test/puppetlabs/puppetdb/integration/inventory/package_inventory.clj
@@ -1,0 +1,40 @@
+(ns puppetlabs.puppetdb.integration.inventory.package-inventory
+  (:require [clojure.test :refer :all]
+            [puppetlabs.puppetdb.integration.fixtures :as int]
+            [puppetlabs.puppetdb.cheshire :as json]
+            [me.raynes.fs :as fs]))
+
+(deftest ^:integration package-inventory
+  (with-open [pg (int/setup-postgres)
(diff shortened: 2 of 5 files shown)


mullr:

puppet/lib/puppet/indirector/facts/puppetdb.rb
```diff
@@ -26,7 +26,18 @@ def save(request)
         facts = request.instance.dup
         facts.values = facts.values.dup
         facts.values[:trusted] = get_trusted_info(request.node)
-        {
+
+        inventory = facts.values['_puppet_inventory_1']
+        package_inventory =
+          if inventory and inventory.respond_to?(:keys) and inventory['packages']
+            inventory['packages'].map { |p|
+              [p['package_name'], p['version'], p['provider']]
```
wkalt: not too big a deal, but seeing as this fact is controlled by us, why not just receive it as a list of tuples and skip this step?

wkalt: if we could do that we could also split out the array-valued versions in the fact definition and not have to worry about any of that in the terminus or application

mullr: We talked about this; the reason we have maps in the fact but tuples in the command is to keep the more 'public' part of the data (i.e. the data generated on the agent, and sent from  the agent to the master) self describing. We should still be able to split out the array values in the fact definition - I asked Scott Walker to do just that in a comment on the design doc.

wkalt:
@mullr needs a rebase

test/puppetlabs/puppetdb/integration/inventory/package_inventory.clj
```diff
@@ -0,0 +1,41 @@
...
+      (is (= 0 (count (int/pql-query pdb "facts { name = '_puppet_inventory_1' }")))))
+
+    (comment
+      (testing "Package inventory can be queried"
+        (is (= [{:package_name "openssl"
+                 :version "1.0.2g-1ubuntu4.6"
+                 :provider "apt"}]
+               (int/pql-query pdb "packages { certname = 'agent-with-packages' }")))))))
+
+
```
wkalt: extra linez

puppet/lib/puppet/indirector/facts/puppetdb.rb
```diff
@@ -26,7 +26,15 @@ def save(request)
         facts = request.instance.dup
         facts.values = facts.values.dup
         facts.values[:trusted] = get_trusted_info(request.node)
-        {
+
+        inventory = facts.values['_puppet_inventory_1']
+        package_inventory = nil
+        if inventory && inventory.respond_to?(:keys)
+          package_inventory = inventory['packages']
+          facts.values.delete('_puppet_inventory_1')
```
wkalt: we should always be deleting this if it exists -- respond_to? keys is irrelevant

test/puppetlabs/puppetdb/integration/inventory/package_inventory.clj
```diff
@@ -0,0 +1,36 @@
...
+                       :env {"FACTER__puppet_inventory_1"
+                             (json/generate-string inventory-fact-value)}}))
+
+    (testing "Ensure that facts were stored"
+      (is (pos? (count (int/pql-query pdb "facts { }")))))
+
+    (testing "Ensure the inventory fact was removed"
+      (is (= 0 (count (int/pql-query pdb "facts { name = '_puppet_inventory_1' }")))))
+
+    (comment
```
mullr: ;; no!

puppet/lib/puppet/indirector/facts/puppetdb.rb
```diff
@@ -26,7 +26,15 @@ def save(request)
         facts = request.instance.dup
         facts.values = facts.values.dup
         facts.values[:trusted] = get_trusted_info(request.node)
-        {
+
+        inventory = facts.values['_puppet_inventory_1']
+        package_inventory = nil
+        if inventory
```
wkalt: if _puppet_inventory_1 is null, we still want to delete it I think. Does this guard against that?

PR #2192, opened 2017-03-16
(maint) Fix doc issue with `JSON` example

petems:
* Without the extra newline the JSON isn't 
highlighted


petems:
Example:
screenshot 2017-03-16 22 29 56

PR #2196, opened 2017-03-20
(MAINT) Consolodated report integration tests

senior:


test/puppetlabs/puppetdb/integration/reports.clj
```diff
@@ -0,0 +1,166 @@
+(ns puppetlabs.puppetdb.integration.reports
```
mullr: commit title: consolodated -> consolidated

PR #2197, opened 2017-03-20
(PDB-3354) Hash strings directly

senior:
Before this commit, string values were converted ot JSON and the
results of that were hashed. We can just hash the string directly and
avoid the JSON roundtrip.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/hash.clj
+++ b/src/puppetlabs/puppetdb/scf/hash.clj
@@ -10,8 +10,11 @@
   instead."
   [data]
   {:post [(string? %)]}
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/scf/hash.clj
```diff
@@ -8,8 +8,11 @@
   instead."
   [data]
   {:post [(string? %)]}
-  (-> (kitchensink/sort-nested-maps data)
-      (json/generate-string)))
+  (if (string? data)
```
mullr: can you add this to the release notes?

PR #2198, opened 2017-03-20
(PDB-3356) Use a CTE for inactive nodes

wkalt:
Switch from using a select_nodes subquery for restrictions to active
nodes, to a dedicated CTE containing only inactive nodes. This approach
is faster and less sensitive to memory settings than the previous
approach, particularly on the prospective query for package inventory.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -909,6 +909,15 @@
                :entity :events
                :source-table "resource_events"}))
 
+(def inactive-nodes-query
+  (map->Query {:projections {"certname" {:type :string
+                                         :queryable? true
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1077,8 +1095,10 @@
                                    (mapv (fn [[name {:keys [field]}]]
                                            [field name])))
                               fs)))
-        new-selection (cond-> (assoc selection :select select)
-                        group-by (assoc :group-by group-by))]
+        new-selection (-> selection
+                          (assoc :select select)
```
wkalt: this is just existing logic. no chance of it having a with though.

mullr:
I'm +1 here. @senior?

PR #2202, opened 2017-03-21
(QENG-2198) Move PuppetDB acceptance over to vmpooler

smcelmurry:
* Add sqa-utils gem.
* Rewrite `acceptance/setup/pre_suite/80_add_dev_repo.rb` to use
  `install_puppetlabs_dev_repo` instead of the special stuff it was doing.
* Add logic to `ext/jenkins/beaker-tests.sh` that will allow PR and merge tests
  to run without making changes to ci-job-configs for vmpooler (those changes
  are coming though!)
* In `ext/jenkins/packaging.sh` provide `PUPPETDB_PACKAGE_BUILD_VERSION` instead
  of `PACKAGE_BUILD_VERSION` to avoid having to convert to that variable in
  `ext/jenkins/beaker-tests.sh`.
* Remove S3 packaging/cleanup stuff.
* Set default beaker type to `aio`. This isn't really necessary, but the test
  suite doesn't work with `foss`, at least not on stable branch.
* BEAKER_CONFIG should take a filename. Same applies to BEAKER_OPTIONS. This
  change matches up with the way we treat variables of the same name in CI.

Thanks to @waynr for this work. This will need to go in combination with puppetlabs/ci-job-configs#2319

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,4 +56,5 @@ group :acceptance do
     # use the pinned version
     gem 'beaker', '~> 3.4'
   end
+  gem 'beaker-hostgenerator', '0.8.2'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -57,5 +57,5 @@ group :acceptance do
     gem 'beaker', '~> 3.4'
   end
   gem 'beaker-hostgenerator', '0.8.2'
-  gem 'beaker-abs', *location_for(ENV['BEAKER_ABS_VERSION'] || '~> 0.1')
+  gem 'beaker-abs', *location_for(ENV['BEAKER_ABS_VERSION'] || '~> 0.2')
 end
--- a/Gemfile
+++ b/Gemfile
@@ -57,4 +57,5 @@ group :acceptance do
     gem 'beaker', '~> 3.4'
   end
   gem 'beaker-hostgenerator', '0.8.2'
+  gem 'beaker-abs', *location_for(ENV['BEAKER_ABS_VERSION'] || '~> 0.1')
 end
--- a/Rakefile
+++ b/Rakefile
@@ -3,26 +3,29 @@ require 'rake'
 RAKE_ROOT = File.dirname(__FILE__)
 
 def run_beaker(test_files)
-  config = ENV["BEAKER_CONFIG"] || "ec2-west-dev"
+  config = ENV["BEAKER_CONFIG"] || "acceptance/config/ec2-west-dev.cfg"
   options = ENV["BEAKER_OPTIONS"] || "acceptance/options/postgres.rb"
   preserve_hosts = ENV["BEAKER_PRESERVE_HOSTS"] || "never"
   no_provision = ENV["BEAKER_NO_PROVISION"] == "true" ? true : false
   color = ENV["BEAKER_COLOR"] == "false" ? false : true
   xml = ENV["BEAKER_XML"] == "true" ? true : false
-  type = ENV["BEAKER_TYPE"] || "git"
+  type = ENV["BEAKER_TYPE"] || "aio"
   keyfile = ENV["BEAKER_KEYFILE"] || nil
+  collect_perf_data = ENV["BEAKER_COLLECT_PERF_DATA"] || nil
 
   beaker = "bundle exec beaker " +
-     "-c '#{RAKE_ROOT}/acceptance/config/#{config}.cfg' " +
+     "-c '#{config}' " +
      "--type #{type} " +
      "--debug " +
      "--tests " + test_files + " " +
-     "--options-file " + options + " " +
+     "--options-file '#{options}' " +
      "--root-keys " +
      "--preserve-hosts #{preserve_hosts}"
 
+  beaker += " --keyfile #{keyfile}" if keyfile
   beaker += " --no-color" unless color
   beaker += " --xml" if xml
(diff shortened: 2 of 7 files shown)


smcelmurry:
Here is the experimental pipeline on CI.Next: https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/view/puppetdb/view/master/

ext/jenkins/packaging-cleanup.sh
```diff
@@ -1,33 +1,3 @@
...
-
-set -e
-set -x
-
-echo "BUCKET_NAME IS: ${BUCKET_NAME}"
-
-time s3cmd del --recursive  ${S3_BRANCH_PATH}/
-
-PUBLISH_PACKAGES
+echo "Nothing to do, no S3 here."
```
mullr: can we just delete the file?

ext/jenkins/beaker-tests.sh
```diff
@@ -19,14 +21,27 @@ export PUPPETDB_USE_PROXIES=false
...
-export PUPPETDB_PACKAGE_REPO_HOST="puppetdb-prerelease.s3.amazonaws.com"
-export PUPPETDB_PACKAGE_REPO_URL="http://puppetdb-prerelease.s3.amazonaws.com/puppetdb/${PACKAGE_BUILD_VERSION}"
-export PUPPETDB_PACKAGE_BUILD_VERSION=$PACKAGE_BUILD_VERSION
-export BEAKER_CONFIG=$LAYOUT
-export INSTALL_MODE=$INSTALL_TYPE
+# Transform LAYOUT from old static host config style to genconfig2 style
+# Once the necessary PRs (puppetlabs/puppetdb#1636 and
+# puppetlabs/ci-job-configs#529) for both this repo and ci-job-configs have been
+# merged, this logic can be removed.
+[ "$LAYOUT" = "ec2-west-el7-64mda-el7-64a" ] \
```
mullr: What's the meaning of 'layout' now that we're going to be on vmpooler? Clearly the ec2-* business doesn't have meaning anymore.

acceptance/setup/pre_suite/80_add_dev_repo.rb
```diff
@@ -1,40 +1,8 @@
...
-        yum_repo_file_path = "/etc/yum.repos.d/puppetlabs-prerelease.repo"
-        on database, "curl \"#{yum_repo_url}\" | #{sed_cmd} > #{yum_repo_file_path}"
-
-        result = on database, "cat #{yum_repo_file_path}"
-        Log.notify("Yum REPO DEFINITION:\n\n#{result.stdout}\n\n")
-      else
-        raise ArgumentError, "Unsupported OS '#{os}'"
-      end
-    end
+    install_puppetlabs_dev_repo database, 'puppetdb', test_config[:package_build_version],
```
mullr: What does the 'puppetdb' parameter here mean? Does it automatically install puppetdb? iirc we use the puppetdb module to do the installation, which we'd like to keep doing.

Rakefile
```diff
@@ -3,26 +3,29 @@ require 'rake'
 RAKE_ROOT = File.dirname(__FILE__)
 
 def run_beaker(test_files)
-  config = ENV["BEAKER_CONFIG"] || "ec2-west-dev"
+  config = ENV["BEAKER_CONFIG"] || "acceptance/config/ec2-west-dev.cfg"
```
mullr: maybe vmpooler.cfg now?

ext/jenkins/beaker-tests.sh
smcelmurry: `LAYOUT` is still used by the beaker-hostgenerator to make the host file. This was initially to use in-conjunction with current ci-job-configs settings, but yeah we can rip this part out.

acceptance/setup/pre_suite/80_add_dev_repo.rb
smcelmurry: This isn't installing puppetdb but rather gathering the `.repo` file as part of a function set up in beaker. This is instead of doing the custom functions removed above. The `puppetdb` parameter is telling the function what component to get. This is the same way it is done in [puppet server](https://github.com/puppetlabs/puppetserver/blob/master/acceptance/suites/pre_suite/foss/30_install_dev_repos.rb#L9-L10). [Beaker docs](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/InstallUtils/FOSSUtils#install_puppetlabs_dev_repo-instance_method)

ext/jenkins/beaker-tests.sh
smcelmurry: Oh wait, I'm an idiot. This section was so we didn't have to sync the ci-job-configs PR until it was in all the necessary branches :man_facepalming:

Gemfile
```diff
@@ -56,4 +56,6 @@ group :acceptance do
     # use the pinned version
     gem 'beaker', '~> 3.4'
   end
+  gem 'beaker-hostgenerator', '0.8.2'
+  gem 'beaker-abs', *location_for(ENV['BEAKER_ABS_VERSION'] || '~> 0.1')
```
sbeaulie: nate told me we should use `~> 0.2`

smcelmurry:
Closing this in favor of #2252

PR #2203, opened 2017-03-21
(maint) add package inventory benchmark data

wkalt:
This includes two package sets from a barebones PE master and one from my laptop.


mullr:
this works fine for me, may it needs to go in after the query pr

PR #2205, opened 2017-03-21
(maint) use a common set of certname-based implicit relations

wkalt:
Centralize definition of certname-based implicit relationships. This
will make us less likely to leave them out or create inconsistencies in
the future.


wkalt:

mullr:
do we need to update the docs to point out all of the new relationships?

wkalt:
yeah good call, I'll update those

PR #2211, opened 2017-03-23
(maint) allow paging options to be specified in two places

wkalt:
Previously paging options supplied in a query would cause paging options
supplied outside the query to be ignored.


wkalt:
this has an issue

wkalt:
I was wrong about this having an issue -- the problem was copy pasting some weird characters. I think it's good to go.

PR #2214, opened 2017-03-28
(PDB-3359) fix regex queries on nested resource params

wkalt:
Corrects an issue where resource parameters specified via dot notation
could not be queried via regex.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -92,6 +92,11 @@
    column-data :- column-schema
    value])
 
+(s/defrecord JsonbRegexExpression
+  [field :- s/Str
+   column-data :- column-schema
+   value])
+
 (s/defrecord InExpression
     [column :- [column-schema]
      ;; May not want this if it's recursive and not just "instance?"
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1220,6 +1229,15 @@
      :state (conj state (su/munge-jsonb-for-storage
                           (path->nested-map path value)))}))
 
+(defn parse-json-regex-query
```
mullr: I'm really confused about what this is doing. Can you explain?

wkalt: This is the param extractor for a JsonbRegexExpression. At a high level these things zip through the query plan, pulling parameters out of the plan into the state vector, and replacing them with placeholders (e.g '?').

In this case, the input looks like

```
[{:field "parameters.message.foo" :value "bar"} ["a" "b" "c"]]
```

and the output looks like

```
[{:field "parameters" :value ["?" "?" "?"]} ["a" "b" "c" "message" "foo" "bar" "message"]]
```

that field and value are eventually passed to su/jsonb-regex, which turns them into proper SQL. Does that clarify things any? Would it help if I renamed it to jsonb-param-extractor?

wkalt:
reopened under the correct name https://github.com/puppetlabs/puppetdb/pull/2218

PR #2215, opened 2017-03-28
(maint) fix spec test failures against puppet master

wkalt:
A recent patch to Puppet broke several of our spec tests. This corrects
the ones that are due to a changed error message. There are some other
failures that are due to what seems like a Puppet bug to me. I'll wait
to get confirmation before changing those ones.


mullr:
@wkalt This is failing spec tests in travis

PR #2216, opened 2017-03-28
(PDB-3371) Deduplication of package data

mullr:


documentation/api/query/v4/packages.markdown
```diff
@@ -1,5 +1,5 @@
 ---
-title: "PuppetDB 4.4: Packages endpoint"
+title: "PuppetDB 4.4: Package endpoints"
 layout: default
 canonical: "/puppetdb/latest/api/query/v4/packages.html"
```
wkalt: I'm not necessarily against squishing these into one doc, but I think it'll be the first time we do this. Things like line 4 here make me wonder if the presentation will make sense on the site.

documentation/api/query/v4/packages.markdown
```diff
@@ -20,15 +20,48 @@ canonical: "/puppetdb/latest/api/query/v4/packages.html"
 [resources]: ./resources.html
 [inventory]: ./inventory.html
 
-> **PE feature**: Package metadata collection, storage, and query is primarily
-> a Puppet Enterprise feature. This data is queryable from PuppetDB if it is
-> submitted via the `package_inventory` key in the payload of the `store facts`
-> command, as described in the [facts format documentation.][facts-format]
+> **PE feature**: Package metadata collection, storage, and query is
```
wkalt: I think "querying" reads better here

documentation/api/query/v4/packages.markdown
mullr: What about line 4 gives you concern?

documentation/api/query/v4/packages.markdown
```diff
@@ -71,6 +104,6 @@ exactly like a call to `/pdb/query/v4/packages` with a query string of `["=",
 
 ## Paging
 
-This query endpoint supports paged results via the common PuppetDB paging
+These query endpoints support paged results via the common PuppetDB paging
 URL parameters. For more information, please see the documentation
 on [paging][paging].
```
wkalt: we will also need to update the nav

documentation/api/query/v4/packages.markdown
```diff
@@ -56,13 +89,13 @@ The array is unsorted by default.
...
 ### Example
 
-[You can use `curl`][curl] to query information about nodes:
+[You can use `curl`][curl] or `puppet query` to query information about nodes:
 
-    curl 'http://localhost:8080/pdb/query/v4' -d  'query=packages{ certname = "agent1" }'
+    puppet query "package_inventory{ certname = 'agent1' }"
 
-    curl http://localhost:8080/pdb/query/v4/packages -d 'query=["=", "certname", "agent1"]'
+    curl -G http://localhost:8080/pdb/query/v4/package-inventory -d 'query=["=", "certname", "agent1"]'
```
wkalt: I think another illustrative example here would be to list all certnames on which a given package name/version are installed (i.e for tracking down vulnerable nodes or whatever)

```
package_inventory[certname]{package_name = "openssl" and version = "1.1.0e" }
```

src/puppetlabs/puppetdb/facts.clj
```diff
@@ -30,6 +30,15 @@
...
    (s/one s/Str "provider")])
 
+(def hashed-package-tuple
+  (conj package-tuple
+        (s/one s/Str "package_hash")))
+
+(defn package-tuple-name [package-tuple] (nth package-tuple 0))
+(defn package-tuple-version [package-tuple] (nth package-tuple 1))
+(defn package-tuple-provider [package-tuple] (nth package-tuple 2))
+(defn package-tuple-hash [hashed-package-tuple] (nth hashed-package-tuple 3))
```
wkalt: are these functions used?

mullr: the last one is used. I put the first ones there in a fit of sicp-fueled encapsulation.

src/puppetlabs/puppetdb/http/handlers.clj
```diff
@@ -252,12 +252,19 @@
...
+    (cmdi/ANY "" []
+              (create-query-handler version "packages")))))
+
+(pls/defn-validated package-inventory-routes :- bidi-schema/RoutePair
+  [version :- s/Keyword]
+  (extract-query
     (cmdi/routes
       (cmdi/ANY "" []
-                (create-query-handler version "packages"
+                (create-query-handler version "package_inventory"
```
wkalt: what's the significance of the underscore here, relative to the dash in server.clj?

mullr: this is the name of the entity. server.clj has the route names, which appear to use dashes. :man_shrugging:

wkalt: ah, okay

src/puppetlabs/puppetdb/scf/hash.clj
```diff
@@ -95,11 +96,19 @@
...
   that structure"
-  [packages :- [package-tuple]]
-  (-> packages
-      sort
-      json/generate-string
-      kitchensink/utf8-string->sha1))
+  [packages :- [hashed-package-tuple]]
+  (->> packages
+       (map package-tuple-hash)
+       sort
```
wkalt: can we leverage generic-identity-hash for the sort/json/sha?

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1248,6 +1248,41 @@
...
+   ["ALTER TABLE ONLY packages "
+    "ADD CONSTRAINT package_hash_key UNIQUE (hash)"]
+
+   ["insert into packages (name, provider, version) "
+    "select distinct name, provider, version "
+    "from package_inventory"]
+
+   ["create table certname_packages"
+    "  (certname_id bigint not null,"
+    "   package_id integer not null,"
```
wkalt: we probably want an fk constraint on packages(id) here

wkalt: possibly certname_id as well depending on how we're handling that elsewhere

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1283,7 +1318,8 @@
    55 index-certnames-unique-latest-report-id
    56 merge-fact-values-into-facts
    57 add-package-inventory
-   58 add-better-package-inventory-indexing})
+   58 add-better-package-inventory-indexing
+   59 switch-to-deduplicated-packages})
```
mullr: We should squish these two migrations before merge

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: Thought about this a bit yesterday, and I think these constraints should be harmless. They'd only be checked on updates to the join table, and they'd check certnames and packages, both of which are relatively small.

documentation/api/query/v4/packages.markdown
mullr: updated

documentation/api/query/v4/packages.markdown
mullr: We talked about this offline; I think the duplication is silly enough here that it warrants squishing things together. Looked at the site, and I think the presentation will be ok.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -991,45 +999,86 @@
...
    inventory :- [facts/package-tuple]]
-  (let [{certname-id :id package_hash :package_hash} (find-certname-id-and-hash certname)
-        new-package-hash (shash/package-similarity-hash inventory)]
-
-    (when-not (= new-package-hash package_hash)
-      (let [old-packages (create-package-map certname-id)
-            [to-add to-remove _] (clojure.data/diff (set inventory) (set (keys old-packages)))
-            to-remove-ids (map old-packages to-remove)]
+  (let [{certname-id :id package-hash :package_hash} (find-certname-id-and-hash certname)
+        hashed-package-tuples (map shash/package-identity-hash inventory)
```
wkalt: I'm not seeing a guarantee that inventory is coming into this function in sorted order -- have we made that explicit somewhere?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -991,45 +999,86 @@
...
+        new-package-hash (shash/package-similarity-hash hashed-package-tuples)]
+
+    (when-not (= new-package-hash package-hash)
+      (let [just-hashes (map facts/package-tuple-hash hashed-package-tuples)
+            existing-package-hashes (find-package-hashes just-hashes)
+            full-hashes-map (insert-missing-packages existing-package-hashes hashed-package-tuples)
+            new-package-id-set (set (vals full-hashes-map)) ;(set (map full-hashes-map just-hashes))
```
wkalt: we can probably remove this comment I guess

src/puppetlabs/puppetdb/scf/migrate.clj
wkalt: I think we will want a bigint for package_id here. right now it won't actually support the same range of ids as the packages table

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -991,45 +999,86 @@
...
+        (when (seq new-package-ids)
+          (jdbc/insert-multi! :certname_packages
+                              ["certname_id" "package_id"]
+                              (map #(vector certname-id %) new-package-ids)))
+
+        (when (seq old-package-ids)
+          (jdbc/delete! :certname_packages
+                        ["certname_id = ? and package_id = ANY(?)"
+                         certname-id
+                         (sutils/array-to-param "bigint" Long (map long old-package-ids))]))))))
```
wkalt: I'm not sure why we need to map long over these ids

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -991,45 +999,86 @@
...
 
 (defn insert-packages [certname inventory]
```
wkalt: same question as above around sortedness of inventory

src/puppetlabs/puppetdb/scf/storage.clj
mullr: woops, my fault. hack shrapnel.

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: looks like the similarity hash is actually handling this

test/puppetlabs/puppetdb/http/packages_test.clj
```diff
@@ -95,13 +116,29 @@
...
         (is-query-result "/v4" query result)))
 
     (testing "querying for a specific node"
-      (is-query-result "/v4/packages/node2" nil
+      (is-query-result "/v4/package-inventory/node2" nil
                        #{(package-map "node2" "bar" "2.3.4" "apt")
-                         (package-map "node2" "foo" "1.2.3" "apt")}))))
+                         (package-map "node2" "foo" "1.2.3" "apt")}))
+
+    (testing "top level packages endoint"
```
wkalt: endpoint

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -686,7 +687,28 @@
...
+        (is (= [["foo-1" "1.2.3" "apt"]]
+               (package-seq certname)))
+
+        (delete-unassociated-packages!)
+
+        (is (= 1
+               (-> ["SELECT count(*) as c FROM packages"]
+                   query-to-vec
+                   first
+                   :c)))))))
```
wkalt: I don't really understand what this is testing. It seems like nothing is being removed?

wkalt: ah, sorry. I think I just missed what's above.

wkalt: can we put another count assertion above the call to the GC function?

src/puppetlabs/puppetdb/facts.clj
wkalt: I think we should remove the unused ones

wkalt:
@mullr this looks pretty good overall. seems to work fine too.

src/puppetlabs/puppetdb/scf/hash.clj
mullr: That uses `ks/sort-nested-maps` for sorting, which doesn't sort sequential collections. We need to sort the sequence here.

src/puppetlabs/puppetdb/scf/storage.clj
mullr: @wkalt It doesn't need to be sorted, afact

wkalt: yeah, this was erroneous. my concern is addressed with the similarity hash.

src/puppetlabs/puppetdb/scf/storage.clj
mullr: We do the same thing in our array-in queries, but it may not be needed here.

mullr: Actually, I think this is a fairly good idea. It makes sure that we're definitely dealing with primitive longs. We may be implicitly boxing them in array-to-param, unsure though.

mullr: it could be more correct to use Long.TYPE

mullr: Looked at the source of the jdbc driver, and it looks like all their tests are with boxed number types. So we should probably keep doing that.

mullr:
@wkalt good to merge?

PR #2218, opened 2017-03-29
(PDB-3285) fix regex queries on nested resource params

wkalt:
Corrects an issue where resource parameters specified via dot notation
could not be queried via regex.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -92,6 +92,11 @@
    column-data :- column-schema
    value])
 
+(s/defrecord JsonbRegexExpression
+  [field :- s/Str
+   column-data :- column-schema
+   value])
+
 (s/defrecord InExpression
     [column :- [column-schema]
      ;; May not want this if it's recursive and not just "instance?"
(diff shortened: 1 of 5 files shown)


test/puppetlabs/puppetdb/http/resources_test.clj
```diff
@@ -75,9 +75,18 @@ to the result of the form supplied to this method."
     (testing "dot-style querying for parameters"
       (doseq [[query result] [[["=" "parameters.ensure" "file"] #{foo1 bar1}]
                               [["=" "parameters.owner" "root"] #{foo1 bar1}]
+                              [[ "=" "parameters.nested.foo" "bar"] #{foo1 bar1}]
```
senior: extra space here

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1789,7 +1810,8 @@
               (map->NullExpression {:column cinfo :null? value}))
 
             [["~" column-name value]]
-            (let [cinfo (get-in query-rec [:projections column-name])]
+            (let [colname (first (str/split column-name #"\."))
```
senior: What will we do for parameters.foo.bar ~ "baz"? Does it look for a map as foo's value and the bar key inside there?

wkalt: the first value is the actual column. the rest are split on the dot and turned into a -> separated list of placeholders to use in a jsonb query. This might make it more apparent:
https://github.com/puppetlabs/puppetdb/pull/2218/files#diff-c3090734eca858a02839292b86db7464R221

PR #2219, opened 2017-03-30
(PDB-3312) Back off population metric db connection on failure

mullr:
When doing db queries for the population metrics, don't wait for the connection
to time out every time. Instead, fail fast if the connection failed less than 5
seconds ago. This improves the performance of the dashboard data endpoint when
the database is down, as it reduces the number of times it has to wait for
timeouts to occur.


wkalt:
this looks good to me. tested as well.

PR #2221, opened 2017-03-30
(maint) Remove 'install-terminus-into' from integration tests

mullr:
We can work around this by:
- Pointing the MRI agent at our terminus using --vardir
- Copying the puppetdb_query function into the master codedir when starting
  puppetserver

### DIFF
--- a/test/puppetlabs/puppetdb/integration/fixtures.clj
+++ b/test/puppetlabs/puppetdb/integration/fixtures.clj
@@ -195,20 +195,6 @@
        (map :file)
        last))
 
-(defn install-terminus-into [puppet-dir]
-  ;; various pieces of the terminus have to actually be inside of puppet itself to work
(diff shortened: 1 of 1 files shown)


rbrw:
I haven't had a chance to consider the code, but I checked that the integration tests have no effect on ~/.gem if you use `bundle install --path ~/.gem` instead of `bundle install --path "$(pwd)/vendor"`, and I also checked that all the integration tests still pass here.  I tested for changes via `rsync -niaHAX before/ after/`.

rbrw:
This seems fine to me, to the extent that I understand how the puppet integration works.

rlinehan:
Full disclosure - I paired with @mullr on this, but this all looks good to me. I still don't really understand how function loading works, but I'm willing to just accept that our solution of loading from the codedir lets these tests pass.

PR #2225, opened 2017-04-03
(PDB-3396) Pinpoint GC for packages

mullr:
The packages table needs to be kept up-to-date with deletions from the
certnames_packages table, as it's used for coarse-grained counting in places
where accuracy matters.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -1058,10 +1058,18 @@
                               (map #(vector certname-id %) new-package-ids)))
 
         (when (seq old-package-ids)
-          (jdbc/delete! :certname_packages
-                        ["certname_id = ? and package_id = ANY(?)"
-                         certname-id
-                         (sutils/array-to-param "bigint" Long (map long old-package-ids))]))))))
+          (let [old-package-id-sql-param (sutils/array-to-param "bigint" Long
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1058,10 +1058,18 @@
                               (map #(vector certname-id %) new-package-ids)))
 
         (when (seq old-package-ids)
-          (jdbc/delete! :certname_packages
-                        ["certname_id = ? and package_id = ANY(?)"
-                         certname-id
-                         (sutils/array-to-param "bigint" Long (map long old-package-ids))]))))))
+          (let [old-package-id-sql-param (sutils/array-to-param "bigint" Long
+                                                                (map long old-package-ids))]
```
senior: what are they that they need to be coerced to longs?

mullr: You tell me, it's your code :stuck_out_tongue: 

I actually looked into this at wyatt's request for the first round, but I don't think I reached any conclusion. Ended up just leaving it as-is since it worked.

senior: Oh, I know where it's from then. You can get integers from SQL results, those need to flip to longs since we're declaring them to be longs.

👍

PR #2228, opened 2017-04-04
(SERVER-1745) Add integration test for Puppet Server metrics

rlinehan:
Add integration test that the http client metrics specified in the PuppetDB
termini show up in the Puppet Server master status endpoint.

This commit also bumps the Puppet Server version to a `voom` release based off
current master of Puppet Server, which will be what eventually goes into
Puppet Server 5.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,6 +1,6 @@
 (def pdb-version "5.0.0-SNAPSHOT")
-(def puppetserver-version "2.7.2")
-(def clj-parent-version "0.6.0")
+(def puppetserver-version "5.0.0-master-20170329_182047-ga16e296")
+(def clj-parent-version "0.6.1")
 
 (defn deploy-info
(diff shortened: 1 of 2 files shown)


rlinehan:
Ah okay, so my test failed when run against stable branch of Puppet Server, which makes sense since the code that I'm testing is not present there. What's the best way to fix this issue? I could gate the tests on Puppet Server version - that seems probably like the easiest thing to maintain. I could also use a test selector, but anything with test selectors seems likely to get pretty gross pretty fast.

rlinehan:
Updated the commit so that it only tests the metrics present if the `master` status exists. This should (hopefully) pass now.

rlinehan:
jenkins, retest this please

rlinehan:
Huzzah, tests all passed! Ping @camlow325 @jpinsonault @adrienthebo for review.

PR #2231, opened 2017-04-06
(PDB-3395) bypass costly join on paged packages aggregate queries

wkalt:
This patch accomplishes a few things
* allows usage of limit/offset in PQL in combination with a group by
* detects paged aggregate queries against package_inventory like those
used in the console and reformulates them to be fast, i.e transforms

```
package_inventory[package_name, version, provider, count()]
  {version ~ "foo" group by package_name, version, provider limit 10
  offset 10}
```

into

```
package_inventory[package_name, version, provider, count()]
  {[package_name, version, provider] in packages[package_name, version,
  provider]{version ~ "foo" limit 10 offset 10} group by package_name,
  version, provider}
```

* relatedly, changes the query engine to process top-level AST from
clauses, rather than stripping them out in query-eng.clj and only
passing the remaining query. This eliminates the distinction between
queries and subqueries in the engine.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng.clj
+++ b/src/puppetlabs/puppetdb/query_eng.clj
@@ -82,6 +82,49 @@
                     (:queryable? projection-value))]
      (keyword projection-key))))
 
+(defn legacy-engine-query-context
+  "Parse a query and possibly some paging clauses from a query to the legacy
+   engine. Required because top-level paging options supplied in the query are
+   not otherwise handled by the engine, so they must be merged with
+   param-supplied paging options and dealt with at the top level."
+  [version query warn]
+  (cm/match
+    query
+    ["from" (entity-str :guard #(string? %)) & remaining-query]
+    (let [remaining-query (cm/match
+                            remaining-query
+
+                            [(c :guard eng/paging-clause?) & clauses]
+                            (let [paging-clauses (eng/create-paging-map (cons c clauses))]
+                              {:paging-clauses paging-clauses :query []})
+
+                            [(q :guard vector?)]
+                            {:query q}
+
+                            [(q :guard vector?) & clauses]
+                            (let [paging-clauses (eng/create-paging-map clauses)]
+                              {:query q :paging-clauses paging-clauses})
+
+                            []
+                            {:query []}
+                            :else (throw
+                                    (IllegalArgumentException.
+                                      (str
+                                        (tru "Your `from` query accepts an optional query only as a second argument.")
+                                        " "
+                                        (tru "Check your query and try again.")))))
+          entity (keyword (utils/underscores->dashes entity-str))]
+      (when warn
+        (eng/warn-experimental entity))
+      {:remaining-query (:query remaining-query)
+       :paging-clauses (:paging-clauses remaining-query)})
+
+    :else (throw (IllegalArgumentException.
+                   (str
+                     (trs "Your initial query must be of the form: [\"from\",<entity>,(<optional-query>)].")
+                     " "
+                     (trs "Check your query and try again."))))))
+
 (defn query->sql
   "Converts a vector-structured `query` to a corresponding SQL query which will
    return nodes matching the `query`."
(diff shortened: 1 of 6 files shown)


wkalt:

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1666,38 +1667,51 @@
...
+  :- {(s/optional-key :projected-fields) [projection-schema]
+      s/Any s/Any}
+  "Create an explicit subquery declaration to mimic the select_<entity>
+   syntax."
+  [entity expr clauses]
+  (let [query-rec (subquery->query-rec (str "select_" (utils/dashes->underscores entity)))
+        {:keys [limit offset order-by]} (create-paging-map clauses)]
+    (if (extract-expression? expr)
+      (let [[_ columns & clauses] expr
+            [query-expr groupby] (map first (split-with #(not= "group_by" (first %)) clauses))
```
senior: I think this should just be `last`. Our query semantics state group by must be the last thing in the extract.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1971,9 +1993,9 @@
                    :cut true})
 
                 :else
-                (when (instance? clojure.lang.IMeta node)
-                  {:node (vary-meta node assoc :query-context context)
-                   :state state})))))
+                    (when (instance? clojure.lang.IMeta node)
```
senior: accident?

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -2203,20 +2192,20 @@
...
-                                (paging/validate-order-by! allowed-fields)
                                 (paging/dealias-order-by query-rec))
         {:keys [plan params]} (->> user-query
                                    (push-down-context query-rec)
                                    expand-user-query
-                                   (convert-to-plan query-rec paging-options)
+                                   (convert-to-plan paging-options)
                                    extract-all-params)
-        sql (-> plan fix-plan-in-expr-multi-comparisons plan->sql)
+        plan (assoc plan :wrap? true)
```
senior: Did we always wrap this before as well?

senior:
I tested this out and it looks good. One question I had was why the optimization didn't apply to:

```
package_inventory[package_name, version, provider]
                              {version ~ "3" 
                               group by package_name, version, provider 
                               limit 10
                               offset 10}
```

Seems like it should

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -82,6 +82,49 @@
...
+                                        (tru "Check your query and try again.")))))
+          entity (keyword (utils/underscores->dashes entity-str))]
+      (when warn
+        (eng/warn-experimental entity))
+      {:remaining-query (:query remaining-query)
+       :paging-clauses (:paging-clauses remaining-query)})
+
+    :else (throw (IllegalArgumentException.
+                   (str
+                     (trs "Your initial query must be of the form: [\"from\",<entity>,(<optional-query>)].")
```
wkalt: not sure -- to indicate that it's optional you mean? I think the comma inside the parens will look weird but I think we should have spaces after the commas. I'm not sure if the parens are necessary at all (this was just a preexisting message).

src/puppetlabs/puppetdb/query_eng/engine.clj
wkalt: yes, but previously the decision to wrap was based on whether a query was a subquery. Since the subquery concept is gone now, this is just telling the engine to always wrap the top level query (which enters here) in the inactive nodes CTE. Behavior should be the same as before.

src/puppetlabs/puppetdb/query_eng/engine.clj
wkalt: that won't quite work, since this is extracting both query-expr and group-by and accounting for the cases where either might be missing. I can think about cleaner ways to write it though.

wkalt:
this should be good to go now @senior

wkalt:

PR #2232, opened 2017-04-06
(PDB-3421) do-commands: pass vector to db-do-commands, not list

rbrw:
db-do-commands doesn't appear to care right now, but but let's follow
the documented API anyway.

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -32,14 +32,15 @@
 
 (defn do-commands
   "Runs the given commands in a transaction on the database given
-  by (jdbc/db).  If a command is a vector, it will be converted to a
-  string via (apply str v)."
+  by (jdbc/db).  If a command is a seq, converts it to a string
+  via (clojure.string/join command)."
   [& commands]
   (sql/db-do-commands *db* true
-                      (for [c commands]
-                        (if (vector? c)
-                          (apply str c)
-                          c))))
+                      (mapv #(cond
+                               (string? %) %
+                               (seq %) (string/join %)
+                               :else %)
+                            commands)))
 
 (defn do-prepared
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -36,10 +36,10 @@
   string via (apply str v)."
   [& commands]
   (sql/db-do-commands *db* true
-                      (for [c commands]
-                        (if (vector? c)
-                          (apply str c)
-                          c))))
+                      (if (not-any? vector? commands)
```
senior: Why traverse the vector twice?

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -36,10 +36,10 @@
   string via (apply str v)."
```
mullr: Postel's law? I'd rather see this change to accept any other collection, rather than narrowed to just vectors. Unless there's some reason to only take vectors?

src/puppetlabs/puppetdb/jdbc.clj
rbrw: Originally was just half-thinking we might avoid the work if the input is fine, but even if that were plausible, it doesn't really support with my "let's always make it a vector" intent very well.

src/puppetlabs/puppetdb/jdbc.clj
rbrw: No strong opinion - was just being a bit conservative (and suppose there's also the clojure precedents like let).

rbrw:
Changed it to use str/join rather than str, so that `["sql" 4 :the #{win?}])` wouldn't be accepted, and @senior convinced me that there could be plausible uses for sets/maps, removing my concern about allowing all seqs, so they're handled now too.

PR #2234, opened 2017-04-06
(PDB-3427) Ignore order of resource_events in generative tests

mullr:
The database can return these in arbitrary orders, at times.


mullr:
need to rebase to get this to cleanly apply against stable. closing, will reopen.

PR #2238, opened 2017-04-07
(PDB-3426) Add PG references and encourage module install

senior:
This commit adds some references to PG docs on setup. It also
encourages users to install via the module on the package install page
along with letting them know about the module in the database
configuration section.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -260,20 +260,33 @@ PuppetDB stores its data in PostgreSQL.
 ### Using PostgreSQL
 
 Before using the PostgreSQL backend, you must set up a PostgreSQL
-server, ensure that it will accept incoming connections, create a user
-for PuppetDB to use when connecting, and create a database for
-PuppetDB. Completely configuring PostgreSQL is beyond the scope of
-this guide, but if you are logged in as root on a running PostgreSQL
-server, you can create a user and database as follows:
+server. Note that users installing PuppetDB via (the module)[module]
+will already have PostgreSQL configured properly and these steps
+should not be necessary.
+
+At a minimum, you will need to ensure that you have PostgreSQL 9.4 or
+later running that will accept incoming connections, a user and
+an empty database for PuppetDB. Information on
+connection/authentication configuration in PostgreSQL and be found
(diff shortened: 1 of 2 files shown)


documentation/configure.markdown
```diff
@@ -260,11 +260,23 @@ PuppetDB stores its data in PostgreSQL.
 ### Using PostgreSQL
 
 Before using the PostgreSQL backend, you must set up a PostgreSQL
-server, ensure that it will accept incoming connections, create a user
-for PuppetDB to use when connecting, and create a database for
-PuppetDB. Completely configuring PostgreSQL is beyond the scope of
-this guide, but if you are logged in as root on a running PostgreSQL
-server, you can create a user and database as follows:
+server. Note that users installing PuppetDB via [the module][module]
```
mullr: `(the module)[module]` (I think)

documentation/configure.markdown
```diff
@@ -260,11 +260,23 @@ PuppetDB stores its data in PostgreSQL.
...
+connection/authentication configuration in PostgreSQL and be found
+(here)[https://www.postgresql.org/docs/9.4/static/auth-pg-hba-conf.html]. Docs
+on setting up users and databases can be found in the (Getting
+Started)[https://www.postgresql.org/docs/9.4/static/tutorial-start.html]
+section of the (PostgreSQL
+manual)[https://www.postgresql.org/docs/9.4/static/index.html].
+
+Completely configuring PostgreSQL is beyond the scope of this guide,
+but a example setup is described below. First, you can create a user
+and database as follows:
```
mullr: Below we have an "if you are running postgresql" qualifier for creating the trigram extension; that should probably be squished together with the regular instructions.

mullr:
+0.98

documentation/configure.markdown
senior: I thought when using a reference like module you had to define it this way. Looking around in the file we use parens an brackets fairly inconsistently. I like the parens better anyway, so I'll switch it.

documentation/configure.markdown
senior: +1

PR #2244, opened 2017-04-10
(maint) clean up the error messages around unknown fields

wkalt:
These messages are surfaced in the console when invalid PQL is supplied.
This change fixes a comma splice and changes the vector in the error
message to a comma separated string.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -1905,13 +1905,13 @@
   [field allowed-fields query-name error-action error-context]
   (let [invalid-fields (unsupported-fields field allowed-fields)]
     (when (> (count invalid-fields) 0)
-      (format "%s unknown '%s' %s '%s'%s. Acceptable fields are: %s"
+      (format "%s unknown '%s' %s %s%s. Acceptable fields are %s"
               error-action
               query-name
-              (if (> (count invalid-fields) 1) "fields:" "field")
-              (str/join "', '" invalid-fields)
+              (if (> (count invalid-fields) 1) "fields" "field")
+              (utils/comma-separated-keywords invalid-fields)
               (if (empty? error-context) "" (str " " error-context))
-              (json/generate-string allowed-fields)))))
+              (utils/comma-separated-keywords allowed-fields)))))
 
 (defn annotate-with-context
   "Add `context` as meta on each `node` that is a vector. This associates the
@@ -1987,11 +1987,13 @@
                             (contains? (set qfields) field)
                             (some #(re-matches % field) (map re-pattern dotted-fields)))
                 {:node node
-                 :state (conj state
-                              (format "'%s' is not a queryable object for %s, %s" field alias
-                                      (if (empty? qfields)
-                                        (format "%s has no queryable objects" alias)
-                                        (format "known queryable objects are %s" (json/generate-string qfields)))))}))
+                 :state (conj
+                          state
+                          (if (empty? qfields)
+                            (tru "''{0}'' is not a queryable object for {1}. Entity {2} has no queryable objects"
+                                 field alias alias)
+                            (tru "''{0}'' is not a queryable object for {1}. Known queryable objects are {2}"
+                                 field alias (utils/comma-separated-keywords qfields))))}))
 
             ; This validation is only for top-level extract operator
             ; For in-extract operator validation, please see annotate-with-context function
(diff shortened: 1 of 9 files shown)


src/puppetlabs/puppetdb/utils.clj
```diff
@@ -288,6 +288,16 @@
...
+(defn single-quote
+  [s]
+  (str "'" s "'"))
+
+(defn comma-separated-keywords
+  [words]
+  (->> words
+       (map single-quote)
+       (string/join ", ")))
+
```
rbrw: Not sure it matters, but if we want this to be general purpose, and if we want the output to be unambiguous, we could do something like this instead: `(pr-str (name k))`, which would also use the more natural double-quotes.  e.g. for `(keyword "this is a keyword")`.  Of course that assumes that we want to strip namespaces from keywords.  If not, then we'd want `str` (and perhaps remove the leading colon) instead of `name`.

test/puppetlabs/puppetdb/http/resources_test.clj
```diff
@@ -194,17 +194,17 @@ to the result of the form supplied to this method."
             response (query-response method endpoint query)]
         (is (= http/status-bad-request (:status response)))
         (is (= (:headers response) {"Content-Type" http/error-response-content-type}))
-        (is (re-find #"'sourceline' is not a queryable object for resources, known queryable objects are" (:body response))))
+        (is (re-find #"'sourceline' is not a queryable object for resources. Known queryable objects are" (:body response))))
```
wkalt: it is a capital K

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1988,10 +1988,11 @@
...
                  :state (conj state
-                              (format "'%s' is not a queryable object for %s, %s" field alias
-                                      (if (empty? qfields)
-                                        (format "%s has no queryable objects" alias)
-                                        (format "known queryable objects are %s" (json/generate-string qfields)))))}))
+                              (tru (format "''{0}'' is not a queryable object for {1}. %s"
+                                           (if (empty? qfields)
+                                             (format "Entity %s has no queryable objects" alias)
+                                             (format "Known queryable objects are %s"
+                                                     (utils/comma-separated-keywords qfields)))) field alias))}))
```
rbrw: I suspect this may not work translation-wise, i.e. tru won't be able to see all the text.

test/puppetlabs/puppetdb/http/resources_test.clj
rbrw: Ahh, so it is - looks odd in gh's diff rendering here.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1986,12 +1986,14 @@
...
+                                         ["Known queryable objects are {2}"
+                                          (utils/comma-separated-keywords qfields)])]
                 {:node node
                  :state (conj state
-                              (format "'%s' is not a queryable object for %s, %s" field alias
-                                      (if (empty? qfields)
-                                        (format "%s has no queryable objects" alias)
-                                        (format "known queryable objects are %s" (json/generate-string qfields)))))}))
+                              (tru (str "''{0}'' is not a queryable object for {1}. " message)
+                                   field alias string))})))
```
rbrw: Does that work?  i.e. assuming the fixed text of the `message` should be translated too, I suspect the translation infrastructure won't be able to see it.  I'm guessing that we'll need the literal strings visible in their entirety in a trs call for each case.

...e.g.

    (if ...
      (tru "''{0}'' is not a queryable object for {1}. Entity {2} has no queryable objects")
      (tru "''{0}'' is not a queryable object for {1}. Known..."))

or something with the same effect.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -2015,10 +2017,11 @@
...
                  :state (conj state
-                              (format "'%s' is not a queryable object for %s, %s" field alias
-                                      (if (empty? qfields)
-                                        (format "%s has no queryable objects" alias)
-                                        (format "known queryable objects are %s" (json/generate-string qfields)))))}))
+                              (tru (format "''%s'' is not a queryable object for %s. %s" field alias
+                                           (if (empty? qfields)
+                                             (format "Entity %s has no queryable objects" alias)
+                                             (format "Known queryable objects are %s"
+                                                     (utils/comma-separated-keywords qfields))))))}))
```
rbrw: Same issue with respect to translation?

wkalt:

rbrw:

PR #2246, opened 2017-04-12
(PE-20115) backport report exists check

mullr:
This backports both the check and the fix to make it hit the index on the hash column to the LTS branch.


rbrw:
Looks good.

PR #2247, opened 2017-04-12
(PDB-3457) Fix sensitive-params test's pg_dump authorization

rbrw:
Postgres command line tools like pg_dump require that credentials be
provided via the console or a PGPASSFILE, so create a temporary file
containing the pdb_test credentials and point PGPASSFILE to it in
pg_dump's environment.

### DIFF
--- a/test/puppetlabs/puppetdb/integration/sensitive_params.clj
+++ b/test/puppetlabs/puppetdb/integration/sensitive_params.clj
@@ -1,32 +1,50 @@
 (ns puppetlabs.puppetdb.integration.sensitive-params
   (:require
    [clojure.java.shell :refer [sh]]
+   [clojure.java.io :refer [writer]]
+   [clojure.pprint :refer [pprint]]
    [clojure.string :as str]
    [clojure.test :refer :all]
    [puppetlabs.puppetdb.integration.fixtures :as int])
   (:import
-   [java.net URI]))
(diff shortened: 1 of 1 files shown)


test/puppetlabs/puppetdb/integration/sensitive_params.clj
```diff
@@ -1,32 +1,47 @@
...
-    (when-not (zero? (:exit result))
-      (is (= 0 (:exit result)))
-      (binding [*out* *err*]
-        (print (:err result))))
-    (:out result)))
+  (let [{:keys [user password subname]} (:db-config pg)
+        attrs (into-array [(-> "rw-------"
+                               PosixFilePermissions/fromString
+                               PosixFilePermissions/asFileAttribute)])
+        pgpass (Files/createTempFile "tmp-pg-dump-passfile" "" attrs)]
```
senior: Any reason not to make an effort to delete this file when we're done?

mullr:
:+1:  here

test/puppetlabs/puppetdb/integration/sensitive_params.clj
rbrw: Umm, no?  Just forgot to include that when I realized I should rewrite (and rewrote) it to deal with perms.

PR #2248, opened 2017-04-12
(PDB-3458) Change the puppetserver-version to SNAPSHOT

rbrw:
The current version, 5.0.0-master-20170329_182047-ga16e296, prevents
use of lein installed versions, and it sounds like we don't need a
specific jar, just some 5.0.0 snapshot.


senior:
@rlinehan Are you a +1 on this?

mullr:
I asked her about this a few days ago, and she was at the time.

rlinehan:
Looks good to me! 👍 

(For the record, you can do like `lein voom build` or something like that - I forget the specific incantation - to build the voom version locally. But since you don't need it just using the SNAPSHOT version is way easier.)

PR #2250, opened 2017-04-13
(PDB-3058) Disallow null bytes from resource event fields

mullr:
There's a broader issue with null bytes from the input leaking through the
database at times, but this is the specific instance we've seen in the field. A
broader fix will be forthcoming.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -1157,13 +1157,21 @@
     (dissoc row-map :environment_id)
     row-map))
 
+(defn replace-null-bytes [x]
+  (if-not (string? x)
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1157,13 +1157,21 @@
     (dissoc row-map :environment_id)
     row-map))
 
+(defn replace-null-bytes [x]
+  (if (string? x)
+    (let [^String s x]
+      (if (>= (.indexOf s (int \u0000)) 0)
```
senior: I agree with @rbrw on the `-1` here

senior:
👍 will merge once the acceptance tests pass

PR #2260, opened 2017-04-18
(maint) Dynamically allocate puppetserver port in integration test

mullr:
Using port 8140 everywhere makes things start failing when multiple tests are
run on the same machine.


rbrw:
Looks reasonable to me.

PR #2263, opened 2017-04-20
(PDB-3420) Exclude inactive and expired nodes in root query endpoint

mullr:
Previously, queries to the root endpoint (which includes all PQL queries)
included all expired and deactivated nodes. This is different from all the other
query endpoints, and practically never what you want to do.

With this change, the root query endpoint modifies queries in the same way as
the other query endpoints to exclude inactive and expired nodes.

### DIFF
--- a/test/puppetlabs/puppetdb/http/index_test.clj
+++ b/test/puppetlabs/puppetdb/http/index_test.clj
@@ -16,6 +16,7 @@
                      vector-param
                      with-http-app]]))
 
+;; Queries issued at the root query endpoint
 (def endpoints [[:v4 "/v4"]])
 
 (deftest-http-app index-queries
@@ -27,15 +28,19 @@
         facts1  (assoc facts "fqdn" "host1")
         facts2  (assoc facts "fqdn" "host2")
         facts3  (assoc facts "fqdn" "host3")
+        facts4  (assoc facts "fqdn" "host4")
         cat1    (assoc catalog :certname "host1")
         cat2    (assoc catalog :certname "host2")
-        cat3    (assoc catalog :certname "host3")]
+        cat3    (assoc catalog :certname "host3")
+        cat4    (assoc catalog :certname "host4")]
     (scf-store/add-certname! "host1")
     (scf-store/add-certname! "host2")
     (scf-store/add-certname! "host3")
+    (scf-store/add-certname! "host4")
     (scf-store/replace-catalog! cat1 (now))
     (scf-store/replace-catalog! cat2 (now))
     (scf-store/replace-catalog! cat3 (now))
+    (scf-store/replace-catalog! cat4 (now))
     (scf-store/add-facts! {:certname "host1"
                            :values facts1
                            :timestamp (now)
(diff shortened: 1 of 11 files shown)


wkalt:
note that this is not different from all other endpoints -- the other endpoints are still mixed in their behavior, which is even worse.

test/puppetlabs/puppetdb/http/index_test.clj
```diff
@@ -179,13 +190,13 @@
...
     (testing "nodes"
-      (testing "query should return all nodes (including deactivated ones)"
+      (testing "query should return only active nodes"
         (doseq [query [["from" "nodes"]
                        "nodes {}"]]
           (is (= (set (mapv :certname (query-result method endpoint query {})))
                  #{"host1" "host2" "host3"}))))
 
-      (testing "broad regexp query should return all nodes"
+      (testing "broad regexp query should return all active nodes"
```
wkalt: can we add a test demonstrating that inactive/deactivated nodes are still accessible via pql?

mullr: They aren't. Are you saying they need to be?

wkalt: yes, definitely

mullr: Why?

wkalt: it was a common feature request until we allowed it. Disallowing it would be a regression.

wkalt: I think for the new version, the way to go is probably to create a queryable pseudo-entity on nodes called "active" that defaults to true and represents deactivated + expired. This will make it easy for users to see deactivated nodes if they want, and also allow more fine-grained differentiation between deactivated and expired if needed via their respective columns.

wkalt: https://tickets.puppetlabs.com/browse/PDB-3070 is related to this

Sharpie: +1 to having a toggle on this behavior. Some more info in PDB-2642.

In most cases, folks only want results from active nodes. However, there are usecases where folks submit a query like `SELECT * FROM nodes` and expect the results  of _that exact query_. Not `SELECT * FROM nodes WHERE expired IS NOT NULL AND deactivated IS NOT NULL`, because that's not the query they submitted. For the API endpoints that do apply this sort of implicit filtering, folks who want a complete query result have to submit two separate queries: one where `active` is set to true and one where it is set to false. Then they have to figure out how to JOIN the two datasets outside of the Puppet tooling (custom scripting).

A concrete usecase for this is folks that have nodes in their population where Puppet doesn't run on a schedule. Those nodes will perpetually be "inactive" because puppet is only executed during specific, infrequent, change windows. Running reports that include data from every node in this sort of population requires a query where the implicit filter is not used.

This implicit filter is a good thing and probably what the majority of folks want for the majority of usecases. However, there needs to be a way to conditionally disable this behavior (perhaps via a query parameter in the URL). Otherwise, this change will create a regression for the folks who need complete results from the entire population in a single query.

rbrw:
Think this is ready.

PR #2268, opened 2017-04-22
(maint) change varchar columns to text

wkalt:
Updates the varchar columns to test. I believe this should be fast
because varchar and text are binary coercible, but we'll need to verify
to be sure.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1424,6 +1424,18 @@
 
   {::vacuum-analyze #{"facts" "fact_values" "fact_paths"}})
 
+(defn varchar-columns-to-text []
+  (jdbc/do-commands
+    "alter table reports
+     alter column puppet_version type text,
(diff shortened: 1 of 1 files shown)


mullr:
@wkalt does this address a specific problem, or is it something you think is just a generally good idea?

wkalt:
@mullr yeah, it's described in https://tickets.puppetlabs.com/browse/PDB-3400

I can reopen this PR and connect it to that ticket.

rbrw:
Agree with Wyatt's assessment that for postgres at least, this probably shouldn't affect performance, and another argument for the switch might to avoid the "somewhat bizarre" behaviors required by SQL for the fixed length fields (i.e. truncating to fit iff the trailing characters are spaces): https://www.postgresql.org/docs/9.6/static/datatype-character.html.  Though I suppose if any data has been affected by that, this change will be a (subtle) change in behavior for those values.

mullr:
@wkalt If you rebase this, I'll merge it

PR #2269, opened 2017-04-25
(PDB-3467) Bump cheshire version to 5.7.1

mullr:
Fixes an issue where cheshire hold onto the head of a seq while write it to a
stream.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -35,7 +35,7 @@
 
   :dependencies [[org.clojure/clojure "1.8.0"]
                  [puppetlabs/i18n ~i18n-version]
-                 [cheshire "5.6.1"]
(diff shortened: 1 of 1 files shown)


mullr:
This needs a jackson bump as well

mullr:
Actually looks like we're getting it transitively through cheshire in this version:
```
 [cheshire "5.7.1"]
   [com.fasterxml.jackson.core/jackson-core "2.8.6"]
   [com.fasterxml.jackson.dataformat/jackson-dataformat-cbor "2.8.6"]
   [com.fasterxml.jackson.dataformat/jackson-dataformat-smile "2.8.6"]
```
So it should be fine.

PR #2270, opened 2017-04-27
(maint) speed up include total

wkalt:


wkalt:
On a small PE master:

before:
```
root@wyatt-centos-master:~[root@wyatt-centos-master ~]# time curl -X GET http://localhost:8080/pdb/query/v4 -d 'query=reports{limit 50}' > /dev/null -d 'include_total=true'
real  0m4.727s
```

after:
```
real  0m0.249s
```
This should make several pages in the console load faster.

wkalt:
also, depends on https://github.com/puppetlabs/puppetdb/pull/2231

wkalt:

wkalt:
this has a bug

mullr:
@wkalt any update on this? Also, do you have before / after query plans handy?

wkalt:
@mullr unfortunately I've forgotten what I was referring to with my ominous comment. I'll poke around and get back to you in a bit.

wkalt:
current include_total plan for reports: https://gist.github.com/wkalt/7e32c2cdcd1b7d2a760c4e7d4d4b26d3

new plan: https://gist.github.com/wkalt/b2dfd446a357ef9252e82bfa200e26a1

Interestingly the difference seems to go away at some level of scale. If I run the first query on my 60k node dataset, I get the second plan with just a seq scan.

wkalt:
also, I think the bug I mentioned above was me mistakenly connecting to the wrong database and getting confused. This is good now as far as I can see.

mullr:
@wkalt merged the 'costly join' PR, can you rebase?

wkalt:
I think we should unrevert #2231 to fix that issue with group_by and paging

wkalt:
then this should go in cleanly

PR #2273, opened 2017-05-01
(maint) add partial indexes on reports

wkalt:
These indexes help support some queries made by the reports summary page
in the PE console.  They take a couple minutes to run on a dataset with
11M reports. It'll probably worthwhile to test on a larger set somehow.

Space on disk is minimal (couple mb) due to how targeted these are.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1252,6 +1252,12 @@
   (jdbc/do-commands
     "create index resource_params_cache_parameters_idx on resource_params_cache using gin (parameters)"))
 
+(defn reports-partial-indices
+  []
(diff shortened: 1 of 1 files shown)


wkalt:
I put this together a few days ago while working on some other stuff. I'll validate again that they actually get used at some point today and update with the results.

wkalt:

mullr:
lgtm, will go and test it out manually. @wkalt I can guess what kind of queries will be helped by this, but do you have the real ones at hand?

wkalt:
@mullr turns out some of these made each other irrelevant. the only ones we really need are the two remaining. Here are a couple queries generated by the reports page with plans before/after; other queries are helped as well. https://gist.github.com/wkalt/46fa496020cc5e7d70a3882af74c3d62

PR #2275, opened 2017-05-03
(PDB-3488) trapperkeeper-webserver-jetty9 -> 2.0.0

rbrw:


rbrw:

ajroetker:
Looks good to me, will there also need to be any updates to extensions @rbrw?

rbrw:
@ajroetker looks like we'll probably want to pin the version in extensions too, to be safe.  I'll raise a pr for that in a bit.

PR #2276, opened 2017-05-04
(PDB-3501) make dashed queries work

wkalt:
For some ancient reason we were converting all dashes in field names to
underscores. Stop doing that so PQL queries with dashes in fact paths
work.


wkalt:

PR #2277, opened 2017-05-05
(PDB-3503) Add the jruby dev dep puppetserver now expects

rbrw:


rbrw:
Sounds like work Jeremy's doing might make this unnecessary/unwanted, but we'll just back it out if/when appropriate.

PR #2278, opened 2017-05-05
(MAINT) Drop the number of selectors/acceptors for puppetserver testing

camlow325:
This commit drops the number of acceptors and selectors used for Puppet
Server testing.  The bump to the new Jetty version results in an
implicit bump in the number of selectors used internally by Jetty.  This
commit drops that number down in order to avoid having the combination
of selector and acceptor threads exceed the max-threads that are
configured.


camlow325:
@rbrw fyi, I think this may fix the current clj unit test failures that started after the Jetty version bump.

rbrw:
Let's find out, and thanks.

PR #2279, opened 2017-05-08
Revert "(PDB-3503) Add the jruby dev dep puppetserver now expects"

camlow325:
This reverts commit 054eb944a5add455b2526c03e1cfd9c457aa3900.  The
commit should no longer be required now that the latest puppetserver
repo jars include their own jruby dependencies again.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -136,8 +136,6 @@
 
   :profiles {:dev {:resource-paths ["test-resources"],
                    :dependencies [[ring-mock]
-                                  ;; puppetserver now expects jruby-deps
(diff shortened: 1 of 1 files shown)


camlow325:
@rbrw, fyi.  I tested this out locally by doing a 'lein test' run for the PuppetDB tests with the latest Puppet Server snapshot delivered to nexus.  The tests all seemed to be passing so I think this is good to go.

rbrw:
Ran pdb unit and integration tests, and tested against current extensions.  Looks good -- thanks.

PR #2280, opened 2017-05-08
(maint) Use a real scope object for terminus testing

adrienthebo:
Changes coming in Puppet 5.0.0 change how Puppet::Parser::Resource
instances set default values; in specific they do so upon object
construction which requires a functioning(ish) scope object. The
Puppetdb terminus specs were relying on the scope object being
relatively simple and easily stubbed but this change in defaults makes
it much harder to treat the scope object as an easily stubbed object.

To simplify stubbing and potentially reduce future breakage this commit
changes the scope stubbing to partially stub a full scope object instead
of creating a pure stub object and adding methods as necessary. By using
a full scope object we can let the internal APIs of Puppet handle
implementation changes.

### DIFF
--- a/puppet/spec/unit/indirector/resource/puppetdb_spec.rb
+++ b/puppet/spec/unit/indirector/resource/puppetdb_spec.rb
@@ -4,6 +4,8 @@ require 'spec_helper'
 require 'puppet/indirector/resource/puppetdb'
 require 'json'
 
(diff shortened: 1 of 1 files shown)


adrienthebo:
yak shaving

rbrw:
Great.  Thanks for the help.

PR #2281, opened 2017-05-09
(PDB-3506) acceptance: install openjdk 8 on jessie

rbrw:
Install openjdk-8-jre-headless on jessie from backports, since we've
dropped support for openjdk 7, and now require openjdk 8.

### DIFF
--- a/acceptance/setup/pre_suite/15_setup_repos.rb
+++ b/acceptance/setup/pre_suite/15_setup_repos.rb
@@ -1,6 +1,14 @@
 def initialize_repo_on_host(host, os, nightly)
   case os
   when :debian
+
+    # For openjdk8
+    if host['platform'].version == '8'
+      create_remote_file(host,
+                         "/etc/apt/sources.list.d/jessie-backports.list",
+                         "deb http://httpredir.debian.org/debian jessie-backports main")
+    end
+
     if options[:type] == 'aio' then
       if nightly
         ## PC1 repos
@@ -18,8 +26,8 @@ def initialize_repo_on_host(host, os, nightly)
       on host, "curl -O http://apt.puppetlabs.com/puppetlabs-release-$(lsb_release -sc).deb"
       on host, "dpkg -i puppetlabs-release-$(lsb_release -sc).deb"
     end
-      on host, "apt-get update"
-      on host, "apt-get install debian-archive-keyring"
+    on host, "apt-get update"
+    on host, "apt-get install debian-archive-keyring"
   when :redhat
     if options[:type] == 'aio' then
       /^(el|centos)-(\d+)-(.+)$/.match(host.platform)
--- a/acceptance/setup/pre_suite/40_install_deps.rb
+++ b/acceptance/setup/pre_suite/40_install_deps.rb
@@ -47,7 +47,13 @@ unless (test_config[:skip_presuite_provisioning])
       if test_config[:install_type] == :git then
         case os
         when :debian
-          on database, "apt-get install -y --force-yes openjdk-7-jre-headless rake unzip"
+          if database['platform'].variant == 'debian' &&
+             database['platform'].version == '8'
+            on database, "apt-get install -y rake unzip"
+            on database, "apt-get install -y -t jessie-backports openjdk-8-jre-headless"
+          else
+            on database, "apt-get install -y rake unzip openjdk-8-jre-headless"
+          end
         when :redhat
           on database, "yum install -y java-1.7.0-openjdk rubygem-rake unzip"
         when :fedora


acceptance/setup/pre_suite/15_setup_repos.rb
```diff
@@ -1,6 +1,14 @@
 def initialize_repo_on_host(host, os, nightly)
   case os
   when :debian
+
+    # For openjdk8
+    if host.platform.start_with?('debian-8')
+      create_remote_file(host,
+                         "/etc/apt/sources.list.d/jessie-backports.list",
+                         "deb http://httpredir.debian.org/debian jessie-backports main")
+    end
```
rlinehan: Does an `apt-get update` need to happen after this? (comparing with https://github.com/puppetlabs/puppetserver/pull/1383/files#diff-b1f26fc14676f1a802179a822fdd9377R204)

acceptance/setup/pre_suite/40_install_deps.rb
```diff
@@ -47,7 +47,11 @@
       if test_config[:install_type] == :git then
         case os
         when :debian
-          on database, "apt-get install -y --force-yes openjdk-7-jre-headless rake unzip"
+          if database['platform'].start_with?('debian-8-')
```
rlinehan: here you're using `debian-8-` but above you're using `debian-8` (probably not important, but figured I might as well point it out).

acceptance/setup/pre_suite/15_setup_repos.rb
```diff
@@ -1,6 +1,14 @@
 def initialize_repo_on_host(host, os, nightly)
   case os
   when :debian
+
+    # For openjdk8
+    if host.platform.start_with?('debian-8')
```
rlinehan: Rather than doing `platform.start-with` you can probably do `platform.variant` and `platform.version` - that seems like it might be a bit more robust? e.g. https://github.com/puppetlabs/puppetserver/pull/1383/files#diff-b1f26fc14676f1a802179a822fdd9377R202

rbrw: Ahh -- there's actually an (improperly indented?) update at the end of this case category.

acceptance/setup/pre_suite/15_setup_repos.rb
rbrw: Ahh -- there's actually an (improperly indented?) update at the end of this case category.

acceptance/setup/pre_suite/15_setup_repos.rb
rbrw: I wondered about that (from Jeremy's PR), but didn't know if that'd be available to us too here; in any case, I agree it'd be preferable, I was just (conservatively) matching what we already had.  I'll adjust it.

acceptance/setup/pre_suite/40_install_deps.rb
rbrw: Oh, right, thanks.

acceptance/setup/pre_suite/40_install_deps.rb
```diff
@@ -47,7 +47,11 @@
       if test_config[:install_type] == :git then
         case os
         when :debian
-          on database, "apt-get install -y --force-yes openjdk-7-jre-headless rake unzip"
+          on database, "apt-get install -y rake unzip"
```
rlinehan: do you need `--force-yes`? also, it looks like now we're only installing openjdk on debian 8? what about other debian variants?

rbrw: I wanted to leave off the force-yes because I don't know why it's there, and I suspect it shouldn't be (I'd planned to add it back if the install doesn't work without it.)

rlinehan: cool, that makes sense about the force-yes

rbrw: Fixed the "other :debians" issue -- thanks.

rlinehan:
LGTM! 💯

acceptance/setup/pre_suite/15_setup_repos.rb
```diff
@@ -1,6 +1,14 @@
 def initialize_repo_on_host(host, os, nightly)
   case os
   when :debian
+
+    # For openjdk8
+    if host['platform'].version == '8'
+      create_remote_file(host,
+                         "/etc/apt/sources.list.d/jessie-backports.list",
+                         "deb http://httpredir.debian.org/debian jessie-backports main")
```
camlow325: For puppetserver, we've been using `deb http://ftp.debian.org/debian jessie-backports main`.  That's apparently been working fine but would you suggest that we change to use what you've done here instead?

acceptance/setup/pre_suite/40_install_deps.rb
```diff
@@ -47,7 +47,13 @@
       if test_config[:install_type] == :git then
         case os
         when :debian
-          on database, "apt-get install -y --force-yes openjdk-7-jre-headless rake unzip"
+          if database.['platform'].variant == 'debian' &&
```
camlow325: Looks like Jenkins CI on the PR is failing because of this line:

```
SyntaxError: /tmp/jenkins/workspace/enterprise_puppetdb_integration-system-puppetdb_full-master/HYPERVISOR/abs/INSTALL_TYPE/install/LAYOUT/centos7-64mda-64a/NIGHTLY/false/PUPPETDB_DATABASE/postgres/label/beaker/acceptance/setup/pre_suite/40_install_deps.rb:50: syntax error, unexpected '[', expecting '('
          if database.['platform'].variant == 'debian' &&
```

Maybe just need to change this to `if database['platform'].variant ...` -- removing the "." after "database"?

acceptance/setup/pre_suite/15_setup_repos.rb
rbrw: Yeah, I think I'd recommend httpredir these days, but I don't know that it's been used heavily inside our CI infrastructure yet, so we might want to wait a bit and see how this goes first.

PR #2283, opened 2017-05-10
(PDB-3506) acceptance: install openjdk 8 on jessie

rbrw:


rbrw:
This includes one patch already on master, and a second that should end up there next.

PR #2285, opened 2017-05-12
(PDB-3511) Invalid commands prevent KahaDB GC

mullr:


puppet/lib/puppet/reports/puppetdb.rb
```diff
@@ -159,8 +159,13 @@ def run_duration
   # @api private
   def event_to_hash(event)
     corrective_change = defined?(event.corrective_change) ? event.corrective_change : nil
+
+    # PUP-1542 sometimes causes status to be nil; fix it up here so we
+    # can store the rest of the report
+    status = event.status || 'failure'
```
wkalt: I'm not sure where the agent team landed with this (maybe this is what we want), but it still feels like we're opening ourselves up to bad data here. The current context where this can happen is understood, but we could end up in other situations down the road where this becomes nil for other reasons.

Maybe we should add a note on the PUP ticket to remove this defaulting when it's resolved?

test/puppetlabs/puppetdb/mq_test.clj
```diff
@@ -168,3 +170,33 @@
...
+(deftest message-exception-handling
+  (testing "on-retry is called for java exceptions"
+    (let [called (atom false)]
+      ((mq-listener/wrap-with-exception-handling
+        (fn [msg] (throw (Exception.)))
+        (fn [msg ex] (reset! called true))
+        (fn [msg cause] nil))
+       "msg")
+      (is @called))
+    )
```
wkalt: hanging paren here

wkalt:
this looks good to me aside from a couple minor comments

puppet/lib/puppet/reports/puppetdb.rb
mullr: We've walked through the code where this can happen; it's a pretty narrow case involving an exception being raised inside a rescue block. So I'm confident that this is indeed a failure. But a note to take out the workaround when (if) we fix it is a good idea, I'll do that.

rbrw: ...and I think where we settled later was that the agent may eventually put a final "status = 'failure'" in the enclosing "ensure" block (presumably when nil).  Assuming ruby guarantees that will always fire, then nil should become impossible.

rbrw: But to be clear, I completely agree with the plan above.

rbrw:
lgtm

PR #2286, opened 2017-05-12
(PDB-3506) acceptance: actually install openjdk 8 on jessie earlier

rbrw:


acceptance/helper.rb
```diff
@@ -838,9 +838,9 @@ def install_puppet_conf
 
   def install_puppet
     hosts.each do |host|
-      if database['platform'].variant == 'debian' &&
-         database['platform'].version == '8'
-        on database, "apt-get install -y -t jessie-backports openjdk-8-jre-headless"
+      if host['platform'].variant == 'debian' &&
+         host['platform'].version == '8'
+        on host, "apt-get install -y -t jessie-backports openjdk-8-jre-headless"
```
wkalt: this doesn't seem quite right to me. We only want to install openjdk on the database and the master, whereas the existing code is attempting to install it on the database once per host, which makes no sense, and this code will install it on standalone agents, which isn't necessary.

It seems like what we really want is to drop the for loop and just have a couple if statements, at which point splitting out a separate host.is_jessie? method may make sense.

mullr: it's unnecessary, but not the end of the world.

PR #2291, opened 2017-05-22
(PDB-3515) Add edges_certname_fkey constraint

mullr:
We had this before, but it was lost in a migration which rewrote the certnames
table but neglected to restore all of the foreign key constraints. This adds it
back in a special startup function; nominally this would be in a migration, but
we've already consumed the following migration numbers in newer versions. The
corresponding change in the new code will use a migration.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1322,6 +1322,30 @@
     (jdbc/do-commands
      "CREATE INDEX fact_values_string_trgm ON fact_values USING gin (value_string gin_trgm_ops)")))
 
+(defn fix-missing-edges-fk-constraint []
+  (when-not (sutils/constraint-exists? "certnames" "edges_certname_fkey")
+    (log/info "Cleaning up orphaned edges")
+
+    (jdbc/do-commands
+     (str "SELECT e.*"
+          "  INTO edges_transform"
+          "  FROM edges e"
+          "  INNER JOIN certnames c ON e.certname = c.certname")
+     (str "ALTER TABLE edges_transform"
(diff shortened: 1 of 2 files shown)


rbrw:
If it's plausible, thought we might want a schema diff migration-test for this.

src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -147,6 +147,24 @@
...
+  ([schema :- s/Str
+    table :- s/Str
+    constraint :- s/Str]
+
+   (let [query (str "select count(*)"
+                    "  from information_schema.constraint_column_usage"
+                    "  where table_schema = ?"
+                    "  and table_name = ?"
+                    "  and constraint_name = ?")
+         results (jdbc/query-to-vec [query schema table constraint])]
```
mullr: I just took it from like 146 above without thinking... and am happy to keep it if it works :)

mullr:
(rebased)

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1322,6 +1322,30 @@
     (jdbc/do-commands
      "CREATE INDEX fact_values_string_trgm ON fact_values USING gin (value_string gin_trgm_ops)")))
 
+(defn fix-missing-edges-fk-constraint []
+  (when-not (sutils/constraint-exists? "certnames" "edges_certname_fkey")
+    (log/info "Cleaning up orphaned edges")
```
rbrw: Missing the trs as compared to stable, if it matters.

mullr: Don't think it matters on LTS

rbrw:
Looks good.

PR #2292, opened 2017-05-22
(PDB-3515) Add edges_certname_fkey constraint

mullr:
We had this before, but it was lost in a migration which rewrote the certnames
table but neglected to restore all of the foreign key constraints. This adds a
migration to recreate the constraint and clean up any orphaned data.


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1252,6 +1252,18 @@
   (jdbc/do-commands
     "create index resource_params_cache_parameters_idx on resource_params_cache using gin (parameters)"))
 
+(defn fix-missing-edges-fk-constraint []
+  (when-not (sutils/constraint-exists? "certnames" "edges_certname_fkey")
+    (log/info "Cleaning up orphaned edges")
+    (jdbc/delete! :edges ["certname not in (select certname from certnames)"])
```
npwalker: I think it might be better to grab the valid rows into a temp table, truncate edges, and then re-insert the valid rows.

If we delete 50% of the rows we'll be leaving behind some pretty bloated tables and indexes.  If we truncate and reinsert we'll be leaving behind a really full not-bloated table.

mullr: Oh indeed. What about a 'vacuum full' on this table, should do the same thing?

npwalker: Something like:

```
SELECT e.* 
INTO edges_new 
FROM edges as e 
INNER JOIN certnames as c 
  ON e.certname = c.certname;

DROP TABLE edges;

ALTER TABLE edges_new RENAME TO edges;

CREATE UNIQUE INDEX edges_certname_source_target_type_unique_key ON edges USING btree (certname, source, target, type);
```

npwalker: I think inserting the valid data into a new table is less work than first deleting the invalid data and then rewriting the table.  

In both cases you rewrite the table but in the DELETE case you have to pay the price of all the deletes before you rewrite.

mullr: Good point. I'll change it to do that.

npwalker: That might all need to be done inside of a transaction.  I'm not sure if the migrations block out incoming traffic or not.

rbrw:
Assuming it's plausible, thought we might want a schema diff migration-test for this.

rbrw:
Also looks good.

PR #2293, opened 2017-05-24
(PDB-3518) fix latest report id index problems

mullr:
First, this backports several indexes to LTS so we can get up to the official version of this index. Then we have a change to the query engine to make queries that don't break the planner's brain.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -263,7 +263,9 @@
                                        [:= :certnames.certname :fs.certname]
 
                                        :reports
-                                       [:= :certnames.latest_report_id :reports.id]
+                                       [:and
+                                        [:= :certnames.certname :reports.certname]
+                                        [:= :certnames.latest_report_id :reports.id]]
 
                                        [:environments :catalog_environment]
                                        [:= :catalog_environment.id :catalogs.environment_id]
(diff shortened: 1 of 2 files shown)


mullr:
Tested the change end-to-end and confirmed that it works. The pdb query to make this happen is:

#+BEGIN_SRC restclient
  POST http://localhost:8080/pdb/query/v4
  Content-Type: application/json

  {
    "limit": 50,
    "offset": 0,
    "order_by": [{"field": "report_timestamp", "order": "desc"}],
    "query": ["from", "nodes", 
               ["extract", ["certname"],
                 ["subquery", "reports", 
                   ["<", "end_time", "2015-10-14"]]]]
  }
#+END_SRC

Where end_time depends on the data set at and.

mullr:
The final commit will need to be pulled up to stable / master

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -263,7 +263,9 @@
                                        [:= :certnames.certname :fs.certname]
 
                                        :reports
-                                       [:= :certnames.latest_report_id :reports.id]
+                                       [:and
+                                        [:= :certnames.certname :reports.certname]
+                                        [:= :certnames.latest_report_id :reports.id]]
```
npwalker: @mullr in addition to here should we be adding the certname join criteria in [latest-report-query](https://github.com/puppetlabs/puppetdb/pull/2293/files#diff-fefe75691a2bbea085773295ea2b7c49R940) as well?

mullr: yes, it should. Why is that even different?

mullr: Since this is just a backport of the actual fix, it should probably be done separately. Will file a ticket.

npwalker: That works.

mullr:
(rebased)

src/puppetlabs/puppetdb/query_eng/engine.clj
mullr: I have no idea what I was thinking before; this is not a backport. Added that condition, and will open this on stable as well.

PR #2294, opened 2017-05-24
(PDB-3527) Use #to_a method, not #to_data_hash when converting TagSets

thallgren:
The `Puppet::Util::TagSet` class has a method named `#to_data_hash`. It
returns an `Array` instead of a `Hash`. PuppetDB assumes that the hash
returned from a call to `Puppet::Resource::Catalog#to_data_hash` contains
tag sets and calls the `#to_data_hash` to convert them into an `Array`.
This now breaks because the bug that caused the hash to contain tag sets
was fixed in PUP-7381.

This commit changes the call to `#to_data_hash` to instead be a call to
`#to_a`. Both the `Puppet::Util::TagSet` and the `Array` class implements
this method which means that PuppetDB will be capable of munging catalogs
produced both before and after PUP-7381.

### DIFF
--- a/puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
+++ b/puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
@@ -709,6 +709,33 @@ describe Puppet::Resource::Catalog::Puppetdb do
           'transaction_uuid', 'environment', 'producer_timestamp', "code_id",
           "job_id", "catalog_uuid", "producer"]
       end
+
+      context 'when dealing with TagSets' do
+        let(:resource_hashes) do
+          {
+            'resources' => [
+              {'type' => 'Stage',
+              'title' => :main,
+              'tags' => Puppet::Util::TagSet.new(['stage']),
+              'exported' => false,
+              'parameters' => {:name => 'main'}},
+             {'type' => 'Class',
+              'title' => 'Settings',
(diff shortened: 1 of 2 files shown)


puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
```diff
@@ -709,14 +709,41 @@ def test_file_require(resource_title, require_title)
...
+              'title' => :main,
+              'tags' => Puppet::Util::TagSet.new(['class']),
+              'exported' => false,
+              'parameters' => {:name => 'main'}}
+            ]
+          }
+        end
+
+        it "should convert TagSet for resource['tags']" do
+          subject.hashify_tags(resource_hashes)['resources'].each { |resource| resource['tags'].should be_an Array }
```
rbrw: Minor concern - I'd prefer to see this wrapped a bit since, if nothing else, it "goes off the side" of github's diff view; don't think we have a pdb standard, but maybe Russ and I can compromise on "normally under 100").

thallgren: Wow. I normally use 150 as the wrap margin. But OK, I'll wrap it.

rbrw:
Best I can tell, this looks OK.

PR #2296, opened 2017-05-25
(PDB-3529) Check latest report on certnames instead of reports

npwalker:
Prior to this commit, we calculated the latest report for a node
by querying the reports table, sorting all the records, and grabbing
the most recent report id.

After this commit, we store the latest report timestamp in the
certnames table with the report-id.  This allows us to compare
each new report with the latest report timestamp as they come in
and decide whether this report is the latest report or not without
querying the reports table.


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1286,16 +1286,21 @@
...
+                                      FROM certnames
                                       WHERE certname = ?
-                                      ORDER BY producer_timestamp DESC
-                                      LIMIT 1" node])))]
-    (jdbc/update! :certnames
-                  {:latest_report_id latest-report}
-                  ["certname = ?" node])))
+                                      " node])))]
+  ;  (if (sql-ts-after? (to-timestamp report-timestamp)
+  ;                   (to-timestamp latest-report-timestamp))
```
npwalker: I don't quite understand how to make the comparison between the timestamp already in certnames and coming in

wkalt: I think you want something like

```
(sql-ts-after? latest-report-timestamp (to-timestamp report-timestamp))
```

since the stamp out of the database is in the correct form already

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1266,6 +1265,11 @@
     "create index reports_job_id_idx on reports(job_id) where job_id is not null"
     "create index catalogs_job_id_idx on catalogs(job_id) where job_id is not null"))
 
+(defn add-latest-report-timestamp-to-certnames []
+  (jdbc/do-commands
+    "ALTER TABLE certnames ADD COLUMN latest_report_timestamp timestamp with time zone"
+    "CREATE INDEX idx_certnames_latest_report_timestamp ON certnames(latest_report_timestamp)"))
```
wkalt: we will want to backfill the column with existing data. I'd guess a simple update would be alright since it's a relatively small table.

npwalker: The latest_report_id can be NULL... so I think we have to change that or we have to allow this new column to also be NULL and if we allow that then we wouldn't need to backfill.  

We'd need to backfill if we can optimize something else to read this field instead of calculating it in other places.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1286,16 +1286,21 @@
 (defn update-latest-report!
   "Given a node name, updates the `certnames` table to ensure that it indicates the
    most recent report for the node."
-  [node]
-  {:pre [(string? node)]}
-  (let [latest-report (:id (first (query-to-vec
-                                    ["SELECT id FROM reports
+  [node report-id report-timestamp]
+  {:pre [(string? node) (integer? report-id)]}
```
npwalker: I wanted to validate the report-timestamp coming in but I'm not sure what datatype it is to check.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1286,16 +1286,21 @@
...
+                                      FROM certnames
                                       WHERE certname = ?
-                                      ORDER BY producer_timestamp DESC
-                                      LIMIT 1" node])))]
-    (jdbc/update! :certnames
-                  {:latest_report_id latest-report}
-                  ["certname = ?" node])))
+                                      " node])))]
+    (if (sql-ts-after? (to-timestamp report-timestamp)
+                     (to-timestamp latest-report-timestamp))
```
npwalker: I'd like to write a unit test that ensures that an older report won't update the latest_report_id and _timestamp.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1286,16 +1286,21 @@
...
+
+      (jdbc/update! :certnames
+                    {:latest_report_id report-id
+                     :latest_report_timestamp report-timestamp}
+                    ["certname = ?" node]))))
```
npwalker: Can I avoid a round trip to the database above by adding 

```
["? > latest_report_timestamp" report-timestamp]
```

This would always send the UPDATE but the where clause would exclude cases where the new report timestamp isn't newer than the existing latest_report_timestamp.

npwalker:
I updated the code to always submit an UPDATE statement with a where clause that can check to make sure that report timestamp being processed is newer than the one that is stored as latest.  

I manually updated the latest_report_timestamp to sometime tomorrow and ran puppet to ensure that it would not override the timestamp or id.  It behaved correctly. 

```
LOG:  duration: 0.046 ms  plan:
	Query Text: UPDATE certnames SET latest_report_id = $1,latest_report_timestamp = $2 WHERE certname = $3 AND latest_report_timestamp < $4
	Update on certnames  (cost=0.41..8.43 rows=1 width=85) (actual time=0.045..0.045 rows=0 loops=1)
	  Buffers: shared hit=4
	  ->  Index Scan using certnames_transform_certname_key on certnames  (cost=0.41..8.43 rows=1 width=85) (actual time=0.044..0.044 rows=0 loops=1)
	        Index Cond: (certname = 'noway.corp.puppetlabs.net'::text)
	        Filter: (latest_report_timestamp < '2017-06-01 16:27:14.903-07'::timestamp with time zone)
	        Rows Removed by Filter: 1
	        Buffers: shared hit=4
```

mullr:
@npwalker we're going to want this on stable, can you rebase it on that branch?

npwalker:
@mullr sure thing.  It's re-targeted at stable but it'll conflict with the other PRs that have a migration 59 when those go in.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1284,16 +1284,13 @@
...
-  (let [latest-report (:id (first (query-to-vec
-                                    ["SELECT id FROM reports
-                                      WHERE certname = ?
-                                      ORDER BY producer_timestamp DESC
-                                      LIMIT 1" node])))]
-    (jdbc/update! :certnames
-                  {:latest_report_id latest-report}
-                  ["certname = ?" node])))
+  [node report-id report-timestamp]
+  {:pre [(string? node) (integer? report-id)]}
```
npwalker: Still need to validate the timestamp coming in

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1284,16 +1284,13 @@
...
+
+  (jdbc/update! :certnames
+                {:latest_report_id report-id
+                 :latest_report_timestamp report-timestamp}
+                ["certname = ? AND latest_report_timestamp < ?" node report-timestamp]))
```
npwalker: I tested that this correctly does not update when the report coming in is not newer but maybe that'd be good to have a unit test for?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1284,16 +1284,13 @@
...
-  {:pre [(string? node)]}
-  (let [latest-report (:id (first (query-to-vec
-                                    ["SELECT id FROM reports
-                                      WHERE certname = ?
-                                      ORDER BY producer_timestamp DESC
-                                      LIMIT 1" node])))]
-    (jdbc/update! :certnames
-                  {:latest_report_id latest-report}
-                  ["certname = ?" node])))
+  [node report-id report-timestamp]
```
mullr: should call this producer-timestamp instead to be as specific as possible (there are a few different timestamps on reports)

src/puppetlabs/puppetdb/scf/storage.clj
mullr: Yes, we should. This could easily be added to `storage-test/latest-report`

mullr:
rebased on top of #

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1258,6 +1258,37 @@
     "DROP INDEX IF EXISTS facts_factset_id_fact_path_id_idx"
     "CREATE INDEX facts_factset_id_fact_path_id_idx ON facts(factset_id, fact_path_id)"))
 
+<<<<<<< HEAD
```
npwalker: missed one

mullr:

PR #2297, opened 2017-05-25
(maint) Use string keys in terminus data munging code

mullr:
Puppet has changed its to_data_hash code to emit json-compatible data. This
means that all map keys are now strings, where some used to be symbols.


rbrw:
lgtm

PR #2301, opened 2017-05-26
(PDB-3173) Add node purge batch limits

rbrw:


rbrw:
Doesn't have the doc updates we'll need yet.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -341,6 +341,33 @@
       result
       (recur db-spec))))
 
+(defn garbage-collector
+  [db clean-lock config]
+  (let [seconds-pos? (comp pos? to-seconds)
+        what (filter identity
+                     [(when-let [node-ttl (:node-ttl config)]
```
mullr: These nested whens can be a bit terser with `some->`, eg. `(when (some-> config :node-ttl to-seconds pos?) ...`

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -341,6 +341,33 @@
...
+                          (let [limit (:node-purge-gc-batch-limit config)]
+                            (if (zero? limit)
+                              "purge_nodes"
+                              ["purge_nodes" {:batch_limit limit}]))))
+                      (when-let [report-ttl (:report-ttl config)]
+                        (when (seconds-pos? report-ttl)
+                          "purge_reports"))
+                      "other"])]
+    (fn []
+      (.lock clean-lock)
```
mullr: https://clojuredocs.org/clojure.core/locking

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -341,6 +341,33 @@
...
+                        (when (seconds-pos? node-purge-ttl)
+                          (let [limit (:node-purge-gc-batch-limit config)]
+                            (if (zero? limit)
+                              "purge_nodes"
+                              ["purge_nodes" {:batch_limit limit}]))))
+                      (when-let [report-ttl (:report-ttl config)]
+                        (when (seconds-pos? report-ttl)
+                          "purge_reports"))
+                      "other"])]
+    (fn []
```
mullr: can we make this a regular function? And use `#()` notation to create the thunk for at-at

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -85,10 +85,38 @@
...
+(def clean-options
+  #{"expire_nodes" "purge_nodes" "purge_reports" "package_gc" "other"})
+
+(def purge-nodes-opts-schema {:batch_limit s/Int})
+
+(def clean-request-schema
+  ;; i.e. a possibly empty collection of elements, each of which must
+  ;; either be a string from clean-options or a purge_nodes command
+  ;; with an options map.
+  [(s/if string?
```
rbrw: Is there a schema `or`?  I thought you were supposed to use `conditional`, `cond-pre`, or, `if` now.

src/puppetlabs/puppetdb/cli/services.clj
rbrw: I'm not sure we can if we want to be able to test exactly what interposed will be running easily (see the test in services-test).  i.e. the #() would need to build the `what` and capture it, and it's nice to be able to test all that.  Though of course there are other ways we could break that up -- this was just a quick transformation.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -239,13 +239,30 @@
...
+      [ts ts])))
+  ([horizon batch-limit]
+   {:pre [(kitchensink/datetime? horizon)]}
+   (let [ts (to-timestamp horizon)]
+     (jdbc/do-prepared
+      (str "with ids as (delete from certnames"
+           "               where id in (select id from certnames"
+           "                              where deactivated < ?"
+           "                                    or expired < ?"
+           "                                limit ?)"
```
rbrw: Unless we think there's some benefit to ordering, I'd be inclined to leave it off so there's no chance it can impose a sorting cost.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -106,18 +134,22 @@
...
+   db :- (s/pred map?)]
   (try
     (kitchensink/demarcate
      (format "purge deactivated and expired nodes (threshold: %s)"
              (format-period node-purge-ttl))
-     (jdbc/with-transacted-connection db
-       (scf-store/purge-deactivated-and-expired-nodes! (ago node-purge-ttl))))
+     (let [horizon (ago node-purge-ttl)]
+       (jdbc/with-transacted-connection db
+         (if-let [limit (and (map? opts) (:batch_limit opts))]
```
rbrw: Ahh, right it has to be a map or true now, given the preconditions -- think that may be vestigial, from a previous iteration.

rbrw: ...though of course that relies on (:batch_limit true) being nil, which is arguably a bit odd.

src/puppetlabs/puppetdb/cli/services.clj
rbrw: Yep - nice.

src/puppetlabs/puppetdb/cli/services.clj
rbrw: Indeed, I know/knew about locking, so I don't know why I didn't use it before; perhaps just forgot.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -321,6 +343,30 @@
...
+                        "expire_nodes")
+                      (when (some-> (:node-purge-ttl config) seconds-pos?)
+                        (let [limit (:node-purge-gc-batch-limit config)]
+                          (if (zero? limit)
+                            "purge_nodes"
+                            ["purge_nodes" {:batch_limit limit}])))
+                      (when (some-> (:report-ttl config) seconds-pos?)
+                        "purge_reports")
+                      "other"])]
+    (fn []
```
mullr: Can we make this a 0th order function? i.e. defn collect-garbage, then wrap it in a #() down below where we need the thunk.

mullr: I guess that's a first order function. I

src/puppetlabs/puppetdb/cli/services.clj
rbrw: See the (arg) schema - opts can be `true`, or a map of options, where `true` means "do the default thing" (i.e. no batch limit) -- that's due to the way we handle the various commands in the caller (`clean-up`), but could of course be changed.

src/puppetlabs/puppetdb/cli/services.clj
rbrw: Oh, there was a reason  I didn't change this; well actually I did locally, but then noticed some issues I was going to mention, and changed it back, but I'll have to remember what they were.

rbrw: Ahh, I think one reason was because the function needs that logic and to capture that state, and so this arrangement facilitates testing.

mullr: ... really? Can the test not just use `#()` in the same way? 

Sorry to harp on this, but higher-order functions are pretty significantly more complex than regular ones.

rbrw: Sure, but now we're copy-pasting the code.

rbrw: I suspect there might be some other way to rework this, though, and I'd be fine with that too -- might be able to break it in to two pieces; I'll poke at it.

mullr: I'll grant you that it's a small bit of copypasta. But I don't think that's sufficient reason to reach for the higher-order-function hammer.

src/puppetlabs/puppetdb/cli/services.clj
mullr: :+1:

PR #2304, opened 2017-05-30
(maint) Workaround PUP-7605

mullr:
Puppet is stringifying all resource titles in catalog.to_data_hash now, but
catalog.resource(type, title) can't accept string titles in all case. It special
cases :main, so apply the inverse special case here.


joshcooper:
This seems surprising... I would expect `catalog.resource` to accept string titles if `to_data_hash` emits them as strings... 

/cc @thallgren

mullr:
Let's merge this once travis passes, can revert it once pup-7605 is fixed

joshcooper:
FYI, the same failure is happening in PE, I assume because the pe-puppetdb-terminus package must be manually promoted into Hoyt? Here's the failure on the master: https://gist.github.com/joshcooper/448889b2e9f573421a3b4feb71f1d8b5

PR #2306, opened 2017-05-30
(PDB-2950) Require PostgreSQL >= 9.6

mullr:
- Update the minimum version vector to [9 6].

- Refactor version check functions to allow the minimum version to be overriden
  if necessary (temporary internal hook, for use by PE)

- Fix a bug that was preventing version-check test assertions from running

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -388,7 +388,7 @@ module PuppetDBExtensions
       # get the pg server up and running
       class { 'postgresql::globals':
           manage_package_repo => true,
-          version             => '9.4',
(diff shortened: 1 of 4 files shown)


mullr:

rbrw:
I'll test this in the morning, but it looks pretty good.

rbrw:
Here at least, this hangs when run against 9.4, after printing the warning.  It might be a local issue, but we should double check, because it's one of the dreaded "C-c and C-\ can't kill it", failures.

rbrw:
Looks good given the ticket to deal with the shutdown hang.

PR #2307, opened 2017-06-01
(PDB-3544) Improve facts_factset_id_index

npwalker:
Prior to this commit, the facts table had an index on factset_id.

After this commit, we update that index to by on factset_id,
fact_path_id because these two columns are often filtered on together
or one is filtered on while the other is used in a join.


mullr:
I've benchmarked this locally and can see essentially zero impact on fact processing speed for our typical test data. Going ahead with the merge as we've observed a benefit in the field.

mullr:
(rebased and merged manually)

PR #2310, opened 2017-06-05
(PDB-3541) Don't hang on unsupported database

rbrw:
Instead of calling System/exit, use the trapperkeeper shutdown service
to shut down when an unsupported database is detected.  Previously pdb
would just hang.

One drawback to this approach is that it looks like the only way to
get trapperkeeper to shut down and exit with a non-zero status is to
throw an exception, and we'd prefer to avoid the backtrace since the
problem has already been described in the log and on stderr.

Hopefully trapperkeeper will add something more suitable, and then we
can use it.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -282,17 +282,21 @@
   context)
 
 (defn initialize-schema
-  "Ensure the database is migrated to the latest version, and warn if
-  it's deprecated, log and exit if it's unsupported.
-
-  This function will return true iff any migrations were run."
+  "Ensures the database is migrated to the latest version, and returns
+  true iff any migrations were run.  Throws
+  {:type ::unsupported-database :current version :oldest version} if
+  the current database is not supported."
   [db-conn-pool config]
   (jdbc/with-db-connection db-conn-pool
-    (scf-store/validate-database-version #(utils/flush-and-exit 1))
-    @sutils/db-metadata
-    (let [migrated? (migrate! db-conn-pool)]
-      (indexes! config)
-      migrated?)))
+    (let [current (:version @sutils/db-metadata)]
+      (when (neg? (compare current scf-store/oldest-supported-db))
+        (throw+ {:type ::unsupported-database
+                 :current current
+                 :oldest scf-store/oldest-supported-db}))
+      @sutils/db-metadata
+      (let [migrated? (migrate! db-conn-pool)]
+        (indexes! config)
+        migrated?))))
 
 (defn init-with-db
   "All initialization operations needing a database connection should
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -281,20 +281,38 @@
...
-(defn initialize-schema
-  "Ensure the database is migrated to the latest version, and warn if
-  it's deprecated, log and exit if it's unsupported.
-
-  This function will return true iff any migrations were run."
-  [db-conn-pool config]
+(defn db-unsupported-msg
+  "Returns a message describing which databases are supported."
+  []
+  (str (trs "PostgreSQL DB versions older than {0}.{1} are no longer supported."
```
mullr: We shouldn't glue trs-ed strings together, instead, there should be one huge string inside a single TRS.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -281,20 +281,38 @@
...
+            (first scf-store/oldest-supported-db)
+            (second scf-store/oldest-supported-db))
+       " "
+       (trs "Please upgrade Postgres and restart PuppetDB.")))
+
+(defn shutdown-if-db-not-supported [shutdown-on-error]
+  (let [oldest scf-store/oldest-supported-db]
```
mullr: `scf-store/oldest-supported-db` can be inlined

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -281,20 +281,38 @@
...
+    (when (sutils/db-version-older-than? oldest)
+      (let [msg (db-unsupported-msg)
+            attn (utils/attention-msg msg)]
+        (utils/println-err attn)
+        (log/error attn)
+        (shutdown-on-error #(throw (Exception. msg)))))))
```
mullr: Yeah, this is really icky. Not much to be done about it now; maybe a comment above it would be good, informing whomever reads this next of just what we were thinking.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -422,14 +441,16 @@
...
+   [:ShutdownService shutdown-on-error]]
   (init [this context]
 
         (doseq [{:keys [reporter]} (vals metrics/metrics-registries)]
           (jmx-reporter/start reporter))
         (assoc context :url-prefix (atom nil)))
   (start [this context]
-         (start-puppetdb context (get-config) this get-registered-endpoints))
+         (start-puppetdb context (get-config) this get-registered-endpoints
+                         (partial shutdown-on-error (service-id this))))
```
mullr: This seems like a pretty idiomatic place to use exceptions. Instead of passing down this 'shutdown-on-error' function, we could have initial-schema check the database version as a precondition and `(throw+ {:type ::unsupported-database, :required-db-version [9 6], :found-pg-version [9 4]}`, then `(try+ ... (catch [:type :scf/unsupported-database] (log blah) (shutdown blarg)))` here at the top level.

src/puppetlabs/puppetdb/cli/services.clj
rbrw: I'd argue it depends on how dependent they are, but perhaps thats plausible here.

src/puppetlabs/puppetdb/cli/services.clj
rbrw: Sure, though I wondered if it might make sense to track that dividing line independently, and in the storage layer -- but I don't feel strongly about it.

src/puppetlabs/puppetdb/cli/services.clj
mullr: You could probably it's convince me it's ok in this case.

PR #2312, opened 2017-06-05
(SERVER-1823) Remove certname from http client metrics

adrienthebo:
When metrics gathering was added for PuppetDB commands, the certname
associated with the command was included as part of the metrics path.
This can cause a large number of individual metrics to be generated in
large environments, which is undesirable for a few reasons. By itself
the number of metrics can be unwieldy, and there's memory overhead for
storing each of these metrics.

This commit reduces the metric proliferation by dropping the certname
from the metrics path when an HTTP request is made.

/cc @camlow325 @rlinehan


rlinehan:
I think it also needs to be removed from the facts indirection and resource indirection and... I think that's it? In those two it's `request.key`.

adrienthebo:
I think that makes sense for finding facts, but for resources I believe that `resource.key` is the resource type, not the certname. It looks like this is the metric_id: `{:metric_id => [:puppetdb, :resource, :search, 'File']}`. It's not the node cername but it might be unnecessary. Shall we just take this out while we're here?

rlinehan:
Yeah, I think it's unnecessary.

adrienthebo:
Certname removed from resource and facts terminii.

rlinehan:
So this is kind of a nitpick, but since it's actually not the certname in the resource terminii, could you update the commit message to say it's the resource name? Otherwise I could see someone going back through the commit log and getting confused.

adrienthebo:
Good by me, commit amended.

puppet/lib/puppet/indirector/facts/puppetdb.rb
```diff
@@ -62,7 +62,7 @@ def find(request)
             Http.multi_arity_get(http_instance,
```
camlow325: @mullr Did y'all make the decision that the next release of puppetdb off of the master branch will require at least Puppet Server 5 to be present?  We had only added the `Http.multi_arity_get` function because we wanted to ensure that this code would work across both Puppet Server 2.x and 5.x (given the underlying arity difference across versions in the http_client get method) but if we no longer have to support Puppet Server 2.x from this code, we could just code to the new arity and make this code a bit simpler.

mullr: @camlow325 Yes, puppetdb 5 (master) will require puppet 5 / puppetserver 5.

camlow325: Thanks, @mullr.  @adrienthebo, @rlinehan - since we're in here already, what would you think about collapsing out the `Http.multi_arity_get` functionality as part of this PR given that?

adrienthebo: Sounds good to me, I'll get on that shortly.

adrienthebo:
@camlow325 @mullr I've removed the `Http.multi_arity_get` method; is there a good way to run acceptance tests with these changes?

mullr:

PR #2316, opened 2017-06-08
(PDB-3549) Update acceptance to use puppet5-nightly

johnduarte:
Replace previous usage of `PC1` nightlies with `puppet5`.

Add support for the use case where the puppetdb build version is not specified. In this case, the version of puppetdb published in the `puppet5-nightly` repo will be installed.

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -265,11 +265,14 @@ module PuppetDBExtensions
   end
 
   def get_package_version(host, version = nil)
-
     if version == 'latest'
       return 'latest'
     elsif version.nil?
       version = PuppetDBExtensions.config[:package_build_version].to_s
+      # If no version was defined, default to latest.
+      if version == ''
+        return 'latest'
+      end
     end
 
     # version can look like:
@@ -335,7 +338,7 @@ module PuppetDBExtensions
       expected_version = get_package_version(host)
 
       Beaker::Log.notify "Expecting package version: '#{expected_version}', actual version: '#{installed_version}'"
-      if installed_version != expected_version
+      if installed_version != expected_version and expected_version != 'latest'
         raise RuntimeError, "Installed version '#{installed_version}' did not match expected version '#{expected_version}'"
       end
     end
--- a/acceptance/setup/pre_suite/80_add_dev_repo.rb
+++ b/acceptance/setup/pre_suite/80_add_dev_repo.rb
@@ -1,6 +1,7 @@
 install_opts = options.merge( { :dev_builds_repos => ["PC1"] })
 repo_config_dir = 'tmp/repo_configs'
-if (test_config[:install_type] == :package and not test_config[:skip_presuite_provisioning])
+if (test_config[:install_type] == :package and test_config[:package_build_version] and not test_config[:skip_presuite_provisioning])
+  # do not install the dev_repo if a package_build_version has not been specified.
   databases.each do |database|
     install_puppetlabs_dev_repo database, 'puppetdb', test_config[:package_build_version],
                                 repo_config_dir, install_opts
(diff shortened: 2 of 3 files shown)


johnduarte:
This has been tested on RedHat 7 and Ubuntu 16.04 using both with and without a specified `PUPPETDB_PACKAGE_BUILD_VERSION`.
With the build specified, the specified build is installed and the tests pass.
```
export PUPPETDB_PACKAGE_BUILD_VERSION=5.0.0.SNAPSHOT.2017.06.06T1646
export GEM_SOURCE=http://rubygems.delivery.puppetlabs.net
export BEAKER_TYPE=aio
export NIGHTLY=true
export LAYOUT=centos7-64mda-64a
export PUPPETDB_DATABASE=postgres
./ext/jenkins/beaker-tests.sh
...
              - Test Case Summary for suite 'tests' -
       Total Suite Time: 154.69 seconds
      Average Test Time: 25.78 seconds
              Attempted: 6
                 Passed: 6
...
Expecting package version: '5.0.0-0.1SNAPSHOT.2017.06.06T1646puppetlabs1', actual version: '5.0.0-0.1SNAPSHOT.2017.06.06T1646puppetlabs1'
```

Without the build specified, the nightly build is installed and the tests pass.
```
export GEM_SOURCE=http://rubygems.delivery.puppetlabs.net
export BEAKER_TYPE=aio
export NIGHTLY=true
export LAYOUT=centos7-64mda-64a
export PUPPETDB_DATABASE=postgres
./ext/jenkins/beaker-tests.sh
...
       Total Suite Time: 138.91 seconds
      Average Test Time: 23.15 seconds
              Attempted: 6
                 Passed: 6
...
Expecting package version: 'latest', actual version: '5.0.0-0.1SNAPSHOT.2017.06.06T2148puppetlabs1'
```

johnduarte:
I should probably add some notes to the {{acceptance/README.md}} about these changes.

acceptance/helper.rb
```diff
@@ -335,7 +338,7 @@ def validate_package_version(host)
       expected_version = get_package_version(host)
 
       Beaker::Log.notify "Expecting package version: '#{expected_version}', actual version: '#{installed_version}'"
-      if installed_version != expected_version
+      if installed_version != expected_version and expected_version != 'latest'
         raise RuntimeError, "Installed version '#{installed_version}' did not match expected version '#{expected_version}'"
```
rbrw: Not sure I completely understand the context yet, but wondered if it might make sense for expected_version to be set to 'latest' when that's appropriate, instead of the extra guard.  Seemed like that might also improve the log message above.

acceptance/setup/pre_suite/80_add_dev_repo.rb
```diff
@@ -1,6 +1,7 @@
 install_opts = options.merge( { :dev_builds_repos => ["PC1"] })
 repo_config_dir = 'tmp/repo_configs'
-if (test_config[:install_type] == :package and not test_config[:skip_presuite_provisioning])
+if (test_config[:install_type] == :package and test_config[:package_build_version] and not test_config[:skip_presuite_provisioning])
```
rbrw: Might wrap this guard since if nothing else, it "goes of the side" of the gh diff window.

acceptance/helper.rb
johnduarte: The installed version value will never be 'latest'. It will be the concrete version number installed. So, setting expected version to 'latest' will always fail this comparison.

I could move the conditional so that it is easier to read:
```
if expected_version == 'latest'
  # we cannot accurately compare to 'latest'
  return
elsif installed_version != expected_version
  raise RuntimeError, "Installed version '#{installed_version}' did not match expected version '#{expected_version}'"
end
```

johnduarte:

acceptance/helper.rb
rbrw: Nevermind - I misread.  I think this is fine.

PR #2321, opened 2017-06-14
(PDB-3558) depend on puppet 5

MosesMendoza:
- Puppet Server, PDB, and Puppet will all be released at version '5' and
  require version 5 of each other. Update the dependency expressed in PDB.
  Support both 'puppet' and 'puppet-agent' in case distros ship a puppet 5 and
  PDB can depend on that.
- for PDB specifically, requiring puppet 5 means that we can rely on puppet 5
  semantics in the various puppetdb termini. This will allow us to remove some
  processing done in the terminus that is no longer necessary, possibly
  optimizing performance, but cetainly optimizing maintainability.

Signed-off-by: Moses Mendoza <moses@puppet.com>

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -1,11 +1,11 @@
 ezbake: {
   pe: {}
   foss: {
-    redhat: { dependencies: ["puppet (>= 5.0.0) | puppet-agent (>= 5.0.0)"],
+    redhat: { dependencies: ["puppet (>= 4.99.0) | puppet-agent (>= 4.99.0)"],
               preinst:  ["if rpm -q puppetdb | grep ^puppetdb-2.* > /dev/null && [ $1 -eq 2 ] ; then tar -czf /tmp/puppetdb-upgrade-config-files.tgz -C /etc/puppetdb/conf.d config.ini database.ini jetty.ini ; fi" ],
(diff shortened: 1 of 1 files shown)


MosesMendoza:
Per comment in PDB-3558, I don't think this will actually work until we tag builds of puppet-agent as 5.0.0. Not sure what the path forward is there but raising this PR so its visible.

mullr:
@MosesMendoza I changed this to >= 4.99 as we discussed

MosesMendoza:
@mullr looks like the travis job failed - expected?

mullr:
At the moment yes, fixing that elsewhere. Will rebase when it's fixed.

mullr:
Rebased on fixed master, builds should be ok now

MosesMendoza:
oops looks like i did it wrong - the rpm dependencies can't be wrapped in parens

MosesMendoza:
```
15:18:45 error: line 130: Dependency tokens must begin with alpha-numeric, '_' or '/': Requires:         puppet (>= 4.99.0) | puppet-agent (>= 4.99.0)
```

mullr:

mullr:
not sure what jenkins is up to. In any case, I previously retriggered the job set up to use the nighly repos and it passed, so I'm going to go ahead and merge this.

PR #2322, opened 2017-06-15
(PDB-3195) Default transaction isolation in connection pool config

mullr:
Before this change, our internal transaction wrapper macro was explicitly
passing a :read-committed isolation level to clojure.jdbc when not otherwise
specified. This causes a call to .setTransactionIsolation on the jdbc driver,
which executes SQL on the server. So it ends up making an extra network round
trip for each query.

This commit sets read-committed transaction isolation in the connection pool
config, so each connection already has that by default. It also allows nil
isolation levels to be passed to clojure.jdbc, which will then use whatever the
connection already has.

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -371,28 +371,33 @@
   `(retry-sql* ~n (fn [] ~@body)))
 
 (defn with-transacted-connection-fn
-  "Calls f within a transaction with the specified clojure.jdbc
-  isolation level.  If isolation is nil :read-committed will be used.
+  "Calls f within a transaction with the specified clojure.jdbc isolation level.
+  If isolation is nil, the connection pool default (read committed) is used.
   Retries the transaction up to 5 times."
   [db-spec isolation f]
-  (retry-sql 5
-             (with-db-connection db-spec
-               (with-db-transaction [:isolation (or isolation :read-committed)]
-                 (f)))))
+  ;; We've set up the connection pool to have read-committed isolation by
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -371,28 +371,27 @@
...
-  "Calls f within a transaction with the specified clojure.jdbc
-  isolation level.  If isolation is nil :read-committed will be used.
+  "Calls f within a transaction with the specified clojure.jdbc isolation level.
+  If isolation is nil, the connection pool default (read committed) is used.
   Retries the transaction up to 5 times."
   [db-spec isolation f]
   (retry-sql 5
              (with-db-connection db-spec
-               (with-db-transaction [:isolation (or isolation :read-committed)]
+               (with-db-transaction [:isolation isolation]
```
rbrw: Offhand,  wonder if it might make make sense to filter out :read-committed too, i.e. if some caller happens to specify it.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -431,7 +436,8 @@
      (doto config
        (.setJdbcUrl (str "jdbc:" subprotocol ":" subname))
        (.setAutoCommit false)
-       (.setInitializationFailFast false))
+       (.setInitializationFailFast false)
+       (.setTransactionIsolation "TRANSACTION_READ_COMMITTED"))
```
wkalt: isn't this the default?

mullr: Huh, turns out it is. I don't feel too bad about setting it here though. The real fix is to not re-set it on every query.

PR #2324, opened 2017-06-19
(PDB-3566) Fix the catch-all for command processing errors

mullr:
The top-level exception handler used in command processing threads missed most
ExceptionInfo exceptions; this includes schema validation errors.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -233,10 +233,12 @@
 
 (defmacro upon-error-throw-fatality
   [& body]
-  `(try+
+  `(try
     ~@body
-    (catch Throwable e#
-      (throw+ (fatality e#)))))
+    (catch Exception e1#
+      (throw+ (fatality e1#)))
+    (catch AssertionError e2#
+      (throw+ (fatality e2#)))))
 
 ;; ## Command submission
 
--- a/test/puppetlabs/puppetdb/command_test.clj
+++ b/test/puppetlabs/puppetdb/command_test.clj
@@ -132,6 +132,14 @@
   (queue/create-command-req "replace catalog" version certname nil "" identity
                             (tqueue/coerce-to-stream payload)))
 
+(s/defn function-that-needs-int-via-schema [x :- s/Int]
+  x)
+
+(defn function-that-needs-int-via-precondition [x]
+  {:pre [(integer? x)]}
+  x)
+
+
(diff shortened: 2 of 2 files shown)


src/puppetlabs/puppetdb/command.clj
```diff
@@ -233,9 +233,9 @@
 
 (defmacro upon-error-throw-fatality
   [& body]
-  `(try+
+  `(try
     ~@body
-    (catch Throwable e#
+    (catch Exception e#
```
rbrw: If we want to catch :pre and post here, I think we might need to catch AssertionError too (it's a Throwable...).

mullr: So what you're saying is that `(catch Throwable` is the right thing to do here?

:duck:

rbrw: Hah, nope, but I imagine that might be how Throwable got here in the first place...

mullr:
@zak-kent care to give this a look?

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -132,6 +132,9 @@
   (queue/create-command-req "replace catalog" version certname nil "" identity
                             (tqueue/coerce-to-stream payload)))
 
+(s/defn function-that-needs-int [x :- s/Int]
+  x)
+
```
rbrw: Might be worth adding a :pre or :post example/test too.

Zak-Kent:
Looks good, so far as I can tell :) 

I do have a few questions around some of the syntax used in the test. 

- I looked up the docs for the `slingshot/throw+` and it looks like a better way to catch errors with more informative messages. Is this correct? 

- Not sure on what the `~@body` part of the test on line 237 is doing. Tried to look up definitions but couldn't find any. 

- Last thing is the # mark after the e. Is this a naming convention we use for exceptions when we catch them?

Thanks for the help!

mullr:
> I looked up the docs for the slingshot/throw+ and it looks like a better way to catch errors with more informative messages. Is this correct?
yes, fsvo 'better'

> Not sure on what the ~@body part of the test on line 237 is doing. Tried to look up definitions but couldn't find any.
> Last thing is the # mark after the e. Is this a naming convention we use for exceptions when we catch them?

Both of these are special macro syntax. See http://www.braveclojure.com/writing-macros/

Zak-Kent:
Sweet! Thanks

PR #2325, opened 2017-06-20
(maint) Remove content-length warning

mullr:
We sometimes don't get the length header now, with gzip compression from
puppetserver.


doug-rosser:
👍

PR #2326, opened 2017-06-20
(RE-8457) Update to ship to puppet5 repos

melissa:


melissa:
This is dependent on https://github.com/puppetlabs/ezbake/pull/420, and will also need an ezbake bump once that change is merged and a new ezbake release is cut

melissa:
we'll also need to update ezbake in this PR once we have a new ezbake release

PR #2327, opened 2017-06-21
(PDB-3564) initial graphql endpoint setup that reroutes queries

Zak-Kent:


src/puppetlabs/puppetdb/http/handlers.clj
```diff
@@ -53,17 +55,44 @@
...
   (cmdi/wrap-routes route #(parent-check % version entity route-param-key)))
 
+(defn wrap-apply-active-node-restriction [handler]
+  (fn [req]
+    (cond
+      (some-> req :puppetdb-query :ast_only http-q/coerce-to-boolean) (handler req)
+      (= ["from" "packages"] (some->> req :puppetdb-query :query (take 2))) (handler req)
+      :else (handler (http-q/restrict-query-to-active-nodes req)))))
+
+(defn req-body->stream [req req-body-map]
```
mullr: Instead of converting from json back into a string, could read into a string and set that aside in case we need it?

src/puppetlabs/puppetdb/http/handlers.clj
```diff
@@ -53,17 +55,44 @@
...
+(defn wrap-handle-graphql-query [handler]
+  ;; Check req, if graphql redirect to new endpoint, else pass on
+  (fn [{:keys [params] :as req}]
+    (let [post-query (when (:body req)
+                       (json/parse-stream (io/reader (:body req))))]
+      (cond
+        (and post-query (= \{ (first (post-query "query"))))
+        (http/json-response {:message "This is experimental, check back later"})
+
+        post-query (handler (req-body->stream req post-query))
```
mullr: comment here about why we need to reconstruct the query would be helpful

src/puppetlabs/puppetdb/http/handlers.clj
```diff
@@ -53,17 +55,44 @@
...
+      (some-> req :puppetdb-query :ast_only http-q/coerce-to-boolean) (handler req)
+      (= ["from" "packages"] (some->> req :puppetdb-query :query (take 2))) (handler req)
+      :else (handler (http-q/restrict-query-to-active-nodes req)))))
+
+(defn req-body->stream [req req-body-map]
+  ;; Convert POST body back into stream for other handlers
+  (-> (json/generate-string req-body-map)
+      (.getBytes "UTF8")
+      (java.io.ByteArrayInputStream.)
+      (->> (assoc req :body))))
```
mullr: `->>` inside `->` is very strange... can you write this another way?

src/puppetlabs/puppetdb/http/handlers.clj
rbrw: ...also, minor, but these guards are complex enough that I might be inclined to pull the bodies onto their own lines:
```clojure
(cond
  (some-> req :puppetdb-query :ast_only http-q/coerce-to-boolean)
  (handler req)
  (= ["from" "packages"] (some->> req :puppetdb-query :query (take 2)))
  (handler req)
  :else (handler (http-q/restrict-query-to-active-nodes req)))
```

mullr: if you do that, please space them out 

```
(cond
  (some-> req :puppetdb-query :ast_only http-q/coerce-to-boolean)
  (handler req)

  (= ["from" "packages"] (some->> req :puppetdb-query :query (take 2)))
  (handler req)

  :else 
  (handler (http-q/restrict-query-to-active-nodes req)))
```

I know rob doesn't care that much about this, but I do :)

rbrw: Hah - actually, I have mixed feelings about it, and in fact had that as an additional suggestion, but decided to limit myself to the first issue.

rbrw:
OK, so once the existing comments have been resolved, the code looks good to me; I'd also be inclined to rework the commits before merge (wrt commit messages and content per-commit), i.e. once things are settled, rework them with an eye toward being useful for those trying to understand the changes later.

PR #2328, opened 2017-06-22
use HTTPS for update check (updates.puppetlabs.com)

hex2a:


mullr:

rbrw:
@ajroetker See any problem switching to https by default for update checks?

ajroetker:
@rbrw Nothing I can think of!

joshcooper:
If we make a secure SSL connection, are we sure the clojure client making the SSL connection has the root cert used to issue the SSL cert for `updates.puppetlabs.com`?

mullr:

PR #2330, opened 2017-06-26
(PDB-3564) add endpoint for GraphQL requests

Zak-Kent:


mullr:
Looks good, but we decided not to do this.

PR #2336, opened 2017-07-07
(RE-8458) Update tests to use the puppet repos

melissa:
In order to test against the right set of packages, we need to update
these tests to hit the new repos. Rather than pin to puppet 5, we're
just going to pin to the rolling puppet repos. This will change
automatically to puppet6 once we change these repos to point to puppet6.

### DIFF
--- a/acceptance/setup/pre_suite/15_setup_repos.rb
+++ b/acceptance/setup/pre_suite/15_setup_repos.rb
@@ -16,8 +16,8 @@ def initialize_repo_on_host(host, os, nightly)
         on host, "dpkg -i puppet5-nightly-release-$(lsb_release -sc).deb"
 
       else
-        on host, "curl -O http://apt.puppetlabs.com/puppetlabs-release-pc1-$(lsb_release -sc).deb"
-        on host, "dpkg -i puppetlabs-release-pc1-$(lsb_release -sc).deb"
+        on host, "curl -O http://apt.puppetlabs.com/puppet-release-$(lsb_release -sc).deb"
+        on host, "dpkg -i puppet-release-$(lsb_release -sc).deb"
       end
     else
       on host, "curl -O http://apt.puppetlabs.com/puppetlabs-release-$(lsb_release -sc).deb"
@@ -38,8 +38,8 @@ def initialize_repo_on_host(host, os, nightly)
         on host, "rpm -i puppet5-nightly-release-#{variant}-#{version}.noarch.rpm"
 
       else
-        on host, "curl -O http://yum.puppetlabs.com/puppetlabs-release-pc1-#{variant}-#{version}.noarch.rpm"
-        on host, "rpm -i puppetlabs-release-pc1-#{variant}-#{version}.noarch.rpm"
+        on host, "curl -O http://yum.puppetlabs.com/puppet/puppet-release-#{variant}-#{version}.noarch.rpm"
+        on host, "rpm -i puppet-release-#{variant}-#{version}.noarch.rpm"
       end
     else
       on host, "yum clean all -y"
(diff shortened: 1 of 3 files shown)


acceptance/setup/pre_suite/80_add_dev_repo.rb
```diff
@@ -1,4 +1,4 @@
-install_opts = options.merge( { :dev_builds_repos => ["PC1"] })
+install_opts = options.merge( { :dev_builds_repos => ["puppet"] })
```
melissa: The only thing is I don't know how beaker will handle this change, if it will do the right thing or now

melissa:
There are some things with Beaker I'm trying to figure out. Please don't merge this until I've finished with that

acceptance/setup/pre_suite/80_add_dev_repo.rb
```diff
@@ -1,4 +1,3 @@
-install_opts = options.merge( { :dev_builds_repos => ["PC1"] })
```
melissa: This option was removed in https://github.com/puppetlabs/beaker/pull/1044/commits/e074def22605b7c5acfdb28a3877ac293547436e, which looks like it was released in beaker 2.32.0, so removing this should not break anything. I'd like to see the tests run before this is merged though

resources/puppetlabs/puppetdb/benchmark/samples/facts/pg2.vm.json
```diff
@@ -9253,8 +9253,8 @@
             "gem"
         ],
         [
-            "puppetlabs-release-pc1",
-            "1.1.0-4jessie",
+            "puppet-release",
+            "1.0.0-1jessie",
```
melissa: I have no idea what this does, but it references the old release packages. I assume you want the new one...

acceptance/setup/pre_suite/15_setup_repos.rb
```diff
@@ -16,8 +16,8 @@ def initialize_repo_on_host(host, os, nightly)
         on host, "dpkg -i puppet5-nightly-release-$(lsb_release -sc).deb"
 
       else
-        on host, "curl -O http://apt.puppetlabs.com/puppetlabs-release-pc1-$(lsb_release -sc).deb"
-        on host, "dpkg -i puppetlabs-release-pc1-$(lsb_release -sc).deb"
+        on host, "curl -O http://apt.puppetlabs.com/puppet-release-$(lsb_release -sc).deb"
+        on host, "dpkg -i puppet-release-$(lsb_release -sc).deb"
```
melissa: So, this enables the new rolling repos, which are linked to the current major series out (currently puppet5). You can definitely pin to a major series, but if you set it here, these puppet repos will be updated every time we switch over to a new series. So, we'll update the link to point to puppet6 when the time comes. I'm not sure if you want this behavior or not, so I figured I'd point it out.

melissa:

melissa:
The travis failures are both `java.lang.RuntimeException: Service ':JRubyMetricsService' not found`, and look unrelated to this change

acceptance/setup/pre_suite/15_setup_repos.rb
melissa: Also, there is a beaker utility for this (`install_puppetlabs_release_repo_on`), but the current version of beaker it will break for the new release package path. I have a PR to fix that in https://github.com/puppetlabs/beaker-puppet/pull/9, but you'd also need to update your beaker version for that change

mullr:
@melissa the build failure should be fixed now; can you rebase?

melissa:
Rebased!

melissa:

melissa:

melissa:
it looks like we're having some transient errors connecting to yum.puppetlabs.com

PR #2337, opened 2017-07-07
(PDB-3611) Rewrite and/or of uniform in/extract to union/intersect 

mullr:
Postgres has a really hard time with queries like this:
```sql
  select x from t1
  where x in (select x from t2)
  and x in (select x from t3)
```
We tend to generate these a lot. But it deals better with
```sql
  select x from t1
  where x in (select x from t2
              intersect
              select x from t3)
```
which is the same. The same goes for or->union. This change does that rewriting.


mullr:
ugh, wrong initial base branch messed up the build matrix. will close and resubmit.

PR #2338, opened 2017-07-07
(PDB-3611) Improve performance of fact queries with multiple conditions

mullr:
This PR improves, in various ways, the performance of queries with multiple conditions along the lines of "fact1=val1 and fact=val2"


Zak-Kent:
+ 1 looks good so far as I can tell.

mullr:
Decided not to take this route; closing.

PR #2343, opened 2017-07-19
(PDB-3611) Deduplicate fact values (again)

mullr:
Previously, we had changed fact storage to put the value in the facts table, in
order to simplify storage and sidestep the problematic garbage collection issues
when fact values had their own table. But this proved to make queries too slow,
especially when dealing with queries which have multiple conditions on facts.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -60,7 +60,8 @@
             [puppetlabs.puppetdb.jdbc :as jdbc :refer [query-to-vec]]
             [puppetlabs.puppetdb.config :as conf]
             [puppetlabs.i18n.core :refer [trs]]
-            [puppetlabs.puppetdb.scf.hash :as hash]))
+            [puppetlabs.puppetdb.scf.hash :as hash])
+  (:import [org.postgresql.util PGobject]))
 
 (defn init-through-2-3-8
   []
@@ -1282,6 +1283,7 @@
      (str "ALTER TABLE ONLY edges"
           "  ADD CONSTRAINT edges_certname_source_target_type_unique_key"
           "  UNIQUE (certname, source, target, type)"))))
+
 (defn add-latest-report-timestamp-to-certnames []
   (jdbc/do-commands
     "DROP INDEX IF EXISTS idx_certnames_latest_report_timestamp"
@@ -1303,7 +1305,7 @@
       value_float double precision,
       value_string text,
       value_boolean boolean,
-      value jsonb
+      value jsonb not null
     );"
 
    "CREATE TABLE facts_transform (
@@ -1312,8 +1314,9 @@
        fact_value_id bigint NOT NULL
     );"
 
-   "INSERT INTO fact_values (value_type_id, value_integer, value_float, value_string, value_boolean, value)
-       SELECT distinct value_type_id, value_integer, value_float, value_string, value_boolean, value FROM facts"
+   "INSERT INTO fact_values (value, value_integer, value_float, value_string, value_boolean, value_type_id)
+       SELECT distinct value, value_integer, value_float, value_string, value_boolean, value_type_id FROM facts"
+
 
    "INSERT INTO facts_transform (factset_id, fact_path_id, fact_value_id)
        SELECT f.factset_id, f.fact_path_id, fv.id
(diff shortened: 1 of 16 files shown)


mullr:
TODO:
- [ ] Shortcut the initial redupe migration for users that haven't done it yet
- [ ] Add a migration diff test

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1281,13 +1283,100 @@
...
+   ;; values for this yet
+   "CREATE TABLE fact_values (
+      id bigint DEFAULT nextval('fact_values_id_seq'::regclass) NOT NULL,
+      value_hash bytea,
+      value_type_id bigint NOT NULL,
+      value_integer bigint,
+      value_float double precision,
+      value_string text,
+      value_boolean boolean,
+      value jsonb not null
```
wkalt: do we not allow null values? I thought that used to be ok

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1281,13 +1283,100 @@
...
+   "INSERT INTO facts_transform (factset_id, fact_path_id, fact_value_id)
+       SELECT f.factset_id, f.fact_path_id, fv.id
+         FROM facts f
+        INNER JOIN fact_values fv
+                ON fv.value_type_id = f.value_type_id
+               AND fv.value_integer IS NOT DISTINCT FROM f.value_integer
+               AND fv.value_float   IS NOT DISTINCT FROM f.value_float
+               AND fv.value_string  IS NOT DISTINCT FROM f.value_string
+               AND fv.value_boolean IS NOT DISTINCT FROM f.value_boolean
+               AND fv.value         = f.value")
```
wkalt: if null values are allowed I think this needs to be "is not distinct from" as well

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1281,13 +1283,100 @@
...
+
+  ;; populate fact_values.value_hash
+  (jdbc/call-with-query-rows
+   ["select id, value::text from fact_values"]
+   (fn [rows]
+     (doseq [batch (partition-all 5000 rows)]
```
wkalt: we need to test this to make sure it won't oom with big facts

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: I had 'is not distinct from', but that made the query take... a really long time. (I ran it overnight with the orch scale dataset, and it didn't finish). That's why I added the not-null constraint above. I believe it's valid, as we have jsonb versions of all the values.

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: ... yes. What's the biggest fact you've heard of?

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: But the current failing unit tests are telling my this isn't a true assumption.

mullr: Even making this explicit, we end up with 

```
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Insert on facts_transform  (cost=332902.86..207785761576.67 rows=1 width=24)
   ->  Hash Join  (cost=332902.86..207785761576.67 rows=1 width=24)
         Hash Cond: (f.value_type_id = fv.value_type_id)
         Join Filter: (((fv.value = f.value) OR ((fv.value IS NULL) AND (f.value IS NULL))) AND (NOT (fv.value_integer IS DISTINCT FROM f.value_integer)) AND (NOT (fv.value_float IS
DISTINCT FROM f.value_float)) AND (NOT (fv.value_string IS DISTINCT FROM f.value_string)) AND (NOT (fv.value_boolean IS DISTINCT FROM f.value_boolean)))
         ->  Seq Scan on facts f  (cost=0.00..1027808.04 rows=40920604 width=98)
         ->  Hash  (cost=319709.05..319709.05 rows=367105 width=162)
               ->  Seq Scan on fact_values fv  (cost=0.00..319709.05 rows=367105 width=162)
(7 rows)
```

Which is going to take stupid amounts of time.

src/puppetlabs/puppetdb/scf/migrate.clj
wkalt: discussed in person but I bet someone has one that's 2mb. Separately, you may be able to do something like this: https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/scf/migrate.clj#L1140,L1147 here. That will use a fetchsize of 500

wkalt: on second thought, that's probably not useful for the batched update we need here

wkalt:
generally seems sane. I'm going to work on a dataset for testing it now.

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: Updated this to handle nulls and non-nulls separately.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1281,13 +1283,121 @@
...
+               AND fv.value IS NOT DISTINCT FROM f.value
+        WHERE f.value IS NULL AND fv.value IS NULL")
+
+  ;; populate fact_values.value_hash
+  (log/info (trs "[4/6] Computing fact value hashes..."))
+  (jdbc/call-with-query-rows
+   ["select id, value::text from fact_values"]
+   (fn [rows]
+     (doseq [batch (partition-all 500 rows)]
+       (let [ids (map :id rows)
```
wkalt:
```
(map :id batch)
```

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1281,13 +1283,121 @@
...
+             hashes (map #(-> (:value %)
+                              json/parse
+                              hash/generic-identity-hash
+                              sutils/munge-hash-for-storage)
+                         rows)]
```
wkalt: batch

mullr:
Closing in favor of https://github.com/puppetlabs/puppetdb/pull/2346

PR #2344, opened 2017-07-19
(maint) Documentation markup fix

smortex:
This fix a rendering problem caused by the missing space:
https://docs.puppet.com/puppetdb/latest/configure.html#conn-keep-alive-1


wkalt:
@smortex thanks!

PR #2346, opened 2017-07-21
(PDB-3611) Deduplicate fact values (again)

mullr:
Previously, we had changed fact storage to put the value in the facts table, in
order to simplify storage and sidestep the problematic garbage collection issues
when fact values had their own table. But this proved to make queries too slow,
especially when dealing with queries which have multiple conditions on facts.


mullr:
got the branch name wrong (again)

PR #2354, opened 2017-08-07
(PDB-3639) Optimize existing-fact-values query

mullr:
This was previously done with the text version of the value_hash column; it's
stored as bytea, so it required a conversion. Change the query to compare in the
bytea domain.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -991,31 +991,31 @@
                                 :fact_path_id pid
                                 :fact_value_id vid})))
 
-(defn existing-row-ids
+(defn existing-fact-value-ids
   "Returns a map from value to id for each value that's already in the
    named database column."
-  [table column values]
-  (into {}
-        (for [{:keys [value id]}
-              (query-to-vec
-               (format
-                "WITH values AS (SELECT unnest(?) as vals)
-                   SELECT %s AS value, id FROM %s
-                   INNER JOIN values ON values.vals = %s"
-                column table column)
-               (sutils/array-to-param "text" String values))]
-          [value id])))
+  [vhashes]
+  (let [vhashes-param (sutils/array-to-param "bytea" PGobject
+                                             (map sutils/munge-hash-for-storage vhashes))
+        rows (query-to-vec "WITH new_hashes AS (SELECT unnest(?) as h)
+                              SELECT id, encode(value_hash::bytea, 'hex') value_hash FROM fact_values
+                              INNER JOIN new_hashes ON new_hashes.h = value_hash"
+                          vhashes-param)]
+    (->> rows
+         (map (fn [{:keys [value_hash id]}]
+                [value_hash id]))
+         (into {}))))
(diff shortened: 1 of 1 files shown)


mullr:
Fact storage benchmark results
Before: 
    medium: 123525.899966 ms
After:
    medium: 87886.204174 ms

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -991,31 +991,31 @@
...
-               (sutils/array-to-param "text" String values))]
-          [value id])))
-
-(defn realize-records!
+  [vhashes]
+  (let [vhashes-param (sutils/array-to-param "bytea" PGobject
+                                             (map sutils/munge-hash-for-storage vhashes))
+        rows (query-to-vec "WITH new_hashes AS (SELECT unnest(?) as h)
+                              SELECT id, encode(value_hash::bytea, 'hex') value_hash FROM fact_values
+                              INNER JOIN new_hashes ON new_hashes.h = value_hash"
```
wkalt: I think we want value_hash on the RHS here to use the index.

wkalt: LHS rather

wkalt:
what's the exact test you're using?

src/puppetlabs/puppetdb/scf/storage.clj
wkalt: I think it's true that the index won't get used if value_hash is on the RHS, but in practice it doesn't seem to be getting used in either case on the limited set I'm looking at. May be worth digging around to confirm that we can use it.

wkalt: poking a little more it looks like the order shouldn't matter, so I retract everything except that we should try and get confirmation that it'll use an index sometimes.

mullr:
@wkalt these timings are https://github.com/puppetlabs/puppetdb/pull/2355, using the 'medium' profile.

src/puppetlabs/puppetdb/scf/storage.clj
mullr:
```
                                                                  QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=0.93..839.51 rows=100 width=40) (actual time=0.033..0.049 rows=3 loops=1)
   CTE new_hashes
     ->  Result  (cost=0.00..0.51 rows=100 width=32) (actual time=0.006..0.008 rows=3 loops=1)
   ->  CTE Scan on new_hashes  (cost=0.00..2.00 rows=100 width=32) (actual time=0.007..0.010 rows=3 loops=1)
   ->  Index Scan using fact_values_value_hash_key on fact_values  (cost=0.42..8.36 rows=1 width=29) (actual time=0.008..0.008 rows=1 loops=3)
         Index Cond: (value_hash = new_hashes.h)
 Planning time: 0.141 ms
 Execution time: 0.075 ms
(8 rows)
```

mullr:

PR #2356, opened 2017-08-14
(PDB-3641) Standardize package GC command on gc_packages

scottyw:


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -168,10 +168,10 @@
     (catch Exception e
       (log/error e (trs "Error while sweeping reports")))))
 
-(defn gc-packages! [ttl db]
+(defn gc-packages! [db]
```
mullr: what happened here? Did we completely mess up when calling this?

scottyw: It looks like it was an arity problem. I figure it wasn't ever seen because the ```when``` that calls this function didn't trigger.

scottyw:
CLA signed.

PR #2358, opened 2017-08-17
(maint) Update EZBake to 1.5.1

aperiodic:
This version of EZBake stores the build metadata (including the deployed
snapshot version and fully-resolved versions of all snapshot
dependencies) in a persistent location, rather than one that gets
deleted after two weeks.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -174,7 +174,7 @@
                                                [puppetlabs/puppetdb ~pdb-version]
                                                [org.clojure/tools.nrepl nil]]
                       :name "puppetdb"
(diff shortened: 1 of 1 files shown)


aperiodic:
I tested this by creating a build job with `lein with-profile ezbake ezbake build` and observing that it completed successfully. According to [the EZBake changelog](ez-change) the only difference between these two versions is the generation of this metadata.

[ez-change]:https://github.com/puppetlabs/ezbake/blob/master/CHANGELOG.md#151---2017-7-13

PR #2359, opened 2017-08-23
(PDB-3643) Fix fact dedupe migration with null values

mullr:
In the flat fact storage code, null fact values were stored as sql null instead
of json null. Update the migration code to account for this.


mullr:
Gonna merge this even with the integration tests failing, as we need to get this fix promoted. a separate pr is up which will hopefully fix those tests.

PR #2364, opened 2017-09-01
(PDB-3648) Rewrite null json chars via jackson for pg

rbrw:


test/puppetlabs/puppetdb/cheshire_test.clj
```diff
@@ -68,3 +69,9 @@
       (spit-json json-out (sorted-map "a" 1 "b" 2))
       (is (= "{\n  \"a\" : 1,\n  \"b\" : 2\n}"
              (slurp json-out))))))
+
+(deftest test-null-replacing-encoder
+  (binding [*json-factory* null-replacing-json-factory]
+    (are [json src] (= json (generate-string src))
+         "\"\\ufffd\"" "\u0000"
+         "[\"\\ufffd\"]" ["\u0000"])))
```
mullr: Can you add a test to verify the particular problem we saw, i.e. to check that it passes through the text "\u0000" (where the backslash is properly escaped) without changing it?

rbrw: Oh, of course.

PR #2368, opened 2017-09-25
(maint) Update to latest ezbake

underscorgan:


underscorgan:
Tests failed because we're in a chicken-and-egg situation with the pipeline PR at https://github.com/puppetlabs/ci-job-configs/pull/3465. That needs the update to ezbake to work, but also changes puppetdb pipelines to be called in a more standard way (not calling `lein build` directly), which is needed for the ezbake updates.

underscorgan:
Tests failed because we're in a chicken-and-egg situation with the pipeline PR at https://github.com/puppetlabs/ci-job-configs/pull/3465. That needs the update to ezbake to work, but also changes puppetdb pipelines to be called in a more standard way (not calling `lein build` directly), which is needed for the ezbake updates.

mullr:
Will get this via mergeups from 5.1.x

PR #2379, opened 2017-10-13
(maint) Update version determination

underscorgan:
With the move to the new ezbake packaging, we've fixed a long-standing
bug where we had packages with the same name but different content as we
didn't include the codename in the release field. This however breaks
the method of puppetdb installations, so update the version
determination.

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -298,8 +298,12 @@ module PuppetDBExtensions
     elsif host['platform'].include?('fedora')
       version_tag = host['platform'].match(/^fedora-(\d+)/)[1]
       "#{version}.fc#{version_tag}"
-    elsif host['platform'].include?('ubuntu') or host['platform'].include?('debian')
-      "#{version}puppetlabs1"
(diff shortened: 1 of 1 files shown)


underscorgan:
Testing and passing with the experimental pipeline [here](https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/view/__experimental%20automatic/job/experimental_auto_puppetdb_init-multijob_githubpr-5.1.x/3/)  deployed based on [this](https://github.com/puppetlabs/ci-job-configs/pull/3465) pull request.

Unfortunately with this one we're in a chicken and egg situation that I don't see any easy way around: merging the ci change will break all tests on deb/ubuntu until this change goes in, and merging this without the ci change will also break all tests on deb/ubuntu. Best bet is probably just pick a time a merge them all at once, does that sound reasonable @mullr ?

Also, guessing this fix needs to go into 4.2.3.x and 4.4.x, and then master will get a merge from 5.1.x?

PR #2380, opened 2017-10-16
(PDB-3692) Handle another variant of null value duplication

mullr:


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1374,21 +1374,26 @@
...
+                                 (-> (jdbc/query-to-vec
+                                      "insert into fact_values (value_type_id, value) values (4, null) returning id")
+                                     first
+                                     :id)))
+        bad-null-value-ids (->> (jdbc/query-to-vec "select id from fact_values
+                                                    where (value_type_id = 5 and value = 'null')
+                                                       or (value_type_id = 5 and value is null)
+                                                       or (value_type_id = 4 and value = 'null')")
+                                (map :id))]
+    (doseq [id bad-null-value-ids]
```
rbrw: Haven't though this through carefully, but wondered if we might be able to use an `update from` or subselect update to let pg do the work, and avoid having to worry about the size of the change set.

mullr: This is post-deduplication, so there will be 3 rows max.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1374,21 +1374,32 @@
...
-    (if (and json-null-value-id real-null-value-id)
-      (do
-        (jdbc/do-prepared "update facts_transform set fact_value_id = (?) where fact_value_id = (?)"
-                          [real-null-value-id json-null-value-id])
-        (jdbc/do-prepared "delete from fact_values where id = (?)"
-                          [json-null-value-id]))
-      (log/info (trs "  ... none found"))))
+  ;; only do this if the DB has some null values
+  (when (pos? (-> (jdbc/query-to-vec "select count(*) from fact_values
+                                      where value_type_id in (4,5)
```
wkalt: is this usage of in actually valid?

wkalt: huh, guess so

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1374,21 +1374,32 @@
...
+                                      and value in (null, 'null')")
```
wkalt: I think this won't work due to null comparison issues

```
puppetdb=# select * from fact_values where value_integer in (null);                                                                                                                                                                           
 id | value_hash | value_type_id | value_integer | value_float | value_string | value_boolean | value                                                                                                                                         
----+------------+---------------+---------------+-------------+--------------+---------------+-------                                                                                                                                        
(0 rows)                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                              
Time: 0.626 ms
```

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1374,21 +1374,32 @@
...
+   (let [real-null-value-id (let [existing-id (some-> (jdbc/query-to-vec
+                                                       "select id from fact_values
+                                                       where value_type_id=4 and value is null")
+                                                      first
+                                                      :id)]
+                              (or existing-id
+                                  (-> (jdbc/query-to-vec
+                                       "insert into fact_values (value_type_id, value) values (4, null) returning id")
+                                      first
+                                      :id)))
```
wkalt: seems like existing-id/real-null-value-id would be nicer as two separate bindings in the top-level let

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1374,21 +1374,32 @@
...
+         bad-null-value-ids (->> (jdbc/query-to-vec "select id from fact_values
+                                                    where (value_type_id = 5 and value = 'null')
+                                                       or (value_type_id = 5 and value is null)
+                                                       or (value_type_id = 4 and value = 'null')")
+                                 (map :id))]
+     (doseq [id bad-null-value-ids]
+       (jdbc/do-prepared "update facts_transform set fact_value_id = (?) where fact_value_id = (?)"
+                         [real-null-value-id id])
+       (jdbc/do-prepared "delete from fact_values where id = (?)"
+                         [id]))))
```
wkalt: I think we can make this run faster by using an array of ids and doing the operations in bulk.

```
update facts_transform set fact_value_id=? where fact_value_id = any(Array[bad-null-ids]);
delete from fact_values where id = any(Array[bad-null-ids])
```

(you'll need the storage-utils array functions to get the array of integers).

src/puppetlabs/puppetdb/scf/migrate.clj
wkalt: I believe this will violate an fk constraint if the install has never seen a null-valued fact

wkalt: ah nvm, last comment is handled

wkalt: actually, it isn't. If you get into this code because you have a string-valued null in a json type, I think this will still violate a constraint

wkalt: forget the last three comments -- the value types are hardcoded in the migration

wkalt:
@mullr this seems reasonable to me. Probably want to fix the null comparison issue and add a test for that case. I'm assuming the situation is such that changing an existing migration won't cause issues for anyone.

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: ARGH good catch

src/puppetlabs/puppetdb/scf/migrate.clj
mullr: There are only ever up to three ids here, so perf shouldn't be much of an issue.

PR #2388, opened 2017-10-24
(PE-22668) Add job_id from the latest report to nodes endpoint.

kbrezina:
In order to be able to display the job-id successfully on the PE Overview table, the job-id needs to be available in the pdb payload when a given report was the result of running from the orchestrator.


mullr:
Can you add a query test for the positive case, i.e. testing that job-id is actually available when it's not null?

kbrezina:
@mullr Just added the missing test.

PR #2389, opened 2017-10-25
(PDB-3734) Ensure that resource indirection uses existing resource type

thallgren:
Before this commit, the resource type would be passed to `Resource#new`
in string form. This is problematic since it may result in the load of
a Ruby resource type which then is in conflict with genereated Pcore
resource type.

This commit ensures that the indirection uses the same logic as Puppet
when creating new resources.


mullr:
Can't claim to understand what's going on with the guts of puppet here, but I'll take your word for it. Does this need to be coordinated with any PRs on the puppet side?

thallgren:
No, there are no PRs on the puppet side. There's a ticket PUP-8002 but this PR is the only thing needed to fix it.

mullr:
:+1: thanks

PR #2390, opened 2017-10-25
(maint) update 5.1.2 release notes

Zak-Kent:


documentation/release_notes.markdown
```diff
@@ -16,6 +16,22 @@ canonical: "/puppetdb/latest/release_notes.html"
 [stockpile]: https://github.com/puppetlabs/stockpile
 [queue_support_guide]: ./pdb_support_guide.html#message-queue
 
+5.1.2
+-----
+
+PuppetDB 5.1.2 is a bugfix and performance release.
```
rbrw: We might well have used this construction before, but might be better to rewrite "performance release" somehow.

documentation/release_notes.markdown
```diff
@@ -16,6 +16,22 @@ canonical: "/puppetdb/latest/release_notes.html"
...
+
+### Bug Fixes
+* This release containes a fix for a uniqueness bug in the recent fact data migration.
```
rbrw: I'd leave off "This release", and also maybe match the other entry, i.e. "Fixed a ..." to match "Added a ...".  Though I might go even further and write both in NEWS style, i.e. "A recent fact data migration should no longer crash when the existing data has unexpected null value representations."

documentation/release_notes.markdown
rbrw: As one possible example "PuppetDB 5.1.2 includes bugfixes and performance improvements".

PR #2392, opened 2017-10-27
 (PDB-3738) Add compound report id index

mullr:
We've encountered some problematic queries when doing historical scans over
reports. This adds an index targeted at those queries.


wkalt:
as discussed, feels like this goes in extensions

PR #2393, opened 2017-11-03
(PDB-3745) Increase autovacuum vacuum scale factor

npwalker:
Prior to this commit, the factsets, catalogs, and certnames tables
were autovacuumed far too frequently.

This commit changes the percentage change before the table
is considered for autovacuuming.  The way these tables are
inserted and updated apparently moves the counter for changes
forward when the rows can't be reclaimed.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1496,11 +1496,12 @@
 
   {::vacuum-analyze #{"factsets"}})
 
-(defn autovacuum-vacuum-scale-factor-factsets-catalogs-certnames []
+(defn autovacuum-vacuum-scale-factor-factsets-catalogs-certnames-reports []
   (jdbc/do-commands
     "ALTER TABLE factsets  SET ( autovacuum_vacuum_scale_factor=0.80 );"
     "ALTER TABLE catalogs  SET ( autovacuum_vacuum_scale_factor=0.75 );"
-    "ALTER TABLE certnames SET ( autovacuum_vacuum_scale_factor=0.75 );"))
+    "ALTER TABLE certnames SET ( autovacuum_vacuum_scale_factor=0.75 );"
+    "ALTER TABLE reports   SET ( autovacuum_vacuum_scale_factor=0.01 );"))
 
 (def migrations
   "The available migrations, as a map from migration version to migration function."
@@ -1546,7 +1547,7 @@
    64 rededuplicate-facts
    65 varchar-columns-to-text
    66 jsonb-facts
-   67 autovacuum-vacuum-scale-factor-factsets-catalogs-certnames})
+   67 autovacuum-vacuum-scale-factor-factsets-catalogs-certnames-reports})
 
 (def desired-schema-version (apply max (keys migrations)))
 
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1496,6 +1496,12 @@
 
   {::vacuum-analyze #{"factsets"}})
 
+(defn autovacuum-vacuum-scale-factor-factsets-catalogs-certnames []
+  (jdbc/do-commands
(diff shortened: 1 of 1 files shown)


npwalker:
Not sure how to add tests that would make sure that future migrations don't accidentally remove these autovacuum settings.  But it'd probably use a query like this to confirm the settings are there. 

https://github.com/npwalker/pe_databases/blob/master/manifests/set_table_attribute.pp#L10

npwalker:
@mullr it looks like all PRs are failing tests... is that an underlying issue with tests?  or is there actually something wrong with my PR?

mullr:
@npwalker iirc @wkalt had some misgivings on this; can we summarize those here?

npwalker:
I believe Wyatt's objection can be summed up as "the PostgreSQL docs say that generally speaking you want to vacuum more often not less often, how do we know that reducing the scale factor won't make things worse?"  

Based on observations and making these changes at customers, I feel confident that reducing the scale factor causes a reduction in CPU use, reduces your chances of maxing out your autovacuum_workers, and does not cause too many dead tuples to pile up.

wkalt:
@mullr yes, @npwalker has summarized my issues. For reference I was talking about the postgres scaling guide here: https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server .

I think we need to replicate the underlying autovacuum issue in-house (in a way that allows us to adjust the dimensions too), and see that the fix works in that environment. None of that is dismissive of the fix, to be clear, it's just the only way we can be confident.

npwalker:
@mullr I rebased this to resolve the merge conflict.  Is there anything I can do to help move this forward?

dylanratcliffe:
@wkalt How is progress going on replicating this? I've just been at an installation with horrific vacuum issues so if you need some advice on how to replicate I think I can be of help. The settings @npwalker has suggested are the default in the [pe_databases](https://forge.puppet.com/npwalker/pe_databases) module which has over 12,000 downloads, combining that with the fact that @npwalker is arguably the most experienced person on this subject of anyone in the world, I think this is ready for merge (assuming the testing hasn't shown anything to the contrary).

er0ck:
@rbrw, have any of the current team looked into this?  especially interesting given @dylanratcliffe 's recent email.  thoughts?

rbrw:
@er0ck no, we haven't yet, but generally speaking, I agree with @wkalt's position, i.e. that ideally we'd like to have the time/resources to reproduce the problem in-house first.  I don't think we're quite ready for that yet, but was hoping to pursue this after we finish some of the current priorities.

npwalker:
@rbrw I think we should weigh the pros and cons of making this change without an in-house reproduction.  I know I saw these same issues when I did scale testing on PE 2015.3 and I've seen them at numerous customers.  

In CS we recommend use of the pe_databases module which implements these same changes so even if these changes we going to cause harm somehow we are already causing that harm at our largest most performance sensitive customers.  

I definitely think we should investigate this further in-house but I don't see why not having an in-house reproduction should preclude inclusion in the product.  It's unclear to me what negative affect we're trying to avoid by waiting on this change.

dylanratcliffe:
@rbrw I'm with @npwalker on this. While imperative testing data is much better than trust/anecdotal evidence (and maybe once we have a proper setup internally we'll be able to improve these values further) the risk/reward ratio for this change should certainly sufficient to get it merged, even if it is a stopgap solution in lieu of better testing.

rbrw:
To clarify - I'm not proposing that we defer this indefinitely (I know it has languished while puppetdb development dramatically slowed for the past year or more).  In any case, my intention is for us to pursue this once we're a bit further along in the process of restoring the resources/people available to puppetdb, and we finish handling some other immediate work.  

At that point, I'd like for at least two of us to make sure we understand the broader context, refresh ourselves on the current state of the pg vacuum machinery, and then get together with you to discuss what you've found in more detail.  After that, we should be ready to proceed (and fwiw, I do suspect we'll want these, or related changes).

While I'm not certain yet, barring any unforeseen escalations, I'd hoped we'd get to that point in the next few months if not sooner (given the holidays) with an eye toward possibly including the changes in the next Y release.

PR #2399, opened 2017-11-13
(MAINT) Enable benchmarking via clojure

HairyMike:
I've moved the `validate-cli` call to the `main` method of the benchmark namespace.

This allows me to call the benchmark function from elsewhere and pass in my options map.  With this change, I can pass an already loaded config without having to read anything from disk.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/benchmark.clj
+++ b/src/puppetlabs/puppetdb/cli/benchmark.clj
@@ -477,8 +477,8 @@
   pass nummsgs), and :stop, function to request termination of the benchmark
   process and and wait for it to stop cleanly. These functions return true if
   shutdown happened cleanly, or false if there was a timeout."
-  [args]
-  (let [{:keys [config rand-perc numhosts nummsgs threads] :as options} (validate-cli! args)
+  [options]
+  (let [{:keys [config rand-perc numhosts nummsgs threads] :as options} options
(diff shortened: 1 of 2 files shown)


mullr:
Hm. The functionality makes sense, but I don't think adding extra arities is the right approach. I've found passing around an option map to be a more sustainable approach. Can you make the changes you need by changing changing the way the args map works? For example, maybe you could add a `:facts-data` key, which would be used instead of loading facts from disk.

HairyMike:
Thanks for the suggestion @mullr.  I'll take a look in the morning.  

Just to be clear, its the trapper-keeper config I want to pass here, not the facts/node or report data.  I'm happy for now that they remain as paths.  (I did want to have them loaded from the classpath, which this doesn't work for me at the moment - I might revisit this in a later PR).

HairyMike:
Hey @mullr,

Do you have an example of any prior art where you've passed the map in an arg vector?  Did you use the kitchensink/cli function for validating the args?  At first glance it looks like it might be tricky.

HairyMike:
Thats far cleaner @mullr,

thanks for the pointer :)

mullr:
Looks like `puppetlabs.puppetdb.cli.benchmark-test` is broken; you may need to make an intermediate function for the test to call into.

HairyMike:
Thanks for that @mullr,

Looks like its working now.

PR #2407, opened 2017-12-06
(PDB-3723) jsonb fact storage

wkalt:
Store factsets as jsonb blobs.


wkalt:
will reopen off fork so CI doesn't do something weird

PR #2410, opened 2017-12-06
(PDB-3723) jsonb fact storage

wkalt:
Store factsets as jsonb blobs.


wkalt:

src/puppetlabs/puppetdb/factsets.clj
```diff
@@ -52,11 +53,13 @@
...
-(pls/defn-validated facts-expanded->wire-v5 :- {s/Keyword s/Any}
+(pls/defn-validated facts-expanded->wire-v5-classic :- {s/Keyword s/Any}
   [facts :- facts-expanded-query-schema]
   (facts-list-to-map
-   (map fact-query->wire-v5
-        (:data facts))))
+   (map fact-query->wire-v5 (:data facts))))
+
+(defn facts-expanded->wire-v5 [facts]
+  (facts-expanded->wire-v5-classic facts))
```
ajroetker: What's the difference between classic and not classic here?

wkalt: I think that's detritus from the initial commits where both forms were supported. I'll consolidate them.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -67,20 +67,112 @@
...
 
+(defn- insert-multi-row
+  "Given a table and a list of columns, followed by a list of column
+  value sequences, return a vector of the SQL needed for the insert
+  followed by the list of column value sequences. The entities
+  function specifies how column names are transformed."
+  [table columns values {:keys [on-conflict] :as opts}]
+  (let [nc (count columns)
+        vcs (map count values)]
+    (if (not (and (or (zero? nc) (= nc (first vcs))) (apply = vcs)))
```
ajroetker: Am I reading this right?
If nc == 0 and vcs are equal, then this won't throw an IllegalArgException? Is that intended?

This might be easier to read as (if I'm understanding the intention)
```
(if (apply = (conj vcs nc)) ...) 
```
I'm not really sure what the (zero? nc) is trying to capture though

wkalt: The new jdbc functions in here are ripped off from https://github.com/clojure/java.jdbc/blob/master/src/main/clojure/clojure/java/jdbc.clj#L1392 . Basically the private functions needed to be modified for on-conflict support. I'll add comments to make that explicit.

Assuming there's no bugs in them I think it's probably better to let them be.

wkalt: there might be a larger question as to whether we want to introduce on-conflict with this work. It's only used in storage/realize-paths, and it would be feasible to do two roundtrips like we previously have.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1168,6 +949,75 @@
...
+
+(defn load-stable-facts [factset-id]
+  (-> (jdbc/query-to-vec "select stable from factsets where id=?"
+                         factset-id)
+      first
+      :stable
+      str
+      json/parse-string))
+
+(def ps-chunksize 6000)
```
ajroetker: A comment here might be useful

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1536,20 +1425,14 @@
...
-   (jdbc/with-transacted-connection' db :serializable
-     (delete-orphaned-paths! *orphaned-path-gc-limit*))
-   (jdbc/with-transacted-connection' db :serializable
-     (delete-orphaned-values! *orphaned-value-gc-limit*))))
+   (do
+     (jdbc/with-transacted-connection db
+       (delete-unassociated-params!)
+       (delete-unassociated-environments!))
+     (jdbc/with-transacted-connection db
+       (delete-unused-fact-paths)))))
```
ajroetker: Do these need to be in separate transactions for a reason?

rbrw: No, I put it in a separate transaction because it didn't need to be coordinated, and assumed that might reduce the potential for contention.

PR #2411, opened 2017-12-06
(PDB-3695) speed up package gc

wkalt:
"delete where not in" is a bad query in postgres. See discussion on SO for
alternatives:
https://stackoverflow.com/questions/15959061/delete-records-which-do-not-have-a-match-in-another-table

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -836,7 +836,7 @@
    (:gc-packages performance-metrics)
    (jdbc/delete!
     :packages
-    ["id NOT IN (SELECT package_id from certname_packages)"])))
+    ["not exists (select * from certname_packages cp where cp.package_id = packages.id)"])))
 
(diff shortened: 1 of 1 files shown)


wkalt:
Testing:

1. Load the database with the benchmark tool
```
lein run benchmark -c postgres.ini -n 2000 -N 1 -r 20
```
2. Create packages to GC
```
puppetdb=# delete from certname_packages where package_id in (select id from packages where random() < 0.05);
DELETE 92854
```
3. Test that old method is slow
```puppetdb=# delete from packages where id not in (select package_id from certname_packages);
DELETE 137
Time: 23146.516 ms
```

4. Orphan more packages
```
puppetdb=# delete from certname_packages where package_id in (select id from packages where random() < 0.05);
DELETE 99228
Time: 495.875 ms
```

5.  Test new method is fast
```
puppetdb=# delete from packages p where not exists (select * from certname_packages cp where cp.package_id = p.id);
DELETE 144
Time: 144.976 ms

```

6. Test old method is still slow
```
puppetdb=# delete from certname_packages where package_id in (select id from packages where random() < 0.05);
DELETE 81936
Time: 262.687 ms
puppetdb=# delete from packages where id not in (select package_id from certname_packages);
DELETE 123
Time: 17433.386 ms
```

PR #2412, opened 2017-12-08
(PDB-3311) Improve error handling for index row size exception

wkalt:
Sometimes users make a puppet error where they try to extract a fact
value with '$facts["foo"]' instead of '${facts["foo"]}' during string interpolation, which causes the
entire factset to be embedded in the catalog. If the factset is large,
this results in data that is too big for a postgres btree index, which
produces a confusing error message.

Intercept those error messages and print something more comprehensible
and divert those messages straight to the DLO, since retrying them
will always fail.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -43,6 +43,7 @@
             [puppetlabs.puppetdb.jdbc :as jdbc :refer [query-to-vec]]
             [puppetlabs.puppetdb.time :refer [to-timestamp]]
             [honeysql.core :as hcore]
+            [slingshot.slingshot :refer [throw+]]
             [puppetlabs.i18n.core :refer [trs]]
             [puppetlabs.puppetdb.package-util :as pkg-util]
             [puppetlabs.puppetdb.cheshire :as json])
@@ -482,7 +483,7 @@
 (s/defn insert-records*
   "Nil/empty safe insert-records, see java.jdbc's insert-records for more "
   [table :- s/Keyword
-   record-coll :- [{s/Keyword s/Any}]]
+   record-coll]
   (jdbc/insert-multi! table record-coll))
 
 (s/defn add-params!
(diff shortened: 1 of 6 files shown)


wkalt:
To verify this manually,

```
dd if=/dev/urandom bs=1M count=1 | base64 > /tmp/output.txt
```

manifest:
```
$ntitle = file("/tmp/output.txt")

notify {$ntitle:
  message => "yo",
}
```

Running the agent with the above will cause PDB to log this line:
```
2017-12-08 08:16:35,868 ERROR [p.p.s.storage] Failed to insert resource for wyatt-work (file: /home/wyatt/.puppetlabs/etc/code/environments/production/manifests/init.pp, line: 3): value too large for btree index. This usually results from
a code error where $facts['my_fact'] is used instead of ${facts['my_fact']}
```

and the message will be kicked to the DLO on attempt 0.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -519,30 +520,45 @@
...
+    (let [record (atom nil)]
+      {:pre [(every? resource-ref? refs-to-insert)]}
+      (update! (:catalog-volatility performance-metrics) (count refs-to-insert))
+      (try
+        (insert-records*
+          :catalog_resources
+          (map (fn [resource-ref]
+                 (let [{:keys [type title exported
+                               parameters tags file line] :as resource} (get refs-to-resources resource-ref)]
+                   (reset! record resource)
```
wkalt: Q: what do we think of this approach for ensuring we have the file/line of the failing resource?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -479,7 +480,7 @@
 (s/defn insert-records*
   "Nil/empty safe insert-records, see java.jdbc's insert-records for more "
   [table :- s/Keyword
-   record-coll :- [{s/Keyword s/Any}]]
+   record-coll]
```
wkalt: this was realizing the seq

wkalt:
This needs to get hooked into the event storage that happens with reports as well.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -519,30 +520,51 @@
...
+
+(defn handle-resource-insert-exception
+  "Handle a java.sql.SQLException from insertion of a resource. This may occur
+   when the inserted resource has a value too big for a postgres btree index."
+  [e certname file line]
+  (case (.getSQLState e)
+    "54000" (do (log/errorf "Failed to insert resource for %s (file: %s, line: %s): value too large for btree index. This usually results from a code error where $facts['my_fact'] is used instead of ${facts['my_fact']} inside a string"
+                            certname file line)
+                (throw+ (utils/fatality "value too large")))
+    :else (throw+ e)))
```
rbrw: Could this just be a regular throw, probably didn't mean the :else for case?

rbrw:
Haven't tested it locally, but other than the one comment above, looks reasonable to me.

PR #2420, opened 2017-12-13
(PDB-3687) Support ordering on fact values

wkalt:
Support order by on fact values, now that we're using jsonb.


robdaemon:
Replaced by #2908

PR #2452, opened 2018-02-05
(PDB-3830) Decrement queue depth when discarding a message

npwalker:
Prior to this commit, when a message is discarded it had no effect
on the command queue depth metrics.

After this commit, when a message is discarded we decrement the
queue depth metric.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -473,7 +473,9 @@
   (-> cmdref
       (queue/cons-attempt ex)
       (dlo/discard-cmdref q dlo))
-  (mark-both-metrics! (:command cmdref) (:version cmdref) :discarded))
+  (let [{:keys [command version]} cmdref]
+    (mark-both-metrics! command version :discarded)
+    (update-counter! :depth command version dec!)))
 
 (defn process-delete-cmdref
   "Processes a command ref marked for deletion. This is similar to


npwalker:
I tested this out by stuffing garbage into puppetdb which it then has to discard.  

```
curl -X POST   -H 'Content-Type:application/json'   -H 'Accept:application/json'   -d '$(cat /tmp/28-1518049341123_facts_5_pe-metrics-puppet-master.json)'   "http://localhost:8080/pdb/cmd/v1?command=replace_facts&version=5&certname=pe-metrics-puppet-master"
```

Doing so with the patch applied shows that the queue does not grow with discards.  That's the first part of the below graph.  I think revert the patch and you can see the command queue does grow when there are discards.  

image

src/puppetlabs/puppetdb/command.clj
```diff
@@ -473,7 +473,9 @@
   (-> cmdref
       (queue/cons-attempt ex)
       (dlo/discard-cmdref q dlo))
-  (mark-both-metrics! (:command cmdref) (:version cmdref) :discarded))
+  (mark-both-metrics! (:command cmdref) (:version cmdref) :discarded)
+  (let [{:keys [command version]} cmdref]
+    (update-counter! :depth command version dec!)))
```
rbrw: This looks good, but let's pull `mark-both-metrics!` into the `let`, so it can use the `command` and `version` bindings too.

rbrw:
I suspect we'll want it on stable, master, and LTS (if it applies), though we might well just cherry-pick it onto the latter.

npwalker:
@rbrw updated but idk why the tests are failing

rbrw:

rbrw:
@npwalker could you rebase this?  Hopefully we've fixed the test issues.

npwalker:
@rbrw rebased

rbrw:
This looks good.  Thanks much.

PR #2453, opened 2018-02-07
(maint) Add project name to build defaults

mwaggett:
This commit adds 'puppetdb' as the project name. Previously, when attempting to ship tarballs, the packaging repo was unable to find any retrieved tarballs because it expected them in the form <project>-<version>.tar.gz, but there was no project set.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -1,4 +1,5 @@
 ---
 packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
 packaging_repo: 'packaging'
+project: 'puppetdb'
 build_tar: FALSE


ext/build_defaults.yaml
```diff
@@ -1,4 +1,5 @@
 ---
 packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
 packaging_repo: 'packaging'
+project: 'puppetdb'
 build_tar: FALSE
```
underscorgan: should this line also be removed?

mwaggett: @underscorgan yes, but I want to get my packaging fix in first.

mwaggett: which only exists in my brain atm

PR #2454, opened 2018-02-08
(maint) Add project name to build defaults

mwaggett:
This commit adds 'puppetdb' as the project name. Previously, when attempting to ship tarballs, the packaging repo was unable to find any retrieved tarballs because it expected them in the form <project>-<version>.tar.gz, but ther was no project set.


mwaggett:
@mullr I'm merging this since tests passed, it has two thumbs, and the same as what you merged into 4.2.3.x yesterday.

PR #2455, opened 2018-02-08
(PDB-3840) Clean up additional orphaned facts

suckatrash:
Prior to this commit the facts table could still contain orphaned facts if an entry on the facts table had no reference in the fact_paths or fact_values table.

After this commit we remove any facts that don't have a matching id on the fact_paths table and the fact_values table.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1315,7 +1315,9 @@
 (defn rededuplicate-facts []
   (log/info (trs "[1/8] Cleaning up unreferenced facts..."))
   (jdbc/do-commands
-   "DELETE FROM facts WHERE factset_id NOT IN (SELECT id FROM factsets)")
+   "DELETE FROM facts WHERE factset_id NOT IN (SELECT id FROM factsets)"
+   "DELETE FROM facts WHERE fact_path_id NOT IN (SELECT id FROM fact_paths)"
+   "DELETE FROM facts WHERE fact_value_id NOT IN (SELECT id FROM fact_values)")
 
   (log/info (trs "[2/8] Creating new fact storage tables..."))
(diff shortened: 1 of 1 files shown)


suckatrash:
Looks like I need to add / modify some tests.  I probably need some guidance here.

rbrw:
Sorry this has languished for so long.  I've been taking a look and the code does rebase fine onto 5.2.x (which is where we'd probably want it if we decide to proceed), but it fails because at that point in the migrations, there is no fact_values table.  And out of curiosity, do you happen to recall if you ran in to orphaned fact paths at this point in the migrations, or was it more a theoretical concern?

In pdb 5.1+, which stores facts via jsonb facts, (some) phantom/orphaned fact paths are expected since the fact_paths table there is just an approximate cache of the paths represented by all of the factsets.

rbrw:
Ahh, I misunderstood -- I see that this is trying to delete facts with orphaned paths, which is what we'd want, and I found the errors you'd recorded previously.  So I think that the first delete here makes sense.

As mentioned, the second one won't work at this point in the migrations because fact_values doesn't exist, so I'm wondering if either we don't need it, or if it's possible this fix isn't quite in the right migration, i.e. that table would exist in an earlier migration (before the one that dededuplicated fact values), or in a later migration (when we reversed that change).

In any case, from the logs, the first delete is clearly fixing that problem in the right place, so we should include it now, and I  we can handle the second delete separately, once we're sure we need it and we know where it should go.

rbrw:
I'll handle the adjustments, rebasing, etc. and we'll get it merged.  Thanks again for the help, and apologies for the delay.

rbrw:
I've rebased this onto current 5.2.x and fixed the fact_values reference here: https://github.com/puppetlabs/puppetdb/pull/2968  So closing this in favor of merging your changes there.  Thanks again for the help.

PR #2457, opened 2018-02-08
Revert "(maint) Add build_tar: FALSE to build_defaults"

mwaggett:
This reverts commit f9733e1b0adcd2fd6b998f6aa817e7bb1cf88488.
We were fixing a symptom of a problem, rather than the problem itself (see
RE-10238).


mwaggett:
https://github.com/puppetlabs/packaging/pull/788 needs to go in first.

PR #2469, opened 2018-03-26
(PDB-3884) Update logback configuration

Sharpie:
This commit updates the logback configuration bundled with puppetdb
packages to implement the following improvements:

  - Set `scanPeriod="60 seconds"` so that settings like log levels can be
    toggled without requiring a service restart.

  - Use RFC 3339 compliant timestamps that include the offset from UTC.

### DIFF
--- a/resources/ext/config/logback.xml
+++ b/resources/ext/config/logback.xml
@@ -1,4 +1,4 @@
-<configuration scan="true">
+<configuration scan="true" scanPeriod="60 seconds">
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
             <pattern>%d %-5p [%c{2}] %m%n</pattern>
--- a/resources/ext/config/logback.xml
+++ b/resources/ext/config/logback.xml
@@ -1,7 +1,7 @@
 <configuration scan="true" scanPeriod="60 seconds">
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
-            <pattern>%d %-5p [%c{2}] %m%n</pattern>
+            <pattern>%d{yyyy-MM-dd'T'HH:mm:ss.SSSXXX} %-5p [%c{2}] %m%n</pattern>
         </encoder>
     </appender>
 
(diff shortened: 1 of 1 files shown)


rbrw:
I'm always a bit wary of automatic config reloads since there's no guarantee the file will be in a consistent state during the reload (e.g. what if someone's editing it in place -- which I imagine would be the norm).  So I wonder if it would be possible, if it's not already the case, to just arrange for pdb to reload this config along with the other bits when it receives a SIGHUP.  Then the policy can be left up to the local admins.

Sharpie:
The problem with SIGHUP is that we're also reloading "the other bits" that are behaving in a way that is causing us to enable DEBUG logging. There's no way to just boost the logging and leave everything else undisturbed.

Additionally, Logback will [guard against incomplete configuration](https://logback.qos.ch/manual/configuration.html#autoScan) by using the last good settings if there is an error:

> As it is easy to make errors while editing a configuration file, in case the latest version of the configuration file has XML syntax errors, it will fall back to a previous configuration file free of XML syntax errors.

Finally, this behavior was already enabled by the existing `scan="true"` attribute. Adding the `scanPeriod` corrects a regression that stopped the reloads from occurring ([TK-426](https://tickets.puppetlabs.com/browse/TK-426)).

Twiddling log settings should be a rare occurrence. On the balance, I think the ability to enable DEBUG logging without disturbing the rest of the application outweighs the potential downsides of an unlucky reload picking up incomplete settings.

Sharpie:
@rbrw Let me know if there's any additional feedback about this change that I can address.

rbrw:
@Sharpie apologies for the delay -- and although I still have general qualms about automatic updates, for PDB your points sound completely reasonable.  In an ideal world (and in a longer run), perhaps the JVM would believe in fs sockets, and we could just support `echo refresh-log-config > /var/run/puppetdb/control`, but that's not the current world.  So I think the scanPeriod change is fine, and suspect it should be moved back to 5.1.x.

Regarding the other change, assuming I read it right, I'm concerned about altering the timestamp format in a stable release, since I'd guess it could break existing parsers.  

So overall, I'd suggest splitting this into two PRs, one against 5.1.x with the scanPeriod change, and another against master with the new timestamp format.

Thanks

Sharpie:
Submitted a separate PR for getting these changes into PE. Will re-target this PR at the master branch.

Sharpie:
Rebased onto master and removed the do not merge flag.

Looks like two travis tests failed during setup when `ext/travisci/install_puppetserver.sh` was run, which is not likely related to the changes in logging configuration.

Sharpie:
Aw yiss. All green.

@rbrw I think this one is ready to merge.

PR #2471, opened 2018-03-29
(PDB-3808) Remove node_state 'any' nodes during query engine processing

ajroetker:


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1526,6 +1526,19 @@
...
+(defn remove-elided-nodes
+  "This step removes elided nodes (marked with ::elide by expand-user-query) from
+  `and` and `or` clauses."
+  [node]
+  (cm/match [node]
+            [["and" & clauses]]
+            (into ["and"] (remove #(= ::elide %) clauses))
+
+            [["or" & clauses]]
+            (into ["or"] (remove #(= ::elide %) clauses))
```
mullr: Is there any way this can happen outside of an "and" or "or"?

ajroetker: what other case are you thinking of?

mullr: maybe if you just had node_state== any by itself? it seems like that would get ::elide, but not hit iether of these cases.

ajroetker:
@mullr Cleaned this up by simplifying to just remove elided values for all vector nodes!

rbrw:
Russ says go, and AJ says go, so GO!

PR #2478, opened 2018-04-18
(PDB-3910) Create an index on reports (start_time)

jplindquist:
Create an index on reports (start_time)
Resolves [PDB-3910](https://tickets.puppetlabs.com/projects/PDB/issues/PDB-3910)


jplindquist:
Resolved via https://github.com/voxpupuli/puppetboard/pull/467 instead

PR #2479, opened 2018-04-25
(PDB-3911) Add primary key to resource_events

npwalker:
Prior to this commit, resource_events had a unique constraint
on four columns.

After this commit, we add a primary key on those same four columns
and drop the unique constraint.  A primary key is enforced by a
unique constraint behind the scenes so there should be no difference
other than 3rd party tools like pglogical or pg_repack will now
be happy that there is a primary key when they used to fail.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1496,6 +1496,14 @@
 
   {::vacuum-analyze #{"factsets"}})
 
+ (defn add-primary-key-to-resource-events []
+   (jdbc/do-commands
+    ;;As far as we can tell property should never be null but if it is it would
+    ;;cause creating the primary key to fail so we make sure property isn't null
+    "UPDATE resource_events SET property ='' WHERE property is null;"
+    "ALTER TABLE resource_events ADD PRIMARY KEY (report_id, resource_type, resource_title, property);"
+    "ALTER TABLE resource_events DROP CONSTRAINT IF EXISTS resource_events_unique;"))
+
 (def migrations
   "The available migrations, as a map from migration version to migration function."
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1496,6 +1496,14 @@
 
   {::vacuum-analyze #{"factsets"}})
 
+ (defn add-primary-key-to-resource-events []
+   (jdbc/do-commands
+    ;;As far as we can tell property should never be null but if it is it would
+    ;;cause creating the primary key to fail so we make sure property isn't null
```
Sharpie: `property` is nullable to allow the storage of skip events which don't affect any resource property. The current schema definition doesn't make this clear, but the PuppetDB 2.x migration.clj has the following comment:

```
    ;; I wish these next two could be "NOT NULL", but for now we are
    ;; fabricating skipped resources as events, and in those cases we don't
    ;; have any legitimate values to put into these fields.
    ["property" "VARCHAR(40)"]
```

https://github.com/puppetlabs/puppetdb/blob/2.3.8/src/com/puppetlabs/puppetdb/scf/migrate.clj#L265-L268

robdaemon:
@npwalker @rbrw We can't convert the unique constraint to a primary key because of the nullable column. 

Adding a serial column has implications for replication / HA that can be problematic in syncing - think in terms of two postgres instances generating serial ID 1. Who is right in a sync? Another datatype that works for this purpose is `uuid`, but generating the UUID in a `INSERT` or `UPDATE` statement requires the `uuid-ossp` or `pgcrypto` extensions to be installed.

robdaemon:
We picked this up and addressed the concerns raised in #2784, closing in favor of that.

PR #2482, opened 2018-05-04
(PDB-3920) 5.2.x: test appropriate server and puppet versions in CI

rbrw:
For now test the corresponding platform versions, and the previous
platform versions.


Zak-Kent:
LGTM +1

PR #2486, opened 2018-05-25
(CPR-553) Migrate puppetdb docker build scripts from puppet-in-docker

underscorgan:


docker/Gemfile
```diff
@@ -0,0 +1,13 @@
+source ENV['GEM_SOURCE'] || "https://rubygems.org"
+
+def location_for(place)
+  if place =~ /^(git[:@][^#]*)#(.*)/
+    [{ :git => $1, :branch => $2, :require => false }]
```
rbrw: Not sure exactly how this works nor whether the tool(s) really care, but, we have a change pending to the puppet entry lower in the file to use :ref instead of :branch because for the integration testing we want to be able to use hashes, tags, or branches.

docker/puppetdb-postgres/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+      org.label-schema.version="9.6.8" \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppetdb" \
+      org.label-schema.vcs-ref=$vcs_ref \
+      org.label-schema.build-date=$build_date \
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+COPY docker-entrypoint-initdb.d /docker-entrypoint-initdb.d
+
+COPY Dockerfile /
```
rbrw: Hmm, OK, so puppetdb's going to have its own postgresql container -- in any case, not sure how they set up that container, but possibly worth noting that the normal default postgresql.conf settings are known to be very conservative (for portability).  If that's also true here, then I'd guess that most installs will want something a bit less conservative.

docker/puppetdb/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppetdb" \
+      org.label-schema.vcs-ref=$vcs_ref \
+      org.label-schema.build-date=$build_date \
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
+
+RUN apt-get update && \
+    apt-get install -y wget netcat lsb-release && \
+    wget https://apt.puppetlabs.com/puppet5-release-"$UBUNTU_CODENAME".deb && \
+    wget https://github.com/Yelp/dumb-init/releases/download/v"$DUMB_INIT_VERSION"/dumb-init_"$DUMB_INIT_VERSION"_amd64.deb && \
```
rbrw: Wonder if it might eventually be worthwhile to have "trust on first use" style checksum validation in in cases where we don't have more substantial authentication, i.e. like this: https://github.com/puppetlabs/puppetdb/pull/2488/commits/2d97bac78fcde16bec12bbe657e69d9922f3a0f8#diff-a1cdf43481aa7800ceaca164deaf8e3bR39

docker/puppetdb/Dockerfile
```diff
@@ -0,0 +1,57 @@
+FROM ubuntu:16.04
+
+ARG vcs_ref
+ARG build_date
+ENV PUPPETDB_VERSION="5.2.2"
```
rbrw: Don't know if it's appropriate, but we can produce the actual project.clj version via `lein pprint :version`.

docker/puppetdb/conf.d/database.conf
```diff
@@ -0,0 +1,8 @@
+database: {
+  classname: org.postgresql.Driver
+  subprotocol: postgresql
+  subname: ${PUPPETDB_DATABASE_CONNECTION}
+  username: ${PUPPETDB_USER}
+  password: ${PUPPETDB_PASSWORD}
+  log-slow-statements: 10
+}
```
rbrw: Wondered whether this ends up with appropriately restrictive permissions, i.e. only readable by the puppetdb group or whatever makes sense.

docker/puppetdb/docker-entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ]; then
+  while ! nc -z puppet 8140; do
+    sleep 1
+  done
+  set -e
+  /opt/puppetlabs/bin/puppet config set certname ${HOSTNAME}
```
rbrw: Wondered about quoting the `HOSTNAME` expansion, i.e. `"$HOSTNAME"`.

docker/puppetdb/puppetdb
```diff
@@ -0,0 +1,22 @@
+###########################################
+# Init settings for puppetdb
+###########################################
+
+# Location of your Java binary (version 7 or higher)
+JAVA_BIN="/usr/bin/java"
+
+# Modify this if you'd like to change the memory allocation, enable JMX, etc
+JAVA_ARGS=$PUPPETDB_JAVA_ARGS
```
rbrw: Perhaps quote as "`$PUPPETDB_JAVA_ARGS`".

docker/puppetdb/Dockerfile
underscorgan: that command doesn't work for me:

```
morgan@river.corp.puppetlabs.net % lein pprint :version
Tried to load org.apache.httpcomponents/httpcore version 4.1.2 but 4.4.4 was already loaded.
Tried to load org.apache.httpcomponents/httpclient version 4.1.2 but 4.5.3 was already loaded.
Tried to load org.apache.maven.wagon/wagon-http version 2.2 but 2.12 was already loaded.
Tried to load commons-codec version 1.4 but 1.9 was already loaded.
Tried to load org.codehaus.plexus/plexus-utils version 3.0 but 3.0.24 was already loaded.
Tried to load org.codehaus.plexus/plexus-component-annotations version 1.5.5 but 1.7.1 was already loaded.
Tried to load org.jsoup/jsoup version 1.6.1 but 1.7.2 was already loaded.
Tried to load org.tcrawley/dynapath version 0.2.3 but 0.2.5 was already loaded.
Tried to load org.codehaus.plexus/plexus-interpolation version 1.14 but 1.24 was already loaded.
Tried to load bultitude version 0.2.0 but 0.2.8 was already loaded.
Tried to load commons-io version 2.0.1 but 2.5 was already loaded.
Tried to load com.cemerick/pomegranate version 0.3.1 but 0.4.0-alpha1 was already loaded.
Tried to load org.apache.maven/maven-repository-metadata version 3.0.4 but 3.5.0 was already loaded.
Tried to load org.apache.maven.wagon/wagon-provider-api version 2.2 but 2.12 was already loaded.
Tried to load commons-logging version 1.1.1 but 1.2 was already loaded.
Tried to load org.apache.maven/maven-model-builder version 3.0.4 but 3.5.0 was already loaded.
Tried to load org.apache.maven/maven-model version 3.0.4 but 3.5.0 was already loaded.
'pprint' is not a task. See 'lein help'.

Did you mean this?
         parent
```

Working on tooling for version is out of scope for this initial PR, I have another ticket in progress for that, but we want to get this initial work in first.

docker/puppetdb/conf.d/database.conf
underscorgan: I think looking into this is something that should be ticketed but not addressed in this PR, we're mostly worried around the automation for publishing the existing containers more than improving the construction of the containers.

docker/puppetdb/Dockerfile
underscorgan: This should be ticketed for future improvement, but is out of scope for the initial work

docker/puppetdb-postgres/Dockerfile
underscorgan: this should be ticketed for investigation

docker/puppetdb/Dockerfile
rbrw: Oh, sorry, should have said `lein with-profile ci pprint :version` since that plugin's currently guarded by the ci profile.

docker/Gemfile
underscorgan: We shouldn't need that in this case I don't think, but easy enough to change if that comes up. This is mostly for use in local testing, where I'm usually running off a file or a git branch I've pushed to my fork.

docker/puppetdb/Dockerfile
underscorgan: ah, yeah, I don't think this will quite do what we want, we have no standard way to get SNAPSHOT versions (an investigation point I have going forward is how to build containers with, say, latest nightlies or something), but for right now we want to build with the latest released version so I think `git describe` will be a more reliable way to get that info

underscorgan:
@rbrw updated the variable quoting, otherwise I think the other comments are all things that should be ticketed / out of scope for the initial work.

docker/puppetdb/Dockerfile
rbrw: Ahh, ok -- wasn't even sure of the context, just thought I'd mention it in case it was useful.

docker/puppetdb/Dockerfile
rbrw: Oh, sure - not critical right now, just thought it might be worth consideration.

docker/puppetdb/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+ENV PUPPET_AGENT_VERSION="5.5.1"
+ENV DUMB_INIT_VERSION="1.2.1"
+ENV UBUNTU_CODENAME="xenial"
+ENV PUPPETDB_DATABASE_CONNECTION="//postgres:5432/puppetdb"
+ENV PUPPETDB_USER=puppetdb
+ENV PUPPETDB_PASSWORD=puppetdb
+ENV PUPPETDB_JAVA_ARGS="-Djava.net.preferIPv4Stack=true -Xms256m -Xmx256m"
+ENV PATH=/opt/puppetlabs/server/bin:/opt/puppetlabs/puppet/bin:/opt/puppetlabs/bin:$PATH
```
rbrw: Perhaps quote this too, in case the existing path has spaces, i.e. "$PATH".

docker/puppetdb/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+      org.label-schema.vendor="Puppet" \
+      org.label-schema.url="https://github.com/puppetlabs/puppetdb" \
+      org.label-schema.name="PuppetDB" \
+      org.label-schema.license="Apache-2.0" \
+      org.label-schema.version=$PUPPETDB_VERSION \
+      org.label-schema.vcs-url="https://github.com/puppetlabs/puppetdb" \
+      org.label-schema.vcs-ref=$vcs_ref \
+      org.label-schema.build-date=$build_date \
+      org.label-schema.schema-version="1.0" \
+      org.label-schema.dockerfile="/Dockerfile"
```
rbrw: Suggest quoting for vcs_ref, PUPPETDB_VERSION, build_date, etc. here too unless the LABEL command does something special, just to be safe wrt accidental injection, etc.

rbrw:
OK, happy to merge once we decide what we want to do about those last bits.  Thanks.

docker/puppetdb/conf.d/database.conf
underscorgan: tickets.puppetlabs.com/browse/CPR-587

docker/puppetdb-postgres/Dockerfile
underscorgan: tickets.puppetlabs.com/browse/CPR-586

docker/puppetdb/Dockerfile
underscorgan: https://tickets.puppetlabs.com/browse/CPR-588

underscorgan:
@rbrw ok, I think I've quoted all the things and opened tickets for the remaining work!

PR #2487, opened 2018-05-29
(CPR-557) Add scripts to build puppetdb docker images

underscorgan:


docker/ci/build
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+
+set -eux
+
+BUNDLER_PATH=.bundle/gems
+
+function build_and_test_image() {
+  container_name=$1
```
rbrw: Perhaps quote `"$1"`.

docker/ci/build
```diff
@@ -0,0 +1,58 @@
...
+
+  : ===
+  : === run linter on the docker files
+  : ===
+   bundle exec puppet-docker local-lint $container_name
```
rbrw: Same here (and for any other vars where expansion to multiple args wouldn't be acceptable), i.e. `"$container_name"`

docker/ci/build
```diff
@@ -0,0 +1,58 @@
...
+: === pull updated base images
+: ===
+bundle exec puppet-docker update-base-images ubuntu:16.04 postgres:9.6.8
+
+container_list=(puppetdb puppetdb-postgres)
+
+: ===
+: === build and test all the images before we push anything
+: ===
+for container in ${container_list[@]}; do
```
rbrw: I believe quoting this will do the same per-item expansion/quoting as for "$@": `"${container_list[@]}"`

underscorgan:
@rbrw quoted all the things!

docker/ci/build
```diff
@@ -0,0 +1,58 @@
...
+function build_and_test_image() {
+  container_name="$1"
+
+  : ===
+  : === run linter on the docker files
+  : ===
+   bundle exec puppet-docker local-lint "$container_name"
+
+  : ===
+  : === build and test $container_name
```
rbrw: Hah, I wondered about injections here (suspected it might be ok), and offhand looks like the trivial case at least is ok, i.e. given something like

```sh
container_name='foo;
rm -rf ...
'
```
it's still fully subsumed by the `:` command.

rbrw:
Looks good.

underscorgan:
Realized this was accidentally against master, so fixed that, and also fixing an issue I ran into setting up the puppetserver pipelines.

rbrw:
Thanks -- rebased and merged as ab9ad9b1550a3a81d002e7101fc14a439ad1f2f3

PR #2489, opened 2018-06-05
(docs) Add 5.2.3 release notes

gguillotte:
No tickets in PDB 5.2.3 are marked as requiring release notes, so in lieu of notes I'm linking to the JIRA search for issues with that fixVersion.


rbrw:
Looks good -- thanks.

PR #2490, opened 2018-06-05
(docs)(DOC-3617) Fix broken package repo installation link

gguillotte:
The docs to install via module link to Puppet Collection, which has been replaced by Puppet Platform. The redirect from puppet.com/docs/puppet/latest/puppet_collections.html to puppet.com/docs/puppet/latest/puppet_platform.html is not reliable since Puppet 5.4, causing the link the break.

This updates the text and link to refer to Puppet Platform instead of Puppet Collections.


rbrw:
Merged as 6dc360197967dcfbd37b27e48947bd2ca9610b39 #2491

rbrw:
Thanks.

PR #2501, opened 2018-06-27
(PDB-2700) Migrate log test utils to trapperkeeper

austb:
Prior to this commit, testutil functions related to logging
were duplicated in trapperkeeper and puppetdb.

This commit removes the duplicated functions and switches to
the ones provided by trapperkeeper.


test/puppetlabs/puppetdb/amq_migration_test.clj
```diff
@@ -134,8 +134,8 @@
                         (:transaction_uuid (first results))))))))))))
 
   (testing "an upgrade with a bad catalog"
-    (with-log-suppressed-unless-notable (every-pred tlog/critical-errors
-                                                    (tlog/starting-with "Output of coerce-to-new-command does not"))
+    (tlog/with-log-suppressed-unless-notable (every-pred log-utils/notable-pdb-event?
+                                                    (log-utils/starting-with "Output of coerce-to-new-command does not"))
```
rbrw: I'd adjust the indentation here.  Normally the `(log-utils` should line up with the `log-utils` above it, but this line's also awfully long.  While we're not strict about 80-columns, we generally try to keep them closer to that when possible, and (excepting internationalization-related issues) usually try to at least stay under about 100.

So I'd probably either just bind that every-pred in a let `(let [foo (every-pred ...` and use foo in the with, or even just

```clojure
    (with-log-suppressed-unless-notable
      (every-pred tlog/critical-errors
                  (tlog/starting-with "Output of coerce-to-new-command does not"))
      (with-test-db
```
Personally, I'd likely also keep the `:refer` for that symbol -- I sometimes favor unprefixed with- forms, but if you prefer the prefixed version, that's fine.

test/puppetlabs/puppetdb/amq_migration_test.clj
```diff
@@ -96,7 +96,7 @@
 
 (deftest shovel-upgrade-test
   (testing "an upgrade with existing catalogs"
-    (with-log-suppressed-unless-notable tlog/critical-errors
+    (tlog/with-log-suppressed-unless-notable log-utils/notable-pdb-event?
```
rbrw: Hmm, was this change intentional (here and below)?

test/puppetlabs/puppetdb/testutils/log.clj
```diff
@@ -207,5 +27,6 @@
   [string]
   (fn [^ILoggingEvent event]
     (-> event
-        .getMessage
+        tlog/event->map
+        :message
```
rbrw: Perhaps stick with `.getMessage` now that we're sticking with Events?

test/puppetlabs/puppetdb/threadpool_test.clj
```diff
@@ -4,13 +4,11 @@
             [clojure.test :refer :all]
             [puppetlabs.puppetdb.threadpool :refer :all]
             [puppetlabs.puppetdb.testutils :as tu]
-            [puppetlabs.trapperkeeper.testutils.logging :refer [atom-logger
-                                                                with-logging-to-atom
-                                                                with-log-suppressed-unless-notable]]
```
rbrw: As mentioned, I sometimes favor the refers, since in some cases, I find the unprefixed code a bit easier on the eyes.  In any case, while I'm often fine with personal preference on that front, I think I'd keep all the refers in this pr because I think minimizing the number of superfluous changes (line-wise) is the more important consideration (i.e. easier to review, less likely to produce merge conflicts now or in any future merges/cherry-picks, etc.).

test/puppetlabs/puppetdb/amq_migration_test.clj
austb: Ah yes, that's a mistake. I initially thought critical-errors was gone after the tk migration. But I brought it back

PR #2503, opened 2018-06-29
(PDB-3942) Add JDK 10 testing

austb:


src/puppetlabs/puppetdb/core.clj
```diff
@@ -22,10 +24,11 @@
...
   {:post [(set? %)]}
-  (set (for [namespace (ns/find-namespaces-on-classpath)
-             :let [ns-str (name namespace)]
-             :when (.startsWith ns-str ns-prefix)]
-         namespace)))
+  (set (filter
+         #(str/starts-with?
+            (name %)
+            ns-prefix)
+         (find-namespaces (cp/classpath)))))
```
rbrw: Might suggest condensing this a bit, say something like
```clojure
(set (filter #(str/starts-with? (name %) ns-prefix)
             (find-namespaces (cp/classpath))))
```
or
```clojure
(->> (find-namespaces (cp/classpath))     
     (filter #(str/starts-with? (name %) ns-prefix))
     set)
```

.travis.yml
```diff
@@ -26,6 +27,14 @@ jobs:
       services: postgresql
       script: jdk_switcher use oraclejdk8 && ext/travisci/run-core-tests
 
+    # OracleJDK9
```
rbrw: Don't feel strongly about it, but I'd be tempted to either leave this header comment and the JDK8 one off, under the argument that the PDB_TEST indicators are prominent enough, or to generalize them to "JDK8" and "JDK9" "group" headers.

.travis.yml
```diff
@@ -26,6 +27,14 @@ jobs:
...
+    - stage: ❧ pdb tests
+      env: PDB_TEST=core/oraclejdk9
+      addons:
+        postgresql: *pg-ver
+      services: postgresql
+      script: jdk_switcher use oraclejdk9 && ext/travisci/run-core-tests
+
     # === integration tests
     - stage: ❧ pdb tests
       env: PDB_TEST=int/openjdk8/pup-master/srv-master
```
rbrw: Did you try adding JDK9 integration tests too?  If those also work now, think we'll want them.

rbrw:
Overall looks good.

.travis.yml
austb: Ah I didn't mean to leave those in, just had those in there while I was making sure I ran all the relevant tests with jdk9

.travis.yml
austb: I'll add them in the next commit

src/puppetlabs/puppetdb/core.clj
```diff
@@ -13,6 +13,8 @@
    `[arg1 arg2 arg3]`."
   (:require [puppetlabs.trapperkeeper.logging :as logging-utils]
             [clojure.tools.namespace :as ns]
```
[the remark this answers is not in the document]
austb: Yes, I believe we can

src/puppetlabs/puppetdb/core.clj
austb: I like how the `->>` version reads, I'll change it to that.

ext/bin/require-jdk
```diff
@@ -0,0 +1,53 @@
...
+
+set -uexo pipefail
+
+script_home="$(cd "$(dirname "$0")" && pwd)"
+
+default=openjdk10
+
+cmdname="$(basename "$0")"
+
+usage() { echo "Usage: $cmdname VERSION INSTALLDIR_IF_NEEDED"; }
```
rbrw: Not sure, but now that I think about it, wonder if it might be worth a trivial elaboration here, to give an indication of what's expected, i.e. something like this for the usage:

    Usage: $cmdname (openjdk10|default|...) INSTALLDIR_IF_NEEDED

.travis.yml
```diff
@@ -20,7 +20,13 @@ aliases:
...
     set -e
     jdk="$(ext/travisci/jdk-from-spec "$PDB_TEST")"
-    if test "$(uname)" = Linux; then jdk_switcher use "$jdk"; fi
+    jdkver="${jdk##*jdk}"
+    if test "$(uname)" = Linux -a "$jdkver" -lt 9 ; then
+        jdk_switcher use "$jdk";
+    elif test "$(uname)" = Linux; then
+        export JAVA_HOME="$(ext/bin/require-jdk "$jdk" ext/travisci/local)"
+        export PATH="$JAVA_HOME/bin:$PATH"
+    fi
```
rbrw: I'm inclined to just assert that what require-jdk does is create a valid JAVA_HOME in the specified dir.  Then we can avoid needing/caring-about the output. i.e.:

    ext/bin/require-jdk "$jdk" ext/travisci/local
    export JAVA_HOME="$(pwd)/ext/travisci/local"
    export PATH="$JAVA_HOME/bin:$PATH"

And then we also wouldn't need the tail at the end of require-jdk, etc.

rbrw:
Looks good.

PR #2504, opened 2018-07-05
(CPR-599) auto-version puppetdb containers based on git describe

underscorgan:


underscorgan:
Have this working at https://pipelines.puppet.com/release-engineering/apps/puppetdb

docker/Gemfile
```diff
@@ -1,7 +1,7 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
 def location_for(place)
-  if place =~ /^(git[:@][^#]*)#(.*)/
+  if place =~ /^((?:git|https?)[:@][^#]*)#(.*)/
```
mwaggett: What is the first `?` doing in this?

underscorgan: `(?:)` in a ruby regex means this is a non-capturing group, so `$1` will still be the whole url, and `$2` will still be the branch

mwaggett: oooh, cool, thanks!

docker/puppetdb-postgres/Dockerfile
```diff
@@ -1,14 +1,17 @@
-FROM postgres:9.6.8
+ARG postgres_version="9.6.8"
+FROM postgres:"$postgres_version"
 
 ARG vcs_ref
 ARG build_date
 
+ARG postgres_version="9.6.8"
```
mwaggett: I'm guessing you don't need this a second time.

docker/puppetdb/Dockerfile
```diff
@@ -2,8 +2,9 @@ FROM ubuntu:16.04
 
 ARG vcs_ref
 ARG build_date
-ENV PUPPETDB_VERSION="5.2.2"
-ENV PUPPET_AGENT_VERSION="5.5.1"
+ARG version="5.1.5"
```
mwaggett: Why an older version now?

docker/puppetdb-postgres/Dockerfile
underscorgan: Yep, you do. Anything before the `FROM` is only in scope for the `FROM` and needs to be redeclared afterwards

docker/puppetdb/Dockerfile
underscorgan: Because the team merged the initial work into the 5.1.x, so this is setting it to the latest version in the 5.1.x series

docker/puppetdb-postgres/Dockerfile
mwaggett: well that's.. silly.

docker/ci/build
```diff
@@ -15,23 +16,37 @@ function build_and_test_image() {
...
+: ===
+: === If we do not git pull --unshallow we get really weird results with git describe
+: ===
+git pull --unshallow
+
+: ===
+: === make sure we fetch tags for versioning
+: ===
+git fetch origin 'refs/tags/*:refs/tags/*'
+git_describe=`git describe`
```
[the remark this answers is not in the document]
underscorgan: I don't think that's going to be relevant here, it looks like `git describe --always` will either match the output of `git describe`, or it will give the git sha if it doesn't find a tag. The build will fail if the version is set to a git sha (since that's used to pick the version of puppetdb installed from repos), I think failing at the `git describe` would make it more straightforward to debug than failing in the container build

docker/ci/build
```diff
@@ -2,6 +2,7 @@
 
 set -eux
 
+PATH=$PATH:/usr/local/bin
```
rbrw: Suggest double quotes here so spaces in the PATH couldn't be an issue.

docker/ci/build
```diff
@@ -15,23 +16,37 @@ function build_and_test_image() {
...
+  bundle exec puppet-docker push "$container_name" --repository pcr-internal.puppet.net/release-engineering --version "$container_version" --no-latest
 }
 
 : ===
 : === bundle install to get ready
 : ===
 bundle install --path "$BUNDLER_PATH"
 
+: ===
+: === If we do not git pull --unshallow we get really weird results with git describe
```
rbrw: Yeah, I imagine you might be able to get away with a clone --depth, but guessing it'd have to be "deep enough" to include the most recent tag, and even if that worked, offhand I'm not sure if you could determine it cheaply (or simply) enough to be worth it.

underscorgan: Yeah, I don't think there's a reliable-enough way to do this, so I'd rather stick with `--unshallow`

docker/ci/build
underscorgan: Added quotes!

PR #2510, opened 2018-07-13
Update deps for 18.04 support against 5.1.x

Zak-Kent:


rbrw:
Looks good -- thanks.

PR #2519, opened 2018-07-20
(PDB-3938) Update to clj-parent 2.0.2 for pdbext rbac-client dep

rbrw:
pdbext needs rbac 0.9.0 in order to be able to upgrade to clj-parent
2.0, clojure 1.9, etc.


rbrw:
Merging even though the integration tests are broken (expected: https://tickets.puppetlabs.com/browse/PUP-8988 -- should be resolved soon), so that we can finish fixing pdbext.

PR #2520, opened 2018-07-20
(PDB-3268) Don't migrate on schema_migration query failures

austb:
Previously, we were catching an old hsqldb error as well
as a regex match on any error containing 'schema_migrations'.
We only intended to catch the error that says the database does
not exist so that we know when to apply the first migration to
initialize the database.

This commit removes the old checks for errors that we don't want to
catch.


rbrw:
This looks good, though I'm tempted to suggest making the commit message a bit more explicit about the motivation, e.g. "Don't attempt migrations anyway, if the schema_migrations query fails", or similar.

rbrw:
Cherry picked to master as 5317dacadc0d90744c133c2c9c69e91616f9bd4e since it wasn't quite up to date, and pushed.

PR #2521, opened 2018-07-23
(PDB-3982) Specify puppet vendormoduledir for integration tests

rbrw:
Given the introduction of vendormoduledir in PUP-8988, the tests are
likely to crash when puppet tries to create
/opt/puppetlabs/puppet/vendor_modules, which is also a directory we
definitely don't want the tests to use.

To avoid that, specify a target/ vendormoduledir in all the relevant
puppet config files.

### DIFF
--- a/test-resources/puppetserver/puppet.conf
+++ b/test-resources/puppetserver/puppet.conf
@@ -9,3 +9,5 @@ storeconfigs = true
 storeconfigs_backend = puppetdb
 reports = puppetdb
 autosign = true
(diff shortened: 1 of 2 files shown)


rbrw:
I'll leave it open for the moment, but this is likely to just be closed in favor of https://github.com/puppetlabs/puppet/pull/6939.

joshcooper:
PDB tests are passing now that the puppet fix was merged, closing.

PR #2523, opened 2018-07-24
(PDB-3953) Add ext/bin/clj-1.9 and use it in dep-diff

rbrw:
...since pdb no longer depends on or uses clojure 1.8, and since
clojure 1.9 is multi-jar.


rbrw:
I'll leave the "don't merge" on here until I finish the pgbox work, since I think it's plausible that this will be incorporated there.

Zak-Kent:
LGTM + 1

PR #2527, opened 2018-07-26
(PDB-3947) Drop remaining activemq code, including upgrade shovel

rbrw:
Remove activemq entirely for version 6.  This means that (as allowed
by the versioning policy), it will not be possible to skip upgrading
to pdb 6 from pdb 4 without running pdb 5 without losing the content
of any existing queue.  See the documentation updates for further
information.


resources/ext/config/logback.xml
```diff
@@ -20,16 +20,9 @@
         </encoder>
     </appender>
 
-    <!-- Supress internal ActiveMQ logging -->
-    <logger name="org.apache.activemq" level="warn"/>
-
     <!-- Supress internal Spring Framework logging -->
     <logger name="org.springframework.jms.connection" level="warn"/>
```
austb: I don't see Spring Framework in our dep tree anymore, can this be removed too?

rbrw: Not sure -- if it's not anywhere in the dep tree, possibly, though wonder if it might be loaded dynamically.  I can do some digging -- see why it was added in the first place.

PR #2531, opened 2018-08-08
(PDB-3470) Add pdbbox and use it in tests; test osx in travis

rbrw:
See the documentation/ and .travis.yml changes for further details.

Note that this adds additional inter-dependent utility scripts to
ext/bin, but we don't want to add ext/bin to the PATH because so far,
the command names in ext/bin haven't been chosen with that in
mind (i.e. sha256sum would shadow the Linux version), so we refer to
them with explicit paths (sometimes generated via basename).

In the longer run we might want to adjust the relevant command names,
or perhaps move "utility" commands to ext/util/, to make ext/bin
suitable for inclusion in the PATH.

Remove the ulimit -u settings because it removes another platform
variance (disallowed on travis osx hosts), and it looks like we may no
longer need it, but if we see the OutOfMemoryError exceptions
return (2080a332b40f95cdf564a65a6bf540704273b583), we can restore it.


rbrw:
Added the "don't merge" for now while we make sure the (upcoming) pdbext pr is set, etc.

ext/bin/pdbbox-init
```diff
@@ -0,0 +1,177 @@
...
+<configuration scan="true">
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d %-5p [%thread] [%c{2}] %m%n</pattern>
+        </encoder>
+    </appender>
+
+    <!-- Silence particularly noisy packages -->
+    <logger name="org.apache.activemq" level="warn"/>
+    <logger name="org.apache.activemq.store.kahadb.MessageDatabase"
```
austb: We should be sure to remove these once activemq removal is merged.

ext/travisci/pgbin-from-spec
```diff
@@ -0,0 +1,24 @@
...
+
+def misuse():
+    usage(stderr)
+    exit(2)
+        
+len(sys.argv) == 2 or misuse()
+specs = sys.argv[1]
+specs = specs.split('/')
+specs = [x for x in specs if x.startswith('pg-')]
+len(specs) == 1 or misuse()
```
austb: We could save ourselves some duplication if we got the pg version using prefixed-ref-from-spec and then just did the os specific stuff in this file.

ext/bin/pdbbox-init
rbrw: Yep - depends on which goes in first (and then I'll rebase).  Though it won't hurt if we accidentally leave them for a while.

ext/travisci/pgbin-from-spec
rbrw: We could, but this seemed simple enough to be worth avoiding the external dep (finding the other tool wrt PATH or script-home or...), and not worth forking a second process.

ext/bin/pdbbox-init
```diff
@@ -0,0 +1,177 @@
...
+    <root level="info">
+        <appender-ref ref="STDOUT" />
+    </root>
+</configuration>
+EOF
+
+cat > "$pdbbox/pdb.ini" <<EOF
+
+[global]
+disable-update-checking = true
```
Zak-Kent: I think this may need to be in the [puppetdb] section. https://puppet.com/docs/puppetdb/5.2/configure.html#puppetdb-settings

PR #2535, opened 2018-08-09
Update acceptance tests

Zak-Kent:
This pr changes the way we install versions in our upgrade tests that accounts for the versions we need being different from one branch to another. It also adds logic to support cases where we're testing against newly supported platforms that don't have older official releases of pdb.


acceptance/helper.rb
```diff
@@ -264,6 +265,36 @@ def sleep_until_started(host)
...
+    return test_config[:os_families].has_key? 'ubuntu1804-64-1'
+  end
+
+  def oldest_supported
+    # account for special case where bionic doesn't have builds before 5.2.4
+    return is_bionic ? '5.2.4' : '4.2.3.8'
+  end
+
+  def get_testing_branch(version)
+    branch_name = version[0..3] + 'x'
```
austb: This will break if we hit double digits in either the major or minor version(s). We could do something with regex like `/^(\d+.\d+.)(?:[\d]+.)*\d$/.match(version)[1] + 'x'`. As a side note, I use https://regex101.com/ when I'm playing around with constructing regular expressions.

acceptance/helper.rb
```diff
@@ -264,6 +265,36 @@ def sleep_until_started(host)
...
+    if branch_name.chars.first != '5'
+      branch_name = 'master'
```
austb: Would this put LTS on master as well?

acceptance/helper.rb
```diff
@@ -318,6 +349,7 @@ class { 'puppetdb::globals':
     }
     class { 'puppetdb::server':
       manage_firewall => false,
+      disable_update_checking => true,
```
austb: 👍

PR #2538, opened 2018-08-10
 (PDB-4006) Add tests for catalog with rich data

austb:
serialize_catalog.rb: https://gist.github.com/austb/4dd8fcc39412bbe608a64818f7264c0d
example manifest: https://gist.github.com/austb/a665e835afc49dbb42ccf587fc1ca6a0

### DIFF
--- a/test/puppetlabs/puppetdb/integration/smoke.clj
+++ b/test/puppetlabs/puppetdb/integration/smoke.clj
@@ -27,3 +27,30 @@
           (is (= catalog-uuid report-uuid)))))))
 
 
+(deftest ^:integration rich-data-agent-run
+  (with-open [pg (int/setup-postgres)
+              pdb (int/run-puppetdb pg {})
+              ps (int/run-puppet-server [pdb] {:agent {:rich-data true}})]
+
+    (testing "Agent run succeeds"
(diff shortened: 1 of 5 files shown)


test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -27,3 +27,30 @@
           (is (= catalog-uuid report-uuid)))))))
 
 
+(deftest ^:integration rich-data-agent-run
+  (with-open [pg (int/setup-postgres)
+              pdb (int/run-puppetdb pg {})
+              ps (int/run-puppet-server [pdb] {:agent {:rich-data true}})]
+
+    (testing "Agent run succeeds"
+      (let [{:keys [out]} (int/run-puppet-as "my-agent" ps pdb "notify { 'foo':\n  message => {\n  'regex'     => /[a-z]+/,\n  'object'    => Object[{name => 'FindMe', 'attributes' => {'x' => String}}],\n  'version'   => SemVer('3.4.1'),\n  'semver'    => SemVerRange('>=3.4.1'),\n  'timestamp' => Timestamp('1994-03-25T19:30:00'),\n  'timespan'  => Timespan('3-10:00:00'),\n  }\n}" {:rich-data true})]
```
Zak-Kent: maybe break the manifest string across multiple lines so it's easier to read. Something like: 
 (let [manifest "notify { 'foo':\n
                                 message =>
                                 {\n  'regex'     => /[a-z]+/,
                                  \n  'object'    => Object[{name => 'FindMe', 'attributes' => {'x' => String}}],
                                  \n  'version'   => SemVer('3.4.1'),
                                   ... "])

When testing this locally it also looks like it works without the \n newline characters so maybe we can drop those out too.

test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -27,3 +27,30 @@
...
+        (is (re-find #">=3.4.1" out))))
```
Zak-Kent: I'm thinking we might want to add more checks here for the other rich data objects too, but I can help add these if needed.

Zak-Kent:
Closing in favor of PDB-4017

PR #2543, opened 2018-08-16
(maint) Bump to ezbake 1.8.6

mwaggett:
This commit bumps ezbake to 1.8.6, which is required to use packaging as a gem.


mwaggett:
that failure should be fixed by a ci-job-configs update I'm working on: https://github.com/puppetlabs/ci-job-configs/pull/4926

mwaggett:
closing this - there are some bugs we need to sort out, which will result in a new version of ezbake anyway

PR #2545, opened 2018-08-20
(PDB-3928) Add regex to facts-blacklist

Zak-Kent:
This commit adds support for regex in the facts-blacklist config 
option. Given a valid regex pattern it will not store facts where
a pattern matches a fact name. This is controlled by the use of a
:facts-blacklist-type flag which defaults to 'regular'. The default
preserves the existing facts-blacklist behavior. Setting this flag
to 'regex' allows a user to submit valid java regex patterns for use.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -296,12 +296,23 @@
 
 ;; Fact replacement
 
+(defn rm-facts-by-regex [facts-blacklist fact-map]
+  (let [blacklisted? (fn [fact-name]
+                       (some #(re-matches % fact-name) facts-blacklist))]
+    (apply dissoc fact-map (filter blacklisted? (keys fact-map)))))
+
 (defn replace-facts*
-  [{:keys [payload id received] :as command} start-time db facts-blacklist]
+  [{:keys [payload id received] :as command} start-time db
+   {:keys [facts-blacklist facts-blacklist-type] :as blacklist-config}]
   (let [{:keys [certname package_inventory] :as fact-data} payload
         producer-timestamp (:producer_timestamp fact-data)
+        blacklisting? (seq blacklist-config)
+        rm-blacklisted (when blacklisting?
+                         (case facts-blacklist-type
+                           "regex" (partial rm-facts-by-regex facts-blacklist)
+                           "literal" #(apply dissoc % facts-blacklist)))
         trimmed-facts (cond-> fact-data
-                        (seq facts-blacklist) (update :values #(apply dissoc % facts-blacklist))
+                        blacklisting? (update :values rm-blacklisted)
                         (seq package_inventory) (update :package_inventory distinct))]
     (jdbc/with-transacted-connection' db :repeatable-read
       (scf-storage/maybe-activate-node! certname producer-timestamp)
@@ -309,12 +320,12 @@
     (log-command-processed-messsage id received start-time :replace-facts certname)))
 
 (defn replace-facts
-  [{:keys [payload version received] :as command} start-time db facts-blacklist]
+  [{:keys [payload version received] :as command} start-time db blacklist-config]
   (replace-facts* (upon-error-throw-fatality
                     (assoc command :payload (fact/normalize-facts version received payload)))
                   start-time
                   db
-                  facts-blacklist))
+                  blacklist-config))
 
 ;; Node deactivation
 
@@ -387,12 +398,12 @@
 (defn process-command!
   "Takes a command object and processes it to completion. Dispatch is
    based on the command's name and version information"
-  [{command-name :command version :version delete? :delete? :as command} db facts-blacklist]
+  [{command-name :command version :version delete? :delete? :as command} db blacklist-config]
   (when-not delete?
     (let [start (now)]
       (condp supported-command-version? [command-name version]
         "replace catalog" (replace-catalog command start db)
-        "replace facts" (replace-facts command start db facts-blacklist)
+        "replace facts" (replace-facts command start db blacklist-config)
         "store report" (store-report command start db)
         "deactivate node" (deactivate-node command start db)))))
 
@@ -443,11 +454,11 @@
 (def quick-retry-count 4)
 
 (defn process-command-and-respond!
-  [{:keys [callback] :as cmd} db response-pub-chan stats-atom facts-blacklist]
+  [{:keys [callback] :as cmd} db response-pub-chan stats-atom blacklist-config]
   (try
     (let [result (call-with-quick-retry quick-retry-count
                   (fn []
-                    (process-command! cmd db facts-blacklist)))]
+                    (process-command! cmd db blacklist-config)))]
       (swap! stats-atom update :executed-commands inc)
       (callback {:command cmd :result result})
       (async/>!! response-pub-chan
@@ -491,7 +502,7 @@
   "Parses, processes and acknowledges a successful command ref and
   updates the relevant metrics. Any exceptions that arise are
   unhandled and expected to be caught by the caller."
-  [cmdref q scf-write-db response-chan stats facts-blacklist]
+  [cmdref q scf-write-db response-chan stats blacklist-config]
   (let [{:keys [command version] :as cmd} (queue/cmdref->cmd q cmdref)
         retries (count (:attempts cmdref))]
     (if-not cmd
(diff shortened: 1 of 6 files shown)


Zak-Kent:
Still need to add notes to the docs. Will update pr after review

src/puppetlabs/puppetdb/facts.clj
```diff
@@ -134,6 +136,23 @@
    (nil? data) 4
    (coll? data) 5))
 
+(defn fact-splitter [fact]
+  "Split fact-blacklist facts if they contain supported globbing chars: * or ?
+   Example: 'hell*0' outputs ['hell' '*' 'o']"
```
Zak-Kent: fix typo

src/puppetlabs/puppetdb/command.clj
```diff
@@ -296,25 +296,39 @@
...
+                        (seq facts-blacklist) (update :values remove-blacklist-facts)
                         (seq package_inventory) (update :package_inventory distinct))]
     (jdbc/with-transacted-connection' db :repeatable-read
       (scf-storage/maybe-activate-node! certname producer-timestamp)
       (scf-storage/replace-facts! trimmed-facts))
     (log-command-processed-messsage id received start-time :replace-facts certname)))
 
 (defn replace-facts
-  [{:keys [payload version received] :as command} start-time db facts-blacklist]
+  [{:keys [payload version received] :as command} start-time db fact-blist-map]
```
austb: Maybe this should be named something like `facts-blacklist-config`. I know it's wordy but bits are cheap and it may help prevent some confusion. There are numerous usages below as well, if we decide we want to change it, we should update those as well.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -296,25 +296,35 @@
...
 (defn replace-facts*
-  [{:keys [payload id received] :as command} start-time db facts-blacklist]
+  [{:keys [payload id received] :as command} start-time db
+   {:keys [facts-blacklist facts-blacklist-type] :as blacklist-config}]
   (let [{:keys [certname package_inventory] :as fact-data} payload
         producer-timestamp (:producer_timestamp fact-data)
+        rm-blacklist (when (seq blacklist-config)
+                       (case facts-blacklist-type
+                         "regex" (partial rm-facts-regex facts-blacklist)
+                         "literal" #(apply dissoc % facts-blacklist)))
```
rbrw: *If* blacklist-config will always be nil when there isn't one, then realized we could also just do this, also replacing the enclosing `when`: `(case facts-blacklist-type ... "regex"  ... "literal" ... nil identity)` and then below:
```
(cond-> (rm-blacklisted fact-data)
  (seq package_inventory) (update :package_inventory distinct))
```

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -72,6 +72,12 @@
     (fs/delete-dir (:path dlo))
     (#'overtone.at-at/shutdown-pool-now! @(:pool-atom delay-pool))))
 
+;; define blacklist map to allow use of with-redefs later
+(def blacklist-config {:facts-blacklist (map re-pattern ["^blacklisted-fact$"
```
rbrw: As per our conversation the other day, perhaps the anchors are irrelevant here?

rbrw: Can we just use `[#"pre.*"...]` instead of the `(map re-pattern ...)`?

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -721,10 +726,15 @@
...
+  (deftest regex-facts-blacklist
     (dotestseq [version fact-versions
-                :let [command (update facts :values #(assoc % "blacklisted-fact" "val"))]]
-      (testing "should ignore the blacklisted fact"
+                :let [command (update facts :values
+                                      #(assoc % "blacklisted-fact" "val"
+                                              "prefix" "val"
+                                              "facts-suff" "val"
+                                              "glob" "val"
+                                              "puppet" "val"))]]
```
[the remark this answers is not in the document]
rbrw: Since it's cheap, might be worth adding a "suff-nope" and maybe a "*blacklisted-fact*" just those other two cases we expect to pass through (i.e. trailing, and "all").

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -685,6 +690,25 @@
...
+                    cast(fv.value_boolean as text),
+                    cast(fv.value_float as text),
+                    '') as value,
+           fs.certname
+    FROM factsets fs
+         INNER JOIN facts as f on fs.id = f.factset_id
+         INNER JOIN fact_values as fv on f.fact_value_id = fv.id
+         INNER JOIN fact_paths as fp on f.fact_path_id = fp.id
+    WHERE fp.depth = 0
+    ORDER BY name ASC"))
```
rbrw: We don't have to do this right now, but if you're inclined, since we're reindenting anyway, we could convert this to one of the reindentation-friendly line-oriented styles.  As one way:
```
(->> ["select fp.path as name,"
      "       coalesce(fv.value_string,"
      ...]
     (apply str)
     query-to-vec)
```

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -721,50 +745,47 @@
...
-                   WHERE fp.depth = 0
-                   ORDER BY name ASC")
+          (is (= (approx-facts-query)
                  [{:certname certname :name "a" :value "1"}
                   {:certname certname :name "b" :value "2"}
-                  {:certname certname :name "c" :value "3"}]))))))
+                  {:certname certname :name "c" :value "3"}
+                  {:certname certname :name "notprefix" :value "val"}]))))))
+
+    (deftest literal-facts-blacklist
```
rbrw: Wondering why the diff thinks this function is completely new -- I suspect it might be ordering.  If you feel like it, might try putting this test before the regex test and see if that shrinks the diff (and makes what really happened clearer).

test/puppetlabs/puppetdb/config_test.clj
```diff
@@ -61,6 +62,76 @@
...
+      (is (thrown+-with-msg? [:type ::conf/cli-error]
+                             #".*Unclosed character class near index 4\.*"
+                             (-> {:database {:subname "bar"
+                                             :facts-blacklist-type "regex"
+                                             :facts-blacklist ["^foo[" "(bar.*"]}}
+                                 convert-blacklist-config))))
+
+    (testing "convert-blacklist-config is no-op when set to literal"
+      ;; with blacklist-type set to literal all blacklist entries kept as literal strings
+      (is (every? string? (-> {:database {:subname "bar"
```
rbrw: Could this just be `(is (= ["expected-1" "expected-2" ...] (-> ...))`

test/puppetlabs/puppetdb/command_test.clj
rbrw: Nevermind - forgot about indentation changes, if nothing else.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -685,6 +690,26 @@
...
+        "                 cast(fv.value_float as text),"
+        "                 '') as value,"
+        "        fs.certname"
+        " FROM factsets fs"
+        "      INNER JOIN facts as f on fs.id = f.factset_id"
+        "      INNER JOIN fact_values as fv on f.fact_value_id = fv.id"
+        "      INNER JOIN fact_paths as fp on f.fact_path_id = fp.id"
+        " WHERE fp.depth = 0"
+        " ORDER BY name ASC"]
+       (apply str)
```
rbrw: Hmm, wondering about changing this to "(apply str/join "\n")" so that we don't need the leading spaces, and assuming it works, that might also make the sql log output nicer (though not postiive about that).  Or we can just join on " ".

test/puppetlabs/puppetdb/config_test.clj
```diff
@@ -61,6 +62,78 @@
     (testing "should default to half the available CPUs, even on single core boxes"
       (is (= (with-ncores 1) 1)))))
 
+(deftest blacklist-configuration
+  (testing "blacklist-configuration"
+    (testing "convert-blacklist-config produces regex patterns and errors when patterns invalid"
```
rbrw: Not critical, but nesting `testing` forms just appends the strings, so this is going to be a pretty long description.  In any case, I'd be tempted to choose shorter names, and since I don't think there's much per-test overhead here, also tempted to just break these up into smaller deftest-chunks, with meaningful test names instead of the testing strings.  That way when something breaks, we can run smaller pieces, say: `(deftest blacklist-conversion-returns-patterns-when-appropriate ...)`, `(deftest black-conversion-returns-errors-for-bad-patterns ...)`, etc.

rbrw:
Looks good.

PR #2554, opened 2018-08-27
(maint) Add ability to create orphaned facts to dev utils tool

Zak-Kent:
This adds the ability to create orphaned facts so we can test fact_path gc queries. There is still some work that needs to be done to consolidate all of this into one command. For the time being you need to run this tool twice in order to create the orphaned facts. Once like './pdb simulate 100 10 true' where the true flag creates the extra nested facts, followed by './pdb simulate 100 10' which will recreate the facts without the extra nested ones which results in the orphans being created. 

There is also an issue where this script seems to hang after it has submitted it's facts. This appears to be unrelated to this pr and will be addressed later.

### DIFF
--- a/util/pdb/commands.py
+++ b/util/pdb/commands.py
@@ -3,6 +3,8 @@ import time
 import json
 import random
 from collections import deque
+import random
+import copy
 
 
 def update_nums(values):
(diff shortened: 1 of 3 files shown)


Zak-Kent:
Closing because pr is old and we already used this to test the orphaned fact path issue. I'll keep the branch hanging around if we ever need to revisit.

PR #2559, opened 2018-08-28
(PDB-3993) Add upgrade cmd to upgrade db, queue, and then exit

rbrw:


src/puppetlabs/puppetdb/core.clj
```diff
@@ -1,90 +1,49 @@
...
-        allowed?   (available-subcommands)]
-
+  [success-fn fail-fn [subcommand & args]]
+  (let [run (case subcommand
+              "version" #(apply cver/-main args)
+              "services" #(apply svcs/-main args)
+              "benchmark" #(apply benchmark/-main args)
+              "fact-storage-benchmark" #(apply fstore-bench/-main args)
+              (do
+                ;; FIXME: this should be *err*
```
Zak-Kent: Does this need to be changed to *err* ?

PR #2571, opened 2018-08-30
(PDB-4020) Return the last successful reponse with command_broadcast

jarretlavallee:
Prior to this PR, the http submission with command_broadcast=true
always returned the last response. As a result, a `:server_error` would be returned if
the last connection produced a 503 response, even if there was a successful PuppetDB response and the minimum successful responses have been met. 

As a result, any configuration with `command_broadcast=true` and the last PuppetDB in the instance in the `server_urls` in maintenance mode would result in failed agent runs for the duration of the maintenance mode. 

This PR changes the behavior to send the last successful response when the minimum successful submissions have been met.

### DIFF
--- a/puppet/lib/puppet/util/puppetdb/http.rb
+++ b/puppet/lib/puppet/util/puppetdb/http.rb
@@ -162,6 +162,7 @@ module Puppet::Util::Puppetdb
     def self.broadcast_action(path_suffix, server_urls, http_callback)
       response = nil
       response_error = nil
+      last_success = nil
       config = Puppet::Util::Puppetdb.config
       successful_submit_count = 0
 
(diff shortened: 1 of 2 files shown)


Zak-Kent:
@jarretlavallee we recently had an issue with our internal tests which is the cause of the jenkins error you hit. Rebasing this pr on 5.2.x should clear that up, thanks.

jarretlavallee:
@Zak-Kent Thanks! I just rebased it.

Zak-Kent:
LGTM, thanks @jarretlavallee!

Going to close this pr, cherry-pick the commit, and put up a pr against 5.1.x [here](https://github.com/puppetlabs/puppetdb/pull/2578).

PR #2576, opened 2018-09-06
(PDB-3942) OpenJDK10 cacerts do not trust clojars by default

austb:


ext/bin/require-jdk
```diff
@@ -51,3 +54,9 @@ mkdir -p "$install"
 
 # Piping to tail to get the last line is necessary until https://github.com/sormuras/bach/issues/35 is resolved
 $tmpdir/install-jdk.sh --emit-java-home -f "$jdkver" -l "$licensetype" --target "$install/jdk"
+
+# The unpatched version of openjdk10 does not contain the cacert that signed clojars
+if test "${3:-}" = "--travis" -a $licensetype = GPL; then
+  rm "$install/jdk/lib/security/cacerts"
+  ln -s /etc/ssl/certs/java/cacerts "$install/jdk/lib/security/cacerts"
+fi
```
rbrw: Wondering if we might want to just always do this on linux (for all jdks).

.travis.yml
```diff
@@ -24,7 +24,7 @@ aliases:
     if test "$(uname)" = Linux -a "$jdkver" -lt 9 ; then
         jdk_switcher use "$jdk";
     elif test "$(uname)" = Linux; then
-        ext/bin/require-jdk "$jdk" ext/travisci/local
+        ext/bin/require-jdk "$jdk" ext/travisci/local --travis
```
rbrw: Probably can drop the --travis arg and the checking in the script now?

ext/bin/require-jdk
```diff
@@ -49,5 +50,8 @@ fi
 chmod +x "$tmpdir/install-jdk.sh"
 mkdir -p "$install"
 
-# Piping to tail to get the last line is necessary until https://github.com/sormuras/bach/issues/35 is resolved
 $tmpdir/install-jdk.sh --emit-java-home -f "$jdkver" -l "$licensetype" --target "$install/jdk"
```
rbrw: Let's quote `"$tmpdir/install-jdk.sh"` so we don't have to worry about the tmpdir chars.

PR #2577, opened 2018-09-07
(PDB-3952) Fix jdbc shutdown exceptions

rbrw:


test/puppetlabs/puppetdb/testutils/db.clj
```diff
@@ -63,9 +63,8 @@
       (assert (valid-sql-id? name))
       name)))
 
-(defn init-db [db read-only?]
-  (jdbc/with-db-connection db (migrate! db))
-  (jdbc/pooled-datasource (assoc db :read-only? read-only?)))
```
Zak-Kent: Why did we need to remove the pooled-datasource here? I don't really know the context behind what it's doing and if it's ok to remove.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -615,7 +624,11 @@
              :response-chan response-chan
              :response-mult response-mult
              :response-chan-for-pub response-chan-for-pub
-             :response-pub (async/pub response-chan-for-pub :id))))
+             :response-pub (async/pub response-chan-for-pub :id)
+             ;; This coordination is needed until we no longer
+             ;; support jdk < 10.
+             ;; https://bugs.openjdk.java.net/browse/JDK-8176254
+             :activities (atom {:executing-delayed 0}))))
```
Zak-Kent: maybe give the two activities atoms names that are specific to what they're keeping track of

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -470,3 +470,14 @@
...
   [m]
   (into {} (filter val m)))
+
+(defn wait-for-ref-state [ref ms pred]
+  (let [watch-key wait-for-ref-state
+        finished? (promise)
+        handle-state #(when (pred %) (deliver finished? true))]
+    (add-watch ref watch-key (fn [_ _ _ new] (handle-state new)))
+    (try
+      (handle-state @ref)
```
Zak-Kent: I'm wondering if there is anyway for this to return true even though there is a gc or other event we're wanting to wait for in progress before the timeout expires. In the gc case when we set the atom to :stopping right before calling wait-for-ref-state won't that cause our atom to have :stopping in the set so that the handle-state call in the try block will deliver true everytime without waiting for the timeout, or at least in most cases?

Zak-Kent: On second thought, maybe this is already handled because if gc was going on you would get a set #{:stopping :collecting-garbage} and then the pred would return false and we'd still be waiting on an update from the watcher.

Zak-Kent:
LGTM

src/puppetlabs/puppetdb/command.clj
```diff
@@ -652,11 +666,17 @@
...
-      (stop-and-reset-pool! delay-pool))
+  (stop [this {:keys [activities consumer-threadpool command-chan delay-pool]
+               :as context}]
+    (some-> command-chan async/close!)
+    (some-> consumer-threadpool gtp/shutdown)
+    (some-> delay-pool stop-and-reset-pool!)
+    ;; Wait up to ~5s for https://bugs.openjdk.java.net/browse/JDK-8176254
+    (swap! activities #(assoc % :stopping true))
+    (if (utils/wait-for-ref-state activities 5000
+                                  #(= % {:stopping true :executing-delayed 0}))
```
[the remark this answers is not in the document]
rbrw: Yep.

src/puppetlabs/puppetdb/utils.clj
rbrw: Right -- all of this rests on the fact that the atom's state changes are...atomic, so that the state can only be #{}, #{:stopping}, or #{:stopping :collecting-garbage}, and also the fact that the :stopping state is "terminal", i.e. once added, it's never removed -- and if the garbage collection task sees :stopping, it bails out.

There is still a narrow period between when the gc task decides not to run (the if/when), and when the top-level enclosing function returns (presumably killing the lingering thread, given the bug), but we're ignoring that since it's not causing any noise, and it shouldn't be able to do any real harm, since all our state is in durable stores.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1776,3 +1778,54 @@
...
+                                                  ;; do nothing
+                                                  nil)]
+        (with-pdb-with-no-gc
+          (let [pdb (get-service *server* :PuppetDBServer)
+                q (-> pdb cli-svc/shared-globals :q)
+                dispatcher (get-service *server* :PuppetDBCommandDispatcher)
+                stop-status (-> dispatcher service-context :stop-status)
+                v5-catalog (get-in wire-catalogs [5 :empty])
+                cmdreqs (repeatedly msg-count #(catalog->command-req 5 v5-catalog))
+                dummy-ex (Exception.)]
```
Zak-Kent: Are we using this anywhere?

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1776,3 +1778,54 @@
...
+            (doseq [cmdreq cmdreqs
+                    :let [cmdref (queue/store-command q cmdreq)
+                          {:keys [command version]} cmdref]]
+              ;; schedule-delayed-message assumes the metrics already exist
+              (cmd/create-metrics-for-command! command version)
+              (utils/noisy-future
+               (cmd/schedule-delayed-message cmdref (Exception.)
```
Zak-Kent: maybe the exception here was supposed to be the dummy-ex?

Zak-Kent:
LGTM

test/puppetlabs/puppetdb/command_test.clj
rbrw: Yeah, just removed it -- decided it didn't add enough information to be worth the indirection.

PR #2582, opened 2018-09-10
(PDB-4017) Accept and test rich data in terminus

austb:


.travis.yml
```diff
@@ -98,7 +104,7 @@ jobs:
       script: *run-integration-tests
 
     - stage: ❧ pdb tests
-      env: PDB_TEST=int/oraclejdk10/pup-master/srv-master/pg-9.6
+      env: PDB_TEST=int/rich/oraclejdk10/pup-master/srv-master/pg-9.6
```
rbrw: Hmm, I'm not sure this is the set of configurations we want, but we can figure that out and adjust if needed later/soon.

test/puppetlabs/puppetdb/integration/reports.clj
```diff
@@ -101,6 +101,86 @@
...
+     {:name "string_with_double_quote" :code "\"he said \\\"hi\\\"\"" :expected "he said \"hi\""}
+     {:name "regexp" :code "/[a-z]+/" :expected "/[a-z]+/"}
+     {:name "deferred" :code "Deferred('join', [[1, 2, 3], ':'])" :expected "1:2:3"}
+     {:name "sensitive_deferred" :code "Sensitive(Deferred('join', [[1, 2, 3], ':']))" :expected-old-value "[redacted]" :expected "[redacted]"}
+     {:name "sensitive" :code "Sensitive('password')" :expected-old-value "[redacted]" :expected "[redacted]"}
+     {:name "timestamp" :code "Timestamp('2012-10-10')" :expected "2012-10-10T00:00:00.000000000 UTC"}
+     {:name "hash_and_array" :code "{'a' => [1, 2, 3], 'b' => 'hello'}" :expected "{\"a\"=>[1, 2, 3], \"b\"=>\"hello\"}"}
+     {:name "special_key" :code "{ ['special', 'key'] => 10 }" :expected "{\"[\\\"special\\\", \\\"key\\\"]\"=>10}"}
+     {:name "hash_with_sensitive_val" :code "{ x => Sensitive(hush) }" :expected #"\{\"x\"=>\"#<Sensitive \[value redacted\]:[0-9]+>\"}"}
+     {:name "hash_with_sensitive_key" :code "{Sensitive(hush) => 42 }" :expected #"\{\"#<Sensitive \[value redacted\]:[0-9]+>\"=>42\}"}
```
rbrw: Perhaps make these longer ones at least two lines so they're easily visible in gh, etc.

test/puppetlabs/puppetdb/integration/reports.clj
```diff
@@ -101,6 +101,86 @@
...
+     {:name "literal_integer" :code "47" :expected "47"}
+     {:name "literal_float" :code "3.14" :expected "3.14"}
+     {:name "literal_true" :code "true" :expected "true"}
+     {:name "literal_string" :code "\"hello\"" :expected "hello"}
+     {:name "string_with_single_quote" :code "\"ta'phoenix\"" :expected "ta'phoenix"}
+     {:name "string_with_double_quote" :code "\"he said \\\"hi\\\"\"" :expected "he said \"hi\""}
+     {:name "regexp" :code "/[a-z]+/" :expected "/[a-z]+/"}
+     {:name "deferred" :code "Deferred('join', [[1, 2, 3], ':'])" :expected "1:2:3"}
+     {:name "sensitive_deferred" :code "Sensitive(Deferred('join', [[1, 2, 3], ':']))" :expected-old-value "[redacted]" :expected "[redacted]"}
+     {:name "sensitive" :code "Sensitive('password')" :expected-old-value "[redacted]" :expected "[redacted]"}
```
rbrw: Maybe change :expected-old-value to :expected-old to match :expected?

test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -27,3 +28,94 @@
           (is (= catalog-uuid report-uuid)))))))
```
rbrw: Is the link to the friendly format doc (examples, etc.) "stable"?  If so, let's put it in a comment here.

test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -27,3 +28,94 @@
...
+                ps (int/run-puppet-server [pdb] {:agent {:rich-data true}})]
+
+      (testing "Agent run succeeds"
+        (let [manifest (str "type Car = Object[attributes => {regnbr => String}];\n"
+                            (apply str (for [{:keys [name code]} rich-data-tests]
+                                         (str "notify { '" name "':"
+                                              "  message => " code
+                                              "}"))))
+              {:keys [out]} (int/run-puppet-as "my-agent" ps pdb manifest {:rich-data true})]
+          (is (re-find #"a-z" out))))
```
rbrw: This is actually looking for the three literal characters in sequence "a-z"?

test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -27,3 +28,94 @@
...
+
+      (testing "Agent run data can be queried"
+        (are [query result] (= result (int/pql-query pdb query))
+          "nodes[certname] {}" [{:certname "my-agent"}]
+          "catalogs[certname, environment] {}" [{:certname "my-agent", :environment "production"}]
+          "factsets[certname, environment] {}" [{:certname "my-agent", :environment "production"}]
+          "reports[certname, environment] {}" [{:certname "my-agent", :environment "production"}]))
```
rbrw: If you prefer this, no worries, but these days I'm tempted to use "is" instead of "are" even if it's a little chattier so that the line numbers will be right when something goes wrong.

test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -27,3 +28,94 @@
...
+
+      (testing "Catalog resource parameters stored in proper format"
+        (let [[{{all-resources :data} :resources} :as result] (int/pql-query pdb "catalogs[certname, resources] { resources { type = 'Notify' and title = 'regexp' } }")
```
rbrw: Let's wrap this, perhaps either with some additional (intermediate) let bindings, or some threading forms, or even just

    (let [[...........something long............]
           (some-long-call ...)

           next-binding ...

test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -27,3 +28,94 @@
...
+              filtered-resources (filter
+                                   #(not (or (= "Stage" (:type %))
+                                             (and (= "Class" (:type %))
+                                                  (or (= "main" (:title %))
+                                                      (= "Settings" (:title %))))))
```
rbrw: I'd need to think about the logic for a minute, but could this be a bit shorter as a remove?  i.e. (remove #(or ...

rbrw: Also, not critical, but I'd be tempted to just shorten this to "resources", since it's the resources of interest, the scope is localized, and we already have all-resources as a comparative term.

test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -27,3 +28,94 @@
...
+              (if (:valid? test-resource)
+                (is ((:valid? test-resource) stored-value))
+                (is (= (:expected test-resource) stored-value)))))))
+
+      (testing "transaction-uuid"
+        (let [catalog-uuid (first (int/pql-query pdb "catalogs [transaction_uuid] {}"))
+              report-uuid (first (int/pql-query pdb "reports [transaction_uuid] {}"))]
+          (testing "is available from puppetdb"
+            (is (not (nil? catalog-uuid)))
+            (is (not (nil? report-uuid))))
```
rbrw: Are we specifically looking for `nil` if so, then this is just right, but if not, perhaps change this to just `(is (not catalog-uuid)), etc.

test/puppetlabs/puppetdb/integration/smoke.clj
austb: That one is straight copy-pasta from the above test, so I'm tempted to leave it unless you want to change both

test/puppetlabs/puppetdb/integration/smoke.clj
austb: this is another one taken from above, so I'm tempted to leave it as is

test/puppetlabs/puppetdb/integration/smoke.clj
austb: Ya, that's left over from when the test only used Regex, so the notify was "/[a-z]+/" and that made sure that it had been notified to the stdout of the puppet run

test/puppetlabs/puppetdb/integration/smoke.clj
```diff
@@ -27,3 +28,94 @@
           (is (= catalog-uuid report-uuid)))))))
 
 
+(when test-rich-data?
+  (def rich-data-tests
```
Zak-Kent: Are we holding off on testing the alien objects for the time being? I just noticed them in the doc but I'm not sure what they are exactly and if we need to be looking out for them.

austb: Yeah we are waiting on that (but I need to open a ticket to add those), they would require also loading some ruby code to bootstrap those objects, which we don't have the infrastructure for in place right now

test/puppetlabs/puppetdb/integration/smoke.clj
austb: I'm hesistant to link to the doc because it's inaccurate, but I can create a new one that's accurate and link it here if you want

rbrw: Oh, I meant *their* doc.  If puppet doesn't have an official one yet for this format, then we can wait, but we should eventually have something to refer to.

test/puppetlabs/puppetdb/integration/smoke.clj
rbrw: Not critical either way.

PR #2585, opened 2018-09-12
(PDB-4080) Fix facts blacklist->vector type checking

Zak-Kent:


src/puppetlabs/puppetdb/schema.clj
```diff
@@ -50,12 +50,12 @@
...
+  [fl]
   (cond
-    string? (->> (str/split fact-list #",")
-                 (map str/trim)
-                 (apply vector))
-    sequential? (vec fact-list)
+    (string? fl) (->> (str/split fl #",")
+                      (map str/trim)
+                      (apply vector))
+    (and (coll? fl) (every? string? fl)) (vec fl)
```
rbrw: Not that it really matters, but do we really need it to be a vector instead of say a list or whatever?

test/puppetlabs/puppetdb/schema_test.clj
```diff
@@ -44,6 +44,15 @@
        :foo :foo
        :foo (s/required-key :foo)))
 
+(deftest blacklist->vector-test
+  (are [x y] (= x (blacklist->vector y))
+    ["baz" "bar"] "baz, bar"
+    ["foo" "bar"] ["foo" "bar"])
```
rbrw: If you'd rather keep "are", that's fine, but I tend to favor just writing out the "is" statements these days so that the line numbers will be right when they fail.

src/puppetlabs/puppetdb/schema.clj
Zak-Kent: We expect a vector in the schema conversion/validation stuff so this has to return a vector. We could change the validation code on a second pass at this but for now it should be good.

test/puppetlabs/puppetdb/schema_test.clj
Zak-Kent: I can rebase and do these with is

src/puppetlabs/puppetdb/schema.clj
rbrw: No, that reason's plenty good.

PR #2593, opened 2018-09-13
(PDB-4082) Add preliminary rich data documentation

rbrw:


Zak-Kent:
LGTM

documentation/api/query/v4/query.markdown
```diff
@@ -95,6 +95,19 @@ Most PuppetDB query endpoints support paged results via a set of shared URL para
 
 All queries return data with a content type of `application/json`. Each endpoint's page describes the format of its return data.
 
+### Rich data
+
+Puppet 6 supports
+[rich_data][https://github.com/puppetlabs/puppet-specifications/blob/master/language/types_values_variables.md#richdata]
```
austb: The URL needs `( )`  around it, but has `[ ]`

documentation/api/query/v4/query.markdown
```diff
@@ -95,6 +95,19 @@ Most PuppetDB query endpoints support paged results via a set of shared URL para
 
 All queries return data with a content type of `application/json`. Each endpoint's page describes the format of its return data.
 
+### Rich data
+
+Puppet 6 supports
+[rich_data](https://github.com/puppetlabs/puppet-specifications/blob/master/language/types_values_variables.md#richdata)
+types like Timestamp and SemVer, and enables rich data by default.
+When rich data is enabled, readable string representations of rich
+data values may appear in the report resource event `old_value` and
```
austb: Maybe this should be "will appear"? "may appear" may give them the impression that they have some way of controlling it

documentation/api/query/v4/query.markdown
```diff
@@ -95,6 +95,19 @@ Most PuppetDB query endpoints support paged results via a set of shared URL para
...
+`new_value` fields, and in catalog parameter values.
+
+For example, a Timestamp value would be recorded in PuppetDB as a
+string like "2012-10-10T00:00:00.000000000 UTC", and a custom Sensor
+object would be recorded as a string like "Sensor({'temp' => '43'})".
```
austb: I would take out the Sensor example since the wire format for custom objects is not functional. You could instead use Binary or Sensitive/Deferred

documentation/api/query/v4/query.markdown
rbrw: Hmm, well I went with "may" because even if rich data's enabled, most people won't see any values right now?  i.e. just because it's turned on doesn't mean it'll show up in any given place, and the other stuff will look the way it always has, or is that also not quite right?

PR #2602, opened 2018-09-20
(PDB-4108) Fix AOT issues and compile uberjar

rbrw:


project.clj
```diff
@@ -213,12 +225,6 @@
   :deploy-repositories [["releases" ~(deploy-info "https://artifactory.delivery.puppetlabs.net/artifactory/list/clojure-releases__local/")]
                         ["snapshots" ~(deploy-info "https://artifactory.delivery.puppetlabs.net/artifactory/list/clojure-snapshots__local/")]]
 
-  ;; By declaring a classifier here and a corresponding profile below we'll get an additional jar
-  ;; during `lein jar` that has all the code in the test/ directory. Downstream projects can then
-  ;; depend on this test jar using a :classifier in their :dependencies to reuse the test utility
-  ;; code that we have.
-  :classifiers  [["test" :testutils]]
```
austb: Is this going to break compatibility with pe extensions? https://github.com/puppetlabs/pe-puppetdb-extensions/blob/master/project.clj#L88

src/puppetlabs/puppetdb/cli/util.clj
```diff
@@ -0,0 +1,41 @@
...
+;; Testing hook
+(defn java-version [] (System/getProperty "java.version"))
+
+(defn jdk-support-status [version]
+  "Returns :official, :tested, or :unknown, or :no."
+  (cond
+    (re-matches #"1.[123456]($|(\..*))" version) :no
+    (re-matches #"1.7($|(\..*))" version) :unknown
+    (re-matches #"1.8($|(\..*))*" version) :official
+    (re-matches #"1.10($|(\..*))*" version) :tested
```
austb: I believe they switched to using the java version with Java 10, so instead of `1.10.2` it's `10.0.2`. Maybe that's just in the cli --version flag. I'll check on that.

src/puppetlabs/puppetdb/cli/util.clj
```diff
@@ -0,0 +1,41 @@
...
+    :else :unknown))
+
+(defn jdk-unsupported-msg [version]
+  (let [status (jdk-support-status version)]
+    (case status
+      (:official :tested :unknown) nil
```
austb: Would it be difficult to log a warning when the java version is untested since we don't want to load the logging libraries?

src/puppetlabs/puppetdb/core.clj
```diff
@@ -24,39 +21,33 @@
...
-        (catch Throwable e
-          (logging-utils/catch-all-logger e)
-          (fail-fn))))))
+    (do (usage *out*) 0)
+    (do (usage *err*) err-exit-status)))
+
+;; Resolve the subcommands dynamically to avoid loading the world just
+;; to print the version.
+(defn run-resolved [cli-name fn-name args]
+  (let [namespace (-> (symbol (str "puppetlabs.puppetdb.cli." cli-name)))]
```
austb: Is the `->` doing anything here?

project.clj
```diff
@@ -100,6 +100,18 @@
          "puppetserver"
          "vendor"]))
 
+(def pdb-aot-namespaces
+  (apply vector
+         #"puppetlabs.puppetdb\..*"
```
austb: Only one of the `.` is escaped here

rbrw: Yeah, that's intentional, i.e. we're looking for all the namespaces that start with "puppetlabs.puppetdb.".

project.clj
rbrw: Umm, how about I actually test this pr against pdbext?  And nice catch.  *That's* the case we weren't thinking of when we were trying to figure out who uses that jar.  We'll need to resolve this before we can proceed -- I'll investigate.

src/puppetlabs/puppetdb/cli/util.clj
rbrw: Hmm, first let me consider what I think about warning here, i.e. is that common, and/or what we'd want regardless, etc.

src/puppetlabs/puppetdb/core.clj
rbrw: Nope, think it's probably vestigial.

project.clj
rbrw: Tested against pdbext, and it was fine (after restoring the test jar).

src/puppetlabs/puppetdb/cli/util.clj
rbrw: (I think we decided for now to make sure the docs are clear, and then maybe revisit this question later, if we like...)

src/puppetlabs/puppetdb/cli/util.clj
austb:
```
Java HotSpot(TM) 64-Bit Server VM 10.0.2+13

user=> (System/getProperty "java.version")
"10.0.2"
user=>
```

project.clj
rbrw: Hah, oh, uh, right.  Thanks.

rbrw: Looks like the remove below it might be wrong too.

rbrw:
Forgot to include cli.util-test (added).

PR #2605, opened 2018-09-21
(maint) 5.2.x mergeup to 6.0.x

Iristyle:


Iristyle:
Self-approving PR to keep the ball moving given the same limited changeset as the previous merge-ups.

PR #2606, opened 2018-09-21
6.0.x mergeup to master

Iristyle:
@rbrw / @Zak-Kent we've merged these changes up through 6.0.x... and this is the final PR up to master to get some of the changes necessary to build Docker containers on Windows.

There was one merge conflict in `project.clj`, but I took what was on master - `(def pdb-version "6.1.0-SNAPSHOT")` from https://github.com/puppetlabs/puppetdb/blob/master/project.clj#L1


Zak-Kent:
@Iristyle that was the right one to keep, thanks

PR #2609, opened 2018-09-28
(PDB-3952) Fix wait in test-stop-with-periodic-gc-running

rbrw:
Don't wait when we're testing the case where we intentionally block
the gc.  Otherwise the test takes much longer than it should.


rbrw:
Wrong branch...

PR #2614, opened 2018-10-01
(PDB-4015) throw an error when a date-time is unparseable

robdaemon:
Instead of returning nil when a user provides an invalid date-time, we should return an error instead, as opposed to giving a misleading result set.


src/puppetlabs/puppetdb/time.clj
```diff
@@ -8,7 +8,8 @@
            (org.joda.time Period ReadablePeriod PeriodType DateTime))
   (:require [clj-time.coerce :as tc]
             [clj-time.format :as tf]
-            [schema.core :as s]))
+            [schema.core :as s]
+            [clojure.string :as str]))
```
austb: Is this dep used?

test/puppetlabs/puppetdb/time_test.clj
```diff
@@ -1,4 +1,5 @@
 (ns puppetlabs.puppetdb.time-test
+  (:import (org.joda.time DateTime DateTimeZone))
```
austb: I think we usually have `:import` after the `:require`. Not sure if that has a reason or if we want to try to maintain consistency. Maybe @rlb will know?

src/puppetlabs/puppetdb/time.clj
robdaemon: No, I'll remove it

test/puppetlabs/puppetdb/time_test.clj
rbrw: Yeah, imports normally go second.

robdaemon: Okay, I'll move them in this file and time.clj.

robdaemon:
Closing this in favor of a different approach

PR #2635, opened 2018-10-05
(PDB-4081) adding perf tuning params to postgresql config

robdaemon:
Sets a few tuning parameters to make the default postgresql installation deal
with the write-heavy load that puppetdb has.

This is best combined with disabling fsync during startup, via:

    pg_ctl -o "-F" start

### DIFF
--- a/ext/bin/pdbbox-init
+++ b/ext/bin/pdbbox-init
@@ -176,4 +176,52 @@ ssl-key = $abs_pdbbox/ssl/pdb.key.pem
 disable-update-checking = true
 EOF
 
+# do some tuning on PostgreSQL
+
+# generally accepted configuration parameters:
+# Set the shared_buffers to be 1/4 of the amount of RAM
+
+case "$OSTYPE" in
+    darwin*)
+        total_ram_bytes=$(sysctl hw.memsize | awk '{ print $2 }')
+        total_ram_kb=$(($total_ram_bytes / 1024)) ;;
+    *)
+        total_ram_kb=$(grep MemTotal: /proc/meminfo | awk '{ print $2 }') ;;
+esac
+
+total_ram_mb=$(($total_ram_kb / 1024))
+# these values come from pgtune and the postgresql tuning wiki
+
+# shared buffers should max out at 1 GB of RAM
+shared_buffers=$(($total_ram_mb / 4))
+if (( $shared_buffers > 1024 )); then
+    shared_buffers=1024
+fi
+
+# do not let this be less than the default value
+maintenance_work_mem=$(($total_ram_mb / 15))
+if (( $maintenance_work_mem < 64 )); then
+    maintenance_work_mem=64
+fi
+
+cat >> "$pdbbox/pg/data/postgresql.conf" <<EOF
+
+# tuning parameters
+shared_buffers = ${shared_buffers}MB
+
+default_statistics_target = 50
+
+maintenance_work_mem = ${maintenance_work_mem}MB
+
+# aim to finish by the time 90% of the next checkpoint is here
+# this will spread out writes
+checkpoint_completion_target = 0.9
+
+work_mem = 96MB
+
+# Increasing wal_buffers from its tiny default of a small number of kilobytes
+# is helpful for write-heavy systems, 16MB is the effective upper bound
+wal_buffers = 8MB
+EOF
+
 touch "$pdbbox/pdbbox"


ext/bin/pdbbox-init
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -ueo pipefail
+set -uexo pipefail
```
Zak-Kent: Maybe set -x after we do the pg password setup? Might not matter because we just use this in testing, but something I wondered about.

robdaemon: Ooooops I did not mean to leave the -x in there.

ext/bin/pdbbox-init
```diff
@@ -176,4 +176,40 @@ ssl-key = $abs_pdbbox/ssl/pdb.key.pem
...
+# Set the shared_buffers to be 1/4 of the amount of RAM
+
+# for development only (considered unsafe in production):
+# Turn off fsync
+# Turn off synchronous commit
+
+case "$OSTYPE" in
+    darwin*)
+        total_ram_bytes=$(sysctl hw.memsize | awk '{ print $2 }') 
+        total_ram_kb=$(expr $total_ram_bytes / 1024) ;;
```
austb: In the shell `$( )` will spin off a subshell to evaluate the command, whereas (I believe) arithmetic expansion `$(( ))` will not.

ext/bin/pdbbox-init
```diff
@@ -176,4 +176,40 @@ ssl-key = $abs_pdbbox/ssl/pdb.key.pem
...
+# generally accepted configuration parameters:
+# Set the shared_buffers to be 1/4 of the amount of RAM
+
+# for development only (considered unsafe in production):
+# Turn off fsync
+# Turn off synchronous commit
+
+case "$OSTYPE" in
+    darwin*)
+        total_ram_bytes=$(sysctl hw.memsize | awk '{ print $2 }')
```
rbrw: Not critical, but suspect we might be able to avoid the external awk via "${parameter##word}", etc. expansions.  Though not worth it if it gets too complicated...

ext/bin/pdbbox-init
```diff
@@ -176,4 +176,40 @@ ssl-key = $abs_pdbbox/ssl/pdb.key.pem
...
+# for development only (considered unsafe in production):
+# Turn off fsync
+# Turn off synchronous commit
+
+case "$OSTYPE" in
+    darwin*)
+        total_ram_bytes=$(sysctl hw.memsize | awk '{ print $2 }') 
+        total_ram_kb=$(expr $total_ram_bytes / 1024) ;;
+    *)
+        total_ram_kb=$(cat /proc/meminfo | grep MemTotal: | awk '{ print $2 }')
```
rbrw: Suggest omitting the `cat`, i.e. `grep MemTotal: /proc/meminfo | ...`

ext/bin/pdbbox-init
```diff
@@ -176,4 +176,40 @@ ssl-key = $abs_pdbbox/ssl/pdb.key.pem
...
+# Turn off fsync
+# Turn off synchronous commit
+
+case "$OSTYPE" in
+    darwin*)
+        total_ram_bytes=$(sysctl hw.memsize | awk '{ print $2 }') 
+        total_ram_kb=$(expr $total_ram_bytes / 1024) ;;
+    *)
+        total_ram_kb=$(cat /proc/meminfo | grep MemTotal: | awk '{ print $2 }') 
+esac
```
rbrw: Might not matter, but tempted to include the `;;` termination line for the final clause.

ext/bin/pdbbox-init
```diff
@@ -176,4 +176,40 @@ ssl-key = $abs_pdbbox/ssl/pdb.key.pem
...
+
+total_ram_mb=$(expr $total_ram_kb / 1024)
+# these values come from pgtune and the postgresql tuning wiki
+shared_buffers=$(expr $total_ram_mb / 4)
+maintenance_work_mem=$(expr $total_ram_mb / 15)
+
+cat >> "$pdbbox/pg/data/postgresql.conf" <<EOF
+
+# tuning parameters
+fsync = off
```
rbrw: I think I'm inclined to leave this off and just make sure we suggest using `pg_ctl start -F ...` in our docs, so that the default is still safe.

ext/bin/pdbbox-init
```diff
@@ -176,4 +176,40 @@ ssl-key = $abs_pdbbox/ssl/pdb.key.pem
...
+synchronous_commit = off
```
rbrw: If we don't yet know that this actually gives a substantial boost to test speed or something, then I lean toward leaving it off for now too.  i.e. I'm somewhat inclined to just start with options that ensure we're at least not using the overly conservative default configuration that pg is famous for.

ext/bin/pdbbox-init
robdaemon: done

ext/bin/pdbbox-init
```diff
@@ -176,4 +176,40 @@ ssl-key = $abs_pdbbox/ssl/pdb.key.pem
 disable-update-checking = true
 EOF
 
+# do some tuning on PostgreSQL
+
+# generally accepted configuration parameters:
+# Set the shared_buffers to be 1/4 of the amount of RAM
```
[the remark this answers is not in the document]
robdaemon: implemented a cap at 1GB of RAM for shared_buffers

rbrw:
lgtm - just needs a rebase.

PR #2640, opened 2018-10-09
(PDB-4135) Support PUPPET_SUPPRESS_INTERNAL_LEIN_REPOS

rbrw:
Until we're able to remove the internal :repositories from the default
repository list, allow them to be removed by setting
PUPPET_SUPPRESS_INTERNAL_LEIN_REPOS to true, yes, or 1 in the
environment.

### DIFF
--- a/documentation/CONTRIBUTING.md
+++ b/documentation/CONTRIBUTING.md
@@ -55,6 +55,12 @@ top of things.
 
 ### Testing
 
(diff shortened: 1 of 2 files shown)


rbrw:
Oh, and I used a PUPPET_* prefix on the var in case it ended up being useful for other projects, but PDB_* would be just fine too.

PR #2641, opened 2018-10-09
(PDB-4015) Validate timestamps passed in queries

robdaemon:
When a timestamp fails to validate, it is currently treated as a nil. This
causes a query to be executed with the wrong value for the specified timestamp.
Instead, we should perform validation of the timestamps and return an HTTP 400
when the timestamp is invalid.


robdaemon:
Working on tests for this now.

test/puppetlabs/puppetdb/http/nodes_test.clj
```diff
@@ -567,7 +567,11 @@
 
       (is-query-result' ["=" "report_timestamp" web1-report-ts] [web1])
       (is-query-result' [">" "report_timestamp" web1-report-ts] [db puppet])
-      (is-query-result' [">=" "report_timestamp" web1-report-ts] [web1 db puppet]))))
+      (is-query-result' [">=" "report_timestamp" web1-report-ts] [web1 db puppet])))
+
+  (let [{:keys [body status]} (query-response method endpoint ["<" "catalog_timestamp" "'2018-08-15 21:11:21 UTC'"])]
```
rbrw: PR looks good, though this line's getting a bit long -- suggest maybe one of these, or some other variation/accommodation:
```
  (let [{:keys [body status]} (->> ["<" "catalog_timestamp" "'2018-08-15 21:11:21 UTC'"]
                                   (query-response method endpoint))]
    ...)
  
  (let [{:keys [body status]}
        (query-response method endpoint ["<" "catalog_timestamp" "'2018-08-15 21:11:21 UTC'"])]
    ...)

  ;; (Also plausible to not use :keys here, since we only refer to each field once....)
  (let [res (->> ["<" "catalog_timestamp" "'2018-08-15 21:11:21 UTC'"]
                 (query-response method endpoint))]
    (is (= 400 (:status res))))
```

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1797,6 +1797,13 @@
           (create-extract-node* cols expr))
       (create-extract-node* query-rec cols expr))))
 
+(defn try-parse-timestamp
+  "Try to convert a string to a timestamp, throwing an exception if it fails"
+  [ts]
+  (if-let [result (to-timestamp ts)]
+    result
+    (throw (IllegalArgumentException. (tru "''{0}'' is not a valid timestamp value" ts)))))
```
rbrw: Just noticed we could also write this as:
```
(or (to-timestamp ts)
    (throw ....)
```
But if you have a preference, that's fine.

rbrw:
Looks good.  If you don't mind - let's do a trivial rebase, and then when that's green, I'll merge it.

robdaemon:
@rbrw rebased :)

PR #2642, opened 2018-10-10
(MAINT) Fix rspec tests

austb:
run-rspec-tests took an argument instead of respecting the PDB_TEST env
variable


.travis.yml
```diff
@@ -94,7 +94,7 @@ jobs:
     # === rspec tests
     - stage: ❧ pdb tests
       env: PDB_TEST=rspec/pup-5.3.x
-      script: ext/travisci/run-rspec-tests 5.3.x
+      script: ext/travisci/run-rspec-tests
```
rbrw: Not sure it's important, but so far I'd avoided using the PDB_TEST var outside of .travis.yml (as we do in the unit and integration tests), and instead passed the relevant values as arguments.  And along those lines, run-rspec-tests does have an argument for the ref, so we could do that here (and could also make it an alias like the others), i.e.:

```
  - &run-spec-tests |
    set -e
    puppet="$(ext/travisci/prefixed-ref-from-spec "$PDB_TEST" pup-)"
    ext/travisci/run-rspec-tests "$puppet"
```

Part of the rationale for keeping the travis bits in .travis.yml was a hope that as much of the test code as possible could be travis independent, and in ext/bin, and we might eventually not even need ext/travisci (or it'd be minimal), but of course if nothing else, the jdk_switcher issue has kept us further from that than I'd hoped.

Not necessarily a strong argument here either way, but that's what I was originally thinking...

austb: I don't feel too strongly either way, although I'd be inclined to adopt that mentality for `ext/bin` but allow using `PDB_TEST` in `ext/travisci` to keep those aliases a little cleaner.

rbrw: Yeah, fair enough -- I'd originally hoped to get away without ext/travisci at all, but we're not there yet.  But indeed, the key interest was just to favor "not travis specific" when reasonable.

rbrw: OK, sounds fine, though how would you feel about passing the PDB_TEST var as the argument -- i.e. perhaps makes it a little more obvious what's going on.  e.g.: `ext/travisci/run-rspec-tests "$PDB_TEST".

rbrw:
Could use a rebase.

PR #2643, opened 2018-10-10
(PDB-4118) Test against OpenJDK11

austb:


rbrw:
Let's add the doc updates to this, or create a separate ticket -- i.e. `git grep -i jdk documentation`.  I think there will at least be a faq update.  And we'll also want to update the jdk-support-status function, etc., but we might want to hold off on the latter until we merge the aot pr, if we think we're about to do that, then rebase this (since the function moved).

robdaemon:
we need to pair this with an upgrade to `dynapath` to 1.0.0, as we currently get a reflective access error in it under jdk 10+. `dynapath` was reverted in puppetlabs/clj-parent#156

austb:
Dynapath does need to get updated, but the failure is identical on this PR and master, should we gate this PR on fixing that?

I'd be in favor of getting it in and creating another ticket to track that

PR #2648, opened 2018-10-11
(PDB-2001) Raise an error when an invalid date is sent on a command

robdaemon:
This adds validation of the producer_timestamp value inside a command body.
    
This supports all the common date-time formats, and adds support for UNIX
timestamps as strings (as is used in a lot of tests)
    
When validation fails, the server returns an HTTP 500, matching the other
behavior of failures for missing required parameters and such.

### DIFF
--- a/src/puppetlabs/puppetdb/http/command.clj
+++ b/src/puppetlabs/puppetdb/http/command.clj
@@ -20,7 +20,8 @@
             [slingshot.slingshot :refer [try+ throw+]]
             [puppetlabs.i18n.core :refer [trs tru]]
             [puppetlabs.puppetdb.time :as pdbtime]
-            [puppetlabs.puppetdb.utils :as utils])
+            [puppetlabs.puppetdb.utils :as utils]
+            [clj-time.format :as tf])
   (:import [org.apache.commons.io IOUtils]
            [org.apache.commons.fileupload.util LimitedInputStream]))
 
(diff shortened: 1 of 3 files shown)


robdaemon:
I don't really like how this API works - when any validation error occurs, the API returns a 500 with no body.

Example where `certname` query parameter is missing:

```
$ curl -D - -X POST   -H 'Content-Type:application/json'   -H 'Accept:application/json'   -d '{"certname":"test1","environment":"DEV","producer_timestamp":"2015-01-01", "producer":"master1","values":{"myfact":"myvalue"}}'   "http://localhost:8080/pdb/cmd/v1?command=replace_facts&version=5"
HTTP/1.1 500 Server Error
Cache-Control: must-revalidate,no-cache,no-store
Connection: close
Server: Jetty(9.4.4.v20170414)
```

Example with this change, with an invalid `producer_timestamp`:

```
$ curl -D - -X POST   -H 'Content-Type:application/json'   -H 'Accept:application/json'   -d '{"certname":"test1","environment":"DEV","producer_timestamp":"01-01-2015", "producer":"master1","values":{"myfact":"myvalue"}}'   "http://localhost:8080/pdb/cmd/v1?command=replace_facts&version=5&certname=test1"
HTTP/1.1 500 Server Error
Cache-Control: must-revalidate,no-cache,no-store
Connection: close
Server: Jetty(9.4.4.v20170414)
```

This means all validation errors are handled the same, but I'd prefer if this API could return a proper HTTP 400 with a body like the query API does.

PR #2649, opened 2018-10-11
(PDB-4139) travis: update test set for 6.0.x

rbrw:


.travis.yml
```diff
@@ -87,49 +87,49 @@ jobs:
...
 
     - stage: ❧ pdb tests
-      env: PDB_TEST=int/oraclejdk8/pup-master/srv-master/pg-9.6
+      env: PDB_TEST=int/oraclejdk8/pup-6.0.x/srv-6.0.x/pg-9.6/rich
       script: *run-integration-tests
 
+    # === integration with previous platform
     - stage: ❧ pdb tests
-      env: PDB_TEST=int/openjdk10/pup-master/srv-master/pg-9.6
+      env: PDB_TEST=int/openjdk10/pup-5.5.x/pup-5.3.x/pg-9.6
```
austb: I don't believe this will work because srv-5.3.x won't run under jdk >= 8

.travis.yml
```diff
@@ -87,49 +87,49 @@ jobs:
...
       script: *run-integration-tests
 
     - stage: ❧ pdb tests
-      env: PDB_TEST=int/rich/oraclejdk10/pup-master/srv-master/pg-9.6
+      env: PDB_TEST=int/openjdk10/pup-5.5.x/pup-5.3.x/pg-9.6/rich
```
austb: I don't believe this will work because srv-5.3.x won't run under jdk >= 8

PR #2654, opened 2018-10-15
(maint) Use shipped version of puppet_docker_tools

Iristyle:


Iristyle:
Moved this into #2595 and closing as a result.

PR #2656, opened 2018-10-16
(PDB-2001) Validate producer-timestamp in query string

robdaemon:
Adds validation to the producer-timestamp in the query string of a command. This
ignores the producer_timestamp in the body, as we do not do validation of
command bodies until the command is popped off the queue.

Also modifies tests that submit producer_timestamp to reflect the proper 'Z'
format of ISO-8601. The command processors fail when this value is a UNIX
timestamp instead of an ISO-8601 string, so updating the tests improves our
"example" code for this field to match the wire format.


robdaemon:
The Ruby code for puppet server relies on sending UNIX timestamps in queries in the `producer-timestamp` parameter, which doesn't match our wire format. The scope of this change / risk has gotten too large, going to abandon this as a "by design" - in v7, we can talk about where we do this "right"

PR #2671, opened 2018-10-22
(PDB-4176) Travis: Fix openjdk8 testing on OSX

austb:
AdoptOpenJDK 8 is deprecated and no longer has a proper cacert bundle

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,22 @@ aliases:
   - &run-core-and-ext-tests |
     set -e
     jdk="$(ext/travisci/jdk-from-spec "$PDB_TEST")"
-    if test "$(uname)" = Linux; then jdk_switcher use "$jdk"; fi
+    jdkver="${jdk##*jdk}"
     ext/travisci/prep-os-essentials-for "$PDB_TEST"
+    case "$OSTYPE" in
+      linux*)
+          jdk_switcher use "$jdk";
+        ;;
+      darwin*)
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -20,8 +20,16 @@ aliases:
   - &run-core-and-ext-tests |
     set -e
     jdk="$(ext/travisci/jdk-from-spec "$PDB_TEST")"
-    if test "$(uname)" = Linux; then jdk_switcher use "$jdk"; fi
     ext/travisci/prep-os-essentials-for "$PDB_TEST"
+    if test "$(uname)" = Linux; then
```
rbrw: Suggest using case and/or OSTYPE (since we know it's bash) here the way we do in some of the other scripts.

.travis.yml
```diff
@@ -35,8 +43,16 @@ aliases:
   - &run-integration-tests |
     set -e
     jdk="$(ext/travisci/jdk-from-spec "$PDB_TEST")"
-    if test "$(uname)" = Linux; then jdk_switcher use "$jdk"; fi
     ext/travisci/prep-os-essentials-for "$PDB_TEST"
+    if test "$(uname)" = Linux; then
+      jdk_switcher use "$jdk"
+    elif test "$(uname)" = Darwin; then
+      JAVA_HOME=$(/usr/libexec/java_home -v $1);
```
rbrw: I think it'd be worth elaborating on what this is/does in the commit message.  And is it just finding some built-in java?  If so, then do we still need the cask?

austb: This is specific to OSX but it will find any installed JDK version whether it's Oracle or OpenJDK. I'll add a comment.

PR #2673, opened 2018-10-22
(PDB-4165) Duplicate dependency pins in :ezbake profile

rbrw:
It looks like the use of ^:replace :dependencies in the ezbake profile
causes all pins in the normal :dependencies section to be ignored.  So
for now, just repeat the pins for the recent CVEs in the :ezbake
profile, which does cause them to show up here:

  lein with-profiles ezbake deps :tree

In addition, stop pinning hikaricp and clojure/jdbc so that we won't
fall afoul of :pedantic, *and* so that we'll actually be testing the
versions that we ship (verified by spot-checking the relevant
META-INF/maven/ pom.xml versions in the 5.1.5 debian/stretch jar).

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,6 @@
 (def pdb-version "5.1.6-SNAPSHOT")
 (def clj-parent-version "1.4.3")
+(def tk-jetty9-ver "2.2.0")
 
 (defn true-in-env? [x]
   (#{"true" "yes" "1"} (System/getenv x)))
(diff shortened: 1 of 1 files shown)


austb:
New diff between `lein deps :tree` and `lein with-profiles ezbake deps :tree` alleviates the issues we had previously (had to add .txt so that Github wouldaccept the file type... because patches don't make sense to support apparently...)

[diff.patch.txt](https://github.com/puppetlabs/puppetdb/files/2503820/diff.patch.txt)

robdaemon:
So doing this instead of the clj-parent changes? I'd really like to see clj-parent have the proper deps too, since it's part of the cause of this, but that can be done in a future release.

PR #2682, opened 2018-10-30
(PDB-2276) Replace robert.hooke with explicit config map

robdaemon:
This removes the use of the robert.hooke library to interrupt normal
trapperkeeper config processing flow. Now we allow tk to parse the config, then
we manipulate it with our defaults and conversion and retirement warnings, then
call a new function in tk to boot with that modified config.

Depends on puppetlabs/trapperkeeper#278


robdaemon:
I expect the CI failures due to the dependency on a new trapperkeeper snapshot.

PR #2690, opened 2018-11-01
(maint) Make check in container whether puppetmaster is up more robust

lutter:
Previously, docker-entrypoint.sh checked whether the puppetmaster is up and
running by using 'nc -z'. That only checks that something is listening on
the given port, but not whether that service is actually ready. This is a
problem when the puppetmaster is behind a load balancer since it only
checks that the loadbalancer is listening, even though the puppetmaster
might take quite a while longer to become ready; in such a scenario, the
PuppetDB startup fails.

This change replaces the 'nc -z' check with a check for the puppetmaster's
status, and only proceeds when the master is fully up and ready to handle
requests.

### DIFF
--- a/docker/puppetdb/docker-entrypoint.sh
+++ b/docker/puppetdb/docker-entrypoint.sh
@@ -1,8 +1,13 @@
 #!/bin/bash
 
+master_running() {
+    curl -sf "http://${PUPPETSERVER_HOSTNAME}:8140/production/status/test" \
+        | grep -q '"is_alive":true'
+}
+
 PUPPETSERVER_HOSTNAME="${PUPPETSERVER_HOSTNAME:-puppet}"
 if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ]; then
-  while ! nc -z "$PUPPETSERVER_HOSTNAME" 8140; do
+  while ! master_running; do
     sleep 1
   done
   set -e


docker/puppetdb/docker-entrypoint.sh
```diff
@@ -1,8 +1,13 @@
 #!/bin/bash
 
+master_running() {
+    curl -sf http://master:8140/production/status/test \
```
underscorgan: should this be `$PUPPETSERVER_HOSTNAME` instead of `master`?

lutter: Ugh, yes, it should. Nice catch. Updating the PR.

lutter: Ugh, yes it should. Nice catch! Updating the PR.

underscorgan:
THough @Iristyle may have more context on the feasibility of this for windows, I know he's been fighting with that a bit

justinstoller:
fyi, the [agent in 6.0](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/configurer.rb#L393-L411) and above uses puppetserver's simple [status endpoint](https://github.com/puppetlabs/trapperkeeper-status/blob/master/documentation/query-api.md) to determine what master to use in the server_list for HA. I believe the puppetserver's status endpoint (though maybe not the "simple" segment) is also used for the status output in the PE console and for support's metrics collection.

PR #2699, opened 2018-11-12
(CPR-648) Run puppetdb docker container tests

Iristyle:
- Previous tests checked only a few issues with the Dockerfile.

   Add new tests that start both postgres and puppetdb containers


docker/puppetdb/spec/puppetdb_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+    #{postgres_name}).chomp
+end
+
+def run_puppetdb_container
+  %x(docker run --rm --detach
+    --env PUPPETDB_PASSWORD=puppetdb
+    --env PUPPETDB_USER=puppetdb
+    --name puppetdb
+    --hostname puppetdb
+    --link puppet
```
underscorgan: I don't see where this is spinning up a puppet container

docker/puppetdb/spec/puppetdb_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+  end
+end
+
+def run_postgres_container
+  postgres_name = File::ALT_SEPARATOR.nil? ?
+    'postgres:9.6' :
+    'stellirin/postgres-windows:9.6'
+  %x(docker pull #{postgres_name})
+  postgres_custom_dir = File::ALT_SEPARATOR.nil? ?
+    '/docker-entrypoint-initdb.d',
```
underscorgan:
```suggestion
    '/docker-entrypoint-initdb.d' :
```

PR #2700, opened 2018-11-12
(maint) Move to makefile for builds

underscorgan:
This moves more to standard tools for unix-y builds.


underscorgan:
This doesn't include steps for the puppetdb-postgres container as we're moving to using upstream postgres soon. Can add that in if needed.

docker/distelli-manifest.yml
```diff
@@ -1,4 +1,9 @@
 pe-and-platform/puppetdb:
+  PreBuild:
+    - make lint
   Build:
+    - make build
+    - bundle install --path .bundle/gems && bundle exec make test
```
nwolfe: I might just put the bundler calls into the Makefile itself, probably as an additional pre-step to `test`, so we can run tests locally just with `make test` and make sure it's "doing the same thing as CI".

underscorgan: Updated!

PR #2705, opened 2018-11-15
(PDB-4186) Redirect all output to stdout when a pdb container

rbrw:


rbrw:
@nwolfe yeah, sorry, should probably have mentioned, but, that's one reason I put the don't merge on it, I realized that it might not be what we want, and I might well just close it, but hadn't had a chance to double-check yet.

rbrw:
Closing because docker already handles this in the cases we're likely to care about for now: https://docs.docker.com/config/containers/logging/

PR #2707, opened 2018-11-15
(WIP) PuppetDB alpine image from source

nwolfe:
@puppetlabs/puppetdb thoughts on this approach? Our goal is to reduce the image size.

Building from source rather than packages also means developers (external; community) can build an image with their local changes, which has its own uses.

This builds an image of size 212MB, and the current puppetdb image is 412MB. Would be nice to get it down further. The Ruby+Puppet bits are what we'd like to cut out.

### DIFF
--- a/Dockerfile.slim
+++ b/Dockerfile.slim
@@ -1,0 +1,57 @@
+FROM clojure:lein-alpine
+RUN apk add --no-cache make
+# Install only dependencies
+WORKDIR /app
+COPY project.clj .
+COPY resources/puppetlabs/puppetdb/bootstrap.cfg \
+    ./resources/puppetlabs/puppetdb/bootstrap.cfg
+RUN lein with-profile uberjar deps
+# Build uberjar -- see .dockerignore
+COPY . .
+RUN lein with-profile uberjar uberjar
+
+
+# FROM openjdk:8-jre-alpine
+FROM alpine-jre-puppet
+RUN apk add --no-cache tini
+COPY --from=0 /app/target/puppetdb.jar /
+ENV PUPPETDB_DATABASE_CONNECTION="//postgres:5432/puppetdb"
+ENV PUPPETDB_USER=puppetdb
+ENV PUPPETDB_PASSWORD=puppetdb
+ENV PUPPETDB_NODE_TTL=7d
+ENV PUPPETDB_NODE_PURGE_TTL=14d
+ENV PUPPETDB_REPORT_TTL=14d
+ENV PUPPETDB_JAVA_ARGS="-Djava.net.preferIPv4Stack=true -Xms256m -Xmx256m"
+
+# Values from /etc/default/puppetdb
+ENV JAVA_BIN="/usr/bin/java"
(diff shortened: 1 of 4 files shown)


Dockerfile.slim
```diff
@@ -0,0 +1,57 @@
...
+COPY resources/puppetlabs/puppetdb/bootstrap.cfg \
+    ./resources/puppetlabs/puppetdb/bootstrap.cfg
+RUN lein with-profile uberjar deps
+# Build uberjar -- see .dockerignore
+COPY . .
+RUN lein with-profile uberjar uberjar
+
+
+# FROM openjdk:8-jre-alpine
+FROM alpine-jre-puppet
```
nwolfe: This is the `puppet/puppet-agent-alpine` image but built from `openjdk:8-jre-alpine` instead of `alpine:3.8` (https://github.com/puppetlabs/puppet-agent/blob/master/docker/puppet-agent-alpine/Dockerfile#L114).

This only exists on my machine at the moment.

docker/puppetdb/docker-entrypoint.sh
```diff
@@ -11,10 +11,13 @@ if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ] && [ "$USE_PUPPETSERVER" = true ]; th
...
   set -e
-  /opt/puppetlabs/bin/puppet config set certname "$HOSTNAME"
-  /opt/puppetlabs/bin/puppet config set server "$PUPPETSERVER_HOSTNAME"
-  /opt/puppetlabs/bin/puppet agent --verbose --onetime --no-daemonize --waitforcert 120
-  /opt/puppetlabs/server/bin/puppetdb ssl-setup -f
+  puppet config set certname "$HOSTNAME"
+  puppet config set server "$PUPPETSERVER_HOSTNAME"
+  puppet agent --verbose --onetime --no-daemonize --waitforcert 120
+  # /opt/puppetlabs/server/bin/puppetdb ssl-setup -f
+  /ssl-setup.sh -f
```
nwolfe: These 3 lines are the most problematic thing here, since this means we need puppet+ruby in the image.

Sounds like we'll need to figure out a way to do this stuff differently, not sure when we'll need to do that though.

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,420 @@
...
+  done
+else
+  # This should be run on the host with PuppetDB
+  PATH=/opt/puppetlabs/bin:/opt/puppet/bin:$PATH
+  agent_confdir=`puppet agent --configprint confdir`
+  agent_vardir=`puppet agent --configprint vardir`
+  # user=<%= EZBake::Config[:user] %>
+  # group=<%= EZBake::Config[:group] %>
+  user=puppetdb
+  group=puppetdb
```
nwolfe: This entire file is just the current `ssl-setup.erb` file, but with the 2 lines of EZBake replaced with hardcoded FOSS values.

nwolfe:
Replaced by https://github.com/puppetlabs/puppetdb/pull/2738

PR #2710, opened 2018-11-17
Revert "(maint) Make check in container whether puppetmaster is up mo…

austb:
…re robust"

This reverts commit 8759635ce598522a10360aae5be68110a96d96a3.

[facebones ignore]


nwolfe:
We can't just revert the commit unfortunately.

See https://github.com/puppetlabs/puppetdb/pull/2714 instead.

PR #2713, opened 2018-11-19
(PDB-4177) Fix bash in place

Zak-Kent:
Prior to these commits the producer-timestamp wasn't sent in the correct format from the terminus or pulled out of the query parameters. This caused all cmdrefs to have nil values for producer-ts.

### DIFF
--- a/puppet/spec/unit/util/puppetdb/command_spec.rb
+++ b/puppet/spec/unit/util/puppetdb/command_spec.rb
@@ -26,11 +26,12 @@ describe Puppet::Util::Puppetdb::Command do
         httpok.stubs(:body).returns '{"uuid": "a UUID"}'
         http.expects(:post).with() do | path, payload, headers, options |
           param_map = CGI::parse(URI(path).query)
-          param_map['certname'].first.should == 'foo.localdomain' &&
+          assert_valid_producer_ts(path) &&
+            param_map['certname'].first.should == 'foo.localdomain' &&
             param_map['version'].first.should == '1' &&
-            param_map['command'].first.should == 'OPEN_SESAME'
-          options[:compress].should == :gzip
-          options[:metric_id].should == [:puppetdb, :command, 'OPEN_SESAME']
+            param_map['command'].first.should == 'OPEN_SESAME' &&
+            options[:compress] == :gzip &&
+            options[:metric_id] == [:puppetdb, :command, 'OPEN_SESAME']
         end.returns(httpok)
 
         subject.submit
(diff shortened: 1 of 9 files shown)


src/puppetlabs/puppetdb/client.clj
```diff
@@ -42,7 +43,7 @@
    (let [body (json/generate-string payload)
          url (str (utils/base-url->str base-url)
                   (format "?command=%s&version=%s&certname=%s&producer-timestamp=%s"
-                          (str/replace command #" " "_") version certname (System/currentTimeMillis))
+                          (str/replace command #" " "_") version certname (c/from-long (System/currentTimeMillis)))
```
robdaemon: Does the toString on a datetime instance automatically format it as the expected ISO8601?

src/puppetlabs/puppetdb/command.clj
```diff
@@ -270,10 +270,17 @@
   (let [id (str id)
         received-time (str (tcoerce/to-long received-time))
         duration (str (in-millis (interval start-time (now))))
-        command-name (command-names command-kw)]
-    (if-let [{:keys [puppet-version]} opts]
+        command-name (command-names command-kw)
```
robdaemon: The changes in this file feel like they're unrelated to a producer-timestamp bug?

src/puppetlabs/puppetdb/client.clj
austb: As long this part of puppetdb is also using jodatime, it does

https://www.joda.org/joda-time/apidocs/org/joda/time/base/AbstractDateTime.html#toString--

src/puppetlabs/puppetdb/command.clj
Zak-Kent: You're right these changes are somewhat unrelated to the bug, but they do add logging to the behavior that bug was suppressing so I went ahead and added it in the same pr. I'm hoping it will help add clarity when people start seeing messages get bashed.

puppet/spec/unit/util/puppetdb/command_spec.rb
```diff
@@ -25,6 +25,7 @@
       it "should issue the HTTP POST and log success" do
         httpok.stubs(:body).returns '{"uuid": "a UUID"}'
         http.expects(:post).with() do | path, payload, headers, options |
+          expect(assert_valid_producer_ts(path)).to be_truthy
```
austb: Looking into the rspec tests a little bit more, I realized we are using mocha, so this with block should look like 

```
        http.expects(:post).with() do | path, payload, headers, options |
          param_map = CGI::parse(URI(path).query)
          assert_valid_producer_ts(path) &&
            param_map['certname'].first.should == 'foo.localdomain' &&
            param_map['version'].first.should == '1' &&
            param_map['command'].first.should == 'OPEN_SESAME' &&
            options[:compress] == :gzip &&
            options[:metric_id] == [:puppetdb, :command, 'OPEN_SESAME']
        end
```

Zak-Kent:
will need rebase once logging commit goes in

rbrw:
Looks good.

PR #2714, opened 2018-11-19
(MAINT) Revert back to nc for Docker entrypoint

nwolfe:
The current curl call doesn't seem to work and thus PuppetDB will never
start as this check never succeeds.

We do need the orthogonal USE_PUPPETSERVER change though, so we can't
just revert the commit that changed the netcat call to curl.

For now, revert back to using netcat while we figure out how to better
detect when Puppet Server is up.

The netcat call doesn't work when a load balancer is in place as it
just detects that the LB is up, but we don't actually need this
functionality yet.

### DIFF
--- a/docker/puppetdb/docker-entrypoint.sh
+++ b/docker/puppetdb/docker-entrypoint.sh
@@ -1,8 +1,12 @@
 #!/bin/bash
 
 master_running() {
-    curl -sf "http://${PUPPETSERVER_HOSTNAME}:8140/production/status/test" \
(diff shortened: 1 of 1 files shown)


nwolfe:
I verified the curl call on master branch doesn't work, and verified that the nc call does still work even when moved into this function.

I targeted the PR at 5.1.x since that's what @austb targeted his PR at.

austb:
Merge at 48f1ff52b4ba4ff7c7ccfb231ded71446719cc2a

PR #2718, opened 2018-11-20
(maint) Update puppetdb entrypoint for less dependence on puppetserver

underscorgan:
Previously, on anything but the initial run, if the puppetdb container
started up before puppetserver, the container would fail and exit since
the directory that is checked for existence never exists on startup. By
updating the checks in the entrypoint, there is now no hard dependency
between puppetserver and puppetdb.

### DIFF
--- a/docker/puppetdb/docker-entrypoint.sh
+++ b/docker/puppetdb/docker-entrypoint.sh
@@ -6,14 +6,18 @@ master_running() {
 }
 
 PUPPETSERVER_HOSTNAME="${PUPPETSERVER_HOSTNAME:-puppet}"
-if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ] && [ "$USE_PUPPETSERVER" = true ]; then
+/opt/puppetlabs/bin/puppet config set certname "$HOSTNAME"
+/opt/puppetlabs/bin/puppet config set server "$PUPPETSERVER_HOSTNAME"
+
+if [ ! -f "/etc/puppetlabs/puppet/ssl/certs/${HOSTNAME}.pem" ] && [ "$USE_PUPPETSERVER" = true ]; then
+  # if this is our first run, run puppet agent to get certs in place
   while ! master_running; do
     sleep 1
   done
   set -e
-  /opt/puppetlabs/bin/puppet config set certname "$HOSTNAME"
-  /opt/puppetlabs/bin/puppet config set server "$PUPPETSERVER_HOSTNAME"
   /opt/puppetlabs/bin/puppet agent --verbose --onetime --no-daemonize --waitforcert 120
+fi
+if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ] && [ "$USE_PUPPETSERVER" = true ]; then
   /opt/puppetlabs/server/bin/puppetdb ssl-setup -f
 fi
 


underscorgan:
Built a container locally and ran this through pupperware tests, was able to remove the dependency between puppetserver and puppetdb in the docker-compose.yml

underscorgan:
jenkins, retest this please

docker/puppetdb/docker-entrypoint.sh
```diff
@@ -7,14 +7,18 @@ master_running() {
 }
 
 PUPPETSERVER_HOSTNAME="${PUPPETSERVER_HOSTNAME:-puppet}"
-if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ] && [ "$USE_PUPPETSERVER" = true ]; then
+/opt/puppetlabs/bin/puppet config set certname "$HOSTNAME"
```
Iristyle: So I'm curious what the behavior is if pdb is started with a particular `$HOSTNAME` / `$PUPPETSERVER_HOSTNAME` that later changes?

underscorgan: Hm. I think the only way around that is going to be run puppet each time the container starts since we aren't preserving the pdb config files on restart. Maybe the dependency between puppetdb and puppetserver is better with the new `master_running` function that went in last week, I'll test that out.

nwolfe: I think we'd need to rerun all these steps each time the `$HOSTNAME` changes, since we'd need to generate new certs for that host (puppet config set + puppet agent run), and then we'd need to rerun `ssl-setup -f` since that modifies puppetdb and Jetty config files to reference SSL certs on disk.

underscorgan: Wouldn't we also need to rerun these if the puppetserver_hostname changed? I feel like the check could be updated to check if the existing certs match the puppetdb hostname, but I don't know that we have enough information preserved in the container to figure out if the puppetserver hostname changed.

In other confusing news, even with the updated master_running function using the puppetserver status endpoint, the pdb container still tries to connect to puppetserver before it's up if I start them up without deps in the compose file. No idea what's causing that :|

nwolfe: > Wouldn't we also need to rerun these if the puppetserver_hostname changed?

That would probably be safest, but I'm not sure it'd be required necessarily?

There's nothing tied to the PUPPETSERVER_HOSTNAME once we have the SSL files; the special CA file is always named `ca.pem` so doesn't need to match the hostname value (unlike the puppetdb host cert file).

> but I don't know that we have enough information preserved in the container to figure out if the puppetserver hostname changed.

I think you're right.

But we can probably still detect if the CA has changed, by instead asking for our host cert and checking if it matches the one we have on disk already.  If they don't match for any reason, then we're out of sync with the CA and some remediation is necessary.

underscorgan: I got this updated to work if `$HOSTNAME` changes. @nwolfe do you know how would I go about 'asking for our host cert and checking if it matches what we have on disk'?

underscorgan: Ah, I guess maybe this?
```
output="$(curl --insecure --silent "https://${PUPPETSERVER_HOSTNAME}:8140/puppet-ca/v1/certificate/${HOSTNAME}.pem" | head -1)"
```

nwolfe: Normally we'd piece together some `puppet config print ...` calls to get things like our hostname and hostcert, then do a curl call to the CA with that hostname and compare it with the hostcert.

The work I have would remove the ability to do this since there's no puppet CLI, but in that case I'm hardcoding/assuming the filepaths and relying on `$HOSTNAME`.

Either way, I'd try something like this:
```bash
cert="$(curl -ks https://${PUPPETSERVER_HOSTNAME}:8140/puppet-ca/v1/certificate/${HOSTNAME})"
localcert="$(cat /etc/puppetlabs/puppetdb/ssl/certs/${HOSTNAME})"
if [ "${cert}" = "${localcert}" ] ...
```

underscorgan: We discussed this some in chat, we're not going to try to add more flexibility to handle the cases of the CA being regenerated or the puppetdb cert being regenerated on the CA, automating that runs too high a risk of blowing away good certificates.

If you need to regenerate certs for pdb or the whole stack, it's reasonable that you'll need to manually remove the existing certs from /etc/puppetlabs/puppet/ssl on the affected nodes

PR #2723, opened 2018-11-26
(PDB-4192) Curl Puppet Server status endpoint

nwolfe:
Hitting an endpoint instead of using netcat will work when Puppet Server
is behind a load balancer.

We tried this approach earlier but weren't specifying HTTPS and the
-k flag when curling.

### DIFF
--- a/docker/puppetdb/docker-entrypoint.sh
+++ b/docker/puppetdb/docker-entrypoint.sh
@@ -1,9 +1,8 @@
 #!/bin/bash
 
 master_running() {
-    # This netcat call doesn't work when a load balancer is in place as it just
-    # detects that the LB is up (see PDB-4192)
-    nc -z "$PUPPETSERVER_HOSTNAME" 8140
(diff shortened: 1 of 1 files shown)


docker/puppetdb/docker-entrypoint.sh
```diff
@@ -1,9 +1,8 @@
 #!/bin/bash
 
 master_running() {
-    # This netcat call doesn't work when a load balancer is in place as it just
-    # detects that the LB is up (see PDB-4192)
-    nc -z "$PUPPETSERVER_HOSTNAME" 8140
+    status=$(curl -sfk "https://${PUPPETSERVER_HOSTNAME}:8140/status/v1/simple")
```
underscorgan: can we use long options here, so `curl --silent --fail --insecure`

lutter:
I don't think the `--insecure` flag is an issue for getting the status, and as you noted, there's not really a good way around that.

nwolfe:
@austb Yeah, no way around `--insecure` in this case. This seems fine long-term.

Ready for Merge.

PR #2724, opened 2018-11-27

robdaemon:
* Adds a config parameter: database.resource-events-ttl with a default of 14d
* Adds two admin metrics: resource-events-purges and resource-events-purge-time
* Adds documentation for this new feature

This will allow the resource_events table to be cleaned up at a different interval
than the reports table.

### DIFF
--- a/documentation/configure.markdown
+++ b/documentation/configure.markdown
@@ -387,6 +387,18 @@ runs every `gc-interval` minutes.
 
 If unset, the default value is 14 days.
 
+### `resource-events-ttl`
+
+Automatically delete report events older than the specified time. The reports
+will still be available, but the contained events will not be present.
+
+This allows for more fine-grained control of the expiration of reports. If this
+value is set higher than `report-ttl`, it will have no effect.
+
+If unset, the default value is 14 days.
+
+See `node-ttl` above for the format of this value.
+
 ### `subname`
 
 This describes where to find the database. It should be something like
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -85,7 +85,7 @@
 (def database-metrics-registry (get-in metrics/metrics-registries [:database :registry]))
 
 (def clean-options
-  #{"expire_nodes" "purge_nodes" "purge_reports" "gc_packages" "other"})
+  #{"expire_nodes" "purge_nodes" "purge_reports" "gc_packages" "purge_resource_events" "other"})
 
 (def purge-nodes-opts-schema {:batch_limit s/Int})
 
@@ -114,6 +114,7 @@
        (map {"expire_nodes" "expiring-nodes"
              "purge_nodes" "purging-nodes"
              "purge_reports" "purging-reports"
+             "purge_resource_events" "purging-resource-events"
              "gc_packages" "package-gc"
              "other" "other"})
        (str/join " ")))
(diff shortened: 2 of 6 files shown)


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -251,6 +269,10 @@
         (time! (:package-gc-time admin-metrics)
                (gc-packages! db))
         (counters/inc! (:package-gcs admin-metrics)))
+      (when (request "gc_resource_events")
```
npwalker: The analogous request for deleting reports is `purge_reports` should this be `purge_resource_events`?

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -209,7 +224,9 @@
    :node-purge-time (timer admin-metrics-registry ["node-purge-time"])
    :report-purge-time (timer admin-metrics-registry ["report-purge-time"])
    :package-gc-time (timer admin-metrics-registry "package-gc-time")
-   :other-clean-time (timer admin-metrics-registry ["other-clean-time"])})
+   :other-clean-time (timer admin-metrics-registry ["other-clean-time"])
+   :resource-events-purges (counter admin-metrics-registry "resource-event-purges")
```
austb: Should we move this counter up to the block of counters?

documentation/configure.markdown
```diff
@@ -387,6 +387,18 @@ runs every `gc-interval` minutes.
 
 If unset, the default value is 14 days.
 
+### `resource-events-ttl`
+
+Automatically delete report events older than the specified time. The reports
+will still be available, but the contained events will not be present.
+
+This allows for more fine-grained control of the expiration of reports. If this
+value is set higher than `report-ttl`, it will have no effect.
```
austb: We don't seem to handle the case where report-ttl is less than resource-events-ttl directly, does report gc always happen when a resource event gc happens? If you can trigger them separately, we probably want to overwrite the resource-event-ttl setting if it is greater than the report-ttl setting.

documentation/configure.markdown
```diff
@@ -387,6 +387,18 @@ runs every `gc-interval` minutes.
 
 If unset, the default value is 14 days.
 
+### `resource-events-ttl`
```
austb: Our other ttls are singular, `node-ttl` and `report-ttl` should we maintain that for this one as well?

documentation/configure.markdown
robdaemon: No, it's the other way around - a report gc implies a resource-events gc. It's really harmless if it's larger than report gc - what'll happen is a delete query that doesn't do anything.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -176,6 +177,20 @@
...
+  {:pre [(map? db)
+         (period? resource-events-ttl)]}
+  (try
+    (kitchensink/demarcate
+     (format "sweep of stale resource-events (threshold: %s)"
+             (format-period resource-events-ttl))
+     (jdbc/with-transacted-connection db
+       (scf-store/delete-resource-events-older-than! (ago resource-events-ttl))))
+    (catch Exception e
+      (log/error e (trs "Error while sweeping resource-events")))))
```
rbrw: Wonder if "resource events" might be preferable for the log message...

src/puppetlabs/puppetdb/cli/services.clj
rbrw: That seems reasonable, and might even consider putting these new ones before "other" in each section, but no strong opinion.

documentation/configure.markdown
robdaemon: I'm not sure this is important.

rbrw: Hmm, that sounds plausible.

PR #2725, opened 2018-11-28
(PDB-3857) Make fact path gc compatible with pg 10+

rbrw:
Don't use set returning functions in statement projections.  We were
doing that in the fact path gc operations, and postgres no longer
allows them there (as of version 10).

The approach taken here should also stop pdb from generating a flood
of tiny temp files when running the gc query.

cf. (PDB-3924)

For now I've arranged for pdb to respond to PDB_TEST_PATH_GC_VIA_TEMP_TABLE=1 to enable the temp file fix.  Without that it runs the same query without the temp table, and so you should be able to verify the fix by setting `log_temp_files = 0` in your postgresql.conf, restart pg, and then run

    lein test :only puppetlabs.puppetdb.scf.storage-test/fact-path-gc

without and then with the environment variable set.

I'll remove the "don't merge" after we've finished testing, and I've removed the environment variable.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -1123,38 +1123,37 @@
 (defn delete-unused-fact-paths []
   "Deletes paths from fact_paths that are no longer needed by any
   factset.  In the unusual case where a path changes type, the
-  previous version will linger."
+  previous version will linger.  This requires a parent transaction
+  with at least a repeatable-read isolation level (and may or may not
+  need postgres' \"stronger than the standard\" repeatable-read
+  behavior).  Otherwise paths could be added to fact_paths elsewhere
+  during the gc, not be noticed, and then be deleted at the end."
+  ;; Use a temp table for now until we figure out why pg is creating a
+  ;; vast number of temp files when this is all handled as a single
+  ;; query.  (PDB-3924)
   (jdbc/do-commands
-   ["with recursive live_paths(key, path, value) as"
-    "  (select key, key as path, value"
-    "     from (select (jsonb_each(stable||volatile)).*"
-    "             from factsets) as base_case"
-    "   union all"
-    "   select sub_path as key,"
-    "          sub_paths.path||'#~'||sub_path as path,"
-    "          sub_value as value"
-    "     from (select *"
-    "             from (select path,"
-    "                          case jsonb_typeof(value)"
-    "                            when 'object'"
-    "                              then (jsonb_each(value)).key"
-    "                            when 'array'"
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1120,41 +1120,65 @@
...
+    "        select path||'#~'||sub_level.key as path,"
+    "               sub_level.value"
+    "          from live_paths,"
+    "          lateral (select *"
+    "                     from (select (jsonb_each(value)).*"
+    "                             where jsonb_typeof(value) = 'object') as sub_fields"
+    "                     union (select generate_series(0, jsonb_array_length(value - 1))::text as key,"
+    "                                   jsonb_array_elements(value) as value"
+    "                              where jsonb_typeof(value) = 'array')) as sub_level)"
+    "   select path into temp tmp_live_paths from live_paths"]
```
Zak-Kent: Just to make sure I'm not missing something, it looks like we're throwing the results of the cte above into the tmp_live_paths table and then using that table to compare which facts are still live in the real table deleting those which don't exist in the tmp table, is that right? If that is the case it seems like we're performing the deletion step in roughly the same way in both functions, was this what we wanted to do? I remember us talking about possibly creating a table that was the result of the live paths query and just keeping that live_paths table while dropping the old table wholesale.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1120,41 +1120,65 @@
...
+   "drop table tmp_live_paths"))
+
+(def delete-unused-fact-paths
+  (if (System/getenv "PDB_TEST_PATH_GC_VIA_TEMP_TABLE")
+    delete-unused-fact-paths-via-temp-table
+    delete-unused-fact-paths-laterally))
+
+(if (= delete-unused-fact-paths delete-unused-fact-paths-laterally)
+  (binding [*out* *err*] (println "Pruning fact paths without temp table"))
+  (binding [*out* *err*] (println "Oruning fact paths via temp table")))
```
austb:
```suggestion
  (binding [*out* *err*] (println "Pruning fact paths via temp table")))
```

src/puppetlabs/puppetdb/scf/storage.clj
rbrw: Yeah, that's what the current code intends, i.e. they do the same thing, but using the temp table hopefully avoids the tiny temp file log flood.

rbrw:
Zak's more or less finished testing, and we've discussed/resolved our current questions, so I've removed the debug-related code, and this should now be closer to the final version.

PR #2727, opened 2018-11-28
(PDB-4197) update clj-parent to 2.4.0

robdaemon:
Minor changes in clj-parent that do not directly impact pdb


robdaemon:
The change in 2.4.0 (compared to 2.3.2) that might affect pdb are a minor update to ring-middleware.

The other changes, rbac-client and trapperkeeper-scheduler, do not affect pdb (we don't use them)

PR #2733, opened 2018-12-03
Update tk-webserver-jetty9 to 2.3.1

highb:
Update is in order to pull in a change to our Jetty configuration that 
prevents Jetty from sending the server version in the HTTP response and
error pages.


rbrw:
Looks like this brings in an indirect clj-parent change, but for the record, we checked lein-parent's behavior, and think that it shouldn't disturb the existing 5.2 deps because the parent parent-project specification (e.g. in the root project.clj) should still prevail.

PR #2738, opened 2018-12-06
(REVIEW ME) PuppetDB alpine image from source

nwolfe:
The goal with this work is to reduce the size of the puppetdb image, and secondary is building out new patterns for acquiring SSL certs and running from source.

The ssl.sh script used in this is here: https://github.com/puppetlabs/pupperware/pull/33

This results in a final puppetdb image size of 116MB, down from the current size of 413MB.

To build this image locally, clone the branch and then **from the root of the repo**:
    
    docker build -t puppetdb:alpine -f Dockerfile.slim .

You can then try it out in the pupperware stack by cloning the repo, modifying the puppetdb image here https://github.com/puppetlabs/pupperware/blob/master/docker-compose.yml#L39 to be your newly-built `puppetdb:alpine` image and then doing `docker-compose up`.

Sample pupperware stack output:
* Full output: https://gist.github.com/nwolfe/d4ec77618ce9e6672883bce706b56201
* New ssl.sh and ssl-setup.sh script output is this part: https://gist.github.com/nwolfe/d4ec77618ce9e6672883bce706b56201#file-gistfile1-txt-L107-L140

----

### Things to Note

* ⚠️ There's no more ruby or puppet in the image, so if people expect to be able to `docker exec` into the container and run things manually they'll probably be impacted by this work
* ⚠️ Similarly, if there's extension points or anything optional that might rely on ruby/puppet being available in the container, they'll probably be impacted
* ❤️ Running from source instead of packages means you can build a container with your local changes, and then use that container wherever you're using the official container. This means contributors (internal and external) could now test their PuppetDB changes work in the Pupperware stack, or wherever they're already using the official image.


Dockerfile.slim
```diff
@@ -0,0 +1,60 @@
...
+COPY docker/puppetdb/logging /etc/puppetlabs/puppetdb/logging/
+COPY docker/puppetdb/conf.d /etc/puppetlabs/puppetdb/conf.d/
+COPY resources/puppetlabs/puppetdb/bootstrap.cfg /etc/puppetlabs/puppetdb/
+COPY resources/ext/config/logback.xml /etc/puppetlabs/puppetdb/
+COPY resources/ext/config/request-logging.xml /etc/puppetlabs/puppetdb/
+RUN mkdir -p /opt/puppetlabs/server/data/puppetdb
+
+RUN addgroup $GROUP && adduser -S $USER -G $GROUP
+
+ADD https://raw.githubusercontent.com/puppetlabs/pupperware/puppetless-ssl/shared/ssl.sh /ssl.sh
```
nwolfe: We will change this URL from `puppetless-ssl` (the branch name) to a proper SHA so we "pin" the dependency.

docker/puppetdb/docker-entrypoint.sh
```diff
@@ -11,10 +11,10 @@ if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ] && [ "$USE_PUPPETSERVER" = true ]; th
     sleep 1
   done
   set -e
-  /opt/puppetlabs/bin/puppet config set certname "$HOSTNAME"
-  /opt/puppetlabs/bin/puppet config set server "$PUPPETSERVER_HOSTNAME"
-  /opt/puppetlabs/bin/puppet agent --verbose --onetime --no-daemonize --waitforcert 120
-  /opt/puppetlabs/server/bin/puppetdb ssl-setup -f
+  /ssl.sh
```
nwolfe: This one line replaces the 4 lines above it.

The script is hardcoded to:
* use `$HOSTNAME` (overridable via argument) for the cert subject and filename
* use `$PUPPETSERVER_HOSTNAME` (overridable via ENV variable though), 
* wait 120 seconds for the cert to be signed (overridable via ENV variable though) 

In other words these two invocations are the same:

```bash
/ssl.sh
PUPPETSERVER_HOSTNAME="$PUPPETSERVER_HOSTNAME" WAITFORCERT=120 /ssl.sh "$HOSTNAME"
```

docker/puppetdb/docker-entrypoint.sh
```diff
@@ -11,10 +11,10 @@ if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ] && [ "$USE_PUPPETSERVER" = true ]; th
...
+  /ssl-setup.sh -f
```
nwolfe: @puppetlabs/puppetdb I had to copy/paste/modify the current ssl-setup.erb file so it didn't require EZBake and could run under /bin/sh (alpine doesn't have bash).

This ssl-setup.sh script is specific to puppetdb so would remain here, and then we'd need to remember it when we make future changes to the ssl-setup.erb file.

We could probably refactor the ssl-setup.erb from bash to sh (I did that below) but not sure how to address the EZBake lines.

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+#
+# Example:
+#
+#   backupfile "/etc/myconfig"
+#
+function backupfile() {
+  # Create the global array if it doesn't already exist
+  if [ -z $backupfile_list ]; then
+    # backupfile_list=()
+    backupfile_list=""
```
nwolfe: Arrays aren't supported in `sh` but we can simulate them with a space-separated string.

I had to do this in 5 or 6 places in this script.

I left the original lines commented out so you can see the diff.

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
+#!/bin/sh
```
nwolfe: This is a copy/paste/modification of the existing ssl-setup.erb (https://github.com/puppetlabs/puppetdb/blob/master/resources/ext/cli/ssl-setup.erb)

I had to make this file sh-compatible as there's no bash on alpine.

Most of this file is unchanged, other than:
* arrays replaced with space-separated strings
* EZBake calls for user/group replaced with ENV variables of the same value (puppetdb/puppetdb)
* interactive prompts for files/dirs are commented-out since it wasn't trivial to change (but totally doable)
* puppet config print calls replaced with hardcoded directory paths

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+function contains() {
+#   local n=$#
+#   local value=${!n}
+#   for ((i=1;i < $#;i++)); do
+#     if [ "${!i}" == "${value}" ]; then
+#       return 0
+#     fi
+#   done
+#   return 1
+  echo "$1" | grep -q "$2"
```
nwolfe: Array refactoring here; made "contains" simpler

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+
+# This function copies the necessary PEM files from Puppet to the PuppetDB
+# SSL directory.
+#
+# This expects various environment variables to have already been set to work.
+function copy_pem_from_puppet {
+  # orig_files=($orig_ca_file $orig_private_file $orig_public_file)
+  orig_files="$orig_ca_file $orig_private_file $orig_public_file"
+  # for orig_file in "${orig_files[@]}"; do
+  for orig_file in $orig_files; do
```
nwolfe: Array refactoring here

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+done
+
+${interactive:=false}
+${force:=false}
+
+# Deal with interactive setups differently to non-interactive
+if $interactive
+then
+    echo "interactive mode not yet implemented"
+    exit 1
```
nwolfe: Disabled interactive mode here since refactoring to comma-separated strings wasn't as easy as it was in other places

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+  # This should be run on the host with PuppetDB
+  PATH=/opt/puppetlabs/bin:/opt/puppet/bin:$PATH
+  # agent_confdir=`puppet agent --configprint confdir`
+  agent_confdir="/etc/puppetlabs/puppet"
+  # agent_vardir=`puppet agent --configprint vardir`
+  agent_vardir="/etc/puppetlabs/puppet/cache"
+  # user=<%= EZBake::Config[:user] %>
+  # group=<%= EZBake::Config[:group] %>
+  user=$USER
+  group=$GROUP
```
nwolfe: Hardcoded paths and replaced EZBake calls in these lines (235-244)

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+
+# mycertname=`puppet agent --confdir=$agent_confdir --vardir=$agent_vardir --configprint  certname`
+mycertname="${HOSTNAME}"
+
+# orig_public_file=`puppet agent --confdir=$agent_confdir --vardir=$agent_vardir --configprint  hostcert`
+orig_public_file="/etc/puppetlabs/puppet/ssl/certs/${mycertname}.pem"
+# orig_private_file=`puppet agent --confdir=$agent_confdir --vardir=$agent_vardir --configprint hostprivkey`
+orig_private_file="/etc/puppetlabs/puppet/ssl/private_keys/${mycertname}.pem"
+# orig_ca_file=`puppet agent --confdir=$agent_confdir --vardir=$agent_vardir --configprint localcacert`
+orig_ca_file="/etc/puppetlabs/puppet/ssl/certs/ca.pem"
```
nwolfe: More hardcoded paths and $HOSTNAME usage here (lines 251-259)

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+
+# Scan through the old settings to see if any are still set, exiting and
+# prompting the user for the -f switch to force the tool to run anyway.
+if ! ${force}; then
+  # old_settings=('key-password' 'trust-password' 'keystore' 'truststore')
+  # new_settings=('ssl-key' 'ssl-cert' 'ssl-ca-cert')
+  old_settings="key-password trust-password keystore truststore"
+  new_settings="ssl-key ssl-cert ssl-ca-cert"
+  # for old_setting in "${old_settings[@]}"; do
+  for old_setting in $old_settings; do
```
nwolfe: Array refactoring here

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+  #   "${orig_public_file}:${public_file}"
+  # )
+  filediffs="
+    ${orig_ca_file}:${ca_file}
+    ${orig_private_file}:${private_file}
+    ${orig_public_file}:${public_file}
+  "
+
+  # for i in "${filediffs[@]}"; do
+  for i in $filediffs; do
```
nwolfe: Array refactoring here

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+  settings="
+    ssl-host:0.0.0.0
+    ssl-port:8081
+    ssl-key:${private_file}
+    ssl-cert:${public_file}
+    ssl-ca-cert:${ca_file}
+  "
+
+  # for i in "${settings[@]}"; do
+  for i in $settings; do
```
nwolfe: Array refactoring here

docker/puppetdb/ssl-setup.sh
```diff
@@ -0,0 +1,454 @@
...
+        fi
+      fi
+    fi
+  done
+
+  # Check old settings are commented out, and fix or warn
+  # settings=('keystore' 'truststore' 'key-password' 'trust-password')
+  settings="keystore truststore key-password trust-password"
+  # for setting in "${settings[@]}"; do
+  for setting in $settings; do
```
nwolfe: Array refactoring here

Dockerfile.slim
```diff
@@ -0,0 +1,60 @@
...
+
+COPY docker/puppetdb/docker-entrypoint.sh /
+RUN chmod +x /docker-entrypoint.sh
+
+EXPOSE 8080 8081
+
+ENTRYPOINT ["/sbin/tini", "-g", "--", "/docker-entrypoint.sh"]
+CMD ["services"]
+
+# COPY Dockerfile.slim /Dockerfile
```
underscorgan: Can we have a puppetdb-slim directory or something where this can live? At build time we can include all the files in the repo using something like `docker build --file puppetdb-slim/Dockerfile ../`

docker/puppetdb/docker-entrypoint.sh
```diff
@@ -11,10 +11,10 @@ if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ] && [ "$USE_PUPPETSERVER" = true ]; th
...
 fi
 
-exec /opt/puppetlabs/server/bin/puppetdb "$@"
+# exec /opt/puppetlabs/server/bin/puppetdb "$@"
+exec java -cp /puppetdb.jar clojure.main -m puppetlabs.puppetdb.core "$@" \
+    -c /etc/puppetlabs/puppetdb/conf.d/
```
underscorgan: You changed the entrypoint for the main puppetdb container but didn't include the new files in the non-slim dockerfile which breaks things

Dockerfile.slim
nwolfe: I'll look into it but I don't think so.

The reason we were able to do that before is because the Dockerfile was building from packages, which meant it didn't need to be scoped to the entire repo, just the few additional configuration files needed to make the package work inside the container.

In this case since we're building from source, the Dockerfile needs access to everything, and I'm not sure you can reference files that aren't in the same directory as the Dockerfile (i.e. you can't do "COPY ../file")

docker/puppetdb/docker-entrypoint.sh
nwolfe: Thanks.

What's broken?

I don't see any changes to the entrypoint script?

I don't follow.

Dockerfile.slim
underscorgan: They're using this for the puppetserver alpine images, as long as the PATH in the `docker build` is set to the top level of the repo, the dockerfile can live wherever.

docker/puppetdb/docker-entrypoint.sh
underscorgan: You edited the entrypoint in the `docker/puppetdb` directory to call the new ssl scripts and execute the jar, but the existing puppetdb containers install from packages and don't have the new ssl scripts so if I do a `make build` and a `make test` the tests fail (the non-alpine puppetdb container won't start)

nwolfe: Oh. I hadn't addressed CI and the make stuff yet, figured I would do that once the container changes were all done. That's the "WIP" part 😄

nwolfe: Ok started working through the CI/make stuff....

Got "make lint" and "make build" working, but "make test" fails miserably as the tests are tightly-coupled to the old package-based approach.

We'll probably have to re-think what tests make sense in this new puppetdb-from-source container rather than "fixing" the current tests.

underscorgan:
@nwolfe for some of the other projects we got rid of the `dockerfile_spec.rb` entirely since that was more testing that the image contained the expected things whereas the `puppetdb_spec.rb` tests that pdb is up and running, so maybe tests will be ok if we just use the puppetdb spec.

nwolfe:
We should get https://github.com/puppetlabs/puppetdb/pull/2753 merged in first so the Windows Azure CI patterns are up to date before I modify them to support this work.

nwolfe:
Currently I'm updating this to incorporate the CI/build changes from #2753 

@puppetlabs/puppetdb I need review on this, specifically around covering all the bases when building from source, and the copy/paste/modification of the `ssl-setup` script.

nwolfe:
Oh, looks like I need to retarget/rebase this onto `5.2.x` as `5.1.x` is gone and causing CI failures.

robdaemon:
@nwolfe yes, sorry - we EOL'ed `5.1.x` as of the first of the year. It shouldn't be too painful to rebase on `5.2.x`?

nwolfe:
Replaced with https://github.com/puppetlabs/puppetdb/pull/2775

PR #2746, opened 2018-12-12

robdaemon:
Reintroduces this feature


robdaemon:
This is ready, but it should be merged after #2806 and #2800.

PR #2757, opened 2019-01-02
(PDB-4221) upgrade to clojure 1.10

robdaemon:
Upgrades to Clojure 1.10 and Dynapath 1.0.0 to match puppetlabs/clj-parent#150


robdaemon:
This PR is a test to see if the clj-parent upgrade will work in pdb, letting the robots run our tests.

robdaemon:
Okay, learned that this should be fine for us - the needed core.async change was made to the clj-parent PR.

PR #2775, opened 2019-01-08
(MAINT) PuppetDB alpine image from source

nwolfe:
This reduces the final image size from 434MB down to 115MB.

There's no more Ruby or puppet-agent in the container, and it's based
on Alpine rather than Ubuntu.

Since this builds from local source instead of released packages,
developers can now go further in their testing of PuppetDB changes by
simply changing their local clone and running "cd docker && make build".

----

⚠️ See https://github.com/puppetlabs/puppetdb/pull/2738 for context, comments, links ⚠️


Iristyle:
FYI - this nearly doubles how long it takes to build the image. I suspect this is because we've gone from 30 steps to 51. I wonder if there's any way we can reduce the step count down?

Dockerfile.slim
```diff
@@ -40,9 +40,12 @@ COPY resources/ext/config/logback.xml /etc/puppetlabs/puppetdb/
 COPY resources/ext/config/request-logging.xml /etc/puppetlabs/puppetdb/
 RUN mkdir -p /opt/puppetlabs/server/data/puppetdb
 
+RUN addgroup $GROUP && adduser -S $USER -G $GROUP
+
+ADD https://raw.githubusercontent.com/puppetlabs/pupperware/6487f9f9bee24a5330b1709cefee8cc10c4bf6d2/shared/ssl.sh /ssl.sh
```
Iristyle: Now that `ssl.sh` is merged, I think we can update this to a better URL

nwolfe:
> FYI - this nearly doubles how long it takes to build the image in Azure. I suspect this is because we've gone from 30 steps to 51. I wonder if there's any way we can reduce the step count down?

Yeah, the build time wasn't a concern of mine while doing this work.  

I figured we'd actually get better build times because:
1) Pulling deps is in its own layer so that'll be cached rather than happen every time, and
2) More steps mean more opportunities for layer caching

But I didn't keep tabs on whether this actually improved things.

Also, in general, how much of a concern is build time really?

An image is only ever built once, but pulled many times, so favoring a smaller image size over a longer build time seems desirable to me.

On the other hand, slow build times would make PR checks take longer, but then I would hope my points above would come into play.

Either way, good observation.

Dockerfile.slim
nwolfe: Done: https://github.com/puppetlabs/puppetdb/pull/2775/commits/1829762d70210e52b5c9e2bdd8e1545a00e6bc60

docker/puppetdb/docker-entrypoint.sh
```diff
@@ -1,24 +1,28 @@
...
 fi
 if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ] && [ "$USE_PUPPETSERVER" = true ]; then
-  /opt/puppetlabs/server/bin/puppetdb ssl-setup -f
+  # /opt/puppetlabs/server/bin/puppetdb ssl-setup -f
+  /ssl-setup.sh -f
 fi
 
-exec /opt/puppetlabs/server/bin/puppetdb "$@"
+# exec /opt/puppetlabs/server/bin/puppetdb "$@"
+exec java -cp /puppetdb.jar clojure.main -m puppetlabs.puppetdb.core "$@" \
```
robdaemon: This needs a reference to `$JAVA_ARGS`, like:

```
exec java -cp /puppetdb.jar $JAVA_ARGS clojure.main -m puppetlabs.puppetdb.core "$@" \
    -c /etc/puppetlabs/puppetdb/conf.d/
```

robdaemon:
This looks great - just one small bug I saw.

docker/puppetdb/docker-entrypoint.sh
nwolfe: Added and verified here: https://github.com/puppetlabs/puppetdb/pull/2775/files#diff-86cc2c2467f4688ca87a833d8bf198b2R22

nwolfe:
@robdaemon Backlink comment added here: https://github.com/puppetlabs/puppetdb/pull/2775/files#diff-8cc40f9421301d8244ed1efd3145186aR3

PR #2786, opened 2019-01-11
Test merge 6.0.x

austb:
[facebones-ignore]


ext/test/top-level-cli
```diff
@@ -2,11 +2,16 @@
 
 set -uexo pipefail
 
-usage() { echo "Usage: [PDB_JAR=JAR] $(basename "$0")"; }
+usage() { echo "Usage: [PDB_JDK=openjdk10] [PDB_JAR=JAR] $(basename "$0")"; }
```
rbrw: I'm wondering whether we could just detect the java version easily enough for that to be preferable, i.e. via a suitable `case` or something, e.g.:
```shjava -version
java_ver=$(java -version 2>&1 | head -1 | cut -d' ' -f 3 | sed -e 's/^"\(.\+\)"$/\1/')
case "$java_ver" in
    10|11)
        ...
        ;;
    *) ...
        ;;;
esac
```
or whatever.  Might simplify the rest of the commit, and avoid the possibility of receiving an incorrect PDB_JDK value.

And actually, if we're going to do that, suppose we could just add a trivial `ext/bin/jdk-major` script so we can centralize the logic.  I suspect for anything before 10, we can just take the Y from the version, and for 10+ the `java_ver` computed above would be right.

robdaemon: This might work, but could be fragile too. We've seen Oracle change the versioning string on the JDK before, breaking detection of the JDK flavor.

PR #2789, opened 2019-01-15
(MAINT) Introduce docker-entrypoint.d/

nwolfe:
Decompose the single docker-entrypoint.sh script into a directory of
ordered scripts.

This follows the same pattern that was introduced recently in the
puppetserver container.

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/10-configure-ssl.sh
+++ b/docker/puppetdb/docker-entrypoint.d/10-configure-ssl.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+master_running() {
+    status=$(curl --silent --fail --insecure "https://${PUPPETSERVER_HOSTNAME}:8140/status/v1/simple")
+    test "$status" = "running"
+}
+
+PUPPETSERVER_HOSTNAME="${PUPPETSERVER_HOSTNAME:-puppet}"
+
+if [ ! -f "/etc/puppetlabs/puppet/ssl/certs/${HOSTNAME}.pem" ] && [ "$USE_PUPPETSERVER" = true ]; then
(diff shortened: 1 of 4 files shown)


docker/puppetdb/docker-entrypoint.d/10-configure-ssl.sh
```diff
@@ -0,0 +1,20 @@
...
+  # if this is our first run, run puppet agent to get certs in place
+  while ! master_running; do
+    sleep 1
+  done
+  set -e
+  /ssl.sh
+fi
+if [ ! -d "/etc/puppetlabs/puppetdb/ssl" ] && [ "$USE_PUPPETSERVER" = true ]; then
+  /ssl-setup.sh -f
+fi
```
nwolfe: I initially tried breaking this script up into 3 distinct parts:
* 10-wait-for-puppetserver.sh (just does the curl+sleep)
* 20-get-ssl-cert.sh (calls /ssl.sh)
* 30-setup-ssl.sh (calls /ssl-setup.sh -f)

But that ended up ballooning out too much I thought, and considering how tightly coupled these steps are I ended up leaving them as one.

At that point this PR was a little less desirable, as I wasn't really decomposing anything.

But, we'll likely have a new script for analytics that we'll want to drop in here to run first, something like `00-analytics.sh` or whatever.  It'll be nice to have this pattern in place for that work, and just in general to keep the entrypoint logic decoupled from one another.

Iristyle:
You got the `+x` bit it looks like, but I think we need a corresponding `.gitattributes` entry like https://github.com/puppetlabs/puppetdb/blob/master/.gitattributes#L1 to ensure we get `LF` and not `CRLF` inside the Linux containers.

PR #2799, opened 2019-01-22
(PDB-3742) Add atom to track enqueued commands

Zak-Kent:
This commit adds an atom which tracks commands that are enqueued but not
yet processed. The pdbext HA-sync code uses this atom to avoid enqueuing
duplicate commands when performing a sync.

### DIFF
--- a/src/puppetlabs/puppetdb/client.clj
+++ b/src/puppetlabs/puppetdb/client.clj
@@ -9,8 +9,7 @@
             [puppetlabs.puppetdb.schema :refer [defn-validated]]
             [puppetlabs.puppetdb.utils :as utils]
             [puppetlabs.kitchensink.core :as kitchensink]
-            [schema.core :as s]
-            [clj-time.coerce :as c]))
+            [schema.core :as s]))
 
 (defn get-metric [base-url metric-name]
   (let [url (str (utils/base-url->str base-url)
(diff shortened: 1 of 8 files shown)


src/puppetlabs/puppetdb/queue.clj
```diff
@@ -400,36 +400,52 @@
    ;; to get rid of it somehow when we refactor mq-listener and command.clj.
    (SortedCommandBuffer. (TreeMap.) (HashMap.) n delete-update-fn)))
 
+(defn make-q-cmd-map
```
Zak-Kent: rename this func and update doc string, these maps are passed along the channel now and not put into the atom directly

src/puppetlabs/puppetdb/client.clj
```diff
@@ -41,10 +40,12 @@
     payload :- {s/Any s/Any}
     timeout]
    (let [body (json/generate-string payload)
+         ;; is it ok to not account for benchmark sending string keys when everyting else uses :producer_timestamp???
+         producer_timestamp (-> payload :producer_timestamp str)
```
Zak-Kent: Getting "" here for producer_timestamp with benchmark might not be what we want.

austb:
```
         ;; is it ok to not account for benchmark sending string keys when everyting else uses :producer_timestamp???
         producer_timestamp (-> payload :producer_timestamp)
         url_params (str
                      (format "?command=%s&version=%s&certname=%s"
                        (str/replace command #" " "_") version certname)
                      (when producer_timestamp
                        (format "&producer_timestamp=%s" producer_timestamp)))
                  url_params
                  (when timeout (format "&secondsToWaitForCompletion=%s" timeout)))]
```

Something like that will prevent benchmark from triggering bash in place but other optimizations still make benchmark kinda useless as a load testing tool.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -614,18 +622,21 @@
                    ex
                    (trs "[{0}] [{1}] Retrying after attempt {2} for {3}, due to: {4}"
                         id command retries certname ex))
+                  ;; is it correct to assume that retried messages will eventually hit another path
+                  ;; which calls update-cmd-tracking! and causes the cmd in question to get cleared
+                  ;; out of the q-cmds-atom?
```
austb: Drop this! Probably with a note that it _is_ getting retried.

PR #2804, opened 2019-01-25
(MAINT) Add container analytics

nwolfe:
Add an entrypoint script to submit a "start" event to our pupperware
account on Google Analytics.

The script will generate a UUID and reuse it as an anonymous client ID,
in an attempt to more accurately identify the number of unique instances
of the container. This UUID file is kept in /var/tmp, which doesn't
require explicit Docker volumes and is persisted across container
restarts.

The `ec` field is for the Event Category, which can either be "dev" or
"production", and is a configured as a container build argument. Only
the containers that we ship to DockerHub should be built as
"production", and the default is "dev".

Tests should not report as either though, so we specify
PUPPERWARE_DISABLE_ANALYTICS when running tests.

Users can disable analytics using a non-empty value for
PUPPERWARE_DISABLE_ANALYTICS, and also specify which stream to report as
using PUPPERWARE_ANALYTICS_STREAM=sometext.


nwolfe:
Example output:
```
(/docker-entrypoint.d/10-analytics.sh) Sending metrics http://www.google-analytics.com/collect?v=1&t=event&tid=UA-132486246-2&an=puppetdb&av=5.2.7&cid=39cf8f14-b5fc-45d4-9b20-66dfa0867722&ec=production&ea=start
Running /docker-entrypoint.d/20-configure-ssl.sh
...
```

And with analytics disabled:
```
Running /docker-entrypoint.d/20-configure-ssl.sh
...
```

docker/puppetdb/docker-entrypoint.d/10-analytics.sh
```diff
@@ -0,0 +1,27 @@
...
+# Event Category
+ec=${PUPPERWARE_ANALYTICS_STREAM:-dev}
+# Event Action
+ea=start
+
+_params="v=1&t=event&tid=${tid}&an=${an}&av=${av}&cid=${cid}&ec=${ec}&ea=${ea}"
+_url="http://www.google-analytics.com/collect?${_params}"
+
+echo "($0) Sending metrics ${_url}"
+curl --silent --show-error -X POST -H "Content-Length: 0" $_url 1> /dev/null
```
nwolfe: The endpoint returns some garbage text, like `GIF89a�����,D;` (seriously).

I wanted to silence that but I can't distinguish between errors and normal output in this case.
I tried silencing everything __except__ errors but I couldn't get it to work.

I even tried looking up WTF is being returned from this endpoint and why (did I curl wrong?) but couldn't find anything 🤷‍♂️ 

So I just did my best and added `1> /dev/null` to get rid of that weird line in the logs, but this means if the curl ever fails we wouldn't see it (why doesn't it output to stderr?). 

I figured it was fine for now.

docker/puppetdb/Dockerfile
```diff
@@ -18,6 +18,10 @@ COPY --from=builder /app/target/puppetdb.jar /
 ARG vcs_ref
 ARG build_date
 ARG version="5.1.5"
+# used by entrypoint to submit metrics to Google Analytics;
+# published images should use "production" for this build_arg
+ARG pupperware_analytics_stream="dev"
```
Iristyle: I guess since this defaults to `dev`, no need to update on the Azure side?

docker/puppetdb/Dockerfile
```diff
@@ -18,6 +18,10 @@ COPY --from=builder /app/target/puppetdb.jar /
...
+ENV PUPPERWARE_ANALYTICS_STREAM="$pupperware_analytics_stream"
```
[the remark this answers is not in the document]
nwolfe: The difference between pdb and puppetserver was because puppetserver has the parent-child image and I wanted the child image to reuse as much from the parent.

To me it seems ideal to keep as much as possible defined in the `10-analytics.sh` script, so I strived for that and added ENV/ARG values in the Dockerfile only as necessary.

They're not meant to be overridden, and each container has its own analytics.sh script so we're not trying to reuse a generic analytics.sh script (like we do for ssl.sh).

docker/puppetdb/Dockerfile
nwolfe: Right.

The only place it's **not** dev is when Distelli builds it and specifies `make build STREAM=production`

docker/puppetdb/docker-entrypoint.d/10-analytics.sh
nwolfe: I think so, but lemme try again 🤞

nwolfe: Yep, this is what I needed, in addition to `--output /dev/null`, gives us the exact behavior we want.

Updated.

PR #2806, opened 2019-01-25
(PDB-2487) remove events from reports hash

robdaemon:
Removes the report events from the calculation of a report's hash. This
will allow garbage collection of resource events. During a sync between
puppetdb instances, the report's hash will change if the events are
missing, which will lead to duplicate reports in the database.


robdaemon:
one last thing to do here - the metrics json has links in it that include the old hashes. Need to update those also during the migration.

PR #2811, opened 2019-01-30
(PDB-3284) Make PQL !~<=> not need leading ws

austb:
Previously, due to greedy Java regexes, the special character operators
such as =, >, >=, ~, etc all required a leading space. This makes the
leading space optional. The goal is that only spaces required are around
the word operators such as 'in', 'and', and 'or'.


austb:
We'll need to hold off until 7.0.0 since this will break queries against names ending in one of `<=>!~`, which I hope no one is doing, but we do currently support. I'll raise this PR again as PDB 7.0.0 approaches.

PR #2813, opened 2019-02-04
(maint) If consul is enabled, register puppetdb service on startup

underscorgan:


Iristyle:
Retargeted at `master` so that we don't forget in the future.

underscorgan:
This should be good to go!

docker/puppetdb/docker-entrypoint.d/30-consul.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+if [ "$CONSUL_ENABLED" = "true" ]; then
+  ipaddress="$(ifconfig eth0 | grep 'inet addr' | cut -d ':' -f 2 | cut -d ' ' -f 1)"
```
nwolfe: I was wondering about the availability of `ifconfig`, and noticed that it does not exist in the `ubuntu` and `centos` images. It does exist in `alpine` though.

So something else that we install in these images causes this to be brought in?

Just wanted to point it out, not sure there's anything to do about it though.

nwolfe: Looks like you get `ifconfig` via the `net-tools` package on Ubuntu.

Iristyle:
Retriggered PDB container tests (since earlier cells failed) and it passes.

docker/puppetdb/docker-entrypoint.d/30-consul.sh
```diff
@@ -0,0 +1,28 @@
...
+  consul_hostname="${CONSUL_HOSTNAME:-consul}"
```
Iristyle: In the puppetserver PR, we made these values `CONSULT_HOSTNAME` and `CONSUL_PORT` for consistency.

underscorgan: @Iristyle updated

Iristyle:
Kicked again to run the container tests. Still green - merging!

PR #2814, opened 2019-02-05
(MAINT) Configurable image name

nwolfe:
Allow the Docker image namespace "puppet" (i.e. puppet/puppetdb) to be
configurable, which makes it easier for local development of the image.

We don't want to push these images to DockerHub ever, so we don't honor
the namespace override in the `make publish` target.


docker/Makefile
```diff
@@ -1,4 +1,5 @@
 PUPPERWARE_ANALYTICS_STREAM ?= dev
+NAMESPACE ?= puppet
```
Iristyle: Should make this change on Windows side as well

nwolfe: This would never be used in CI, only by devs while they build locally with their changes.

What would we change on the Windows side for this (if that even is desirable)?

Iristyle: If someone is dev'ing on non-Windows, they should be able to dev on Windows as well

nwolfe: Sure, but that's what I was asking you about.

The `build.ps1` file has a `Namespace` parameter for the `Build-Container` function, which seems like enough.

But that's not used the same way the Makefile is, so there's no corresponding change on the Windows side I don't think?

nwolfe:
Replaced by #2822

PR #2815, opened 2019-02-08
(maint) Remove docker builds / assets from 5.x

Iristyle:
- Pupperware is only interested in 6.0.x + master at this point,
   so remove docker assets from old branches to prevent running
   in Azure + TravisCI

 - Merge up through all 5.x series branches, then on the 6.x series
   merge, keep what's in 6.x


Iristyle:
Also FYI - we're going to merge this up through all the 5.x branches... but when we merge up to 6.x, we'll be careful to keep the files in the 6.x branch. Then no future merge ups from 5.x will alter Docker related files.

PR #2823, opened 2019-02-12
(MAINT) Add container shutdown hooks

nwolfe:
Add a directory of scripts to be executed when the container shuts down,
for things like Google Analytics metrics and Consul deregistration.

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/10-analytics.sh
+++ b/docker/puppetdb/docker-entrypoint.d/10-analytics.sh
@@ -1,28 +1,3 @@
 #!/bin/sh
 
-if [ "${PUPPERWARE_DISABLE_ANALYTICS}" = "true" ]; then
-    echo "($0) Pupperware analytics disabled; skipping metric submission"
-    exit 0
-fi
-
-# See: https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters
-# Tracking ID
-tid=UA-132486246-1
-# Application Name
-an=puppetdb
-# Application Version
-av=$PUPPETDB_VERSION
-# Anonymous Client ID
-_file=/var/tmp/pwclientid
-cid=$(cat $_file 2>/dev/null || (cat /proc/sys/kernel/random/uuid | tee $_file))
-# Event Category
-ec=${PUPPERWARE_ANALYTICS_STREAM:-dev}
-# Event Action
-ea=start
-
-_params="v=1&t=event&tid=${tid}&an=${an}&av=${av}&cid=${cid}&ec=${ec}&ea=${ea}"
-_url="http://www.google-analytics.com/collect?${_params}"
-
-echo "($0) Sending metrics ${_url}"
-curl --fail --silent --show-error --output /dev/null \
-    -X POST -H "Content-Length: 0" $_url
+/analytics.sh start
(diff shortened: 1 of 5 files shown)


nwolfe:
@Iristyle @underscorgan What do you think of this approach? Any concerns?

I guess I'm not sold on it, not exactly sure why, maybe because of all that it requires and how fragile it kinda seems?

Either way I learned a lot of wrote up my notes here: https://github.com/puppetlabs/pupperware-commercial/wiki/Container-Shutdown

underscorgan:
I think I like this. What feels fragile to you?

Mostly I like how analogous it is to the docker-entrypoint.d pattern we're using everywhere, and I like the fact that each individual script doesn't have to deal with TRAPs, that's all in the top-level script

nwolfe:
> What feels fragile to you?

Thinking out loud here.....Maybe it's the fact that the behavior is so dependent on what shebang line you have in your entrypoint script (i.e. what shell you're using), and then that it seems easy to accidentally have your hook called multiple times, like if you forget the `trap '' EXIT` line before calling `exit $statuscode`....

And then that you need a package like `su-exec` or `gosu`, and there isn't an `su-exec` package readily available on Ubuntu/Debian distros. I haven't tried `gosu` yet though, but I avoided it because it was a larger package (filesize) and does more than just what su-exec does.

So to call out that point above, now that I'm remembering it.....

⚠️ There's no `su-exec` package on Debian/Ubuntu ⚠️ 

There are requests for a package on the su-exec Github, or a statically linked binary at least. But we'd have to figure out how to use `su-exec` and/or `gosu` on Debian/Ubuntu.

docker/puppetdb/docker-entrypoint.d/10-analytics.sh
```diff
@@ -1,28 +1,3 @@
...
-# Event Action
-ea=start
-
-_params="v=1&t=event&tid=${tid}&an=${an}&av=${av}&cid=${cid}&ec=${ec}&ea=${ea}"
-_url="http://www.google-analytics.com/collect?${_params}"
-
-echo "($0) Sending metrics ${_url}"
-curl --fail --silent --show-error --output /dev/null \
-    -X POST -H "Content-Length: 0" $_url
+/analytics.sh start
```
Iristyle: Hmmmmm... I'm not sure about having one script in `docker-entrypoint.d` and then another script outside, but I understand why you did it (although `stop` is not implemented). Seems like it might confuse us later on (and also don't like we have to add additional `COPY` steps in the `Dockerfile` rather than just using our established pattern).

Will have to think about this a bit more.

nwolfe: Stop is implemented.

If we don't have a shared file at `/analytics.sh` then we'd have to have the exit script call into the entrypoint scripts, which is worse than this situation.

I think we're going to end up with some common files at `/` (or somewhere else, `/pupperware` or whatever) like we have with `/ssl.sh`.

That'd just be another pattern that we establish!

Good points though, no rush.

Iristyle: Oh oh... I see that `stop` is just a value passed in the URI. I thought it was being used to alter behavior in the script - my mistake.

In any event, looks like we're going to close this per our discussion earlier anyhow!

Iristyle:
We've decided to not move forward with this for now - but may revisit in the future if the added complexity is warranted!

PR #2827, opened 2019-02-25
(RE-12078) Bump to ezbake 1.9.6

mwaggett:
This commit bumps the ezbake version to 1.9.6, which adds support for el 8.


robdaemon:
build failure here isn't from the change, it's from the fun with java certs on Mac OS X

PR #2829, opened 2019-02-26
(maint) Use a network instead of `--link` in the container tests

underscorgan:


docker/puppetdb/spec/puppetdb_spec.rb
```diff
@@ -106,6 +107,11 @@ def start_puppetdb
...
   before(:all) do
     @mapped_ports = {}
+    # Windows doesn't have the default 'bridge network driver
+    network_opt = File::ALT_SEPARATOR.nil? ? '' : '--driver=nat'
+
+    @network = %x(docker network create #{network_opt} puppetdb_test_network).chomp
+
     @postgres_container = run_postgres_container
 
     @pdb_image = ENV['PUPPET_TEST_DOCKER_IMAGE']
```
nwolfe: Not something you introduced, but...

Can we move this up to be the first part of `before(:all)` ?

We should fail-fast if the image name isn't provided, before we start creating anything or running containers (or doing anything, really).

underscorgan: updated in edffe1d13c04f56dda2de3922c08780545020f3d

Iristyle:
Retriggered Travis container tests, and they've passed.

PR #2830, opened 2019-02-27
(PDB-4270) Support "configure expiration" command

rbrw:
Add a new "configure expiration" command which can be used to disable
factset expiration for a given certname.  This is intended to support
nodes that are managed more directly, and/or are more ephemeral than
nodes that are managed by the agent.

### DIFF
--- a/documentation/api/command/v1/commands.markdown
+++ b/documentation/api/command/v1/commands.markdown
@@ -15,6 +15,7 @@ canonical: "/puppetdb/latest/api/command/v1/commands.html"
 [reportv7]: ../../wire_format/report_format_v7.html
 [reportv8]: ../../wire_format/report_format_v8.html
 [deactivatev3]: ../../wire_format/deactivate_node_format_v3.html
+[expirev1]: ../../wire_format/configure_expiration_format_v1.html
 
 Commands are used to change PuppetDB's model of a population. Commands are represented by `command objects`,
 which have the following JSON wire format:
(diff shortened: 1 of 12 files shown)


documentation/api/command/v1/commands.markdown
```diff
@@ -204,6 +208,16 @@ The payload is expected to be a report, containing events that occurred on
 Puppet resources. It is structured as a JSON object, conforming to the
 [report wire format v5][reportv5].
 
+### "configure expiration", version 1 (experimental)
+
+The payload should be a JSON format command, conforming to the
+[configure_expiration wire format v1][expirev1], indicating whether or
```
robdaemon: I think you mean `[]()` here, not `[][]`

documentation/api/wire_format/configure_expiration_format_v1.markdown
```diff
@@ -0,0 +1,77 @@
...
+> *Note*: this is an experimental command, which might be altered or
+> removed in a future release, and for the time being, PuppetDB
+> exports will not include this information.
+
+Configure expiration command format
+-----
+
+### Version
+
+This is **version 1** of the catalog interchange format.
```
robdaemon: s/catalog/configure expiration/

documentation/api/command/v1/commands.markdown
rbrw: Hmm, I think [][] may be right for a "reference link" (it's pointing to the top of the file) -- see the other examples above and below.

documentation/api/wire_format/configure_expiration_format_v1.markdown
rbrw: Hah, right -- thanks.

austb:
rebase looks good 👍

PR #2832, opened 2019-02-27
(PDB-3911) adding pk to resource_events table

robdaemon:
This will allow this table to be cleaned with pg_repack instead of relying
on a VACUUM FULL, which avoids locks on the table.


robdaemon:
Supersedes #2800

PR #2834, opened 2019-02-27

robdaemon:
* Adds a config parameter: database.resource-events-ttl with a default of 14d
* Adds two admin metrics: resource-events-purges and resource-events-purge-time
* Adds documentation for this new feature

This will allow the resource_events table to be cleaned up at a different interval
than the reports table.

### DIFF
--- a/test/puppetlabs/puppetdb/integration/resource_events.clj
+++ b/test/puppetlabs/puppetdb/integration/resource_events.clj
@@ -1,0 +1,46 @@
+(ns puppetlabs.puppetdb.integration.resource-events
+  (:require [clojure.test :refer :all]
+            [me.raynes.fs :as fs]
+            [puppetlabs.puppetdb.integration.fixtures :as int]
+            [puppetlabs.puppetdb.testutils.services :as svc-utils]
+            [clj-time.core :refer [now]]
+            [clj-time.core :refer [now days plus]]
+            [metrics.counters :as counters]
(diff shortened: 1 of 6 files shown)


robdaemon:
Supersedes #2746

test/puppetlabs/puppetdb/integration/resource_events.clj
```diff
@@ -0,0 +1,46 @@
...
+    (with-open [pdb (int/run-puppetdb pg {})
+                ps (int/run-puppet-server [pdb] {})]
+      (testing "Run agent once to populate database"
+        (int/run-puppet-as "ttl-agent" ps pdb "notify { 'irrelevant manifest': }"))
+
+      (testing "Verify we have resource events"
+        (is (= 1 (count (int/pql-query pdb "events { timestamp > 0 }"))))))
+
+    (testing "Sleep for one second to make sure we have a ttl to exceed"
+      (Thread/sleep 1000))
```
rbrw: When adding another recent test, I found out that we can probably set `ms` ttls, i.e. `1ms`.  Might want to double-check me on that (and I'll need to fix the other code if I was wrong), but if that's right, then might be worth changing this sleep to say 2ms or something.  Doesn't matter for this one test, but over time, they add up...

robdaemon: Ahhh, I'll give that a shot here

austb:
Obsoleted by puppetlabs/puppetdb#3027

PR #2835, opened 2019-02-27
(PDB-2488) remove events from reports hash

robdaemon:
Removes the report events from the calculation of a report's hash. This
will allow garbage collection of resource events. During a sync between
puppetdb instances, the report's hash will change if the events are
missing, which will lead to duplicate reports in the database.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1575,6 +1575,118 @@
     "     check ((expire is not null and updated is not null)"
     "             or (expire is null and updated is null)))"]))
 
+(defn change-reports-hash-to-summary-hash
+  "Changes the reports table primary key hash to remove the events from
+  the hash calculation. This allows events to be garbage collected separately
+  from reports"
+  []
+  (jdbc/do-commands
+   ["CREATE TABLE reports_transform ("
+    "    id bigint not null,"
+    "    hash bytea not null,"
+    "    transaction_uuid uuid,"
+    "    certname text not null,"
+    "    puppet_version text not null,"
+    "    report_format smallint not null,"
+    "    configuration_version text not null,"
+    "    start_time timestamp with time zone not null,"
+    "    end_time timestamp with time zone not null,"
+    "    receive_time timestamp with time zone not null,"
+    "    noop boolean,"
+    "    environment_id bigint,"
+    "    status_id bigint,"
+    "    metrics_json json,"
+    "    logs_json json,"
+    "    producer_timestamp timestamp with time zone not null,"
+    "    metrics jsonb,"
+    "    logs jsonb,"
+    "    resources jsonb,"
+    "    catalog_uuid uuid,"
+    "    cached_catalog_status text,"
+    "    code_id text,"
+    "    producer_id bigint,"
+    "    noop_pending boolean,"
+    "    corrective_change boolean,"
+    "    job_id text)"])
+
+  (jdbc/call-with-query-rows
+   ["select
+       id, transaction_uuid, certname, puppet_version, report_format, configuration_version,
+       start_time, end_time, receive_time, noop, environment_id, status_id, metrics_json,
+       logs_json, producer_timestamp, metrics, logs, resources, catalog_uuid, cached_catalog_status,
+       code_id, producer_id, noop_pending, corrective_change, job_id
+     from reports"]
+   (fn [rows]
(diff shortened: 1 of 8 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1575,6 +1575,118 @@
...
+       id, transaction_uuid, certname, puppet_version, report_format, configuration_version,
+       start_time, end_time, receive_time, noop, environment_id, status_id, metrics_json,
+       logs_json, producer_timestamp, metrics, logs, resources, catalog_uuid, cached_catalog_status,
+       code_id, producer_id, noop_pending, corrective_change, job_id
+     from reports"]
+   (fn [rows]
+     (reduce (fn [hashes-seen row]
+               (conj hashes-seen
+                     (let [hash (-> (hash/report-identity-hash row)
+                                    (sutils/munge-hash-for-storage))]
```
rbrw: Trivial, and feel free to leave it alone, but I think it may be typical to omit the threading form parens when the function takes no args.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1575,6 +1575,118 @@
...
+                       (when-not (contains? hashes-seen hash)
```
rbrw: If you feel strongly about it, no worries, but I believe it's generally idiomatic to avoid contains? in favor of using the set itself as the function (e.g. `(hashes-seen hash)`.  Though I could see a clarity argument on the other side if the set name's not sufficiently evocative...

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1575,6 +1575,118 @@
...
+                         (jdbc/insert-multi! "reports_transform"
+                                             (list (assoc row :hash hash))))
+                       hash)))
+             #{}
+             rows)))
```
rbrw: Overall I think this (and the pr) looks good, though I had wondered if we might want to handle the insert in chunks.  That said, doing so would make the code more complicated if it doesn't matter, and we'd need to be careful not to make the partition size too big relative to the row sizes (cf. potentially giant factsets when dealing with facts).

In any case, here's the kind of thing I was pondering (untested, and could be structured other ways):

```clojure
(letfn [(insert-row-chunk [hashes-seen chunk-rows]
          (let [[hashes-seen chunk-rows]
                (loop [hashes-seen hashes-seen
                       remainder chunk-rows
                       result []]
                  (if-not (seq remainder)
                    [hashes-seen result]
                    (let [hash (-> (hash/report-identity-hash row)
                                   sutils/munge-hash-for-storage)]
                      (if (hashes-seen hash)
                        (recur hashes-seen (rest remainder) result)
                        (recur (conj hashes-seen hash)
                               (rest remainder)
                               (conj result (assoc row :hash hash)))))))]
            (jdbc/insert-multi! "reports_transform" chunk-rows)
            hashes-seen))]
  (reduce insert-chunk #{} (partition 100 rows)))
```

src/puppetlabs/puppetdb/scf/migrate.clj
robdaemon: I just removed the use of this - I have a concern with the set containing byte arrays. I've previously had trouble with that sort of thing, as the hashcode doesn't end up being what you expect. I just changed this to use the string of the hash in the set, as that definitely does what I expect.

src/puppetlabs/puppetdb/scf/migrate.clj
robdaemon: changed to `(hashes-seen hash-str)`

src/puppetlabs/puppetdb/scf/migrate.clj
robdaemon: I'm not sure this makes much of a difference in the long run - the migration function starts a single transaction, and this code doesn't appear to be hanging on to the head. transaction: https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/scf/migrate.clj#L1726-L1739

I'd rather keep it mess complicated if possible.

rbrw: Yeah, the added complication may not be worth it, particularly since it sounds like the runtime was about the same as it was in the earlier version that didn't have the set and was partitioned.  So I think I'm fine sticking with the simpler version.

robdaemon:
closing in favor of #3161

PR #2837, opened 2019-02-28
(PDB-4277) Add logback logstash encoder

Sharpie:
This commit adds a dependency on net.logstash.logback/logstash-logback-encoder
which is described in the PuppetDB documentation as a tool for transforming
log messages into JSON format:

  https://puppet.com/docs/puppetdb/6.2/logging.html#json-text

Prior to this commit, the required library was missing from the uberjar. The
library is managed by clj-parent as Puppet Server includes it.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -177,6 +177,10 @@
                  [robert/hooke "1.3.0"]
                  [slingshot]
                  [trptcolin/versioneer]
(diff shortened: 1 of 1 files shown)


austb:
Thanks! We'll get this merged in https://github.com/puppetlabs/puppetdb/pull/2840 into the 5.2.x branch

PR #2839, opened 2019-03-01
(PDB-4271) query for fact expiration

robdaemon:
Adds an "expires" field to the node response, that contains two embedded
fields: "facts" and "facts_updated"

If "facts" is true (the default), the node is subject to garbage
collection and expiration. If the value is set to false, the node is
excluded from the garbage collection procedure. If this value has been
explicitly set via a command, the facts_updated field will be a
timestamp representing when the field was set.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -48,7 +48,8 @@
    "fact_contents" {:columns ["certname"]}
    "events" {:columns ["certname"]}
    "edges" {:columns ["certname"]}
-   "resources" {:columns ["certname"]}})
+   "resources" {:columns ["certname"]}
+   "certname_fact_expiration" {:columns ["certid"]}})
 
 (def type-coercion-matrix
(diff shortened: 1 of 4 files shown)


robdaemon:
This is going to need a rebase - it's based on @rbrw's change that adds the table to the database

src/puppetlabs/puppetdb/command.clj
```diff
@@ -387,6 +387,28 @@
...
+ :certname "foo"
+ :version 1
+ :expire {:facts false}}
+
+(defn configure-expiration
+  [{:keys [id received payload]}
+   start-time db]
+  (let [certname (:certname payload)
+        stamp (:producer_timestamp payload (now))
+        expire-facts? (get-in payload [:expire :facts])]
```
Zak-Kent: We'll need behavior similar to how we handle :producer_timestamp with other commands with the :facts_updated timestamp. :facts_updated needs to stay the same across multiple submissions in the sync context so we can make decisions about which "configure expiration" commands are newer than others and need to be synced.

PR #2840, opened 2019-03-01
(PDB-4277) Add logback logstash encoder

austb:
This commit adds a dependency on net.logstash.logback/logstash-logback-encoder
which is described in the PuppetDB documentation as a tool for transforming
log messages into JSON format:

  https://puppet.com/docs/puppetdb/6.2/logging.html#json-text

Prior to this commit, the required library was missing from the uberjar. The
library is managed by clj-parent as Puppet Server includes it.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -177,6 +177,11 @@
                     [robert/hooke "1.3.0"]
                     [slingshot]
                     [trptcolin/versioneer]
+                    ;; We do not currently use this dependency directly, but
+                    ;; we have documentation that shows how users can use it to
+                    ;; send their logs to logstash, so we include it in the jar.
+                    [net.logstash.logback/logstash-logback-encoder]
+
 
                     ;; Filesystem utilities
                     [org.apache.commons/commons-lang3 "3.4"]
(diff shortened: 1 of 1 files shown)


Sharpie:
We also noticed yesterday afternoon that this is broken in pe-puppetdb where the `net.logstash.logback/logstash-logback-encoder` library is present. The reason being that the encoder depends on the `jackson-databind` library that we blacklisted in PDB-4236.

austb:
This PR needs to re-add `jackson-databind`, a dependency we removed for security reasons. Since we have some updated versions available we should be able to do this but we will have to investigate that.

robdaemon:
I'd like to have an alternative here - jackson-databind is going to continue to have CVE's because of it's very nature. Is there a logstash appender that doesn't need databind?

Sharpie:
The only other JSON encoder for logback that immediately seems promising is from logback-contrb: https://github.com/qos-ch/logback-contrib/wiki/JSON

However, it also uses jackson-databind to serialize objects :(

robdaemon:
Well, I guess we'll be adding jackson-databind back. Boo.

PR #2851, opened 2019-03-12
(SERVER-2497) Extract `job_id` from request metadata

Magisus:
Previously, the `save` method of the catalog terminus would only save
`job_id` if it came in as part of the catalog. This commit ensures that
we also extract `job_id` from the extra request data if it is present,
and use that if `job_id` is not present in the catalog.


Magisus:
Let me know if there are any tests you think I should update for this.

Magisus:
I've tested this manually with my new catalog endpoint and seen that `job_id` is now correctly pulled out of the extra request data.

justinstoller:
This has to go out in Kearney, right? Is this targeted correctly for that?

austb:
Yes, we have not yet cut a `6.3.x` branch for Kearney so master is the correct place for it.

PR #2854, opened 2019-03-14
(SERVER-2497) Convert job_id to string

Magisus:
PuppetDB has a schema in place for the wire format of the catalog that
requires `job_id` to be a string. Since it comes from Puppet as an
integer, this commit converts it to a string in the terminus before
passing it on to PDB proper.


Magisus:
I tested this using `puppet agent --job-id 1234` and verified that the catalog could be saved without error.

Magisus:
Updated this to leave `job_id` as `nil` instead of converting `nil` to `""`.

PR #2858, opened 2019-03-19
(PDB-4312) handle duplicates in resource_events table

robdaemon:
The resource_events table can contain duplicate rows because the existing
unique constraint does not work as intended. In PostgreSQL, null values
are not considered equal in an index.

Migration now uses a set to dedupe the resource_events table as its being
written to, and the unique constraint is dropped in favor of the new
primary key on the resource_events table.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1517,18 +1517,33 @@
       containing_class text,
       corrective_change boolean)")
 
+  ;; null values are not considered equal in postgresql indexes,
+  ;; therefore the existing unique constraint did not function
+  ;; as intended. this will replace the unique constraint with
+  ;; a primary key that is a hash of the four columns used to
+  ;; consider uniqueness.
+
+  ;; since the unique constraint could allow duplicates to exist
+  ;; in the table, we have to keep a running set of the hashes
+  ;; we've encountered during the migration to avoid inserting
+  ;; duplicates into the new table.
+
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1517,18 +1517,33 @@
...
-       (jdbc/insert-multi! "resource_events_transform"
-                           (map #(-> %
-                                     (assoc :event_hash (->> (hash/resource-event-identity-pkey %)
-                                                             (sutils/munge-hash-for-storage))))
-                                batch)))))
+     (reduce (fn [hashes-seen row]
+               (conj hashes-seen
+                     (let [hash-str (hash/resource-event-identity-pkey row)]
+                       (when-not (hashes-seen hash-str)
+                         (jdbc/insert-multi! "resource_events_transform"
```
Zak-Kent: Does this mean we're not batching the inserts here anymore? Just want to make sure I know what it's doing.

robdaemon: no, we're not - but it happens in a single transaction, so it evens out.

robdaemon: Yeah, perf remained fairly consistent between the two. Ordinarily I would batch, but that definitely makes the migration code here more complex.

I'll keep `insert!` vs `insert-multi!` in my head for next time. I just left it as `insert-multi!` because it was there prior to this change.

PR #2860, opened 2019-03-21
(maint) If consul is enabled, run the http healthcheck every 10 seconds

underscorgan:
We should run this check every 10 seconds to decrease latency with consul knowing what services/hosts are available

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/40-consul.sh
+++ b/docker/puppetdb/docker-entrypoint.d/40-consul.sh
@@ -14,7 +14,7 @@ if [ "$CONSUL_ENABLED" = "true" ]; then
   "checks": [
     {
       "http": "http://$HOSTNAME:8080/status/v1/services/puppetdb-status",
-      "interval": "30s",
+      "interval": "10s",
       "deregister_critical_service_after": "10m"
     }
   ]


Zak-Kent:
Just a heads up, the osx failures in travis are known and expected.

rbrw:
This might well be fine (not sure I understand the context completely -- i.e. what kind of common delays it's trying to protect against), but if it doesn't need to be that frequent, I might be tempted to slow it a bit.  i.e. if say 5s or 10s would serve just as well.

robdaemon:
I'm with @rbrw on this - this healthcheck involves a database connection, so 5-10s is probably better.

docker/puppetdb/docker-entrypoint.d/40-consul.sh
```diff
@@ -14,7 +14,7 @@ if [ "$CONSUL_ENABLED" = "true" ]; then
   "checks": [
     {
       "http": "http://$HOSTNAME:8080/status/v1/services/puppetdb-status",
-      "interval": "30s",
+      "interval": "1s",
```
Iristyle: Looks like we might want to slow this down @underscorgan ?

underscorgan:
updated!

PR #2861, opened 2019-03-22
(maint) Disable Azure Docker builds on 6.x branch

Iristyle:
- When merging this up, do not carry the change forward to master
 - Note that changes had to be made in the Azure UI so as not override
   the values set in the YAML

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,6 +7,10 @@ pool:
 variables:
   NAMESPACE: puppet
 
+# completely disables the pipeline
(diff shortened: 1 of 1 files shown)


Iristyle:
Threw up a temp commit 85c72d9 that re-enables Azure to make sure that things still build.  They did, so I think we can ship 98a7b58 no problem.

Iristyle:
Intentionally skipped CI running b/c this commit only impacts Azure. Merging accordingly.

PR #2870, opened 2019-03-25
(PDB-4275) import/export configure expiration

robdaemon:
Adds export and import of configure-expiration commands via the nodes
queries

### DIFF
--- a/src/puppetlabs/puppetdb/export.clj
+++ b/src/puppetlabs/puppetdb/export.clj
@@ -11,6 +11,7 @@
             [puppetlabs.puppetdb.catalogs :as catalogs]
             [puppetlabs.puppetdb.factsets :as factsets]
             [puppetlabs.puppetdb.reports :as reports]
+            [puppetlabs.puppetdb.nodes :as nodes]
             [puppetlabs.puppetdb.command :as command]
             [puppetlabs.puppetdb.command.constants :as command-constants]
             [puppetlabs.puppetdb.utils :as utils]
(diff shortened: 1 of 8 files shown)


src/puppetlabs/puppetdb/export.clj
```diff
@@ -11,14 +11,16 @@
...
             [puppetlabs.puppetdb.command :as command]
             [puppetlabs.puppetdb.command.constants :as command-constants]
             [puppetlabs.puppetdb.utils :as utils]
             [clj-time.format :as time-fmt]
             [clj-time.coerce :as time-coerce]
             [puppetlabs.puppetdb.schema :as pls]
             [schema.core :as s]
-            [puppetlabs.i18n.core :refer [trs]]))
+            [puppetlabs.i18n.core :refer [trs]]
+            [flatland.ordered.map :refer [ordered-map]]))
```
austb: Doesn't look like this is used, and it threw a compilation error in jenkins

src/puppetlabs/puppetdb/nodes.clj
```diff
@@ -0,0 +1,51 @@
...
+   :latest_report_corrective_change (s/maybe s/Bool)
+   :latest_report_job_id (s/maybe s/Str)
+   :expires_facts (s/maybe s/Bool)
+   :expires_facts_updated (s/maybe pls/Timestamp)})
+
+(pls/defn-validated nodes-query->configure-expiration-wire-v1
+  :- [configure-expiration-wireformat-schema]
+  [nodes :- [nodes-wireformat-schema]]
+  (->> nodes
+       (filter #(false? (:expires_facts %)))
```
austb: `false?` is only false for the literal value `false`, since `expires_facts` is `nil` this will end up not filtering anything right?

austb:
```suggestion
       (filter :expires_facts_updated %))
```

PR #2872, opened 2019-03-27
(PDB-4315) ignore duplicates in resource_events

robdaemon:
Add a name column to the resource_events table that will be put to use
in a future PR. This column is needed to further guarantee uniqueness of
events.

Add more columns to the resource_events hash: timestamp, status,
old_value, new_value, message, file, line, name

Replace existing migration with a new migration that will rehash
the resource_events table again

Add an ON CONFLICT DO NOTHING to do the resource_events insert-multi to
ignore any duplicate hashes found. This can happen due to an agent bug
where the exec module can send duplicate events.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1513,6 +1513,7 @@
       message text,
       file text DEFAULT NULL::character varying,
       line integer,
+      name text,
       containment_path text[],
       containing_class text,
       corrective_change boolean)")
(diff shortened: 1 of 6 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1631,8 +1631,9 @@
    64 rededuplicate-facts
    65 varchar-columns-to-text
    66 jsonb-facts
-   67 add-resource-events-pk
-   68 support-fact-expiration-configuration})
+   ;; migration 69 replaces migration 67 - we do not need to apply both
```
rbrw: Not sure, but wonder if it'd be worth a little more detail here, perhaps mentioning why it's ok to ignore the previous one, i.e. because this one just rewrites the same table in a slightly different way, or...

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1318,12 +1318,13 @@
...
-                          (map add-event-hash)
-                          (jdbc/insert-multi! :resource_events)
-                          dorun))
+                     (let [inserter (fn [x] (jdbc/insert-multi! :resource_events x {:on-conflict "do nothing"}))]
+                       (->> resource_events
+                            (sp/transform [sp/ALL :containment_path] #(some-> % sutils/to-jdbc-varchar-array))
+                            (map adjust-event)
+                            (map add-event-hash)
+                            (inserter)
+                            dorun)))
```
rbrw: I might be tempted to call this `insert!` or something, rather than `inserter` and it's typical to leave off the parens in a threading form when there are no args, but neither is important.

PR #2878, opened 2019-03-28
(PDB-4321) escape asterisks in example regex

croddy:
This should escape the asterisks in the blacklist example regex so that they render properly on the docs site.


Zak-Kent:
Thanks @croddy! could you target this pr at the 5.2.x branch? Once merged I'll make sure it gets all the places needed for docs on the different versions.

croddy:
Rebased onto `5.2.x`.

Do I need to sign the CLA if I am a Puppet employee? @puppetcla

PR #2882, opened 2019-04-02
(PDB-4326) Allow unicode in resource tags

austb:
In PUP-7579, Puppet added support for Unicode letters in resource tags,
but the supported tags regex was never changed in PuppetDB so any
catalogs using unicode resource tags would not be stored in PuppetDB.

This changes the tag regex in PuppetDB to allow all lower case unicode
letters as tags.

### DIFF
--- a/src/puppetlabs/puppetdb/catalogs.clj
+++ b/src/puppetlabs/puppetdb/catalogs.clj
@@ -302,7 +302,11 @@
 ;; Functions to ensure that the catalog structure is coherent.
 
 (def ^:const tag-pattern
-  #"\A[a-z0-9_][a-z0-9_:\-.]*\Z")
+  ; tags are case insensitive, and Puppet automatically converts
+  ; tags to lower case, so we do not consider uppercase letters in
+  ; our valid tag regex.
+  ; See the doc for more info https://puppet.com/docs/puppet/6.4/lang_reserved.html#tags
+  #"[[\p{L}&&[^\p{Lu}]]\p{N}_][[\p{L}&&[^\p{Lu}]]\p{N}_:.-]*")
 
 (defn validate-resources
   "Ensure that all resource tags conform to the allowed tag pattern."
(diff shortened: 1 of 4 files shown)


src/puppetlabs/puppetdb/catalogs.clj
```diff
@@ -302,7 +302,7 @@
 ;; Functions to ensure that the catalog structure is coherent.
 
 (def ^:const tag-pattern
-  #"\A[a-z0-9_][a-z0-9_:\-.]*\Z")
+  #"\A[[\p{L}&&[^\p{Lu}]]0-9_][[\p{L}&&[^\p{Lu}]]0-9_:\-.]*\Z")
```
rbrw: What's the matching, presumably ruby pattern?  And might be good to include a comment or docstring describing what this has to match on the puppet side (better yet if there's a solid doc link).

rbrw: Oh, and didn't mean that we should include the current ruby pattern here, just wanted to look at it myself.

austb: Current ruby pattern is below (taken from here https://github.com/puppetlabs/puppet/blob/2bec43a16c472f0ef1a6f9b79f7eff929f6e95dd/lib/puppet/util/tagging.rb#L4)

`/\A[[:alnum:]_][[:alnum:]_:.-]*\Z/u`

austb: I need to change the way we match digits, Puppet now allows unicode digits as well as `0-9`

src/puppetlabs/puppetdb/catalogs.clj
```diff
@@ -302,7 +302,8 @@
 ;; Functions to ensure that the catalog structure is coherent.
 
 (def ^:const tag-pattern
-  #"\A[a-z0-9_][a-z0-9_:\-.]*\Z")
+  ; https://puppet.com/docs/puppet/6.4/lang_reserved.html#tags
+  #"\A[[\p{L}&&[^\p{Lu}]]\p{N}_][[\p{L}&&[^\p{Lu}]]\p{N}_:\-.]*\Z")
```
rbrw: Wonder if it would be helpful or harmful to add something like this to the comment:

```clojure
;; This regex is intended to match this ruby pattern which was specified
;; there: \A[[:alnum:]_][[:alnum:]_:.-]*\Z
```

So that we'll know what was originally intended, even if the puppet standard changes, or maybe that'd be better in the commit message, if anywhere.  Not sure.

austb: It may be harmful in the comment, if there's a change on the Ruby side, hopefully someone will follow the link to find an updated regex as opposed to checking the regex in the comment. But adding it in the commit message seems like a fine way to record the actual regex we are matching.

PR #2883, opened 2019-04-02
(maint) Make PDB wait on dependent services

Iristyle:
- Until some of the PDB code is more resilient to Postgres,
   Puppetserver and Consul starting later than it's prepared for,
   add a new script `20-wait-for-hosts.sh` that waits on specific
   hostnames.

   Executing ping against the host will both ensure that DNS resolution
   is working, and that the node is reachable within the network.

 - Code for wtfc.sh is written for /bin/sh and Alpine, and is imported
   from https://github.com/typekpb/wtfc.

   That code is MIT licensed

 - Refactors other health check waiting to the wait script

 - Emits logs on failure


docker/puppetdb/docker-entrypoint.d/30-configure-ssl.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+if [ ! -f "/etc/puppetlabs/puppet/ssl/certs/${HOSTNAME}.pem" ] && [ "$USE_PUPPETSERVER" = true ]; then
+  # if this is our first run, run puppet agent to get certs in place
```
underscorgan: [nitpick] I know this isn't related to your changes but this comment is inaccurate since we're using ssl.sh now

PR #2884, opened 2019-04-04
(maint) Fix unquoted Docker PDB health wait

Iristyle:
#2883 contained a bug which this PR addresses

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
+++ b/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
@@ -44,7 +44,7 @@ if [ "$CONSUL_ENABLED" = "true" ]; then
 fi
 
 if [ -n "$HEALTH_COMMAND" ]; then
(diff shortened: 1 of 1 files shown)


nwolfe:
Tests passed but there's this in the logs:
```
#<Thread:0x0000000003144eb8@C:/agent/_work/6/s/docker/puppetdb/spec/spec_helper.rb:9 run> terminated with exception (report_on_exception is true):
C:/agent/_work/6/s/docker/puppetdb/spec/spec_helper.rb:10:in `each': closed stream (IOError)
	from C:/agent/_work/6/s/docker/puppetdb/spec/spec_helper.rb:10:in `block (2 levels) in run_command'
```

Iristyle:
That crops up occasionally @nwolfe ... not sure why yet exactly.

The logs didn't show the actual fake command of `\"echo 'foo' | grep foo\"` being run either because I put it in the postgres container instead of the puppetdb container. Repushing!

PR #2885, opened 2019-04-04
(DO NOT MERGE) trash bad state on build server

Iristyle:
I don't have SSH access to the build server, so using this to run a couple quick commands


Iristyle:
My work here is done :)

PR #2886, opened 2019-04-04
(maint) Properly perform health checks

Iristyle:
- After some fumbling around with @underscorgan, we determined that the
   wtfc code doesn't properly handle command strings that include pipes.

   Even if situations where the command is successful, it will exit with
   a non-0 failing exit status.

   The fix for that is in https://github.com/puppetlabs/wtfc/pull/1

 - With this script updated in the container, nested quotes have to be
   handled differently as well (in the case where the Consul JSON is
   being grepped for "state":"running"). " must first become \", which
   is then escaped as \\"\"".

 - Simplify how quoting is done around hostnames, replacing " with '

 - This fixes code originally introduced in #2883 that was presumed
   fixed in #2884 (but wasn't actually fixed)

### DIFF
--- a/docker/puppetdb/Dockerfile
+++ b/docker/puppetdb/Dockerfile
@@ -72,7 +72,7 @@ RUN chmod +x /ssl-setup.sh
 
 VOLUME /etc/puppetlabs/puppet/ssl/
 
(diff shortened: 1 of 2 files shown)


underscorgan:
cat escaping from dog

Always be escaping

PR #2888, opened 2019-04-05
(maint) Rework tests for Linux Postgres container / Docker Compose

Iristyle:
- Wait on Postgres port to be listening

 - In Pupperware tests, waiting on the puppetserver to be listening
   introduces additional wait time that appears to allow Postgres to
   fully startup (preventing cluster start failures), which we don't
   get the benefit of here

 - This test suite isn't using puppetserver, so should instead rely on
   waiting for Postgres

 - Use docker-compose 1.24.0 in Travis

 - Prune volumes / networks in Azure cleanup step

 - Switch to using Linux Postgres container instead of unmaintained Windows Postgres container given our CI builders are now RS5

 - Randomize the location on disk for Postgres data (to be outside of git checked out area)

 - Refactor test failures modes in PuppetDB to catch all errors properly

 - Remove usage of `Timeout` class for retrying operations and replace with custom timer based solution that doesn't cause weird test suite failure modes

 - Refactor the suite to use `docker- compose` instead of `docker run` (adds additional code to spec_helper)

 - Remove unnecessary tests / move required startup checks to pre-suite `before(:all)`

 - Improve error handling inside `run_command` so that terminating threads reading from stdout / stderr don't tank the suite

 - Support ENV vars in `run_command`

 - Most of the spec_helper code is now generically refactored enough to be extracted to a shared testing library, which is the next step!

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
+++ b/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
@@ -4,9 +4,10 @@
 #
 #
 # Optional environment variables:
-#   PUPPETDB_WAITFORHOST_SECONDS    Number of seconds to wait for host, defaults to 30
-#   PUPPETDB_WAITFORHEALTH_SECONDS  Number of seconds to wait for health
-#                                   checks of Consul / Puppetserver to succeed, defaults to 600
+#   PUPPETDB_WAITFORHOST_SECONDS     Number of seconds to wait for host, defaults to 30
+#   PUPPETDB_WAITFORPOSTGRES_SECONDS Number of seconds to wait on Postgres, defaults to 90
+#   PUPPETDB_WAITFORHEALTH_SECONDS   Number of seconds to wait for health
+#                                    checks of Consul / Puppetserver to succeed, defaults to 600
 
 msg() {
     echo "($0) $1"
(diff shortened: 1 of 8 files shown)


docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
```diff
@@ -24,13 +24,21 @@ wait_for_host() {
...
+}
+
 PUPPETDB_WAITFORHOST_SECONDS=${PUPPETDB_WAITFORHOST_SECONDS:-30}
 PUPPETDB_WAITFORHEALTH_SECONDS=${PUPPETDB_WAITFORHEALTH_SECONDS:-600}
 PUPPETSERVER_HOSTNAME="${PUPPETSERVER_HOSTNAME:-puppet}"
 CONSUL_HOSTNAME="${CONSUL_HOSTNAME:-consul}"
 CONSUL_PORT="${CONSUL_PORT:-8500}"
 
 wait_for_host "postgres"
+wait_for_host_port "postgres" 5432
```
nwolfe: Do we still need the `wait_for_host "postgres"` above if we're doing this now?

I'm good with this, just wondering.

Iristyle: Probably not... but need to understand failure cases for `curl` with DNS resolutions and how that might be different than `ping`

Iristyle:
For `nc` usage - 3 scenarios

### success

```
nc -G 1 -w 1 -z 'puppet.delivery.puppetlabs.net' 8140
Connection to puppet.delivery.puppetlabs.net port 8140 [tcp/*] succeeded!

> echo $?
0
```

### fail (bad host name)

```
nc -G 1 -w 1 -z 'puppet' 8140
nc: getaddrinfo: nodename nor servname provided, or not known

> echo $?
1
```

### fail (port not open)

```
nc -G 1 -w 1 -z 'puppet.delivery.puppetlabs.net' 8141

> echo $?
1
```

PR #2903, opened 2019-04-12
(maint) PDB wait on Puppetserver port

jarretlavallee:
Prior to this PR, PDB waited for Puppetserver by pinging it. In some
environments, the puppserver may not be pingable even if it is online.
This commit changes the wait for puppetserver to start by checking that
the service is listening on port 8140.

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
+++ b/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
@@ -43,7 +43,7 @@ CONSUL_PORT="${CONSUL_PORT:-8500}"
 wait_for_host_port "postgres" "5432" $PUPPETDB_WAITFORPOSTGRES_SECONDS
 
 if [ "$USE_PUPPETSERVER" = true ]; then
-  wait_for_host $PUPPETSERVER_HOSTNAME
+  wait_for_host_port $PUPPETSERVER_HOSTNAME "8140"
   HEALTH_COMMAND="curl --silent --fail --insecure 'https://${PUPPETSERVER_HOSTNAME}:8140/status/v1/simple' | grep -q '^running$'"
 fi
 


jarretlavallee:
Output from a build:

~~~
$ kubectl logs puppetdb-b59bc9795-k9fqb
Running /docker-entrypoint.d/20-wait-for-hosts.sh
wtfc.sh: waiting 90 seconds for nc -v -w 1 -z 'postgres' 5432
wtfc.sh: nc -v -w 1 -z 'postgres' 5432 finished with expected status 0 after 0 seconds
wtfc.sh: waiting 30 seconds for nc -v -w 1 -z 'puppet' 8140
wtfc.sh: nc -v -w 1 -z 'puppet' 8140 finished with expected status 0 after 0 seconds
wtfc.sh: waiting 600 seconds for curl --silent --fail --insecure 'https://puppet:8140/status/v1/simple' | grep -q '^running$'
wtfc.sh: curl --silent --fail --insecure 'https://puppet:8140/status/v1/simple' | grep -q '^running$' finished with expected status 0 after 1 seconds
~~~

Iristyle:
Thanks for the PR and output @jarretlavallee ... since we don't fire up puppetserver in CI here, it would have otherwise required us to vet it. But you've saved us the trouble.

We might want to think about a few things before merging though (now that I'm revisiting this code):

* Previously we waited up to 30 seconds for just a dns resolve + ping response, but now we're waiting for a dns resolve + ping + java services to launch (but not necessarily respond), which will take longer. We were primarily interested in failing fast if the cluster never even DNS resolves the name / starts the service. (We can still extend the wait by setting the `PUPPETDB_WAITFORHOST_SECONDS` but I'm not sure that's the best approach)

* We already have a health check against the status endpoint running on 8140 (which takes even longer to respond) a little later in the file

* There's also a check against Consul which will run into the same problem resolving in k8s


So given all of that, I'm now thinking it might just make more sense to make sure `dig` or `nslookup` is in the container and rely on that to cover the DNS resolution part? I take it ICMP traffic is locked down in k8s which prevents the ping, correct? Are there other reachability checks we can use that are k8s friendly? (Until we get fully rolling with Consul?)

jarretlavallee:
@Iristyle Thanks for taking the time to review this and provide some great feedback. Those are very good points and this solution does not seem like the best way to approach the problem. The service in k8s does not respond to pings. Since the service is named the same as the pod, it will be the first `puppet` DNS record over the POD due to the DNS search path. 

I can solve this problem in the environment by using the `$PUPPETSERVER_HOSTNAME` or changing the DNS search path in the pod. For now, I will close this out as it is an incomplete solution.

PR #2920, opened 2019-04-26
(maint) Update the check when consul is enabled

underscorgan:
We were checking for an incorrect string to tell when puppetserver is up
and running, causing PDB to loop forever then fail.

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
+++ b/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
@@ -50,7 +50,7 @@ fi
 if [ "$CONSUL_ENABLED" = "true" ]; then
   wait_for_host $CONSUL_HOSTNAME
   # with Consul enabled, wait on Consul instead of Puppetserver
-  HEALTH_COMMAND="curl --silent --fail 'http://${CONSUL_HOSTNAME}:${CONSUL_PORT}/v1/health/checks/puppet' | grep -q '\\"\""state"\\\"":\\"\""running\\"\""'"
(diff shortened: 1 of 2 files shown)


docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
```diff
@@ -50,7 +50,7 @@ fi
 if [ "$CONSUL_ENABLED" = "true" ]; then
   wait_for_host $CONSUL_HOSTNAME
   # with Consul enabled, wait on Consul instead of Puppetserver
-  HEALTH_COMMAND="curl --silent --fail 'http://${CONSUL_HOSTNAME}:${CONSUL_PORT}/v1/health/checks/puppet' | grep -q '\\"\""state"\\\"":\\"\""running\\"\""'"
+  HEALTH_COMMAND="curl --silent --fail 'http://${CONSUL_HOSTNAME}:${CONSUL_PORT}/v1/health/checks/puppet' | grep -q '\\"\""Status"\\\"": \\"\""passing\\"\""'"
```
Iristyle: This is consistent with the docs at https://www.consul.io/api/health.html#sample-response-1

```json
[
  {
    "Node": "foobar",
    "CheckID": "service:redis",
    "Name": "Service 'redis' check",
    "Status": "passing",
    "Notes": "",
    "Output": "",
    "ServiceID": "redis",
    "ServiceName": "redis",
    "ServiceTags": ["primary"]
  }
]
```

Iristyle:
We should get some validation for this change in https://github.com/puppetlabs/pupperware/pull/62

Iristyle:
Travis is failing because the Docker specs are running on Ruby 2.4, but we need Ruby 2.5

underscorgan:
Got in a real weird travis state, reopening in #2921

PR #2922, opened 2019-04-30
(maint) PDB wait on Puppetserver port

jarretlavallee:
Prior to this PR, PDB waited for Puppetserver by pinging it. In some environments, the puppserver may not be pingable even if it is online. This PR changes the wait for puppetserver to check if either it is pingable or the service is listening on port 8140.

This should hopefully be a better implementation of #2903. The `nc` ensure the service is listening on the port but not the status of the service, so the status API query should not be fully redundant.

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
+++ b/docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
@@ -33,6 +33,14 @@ wait_for_host_port() {
   fi
 }
 
+wait_for_host_or_port() {
+  # Use either the ping or port connection to determine the connectivity of the host
+  /wtfc.sh --timeout=${3:-$PUPPETDB_WAITFORHOST_SECONDS} --interval=1 --progress "ping -c1 -W1 '${1}' || nc -v -w 1 -z '${1}' ${2}"
+  if [ $? -ne 0 ]; then
+    error "host cannot be resolved or contacted on $1:$2"
+  fi
+}
+
 PUPPETDB_WAITFORHOST_SECONDS=${PUPPETDB_WAITFORHOST_SECONDS:-30}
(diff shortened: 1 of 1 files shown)


jarretlavallee:
Here is the output from a test system. 

~~~
$ kubectl logs pod/puppetdb-7cfdc5cd4b-lb24k
Running /docker-entrypoint.d/20-wait-for-hosts.sh
wtfc.sh: waiting 150 seconds for nc -v -w 1 -z 'postgres' 5432
wtfc.sh: nc -v -w 1 -z 'postgres' 5432 finished with expected status 0 after 0 seconds
wtfc.sh: waiting 90 seconds for ping -c1 -W1 'puppet' || nc -v -w 1 -z 'puppet' 8140
wtfc.sh: ping -c1 -W1 'puppet' || nc -v -w 1 -z 'puppet' 8140 finished with expected status 0 after 1 seconds
~~~

docker/puppetdb/docker-entrypoint.d/20-wait-for-hosts.sh
```diff
@@ -33,6 +33,14 @@ wait_for_host_port() {
   fi
 }
 
+wait_for_host_or_port() {
```
Iristyle: We talked yesterday about improving the DNS query bit... I think we might end up installing `dnsutils` package so that we have `dig` available. Then we can get better error messages when it's a config / DNS problem.

Will try to get some variation of this PR in ASAP for you though - thanks for the submission!

jarretlavallee: That sounds like a great idea. Thanks for looking into this!

Iristyle: I have #2935 up which hopefully does the trick for you!

jarretlavallee:
Closing in favor of #2935

PR #2926, opened 2019-04-30
(maint) Refactor PDB Docker HEALTHCHECK

Iristyle:
- The previous format can confuse LCOW and result in errors like:

   CreateProcess: failure in a Windows system call: Unspecified error (0x80004005)
   [Event Detail: failed to run runc create/exec call for container

 - Puppetserver is using this style with success

### DIFF
--- a/docker/puppetdb/healthcheck.sh
+++ b/docker/puppetdb/healthcheck.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env sh
+
+set -x
+set -e
+
+curl --fail \
+--resolve "puppetdb:8080:127.0.0.1" \
+"http://puppetdb:8080/status/v1/services/puppetdb-status" \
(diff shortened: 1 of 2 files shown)


docker/puppetdb/healthcheck.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
```
Iristyle: `puppetserver` uses `bash` here ... but this container doesn't have `bash`

docker/puppetdb/healthcheck.sh
```diff
@@ -0,0 +1,10 @@
...
+
+set -x
+set -e
+
+curl --fail \
```
Iristyle: puppetserver doesn't use `--silent` so I removed it

PR #2928, opened 2019-05-01
(maint) Make Docker Postgres configurable

Iristyle:
- Prior to this commit, the wait script would wait on the host named
   'postgres' with a fixed port of '5432', which is quite inflexible.

   In testing the Pupperware stack, it's necessary to change the host
   to `postgres.local` to align with other hosts.

   This will allow that configuration change to be made more easily
   at runtime


Iristyle:
Spun up the `pupperware` cluster with this change and a `docker-compose.yml` that sets `PUPPETDB_POSTGRES_HOSTNAME=postgres.local` - pdb queries the right host / logs properly

```
puppetdb_1  | ......................wtfc.sh: nc -v -w 1 -z 'postgres.local' 5432 finished with expected status 0 after 22 seconds
```

The container reaches healthy and the it properly builds the connection string to connect to Postgres.

PR #2931, opened 2019-05-02
(PDB-4308) Notice a shutdown during pdb initialization

rbrw:


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -332,31 +332,53 @@
...
+                         (initialize-schema {:datasource db-pool} config)
+                         true
+                         (catch java.sql.SQLTransientConnectionException ex
+                           ;; When coupled with the 3000ms timout, this
+                           ;; is intended to log a duplicate message no
+                           ;; faster than once per minute.
+                           (when (or (not= (str ex) (str last-ex))
+                                     (zero? (mod i 20)))
+                             (log/error (str ex))
+                             (log/error (trs "Database initialization connection has temporarily failed (retrying)")))
```
robdaemon: can we combine this with the line above so we only create one log entry, like:

```
(log/error ex  (trs "Database initialization connection has temporarily failed (retrying)"))
```

`log/error` will take a throwable as the first parameter in one of its overloaded forms.

rbrw: Ahh, right, perhaps should have included a comment, but I intentionally didn't use that form because I wanted to avoid the backtrace in the log for this case, where we expect the error.  I felt like this was a lot easier to follow, and looked a lot nicer in the log and/or on *err*.  If you want to see what you think, you can trigger it by just changing the password in your pdb.ini and running "services -c pdb.ini" -- it'll immediately hit this loop.

robdaemon: Oh, okay... well, how about combining them like:

```
(log/error (str (trs "Database initialization connection has temporarily failed (retrying)") ex))
```

so they appear on the same line? I just want to avoid someone getting confused and thinking there were two separate issues going on.

rbrw: No, they're on separate lines, and perhaps you're right that semantically they should be one event, but I did that so that they might read a bit better, i.e. you see the error, then you see the message notifying about the pending retry (prompted by the error).

Here's how it looks as it is:

```
2019-05-07 13:03:45,960 ERROR [async-dispatch-2] [p.p.c.services] java.sql.SQLTransientConnectionException: PDBMigrationsPool - Connection is not available, request timed out after 3004ms.
2019-05-07 13:03:45,961 ERROR [async-dispatch-2] [p.p.c.services] Database initialization connection has temporarily failed (retrying)
2019-05-07 13:03:48,962 ERROR [async-dispatch-2] [p.p.c.services] java.sql.SQLTransientConnectionException: PDBMigrationsPool - Connection is not available, request timed out after 3000ms.
2019-05-07 13:03:48,963 ERROR [async-dispatch-2] [p.p.c.services] Database initialization connection has temporarily failed (retrying)
2019-05-07 13:04:42,974 ERROR [async-dispatch-2] [p.p.c.services] java.sql.SQLTransientConnectionException: PDBMigrationsPool - Connection is not available, request timed out after 3000ms.
2019-05-07 13:04:42,975 ERROR [async-dispatch-2] [p.p.c.services] Database initialization connection has temporarily failed (retrying)
...
```

Or course we can change it, but if so, I'll need to figure out some way to put them in the same message that's still fairly clear -- while assuming (I think?) the translated retry bit should be before the exception's string on the line, since we don't know what its content iis going to be.

rbrw: Replacing with:

```
2019-05-07 13:03:45,961 ERROR [async-dispatch-2] [p.p.c.services] Will retry connection to database after temporary failure: java.sql.SQLTransientConnectionException: PDBMigrationsPool - Connection is not available, request timed out after 3004ms.
```

PR #2938, opened 2019-05-03
(DOCS) edits to repair once/since issues

jbondpdx:
Hey folks, I'm just here with some friendly docs pothole repairs! This PR fixes instances of once and since in the puppetdb/documentation directory. Both of these words have several different meanings, making them ambiguous for non-native English speakers, so I've replaced them with less ambiguous words---mostly "because," "after," and in a few cases "when."

I've also fixed a few nearby instances of future tense and lowercased "Puppet".


rbrw:
This looks just fine, but I think we'd like to put it on the 5.2.x branch instead, and then we'll merge it up[1].  If you're comfortable with making that change, great -- otherwise we can help.  I suspect that'll require some adjustment to the release notes diff, but the rest may rebase fairly cleanly.

[1] So we'll have it on all our active branches, and so we'll minimize future merge conflicts between them.

Thanks

jbondpdx:
Eh, I should have just had you all do it, but I thought I would poke at Git a bit. I may need to close this commit and begin again.

PR #2967, opened 2019-05-22
(PDB-4356) switch to sha1 from md5

robdaemon:
Replace use of MD5 in aggregate functions with SHA1. FIPS discourages
the use of MD5, and this aggregate is used in queries (not stored to the
database) in PE.

### DIFF
--- a/test/puppetlabs/puppetdb/scf/migrate_test.clj
+++ b/test/puppetlabs/puppetdb/scf/migrate_test.clj
@@ -506,19 +506,19 @@
                         #"Found an old and unuspported database migration.*"
                         (migrate! *db*))))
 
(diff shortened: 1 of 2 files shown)


test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -576,22 +576,6 @@
                         #"Found an old and unuspported database migration.*"
                         (migrate! *db*))))
 
-(deftest md5-agg-test
```
robdaemon: this was a dupe - this test was in this file twice. I was getting confused running the tests.... "how could this fail? I swear I just fixed that one...."

PR #2969, opened 2019-05-24
(maint) Use explicit hostname/network for puppetdb tests

underscorgan:


docker/docker-compose.yml
```diff
@@ -1,7 +1,8 @@
-version: '3'
+version: '3.5'
 
 services:
   postgres:
+    hostname: postgres.test
```
Iristyle: FWIW - I *think* we're supposed to set this like:

```
hostname: postgres
domain: test
```

https://docs.docker.com/compose/compose-file/#domainname-hostname-ipc-mac_address-privileged-read_only-shm_size-stdin_open-tty-user-working_dir

Sidebar - I think we ran into problems trying to set hostnames in k8s that included `.`

underscorgan: ahhh, ok. will give that a shot, thanks!

PR #2970, opened 2019-05-28
(maint): add "wrtrepo ref-lands-where REF" utility

rbrw:
Given a ref, report if that ref isn't included in the most recent
version of any release branch (any *.x branch) that it's on.

If nothing else, this might make computing fixversions easier, e.g.:

  $ ext/bin/wrtrepo ref-lands-where HASH
  New in release after 5.2.8 of refs/remotes/origin
  New in release after 6.0.3 of refs/remotes/origin
  New in release after 6.3.2 of refs/remotes/origin

  $ ext/bin/wrtrepo ref-lands-where HASH
  Appears to have been released

Note that the tool doesn't try to guess about what the next release on
a branch will be, since it could be an X, Y, or Z.

### DIFF
--- a/ext/bin/test-config
+++ b/ext/bin/test-config
@@ -96,7 +96,7 @@ reset_val = defaultdict(lambda: reset_requested)
 reset_val['puppet-ref'] = reset_puppet_ref
 reset_val['puppetserver-ref'] = reset_puppetserver_ref
 
-if sys.argv == ['--help']:
+if sys.argv[1:] == ['--help']:
     print(usage())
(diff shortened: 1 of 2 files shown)


rbrw:
This is another tool I've been using that might or might not be interesting.  Thought I'd post it for discussion.  And if we do want it, I'm not really sure whether it should be a standalone command, or the subcommand-style program it is right now, but I do have some doc-branch management related work that's not ready to consider yet, and that's currently couched as additional subcommands.

PR #2978, opened 2019-05-31
(PDB-4347) Make fact path gc table temp, not unlogged

rbrw:
Temp tables *are* unlogged tables[1], not vice versa (as had been
assumed), so change the gc table to a temp table.

Since unlogged tables aren't temp tables, you could see a table name
conflict if, for example, multiple gc attempts ran simultaneously:

  WITH recursive live_paths(path, value) AS
  ...
  SELECT path
  INTO   unlogged tmp_live_paths
  FROM   live_paths

  was aborted: ERROR: relation "tmp_live_paths" already exists
  ...

[1] https://rhaas.blogspot.com/2010/05/global-temporary-and-unlogged-tables.html


austb:
That `default` travis job likely failed because the job was reaped while it was running, let's see how `PR Tests` does

PR #2980, opened 2019-06-04
(PDB-4390) Store the uncoerced catalog certname

austb:
If a certname contain invalid characters for the queue metadata, such as
an underscore, or if it is too long it will be coerced and a hash of the
original will be appended to the end. For replace-catalogs, this
certname would end up getting stored in the database (only after PDB-4257).

Now the certname is correctly read out of the command payload (the
uncoerced value) and stored properly.

### DIFF
--- a/test/puppetlabs/puppetdb/command_test.clj
+++ b/test/puppetlabs/puppetdb/command_test.clj
@@ -482,21 +482,33 @@
                    (query-to-vec (format "SELECT certname, %s as catalog FROM catalogs"
                                          (sutils/sql-hash-as-str "hash")))))
             (is (= 0 (task-count delay-pool)))
-            (is (empty? (fs/list-dir (:path dlo)))))
+            (is (empty? (fs/list-dir (:path dlo))))))
 
-          (testing "should store the catalog if the node was deactivated after the message"
+        (testing "should store the catalog if the node was deactivated after the message"
+          (with-message-handler {:keys [handle-message dlo delay-pool q]}
+
+            (scf-store/delete-certname! certname)
+            (jdbc/insert! :certnames {:certname certname :deactivated tomorrow})
+
+            (handle-message (queue/store-command q (make-cmd-req)))
+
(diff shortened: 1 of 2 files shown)


austb:
I found a test that was improperly nested inside another, so I'd recommend reviewing my changes ignoring whitespace

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -482,21 +482,30 @@
...
+            (is (= [{:certname certname :catalog catalog-hash}]
+                   (query-to-vec (format "SELECT certname, %s as catalog FROM catalogs"
+                                         (sutils/sql-hash-as-str "hash")))))
+            (is (= 0 (task-count delay-pool)))
+            (is (empty? (fs/list-dir (:path dlo))))))
+
+        (testing "should store the proper certname for catalogs with underscores"
+          (let [certname "host_0"]
+            (with-message-handler {:keys [handle-message dlo delay-pool q]}
+              (handle-message (queue/store-command q (catalog->command-req version-num (assoc catalog :certname certname))))
```
rbrw: This is nested enough that I think `->>` might be a bit easier to follow?
```
(->> (assoc catalog :certname certname)
     (catalog->command-req version-num)
     (queue/store-command q)
     handle-message)
```
Possibly the same in the replace-facts test...

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -813,6 +822,21 @@
           (let [result (query-to-vec "SELECT certname,environment_id FROM factsets")]
             (is (= result [(with-env {:certname certname})])))))))
 
+  (deftest replace-facts-nonstandard-certname
```
rbrw: Not sure, but thought it might make sense to consider using "unusual" rather than "nonstandard" just to match our usage elsewhere (in the recent commits fixing other issues related to this feature, etc.).

PR #2981, opened 2019-06-06
(PDB-4385) update lein-ezbake version for FIPS

robdaemon:
Update lein-ezbake to 2.0.3 to support FIPS


austb:
We should rebase this and get some more Travis tests to pass before merging

rbrw:
Wondered if the latest docker test failure was transient, so I restarted it (though I suspect it might be something ruby-related wrt the xenial upgrade).

PR #2982, opened 2019-06-06
(maint) travis: switch to xenial and adoptopenjdk

rbrw:


rbrw:
Might want to look at the individual commits -- and for at least "ext/bin/flavor-for-spec: add and use in prep-os-essentials-for", might want "ignore whitespace".  I did that one separately to get things in the right places to (hopefully) make the subsequent commits easier to follow.

rbrw:
Oh, and let's keep the merge commit for this one when we merge it.

ext/bin/config-puppet-test-ref
```diff
@@ -35,9 +35,9 @@ case "$#" in
         exit 2
 esac
 
-bundle install --without acceptance --path vendor/bundle
+bundle install --without acceptance --path vendor/bundle --retry=10
 bundle update
-bundle info puppet
+bundle info puppet || true  # Not all versions appear to have info
```
austb: I think this may be an artifact of Puppet not being installed properly. Can we give this a try without `|| true` now that it is working?

PR #2986, opened 2019-06-10
(PDB-4360) add IPv6 support to pdbbox

robdaemon:
Add IPv6 support to pdbbox. Allows a pdbbox sandbox to initialize with
IPv6 support, IPv4 support, or both. IPv4 defaults to on, IPv6 defaults
to off.

### DIFF
--- a/ext/bin/pdbbox-env
+++ b/ext/bin/pdbbox-env
@@ -10,9 +10,10 @@ if test "$#" -lt 1 -o -z "$PDBBOX"; then
 fi
 
 pgport="$(cat "$PDBBOX/pg/port")"
+pghost="$(cat "$PDBBOX/pg/bind-addrs" | tail -1)"
 
 export PGBOX="$PDBBOX/pg"
-export PDB_TEST_DB_HOST=localhost
+export PDB_TEST_DB_HOST="$pghost"
 export PDB_TEST_DB_PORT="$pgport"
 export PDB_TEST_DB_USER=pdb_test
 export PDB_TEST_DB_USER_PASSWORD="$(cat "$PDBBOX/test-pass")"
--- a/ext/bin/pdbbox-env
+++ b/ext/bin/pdbbox-env
@@ -10,7 +10,7 @@ if test "$#" -lt 1 -o -z "$PDBBOX"; then
 fi
 
 pgport="$(cat "$PDBBOX/pg/port")"
-pghost="$(cat "$PDBBOX/pg/bind-addrs" | tail -1)"
+pghost="$(tail -1 "$PDBBOX/pg/bind-addrs")"
 
 export PGBOX="$PDBBOX/pg"
 export PDB_TEST_DB_HOST="$pghost"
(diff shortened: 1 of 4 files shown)


robdaemon:
Depends on https://gitlab.com/pgbox-org/pgbox/merge_requests/1

ext/bin/pdbbox-init
```diff
@@ -36,20 +37,27 @@ while test $# -gt 0; do
             pgport="$1"
             shift
             ;;
+        --bind-addr)
+            shift
+            test ! -z $# || misuse
```
rbrw: Perhaps `test $# -gt 1`?

ext/bin/pdbbox-init
```diff
@@ -36,20 +37,27 @@ while test $# -gt 0; do
...
 test "$pgport" || misuse
+test "$bind_addr" || misuse
 
 export PGBOX="$pdbbox/pg"
 export PGUSER=postgres
 
 mkdir "$pdbbox"
 
-pgbox init --pgbin "$pgbin" --port "$pgport" -- -E UTF8 --locale=C
+pgbox init --pgbin "$pgbin" --port "$pgport" --bind-addr "${bind_addr[@]/#/}" -- -E UTF8 --locale=C
```
rbrw: Hmm, interesting, what does the `/#/` do?  And I suspect this might actually need to build all the args in an array or something, i.e. with alternating `--bind-addr` and ADDR values, so it'll handle multiple addresses.  

Either way, might also be worth changing at least one of the pgbox-init calls in our tests to specify 2 --bind-addr values, so we'll implicitly test that case.  I suppose it might be fine to just specify `localhost` twice, or if pg won't care if the second address is actually available everywhere we test (it may not -- iirc it may just warn), suppose we might just add ip6-localhost in one of our invocations...

ext/bin/pdbbox-env
```diff
@@ -10,9 +10,10 @@ if test "$#" -lt 1 -o -z "$PDBBOX"; then
 fi
 
 pgport="$(cat "$PDBBOX/pg/port")"
+pghost="$(cat "$PDBBOX/pg/bind-addrs" | tail -1)"
```
rbrw: Perhaps just `$(tail -1 "$PDBBOX/pg/bind-addrs")`

ext/bin/pdbbox-init
robdaemon: This builds it into: `--bind-addr localhost ip6-localhost`.

ext/bin/pdbbox-init
robdaemon: the bash `-z` seems to be the canonical way to do this.

ext/bin/pdbbox-init
robdaemon: At this point, I'm going to defer to you on this PR and feature entirely. Make any changes you feel necessary to implement it your way. I've reworked this enough.

PR #2990, opened 2019-06-11
(maint) require-jdk: unpack in empty tmpdir

rbrw:


ext/bin/require-jdk
```diff
@@ -77,24 +77,28 @@ esac
...
          case "$ver" in
-             openjdk8) mv */Contents/Home jdk ;;
-             *) mv * jdk ;;
+             openjdk10) mv * jdk ;;
+             *) mv */Contents/Home jdk ;;  # At least 8 and 11
          esac)
         ;;
     *)
-        (cd "$install"
+        (cd "$tmpdir/tmp-unpack"
```
robdaemon: Is this a duplicate of line 80-83?

rbrw: I don't think so?  I'm trying not to change the current dir globally, so I cd in a subshell each time. i.e. the case statement here just handles the platform archive specific naming issues.

PR #3001, opened 2019-06-13
PDB-4398 bulk configure expiration cmd

Zak-Kent:


src/puppetlabs/puppetdb/client.clj
```diff
@@ -28,21 +28,21 @@
...
     command :- s/Str
     version :- s/Int
     payload :- {s/Any s/Any}
     timeout]
    (let [body (json/generate-string payload)
          url-params (str
-                     (format "?command=%s&version=%s&certname=%s"
-                             (str/replace command #" " "_") version certname)
+                     (format (str "?command=%s&version=%s" (when certname "&certname=%s"))
+                             (str/replace command #" " "_") version (when certname certname))
```
austb: I think we may be able to cut down the moving pieces w.r.t handling the number of arguments to `format` if we move certname to its own format and let the outside-most `str` call stringify whatever is there, similarly to `producer_timestamp` below.

```
(str
  (format (str "?command=%s&version=%s") (str/replace command #" " "_") version)
  (when certname
    (format "&certname=%s" certname))
  (when-let [producer_timestamp (-> payload :producer_timestamp str)]
    (format "&producer-timestamp=%s" producer_timestamp)))
```

Zak-Kent: yep, that's cleaner for sure. updated

Zak-Kent:
Closing because this has been open for a while and no one has requested the feature. This was close to being finished, iirc. I'll hang on to the branch and we can reopen if there are requests for bulk config expiration commands.

PR #3002, opened 2019-06-14
(docs) 6.3.3 Release notes

hestonhoffman:


rbrw:
Did you intend for this to remove such a large section of the file?  And if so, and if the intention is to remove older release notes, I'd be inclined to at least keep them for all of the supported branches, which I believe still includes 5.2.x.

Offhand, I might also be inclined to keep them in the repository by just moving them to another file -- i.e. release_notes_1.md or release_notes_upto_5.2.x.md or...  I'm thinking of precedents like Emacs NEWS (https://git.savannah.gnu.org/cgit/emacs.git/tree/etc), libc's Changelog.old/ (https://sourceware.org/git/?p=glibc.git;a=tree), etc. -- so that the older notes are still relatively easy to find, link-to (even if just via git) and read, if/when needed.

hestonhoffman:
Hi @rbrw, yes this change is intentional. 

Here is the issue I'm solving by removing the old release notes. Users visiting the [release notes page](https://puppet.com/docs/puppetdb/6.3/release_notes.html) for the latest version of DB have to scroll through a massive TOC to get to the release notes. It's a poor user experience. 

The change also aids consistency across the documentation site. If a user is looking for release notes for other supported versions of DB, they can use the product switcher to select the version they're looking for. This is how the release notes work for all Puppet products. 

If you'd like to keep a historical log of all the changes made to DB in the repo, then I agree that a Changelog is a good solution.

rbrw:
OK, hmm (speculating) what if we just moved all the release notes to release_notes.X or release_notes_X.Y.md files (depending on what granularity you wanted).  Then (unless we wanted to handle it some other way) we might just symlink release_notes.markdown to the correct version in each branch?  Not sure it's sensible yet, just came to mind...

rbrw:
(And of course my assumption there is that the web pages would just pick up the right changes without any extra effort on that side.)

PR #3009, opened 2019-06-20
(PDB-4266) Add include_package_inventory flag to factsets query

austb:
This adds support for the include_package_inventory query option, which
when provided will re-create the original wire format of
package_inventory.

This is intended to support proper syncing of factsets in PE.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -1009,56 +1009,81 @@
                :subquery? false
                :source-table packages"}))
 
+(def factsets-query-base
+  {:projections
+   {"timestamp" {:type :timestamp
+                 :queryable? true
+                 :field :timestamp}
+    "facts" {:type :queryable-json
+             :queryable? true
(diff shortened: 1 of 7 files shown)


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1009,56 +1009,81 @@
...
+  (map->Query factsets-query-base))
+
+(def factsets-with-packages-query
+  "Query for factsets with the package_inventory reconstructed (used for sync)"
+  (let [original-joins (get-in factsets-query-base [:selection :left-join])
+        new-joins (conj original-joins
+                        :certnames
+                        [:= :certnames.certname :fs.certname]
+                        [{:select [[:certname_packages.certname_id :certname_id]
+                                   [:%array_agg.p.triple :packages]]
```
austb: This array aggregate function is producing an `[]` even when there are no packages submitted. The key needs to be omitted entirely if there are no entries in the package inventory tables for this node.

rbrw:
Looks good.

PR #3017, opened 2019-07-02
(PDB-4373) redhat: kill old pdb in preinst given 6.3 main ns change

rbrw:
pdb's main namespace was changed in 6.3, during the work to compile
the uberjar, and we didn't realize that ezbake captures and depends on
it[1].

That alone might not be an issue, but an acceptance test failure
suggests that our rpms are using the new version's namespace[1] when
trying to kill the old server during an upgrade (perhaps via the new
version's cli/stop script).

From looking at the rpm --scripts for the current 5.2 rpm, it appears
that we don't stop and then later start pdb during an upgrade, we only
restart it, and we only do that once, after the new package is
installed, from the old package's "%postun" (postuninstall) script --
step 11 here:

  https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#ordering

At that point, assuming the cli/stop scripts, etc. are those from the
new package, then the pgrep will be looking for the wrong namespace to
kill and the old server won't be stopped.

To fix that, hack the pdb rpm preinst (presumably run at step 2 here)
to always try to kill the old server (i.e. via the old namespace).

[1] ezbake currently embeds the main namespace in various packaging
    scripts, and uses pgrep to find the server to kill, searching in
    part for that namespace, which originates in project.clj's
    [:lein-ezbake :main-namespace]).


rbrw:
Currently pending on the outcome of https://github.com/puppetlabs/ezbake/pull/541

rbrw:
I've manually tested this, and it works, but not always.  There appears to be some kind of `service start` race when upgrading from 5.2 -> 6.3 (with the problem) -> 6.5 SNAPSHOT (with the fix) that we'll need to resolve.  If I put a `sleep 1` where indicated, the upgrade works, and puppetdb is restarted.  Without the sleep, you're sometimes left with no puppetdb running because the start does nothing, but also doesn't produce any error output (haven't checked the exit status yet, though arguably we should be).

When this works, it also appears to be restarting twice during the entire upgrade process, though that may not matter if it's not easy to fix, since the workaround's start should only be triggered once for any given host.

If possible, I'm hoping someone with more ezbake and/or rpm install/update process knowledge can give these changes a closer look to see if they seem mostly plausible, and help me figure out what's causing the race.  Among other things I don't know rpm package script behaviors nearly as well as debs.  I'm also planning look more closely at /sbin/service and our related scripts on Monday.

resources/ext/ezbake.conf
```diff
@@ -2,11 +2,36 @@ ezbake: {
...
+                  "         if kill -0 \"$pid\" 2>/dev/null; then",
+                  "             maybe_dead=1",
+                  "             break",
+                  "         fi",
+                  "         sleep 0.2",
+                  "     done",
+                  "     if test -z \"$maybe_dead\"; then",
+                  "         echo 'Unable to kill puppetdb server cleanly; sending KILL' 1>&2",
+                  "         kill -9 \"$pid\"",
+                  "     fi",
```
rbrw: A `sleep 1` right here, before the `service start`, appears to avoid some kind of race that prevents the start from working.

PR #3022, opened 2019-07-09
(PDB-4370) catalog inputs

Zak-Kent:


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -793,6 +824,9 @@
     received-timestamp :- pls/Timestamp]
    (time! (:replace-catalog performance-metrics)
           (jdbc/with-db-transaction []
+            ;; TODO: figure out if we want/need to include catalog inputs in the similarity hash for catalogs
+            ;; this could cause issues similar to those with package inventory wrt sync
+            ;; it might be better to have catalog inputs sync as a diff command with each catalog entity in sync
```
austb: Yes we probably should not hash the catalog inputs into the catalog hash because that'd mean we'd need to query/reconstruct the inputs on every catalog submission that didn't include them.

This does present sync difficulties that may result in needing a new `"replace catalog-inputs"` command type

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -804,6 +838,7 @@
                         (.after (to-timestamp producer_timestamp)))
                 (log/warn (trs "Not replacing catalog for certname {0} because local data is newer." certname))
 
+                ;; TODO: adjust this check to allow a catalog change to only catalog inputs to trigger an update for inputs
```
austb: To keep in line with how we store package_inventory, we should likely move this up to the same function that calls `replace_catalog!` (from command.clj I believe?) and add something like

```
(when (seq (:inputs catalog))
  (replace-catalog-inputs! (:inputs catalog)))
```

austb:
Actually, now that I've had some time to think about this. I think it might be best if the catalog inputs are stored via their own command. We can possibly coordinate that via the terminus so the world does not need to know we added a new command/endpoint. The over-arching reason behind this is sync, but I think it's also the "correct" way to handle this set of data wrt to the PuppetDB model.

I think this is different than `package_inventory` coming in with `factset` because at its core `pacakge_inventory` was a "fact". It was the custom fact [`_puppet_inventory_1`](https://github.com/puppetlabs/puppet-enterprise-modules/blob/9cf6aa3af09f4eea882f14a8cb328fbf075742e9/modules/puppet_enterprise/lib/facter/inventory.rb), so it is submitted every time and we just munged it in order to optimize its storage/querying. I will acknowledge this argument has a slight issue, "if it's a fact, why doesn't it show up in /fact queries?" To which I'd say 🤷‍♂, but probably because it never showed up there, had a different use-case in mind, and it'd be expensive to recompute for every fact query.

At the core of this feature, I would argue, a catalog's inputs are a part of **_a catalog_**, but not _**the catalog**_ catalog that we store. It's really metadata about a single catalog, that catalog being the last one the inputs were calculated on, not necessarily the catalog we store. Therefore it seems in our best interest to separate it from catalogs, otherwise the hash generation, summary query, and syncing of catalogs will be a nightmare because we are attempting to combine something that is submitted every 30 minutes with something that might only get submitted every 30 days.

By separating it out entirely, we simplify sync, don't risk breaking anything w.r.t storing, querying, or syncing catalogs and allow ourselves to make use of the existing sync strategies instead of adding on some sort of special-cased double query madness to sync a catalog and its inputs in one go.

austb:
Merged in #3028

PR #3025, opened 2019-07-17
Test merge 6.3.x

austb:
[facebones ignore]


austb:
Merge conflicts in the following files

```
src/puppetlabs/puppetdb/http/query.clj 
src/puppetlabs/puppetdb/query_eng.clj 
```

They could probably use a review before merging.

PR #3028, opened 2019-07-18
(PDB-4372) Add "replace catalog inputs" command

austb:
Outstanding work

- [x] Garbage collection
- [x] Tests
- [x] Docs

### DIFF
--- a/documentation/api/wire_format/replace_catalog_inputs_format_v1.markdown
+++ b/documentation/api/wire_format/replace_catalog_inputs_format_v1.markdown
@@ -1,0 +1,73 @@
+---
+title: "PuppetDB: Replace catalog inputs wire format, version 1 (experimental)"
+layout: default
+canonical: "/puppetdb/latest/api/wire_format/replace_catalog_inputs_v1.html"
+---
+
+The v1 `replace catalog inputs` command tells PuppetDB to replace the current
+set of catalog inputs with the new.
+
+Changing the value of this setting for a certname has the same effect
+as receipt of a new catalog with respect to extending the lifetime of
+the node.
+
+Replace catalog inputs command format
+-----
+
+### Version
+
+This is **version 1** of the replace catalog inputs command.
+
+### Encoding
+
+The command is serialized as JSON, which requires strict UTF-8 encoding.
+
+### Main data type: Replace catalog inputs for node
+
+     {
+      "certname": <string>,
+      "producer_timestamp": <datetime>,
+      "catalog_uuid": <string>,
+      "inputs": [[<string>, <string>], ...],
+     }
+
+#### `certname`
+
+String. The name of the node the catalog was compiled for.
+
+#### `producer_timestamp`
+
+DateTime. The time of command submission from the Puppet master to PuppetDB,
+according to the clock on the Puppet master.
+
+`producer_timestamp` is optional but *highly* recommended. When provided, it is
+used to determine the precedence between this command and other commands that
+modify the same node. This field is provided by, and should thus reflect the
+clock of, the Puppet master.
+
+When `producer_timestamp` is not provided, the PuppetDB server's local time is
+used. If another command is received for a node while a non-timestamped
+"deactivate node" command is pending processing, the results are *undefined*.
+
+#### `catalog_uuid`
+
+The uuid of the catalog this input was used for. Since not every catalog will
+provide a record of its inputs, and the catalog is replaced via a different
+command, this may not be the same as the current catalog stored in PuppetDB.
+
+#### `inputs`
+
+The `inputs` to a catalog are a list of tuples, where the first element is its
+`type` (ie. `"hiera"`) and the second is a unique name for the input (for a hiera
+key it would just be its full key, `"puppetdb::globals::version"`.
+
+### Data type: `<string>`
+
+A JSON string. Because the command is UTF-8, these must also be UTF-8.
+
+### Data type: `<datetime>`
+
+A JSON string representing a date and time (with time zone), formatted based on
+the recommendations in ISO 8601. For example, for a UTC time, the string would be
+formatted as `YYYY-MM-DDThh:mm:ss.sssZ`. For non-UTC time, the `Z` may be replaced
+with `±hh:mm` to represent the specific timezone.
--- a/puppet/lib/puppet/indirector/catalog/puppetdb.rb
+++ b/puppet/lib/puppet/indirector/catalog/puppetdb.rb
@@ -10,8 +10,9 @@ class Puppet::Resource::Catalog::Puppetdb < Puppet::Indirector::REST
   def save(request)
     profile("catalog#save", [:puppetdb, :catalog, :save, request.key]) do
       current_time = Time.now
-      catalog = munge_catalog(request.instance, current_time, extract_extra_request_data(request))
+      catalog, inputs = munge_catalog(request.instance, current_time, extract_extra_request_data(request))
       submit_command(request.key, catalog, CommandReplaceCatalog, 9, current_time.clone.utc)
+      submit_command(request.key, inputs, CommandReplaceCatalogInputs, 1, current_time.clone.utc) if inputs
     end
   end
 
@@ -48,6 +49,19 @@ class Puppet::Resource::Catalog::Puppetdb < Puppet::Indirector::REST
     nil
   end
 
+  def munge_catalog_inputs(hash)
+    inputs = hash.delete('inputs')
+
+    return if inputs.nil?
+
+    {
+      certname: hash['certname'],
+      catalog_uuid: hash['catalog_uuid'],
+      producer_timestamp: hash['producer_timestamp'],
+      inputs: inputs
+    }
+  end
+
   def munge_catalog(catalog, producer_timestamp, extra_request_data = {})
     profile("Munge catalog", [:puppetdb, :catalog, :munge]) do
       data = profile("Convert catalog to JSON data hash", [:puppetdb, :catalog, :convert_to_hash]) do
@@ -75,7 +89,9 @@ class Puppet::Resource::Catalog::Puppetdb < Puppet::Indirector::REST
       add_producer(data, Puppet[:node_name_value])
       redact_sensitive_params(data)
 
-      data
+      inputs = munge_catalog_inputs(data)
+
+      [data, inputs]
     end
   end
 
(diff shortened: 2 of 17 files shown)


src/puppetlabs/puppetdb/command.clj
```diff
@@ -427,7 +441,8 @@
         "replace facts" (replace-facts command start db blacklist-config)
         "store report" (store-report command start db)
         "deactivate node" (deactivate-node command start db)
-        "configure expiration" (configure-expiration command start db)))))
+        "configure expiration" (configure-expiration command start db)
+        "replace catalog-inputs" (replace-catalog-inputs command start db)))))
```
rbrw: I'm not sure, but wondering if we might want "replace catalog inputs" here, i.e. without the dash.  Don't think we have any precedents, but I'd tended to think of the wire command names as the plain text version, which we convert in various contexts by either `s/ /_/` or `s/ /-/`, depending.  I suppose that also has the advantage of making the conversion trivially reversible, but not sure that matters.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -302,6 +302,20 @@
...
+  [{:keys [id received payload]}
+   start-time db]
+  (let [{:keys [certname inputs catalog_uuid]
+         stamp :producer_timestamp, :or {stamp (now)}} payload]
+    (when (seq inputs)
+      (jdbc/with-transacted-connection db
+        (scf-storage/maybe-activate-node! certname stamp)
+        (scf-storage/replace-catalog-inputs! certname catalog_uuid inputs stamp))
+      (log-command-processed-messsage id received start-time
+                                      :configure-expiration certname))))
```
rbrw: copy-paste? :configure-expration

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -215,6 +216,7 @@
   "Delete the given host from the db"
   [certname]
   {:pre [certname]}
+  (jdbc/delete! :catalog_inputs ["certname_id in (select id from certnames where certname=?)" certname])
   (jdbc/delete! :certname_packages ["certname_id in (select id from certnames where certname=?)" certname])
   (jdbc/delete! :certnames ["certname=?" certname]))
```
rbrw: May not matter yet, but if we add more certname_id related lines like this and assuming correct transactional protection, wonder if we could avoid some work by returning the id from the first delete and using it for others.  i.e. `delete from foo where ... returning certname_id`, or assuming it's feasible and we can get it right (this likely isn't quite), do it all at once:

```
with ids as (delete from whatever where ... returning certid
                 union delete from ...)
    delete from certnames where id in (select certid from ids);
```

Not entirely sure this would be wise, just wondered.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1404,7 +1446,7 @@
                          [sql-timestamp certname sql-timestamp])))))
 
 (pls/defn-validated expire-stale-nodes [horizon :- Period]
-  "Expires nodes with no activity within the provided horizon (prior
+  "Expires nodes with no activity within the provided horizon (prior/catalog
```
rbrw: Not sure I followed "prior/catalog" -- might be worth elaborating a little more.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -823,6 +825,46 @@
...
+    (let [updated (to-timestamp updated)
+          catalog-uuid (sutils/munge-uuid-for-storage catalog_uuid)
+          certid (-> (jdbc/query
+                       (hcore/format {:select [:id]
+                                      :from [:certnames]
+                                      :where [:= :certname certname]}))
+                     first
+                     :id)]
+      (delete-catalog-inputs! certid)
+      (store-catalog-inputs! certid inputs)
```
rbrw: Perhaps a candidate for an upsert (on conflict replace) once we're ready to pursue them more enthusiastically.  Wonder if we might want to leave a comment (something greppable) to make it easier to find when we're working on that, if we're not ready to do it now.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -823,6 +825,46 @@
...
+
+(defn delete-catalog-inputs! [certid]
+  (jdbc/delete! :catalog_inputs ["certname_id = ?" certid]))
+
+(defn store-catalog-inputs!
+  [certid inputs]
+    (jdbc/insert-multi!
+      :catalog_inputs
+      [:certname_id :type :name]
+      (add-certname-id-to-catalog-inputs certid inputs)))
```
rbrw: Assuming I understood this right, I'd be tempted to just avoid the global helper if we don't need it elsewhere, and just say something like this directly:
```
(defn store-catalog-inputs!
  [certid inputs]
    (jdbc/insert-multi! :catalog_inputs
                        [:certname_id :type :name]
                        (map #(apply vector certid %) inputs)))
```

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -823,6 +825,46 @@
...
+   updated :- pls/Timestamp]
+  (jdbc/with-db-transaction []
+    (let [updated (to-timestamp updated)
+          catalog-uuid (sutils/munge-uuid-for-storage catalog_uuid)
+          certid (-> (jdbc/query
+                       (hcore/format {:select [:id]
+                                      :from [:certnames]
+                                      :where [:= :certname certname]}))
+                     first
+                     :id)]
```
rbrw: Perhaps `(certname-id certname)`.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -302,6 +302,20 @@
         (assoc :payload validated-payload)
         (replace-catalog* start-time db))))
 
+;; Catalog input replacement
+
+(defn replace-catalog-inputs
+  [{:keys [id received payload]}
+   start-time db]
+  (let [{:keys [certname inputs catalog_uuid]
+         stamp :producer_timestamp, :or {stamp (now)}} payload]
```
rbrw: Prefer continuing to omit commas unless others feel strongly about it.

src/puppetlabs/puppetdb/scf/storage.clj
austb: oops that's a accidental paste from the clipboard, I'll remove that

documentation/api/wire_format/replace_catalog_inputs_format_v1.markdown
```diff
@@ -0,0 +1,74 @@
...
+
+The command is serialized as JSON, which requires strict UTF-8 encoding.
+
+### Main data type: Replace catalog inputs for node
+
+     {
+      "certname": <string>,
+      "producer_timestamp": <datetime>,
+      "catalog_uuid": <string>,
+      "inputs": [[<string>, <string>]],
```
rbrw: Let's add the ... to match our other docs, e.g.: `"inputs": [[<string>, <string>], ...]`

documentation/api/wire_format/replace_catalog_inputs_format_v1.markdown
```diff
@@ -0,0 +1,74 @@
...
+     }
+
+#### `certname`
+
+String. The name of the node for which the catalog was compiled for.
+
```
rbrw: Maybe drop the final "for"?

documentation/api/wire_format/replace_catalog_inputs_format_v1.markdown
```diff
@@ -0,0 +1,74 @@
...
+When `producer_timestamp` is not provided, the PuppetDB server's local time is
+used. If another command is received for a node while a non-timestamped
+"deactivate node" command is pending processing, the results are *undefined*.
+
+#### `catalog_uuid`
+
+The uuid of the catalog this input was used for. Since not every catalog will
+provide a record of its inputs, and the catalog is replaced via a different
+command, this may not be the same as the current catalog being stored in
+PuppetDB.
```
rbrw: Suggest "The uuid of the catalog that was generated from the included `inputs`...", and perhaps "being stored"  -> "stored".

documentation/api/wire_format/replace_catalog_inputs_format_v1.markdown
```diff
@@ -0,0 +1,74 @@
...
+
+#### `inputs`
+
+The `inputs` to a catalog are a list of tuples, where the first element is its
+`type` (ie. `"hiera"`) and the second is a unique name for the inputs (for a hiera
```
rbrw: "unique name for the input" (singular), or maybe even "...where the first element is the input type (e.g. "hiera"), and the second is a unique name (for a hiera key ...".

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -823,6 +825,49 @@
     (sutils/sql-hash-as-str "hash") certname]
    (comp :catalog first sql/result-set-seq)))
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;; Catalog inputs updates/changes
+
+(defn add-certname-id-to-catalog-inputs [certid inputs]
+  (map (fn [input] (into [certid] input)) inputs))
```
rbrw: Vestigial now?

rbrw:
After talking with Austin, think we may see if Maggie has time to take a look at the terminus changes.

documentation/api/wire_format/replace_catalog_inputs_format_v1.markdown
Magisus: Gotta replace it with "...for which the catalog..." if you're going to drop it.

puppet/lib/puppet/indirector/catalog/puppetdb.rb
```diff
@@ -48,6 +49,19 @@ def redact_sensitive_params(hash)
...
+
+    return if inputs.nil?
+
+    {
+      certname: hash['certname'],
+      catalog_uuid: hash['catalog_uuid'],
+      producer_timestamp: hash['producer_timestamp'],
+      inputs: inputs
+    }
+  end
```
Magisus: @pcarlisle this is the proposal for how the inputs are getting to PDB, via the catalog indirection. Thoughts?

rbrw:
May need a rebase?

rbrw:
Looks good.

PR #3031, opened 2019-07-31
(docs) Fix header versions and separate release notes into per X.Y files

austb:
Removes prefixes from docs page titles
Removes usage of »
Limits `documentation/release_notes.markdown` file to containing only the release notes for 5.2.x releases


documentation/api/metrics/v1/changes-from-puppetdb-v3.markdown
```diff
@@ -1,5 +1,5 @@
 ---
-title: "PuppetDB 5.2 » Metrics API » v1 » Upgrading to PuppetDB 4 or newer"
+title: "Metrics API v1"
 layout: default
```
rbrw: Wondered if this should keep the upgrade context, say "Metrics API changes in version 4" or something.

documentation/api/query/v4/pql.markdown
```diff
@@ -1,5 +1,5 @@
 ---
-title: "PuppetDB 5.2 » Puppet query language (PQL) » Reference guide"
+title: "PQL Reference guide"
 layout: default
```
rbrw: I'm somewhat tempted to keep the full name here since, if nothing else, it looks like that's what shows up in the search engine match line, and so that one line helps whether you know the full name or the acryonym, but I don't feel strongly about it. i.e. "Puppet query language (PQL) Reference guide".

documentation/trouble_session_logging.markdown
```diff
@@ -1,5 +1,5 @@
 ---
-title: "PuppetDB 5.2 » Troubleshooting » Session Logging"
+title: "Troubleshooting Session Logging"
```
rbrw: Hmm, thought perhaps "Troubleshooting: Session Logging"?

rbrw:
Looks good.

PR #3032, opened 2019-08-01
(maint) Update Docker shared helper usage

Iristyle:


Iristyle:
The Azure failure was unrelated.. just took too long!

Merging this because all of the bits we care about have been validated in Azure already.

PR #3035, opened 2019-08-06
(maint) Specify all Docker lints as data for Azure

Iristyle:
- Don't rely on defaults for Lint-Dockerfile as they will be removed.
   Instead, specify all explicitly.

   This further standardizes the YAML


Iristyle:
Travis failure is unrelated since these are Azure specific changes, so merging.

austb:
:+1: Travis failures seem to be a network issue, they can't even clone puppetdb

PR #3036, opened 2019-08-06
(SERVER-2581) Send hiera key lookups

pcarlisle:
We're storing this in the Context


puppet/lib/puppet/indirector/catalog/puppetdb.rb
```diff
@@ -50,9 +50,12 @@ def redact_sensitive_params(hash)
   end
 
   def munge_catalog_inputs(hash)
-    inputs = hash.delete('inputs')
+    recorder = Puppet.lookup(:lookup_key_recorder) { nil }
+    return unless recorder && recorder.respond_to?(:lookups) && !recorder.lookups.empty?
```
pcarlisle: :rage1:

PR #3037, opened 2019-08-06
(PDB-4372) Log command name for replace catalog inputs

austb:
An incorrect keyword used to look up the command name
for 'replace catalog inputs' commands caused a log message like
`'null' command processed for new-host` to be printed.


austb:
The command name lookup map is in [`src/puppetlabs/puppetdb/command/constants.clj`](http://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/command/constants.clj#L4-L10)

PR #3038, opened 2019-08-08
(MAINT) Fix postgres ENV variable

nwolfe:
Using an ENV variable in the definition of another ENV variable in the
Dockerfile doesn't work as expected, as these values are interpreted at
build time, not run time.

Also bump to the latest ssl.sh which supports DNS_ALT_NAMES, and plumb
that through with an empty default.


docker/puppetdb/docker-entrypoint.d/30-configure-ssl.sh
```diff
@@ -2,7 +2,7 @@
 
 if [ ! -f "/etc/puppetlabs/puppet/ssl/certs/${HOSTNAME}.pem" ] && [ "$USE_PUPPETSERVER" = true ]; then
   set -e
-  /ssl.sh
+  DNS_ALT_NAMES="${DNS_ALT_NAMES}" /ssl.sh "$HOSTNAME"
```
nwolfe: While it's not necessary for us to specify this values here, since ssl.sh will find these ENV variables itself, I wanted to make it clear where they're being used in this container definition since ssl.sh doesn't live in this repo.

docker/puppetdb/docker-entrypoint.d/00-envsubst.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# NOTE: define ENV variables and defaults in the Dockerfile
+
+envsub() {
+    envsubst < "$1" > "${1}.new"
+    mv --force "${1}.new" "$1"
+}
+
+for f in /etc/puppetlabs/puppetdb/conf.d/*.conf; do
```
Iristyle: I don't think we even need to do this, do we? 

I think `database.conf` already understands / uses the env vars defined in the environment, doesn't it?

nwolfe: You'd think it would work but it didn't when I tried it before submitting the PR.

The existence of quotes screws it up, and without them the syntax is incorrect.

nwolfe:
```
subname: \"//${PUPPETDB_POSTGRES_HOSTNAME}:${PUPPETDB_POSTGRES_PORT}/${PUPPETDB_POSTGRES_DATABASE}\"
```

Error:
```

Comment (austb) [reply docker/puppetdb/docker-entrypoint.d/00-envsubst.sh]: That reads to me like the `\`'s aren't necessary, does that work?

Comment (nwolfe) [reply docker/puppetdb/docker-entrypoint.d/00-envsubst.sh]: I'm reproducing all the ways it fails....stay tuned

Comment (nwolfe) [reply docker/puppetdb/docker-entrypoint.d/00-envsubst.sh]: Error:

Comment (nwolfe) [reply docker/puppetdb/docker-entrypoint.d/00-envsubst.sh]:
```
  subname: //${PUPPETDB_POSTGRES_HOSTNAME}:${PUPPETDB_POSTGRES_PORT}/${PUPPETDB_POSTGRES_DATABASE}
```

Error:
```
puppetdb_1  | Exception in thread "main" com.typesafe.config.ConfigException$Parse: /etc/puppetlabs/puppetdb/conf.d/database.conf: 3: Expecting close brace } or a comma, got ':' (if you intended ':' to be part of the value for 'subname', try enclosing the value in double quotes)
```

nwolfe: Ok...see above for the 3 forms that don't work.

nwolfe: The behavior of this .conf file when it's read by PuppetDB seems to be:
- Quotes prevent variable interpolation
- Unquoted is a syntax error (for this particular value)

underscorgan: :sigh: computers are the worst

Iristyle: Should we file a ticket / leave a comment here about why the typical expected PDB behavior doesn't work?

/cc @rbrw @robdaemon @austb

robdaemon: file a ticket, please :)

Iristyle: It looks like this is a HOCON thing per https://github.com/lightbend/config/issues/633

austb: So intuitive... 🤦‍♂

nwolfe: Nice find. That works!

PR #3039, opened 2019-08-09
(maint) Bump clj-parent to 3.1.1

austb:
Breaking changes for `puppetlabs/ssl-utils`, which we only
depend on via `puppetlabs/http-client`.


austb:
I targeted `6.3.x` instead of `6.0.x` because `6.0.x` is EOL at the end of the month(ish).

There is one newer version of clj-parent (`4.0.0`), but I didn't want to do two breaking changes at once and 4 has more breaking changes than 3. It would be good to upgrade to 4 after the next round of releases though.

PR #3040, opened 2019-08-13
(maint) Switch PDB container off of Alpine

Iristyle:
- Alpine seems to still be having issues with DNS resolutions inside
   an LCOW environment. In an effort to reduce these transient
   problems, switch the base container to a non-Alpine platform.

   A ticket has been filed with a repro for Alpine DNS issues under LCOW
   https://github.com/docker/libnetwork/issues/2371

 - While this may increase the image size by about 100MB, the goal here
   is reliability and robustness

   for the builder container:
   clojure:lein-alpine was about 142MB
   clojure:openjdk-8-lein is about 507MB

	 for the target container:
   openjdk:8-jre-alpine was about 85MB
   openjdk:8-buster-slim is about 184MB

 - Ubuntu 18.04 shares a lineage with debian buster, which should
   be a well supported platform for PDB

   All OpenJDK container variants are listed at:
   https://github.com/docker-library/docs/blob/master/openjdk/README.md#supported-tags-and-respective-dockerfile-links

### DIFF
--- a/docker/puppetdb/Dockerfile
+++ b/docker/puppetdb/Dockerfile
@@ -1,5 +1,6 @@
-FROM clojure:lein-alpine AS builder
-RUN apk add --no-cache make
+FROM clojure:openjdk-8-lein AS builder
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends make
 # Install only dependencies
 WORKDIR /app
 COPY project.clj /app/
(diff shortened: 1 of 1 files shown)


Iristyle:
I'm not sure if we can get away with an alpine builder still or not? I would expect not for a C++ application unless it was explicitly cross-compiling, but I'm not sure on our Java code.

PR #3042, opened 2019-08-15
(maint) Fix Docker ssl-setup.sh for Debian / Dash

Iristyle:
Contains a fix for #3040 

 - When the base container image was changed from Alpine Linux to
   Debian Slim Buster in #3040, the /bin/sh implementation changed.

   3c876e5847915fd769976a382b1dc3f890809019 merged this code

   On Alpine, /bin/sh is the standard bare-bones cross-platform
   Bourne Shell.

   On Debian, /bin/sh is a symlink to dash.

 - Update shellscript syntax to be compatible with Dash.

### DIFF
--- a/docker/puppetdb/ssl-setup.sh
+++ b/docker/puppetdb/ssl-setup.sh
@@ -18,7 +18,7 @@ mycertname="${CERTNAME:-$HOSTNAME}"
 #
 #   usage
 #
-function usage() {
+usage() {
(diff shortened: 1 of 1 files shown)


Iristyle:
Ran this locally against the pupperware docker-compose file at https://github.com/puppetlabs/pupperware/blob/master/docker-compose.yml

SSL setup proceeded successfully and all containers fired up.

PR #3043, opened 2019-08-15
(PDB-4482) Suppress duplicate catalog inputs on ingestion

rbrw:
I imagine we'll eventually want an index anyway, that'd end up
requiring uniqueness, but for now suppress them so that they won't be
visible to queries, etc.


rbrw:
Not sure this is what we want.  Thoughts?  Also - think this might be the first time I've used a transducer -- might be be a bit cheaper, if I got it right...

PR #3046, opened 2019-08-17
(maint) Add Postgres pg_isready based healthcheck

Iristyle:
- pupperware stack is using this as a healthcheck, which allows
   replacing any calls to PDB specific waiters with a generic
   health waiter


Iristyle:
Travis failures were unrelated - merging!

PR #3047, opened 2019-08-19
(maint) release notes: add contributors for recent 5.2 releases 

rbrw:


rbrw:
Might be worth spot checking the lists to make sure I did what I think I did.

.mailmap
```diff
@@ -0,0 +1,6 @@
+Claire Cadman <claire.cadman@puppet.com> claire cadman <claire.cadman@puppet.com>
+Claire Cadman <claire.cadman@puppet.com> clairecadman <claire.cadman@puppet.com>
+Heston Hoffman <hestonhoffman@gmail.com> hestonhoffman <hestonhoffman@gmail.com>
+<morgan@puppet.com> <morgan@puppetlabs.com>
```
austb: I have (oh so impressively) identified this line as different than the rest. Is this right?

rbrw: Yeah, I think so?  Check out `git-blame(1)`.  See the "MAPPING AUTHORS" section at the end.  There are multiple formats depending on what you're trying to normalize.

PR #3050, opened 2019-08-21
Update to clj-parent 4.0.4

robdaemon:


test-resources/integration-puppetdb.conf
```diff
@@ -31,7 +31,7 @@ jetty: {
   ssl-cert: ./test-resources/puppetserver/ssl/certs/localhost.pem
   ssl-key: ./test-resources/puppetserver/ssl/private_keys/localhost.pem
 
-  ssl-protocols: ["TLSv1", "TLSv1.1"]
+  ssl-protocols: ["TLSv1", "TLSv1.1", "TLSv1.2"]
```
robdaemon: @austb specifically this change is what I think is fixing it

PR #3052, opened 2019-08-22
(maint) disable tests that cannot pass

robdaemon:
with clj-parent 4.0.4, you cannot run puppetserver 6.4 in the same JVM
as puppetdb. We will address this in a future change to run puppetserver
and puppetdb in separate JVMs.


robdaemon:
container tests are failing but travis-ci tests passed

PR #3055, opened 2019-08-27
(PDB-2634) Hash projection for (some) json fields

austb:
See commit message.

It's ready!

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -37,6 +37,11 @@
                               SqlCall SqlRaw
                               {:select s/Any s/Any s/Any}))
 
+(defn is-dotted-projection?
+  [projection]
+  ;; If the projection has a dot it must be of the form fact.foo
+  (str/includes? projection "."))
+
 ; The use of 'column' here is a misnomer. This refers to the result of
 ; a query and the top level keys used should match up to query entities
 ; and the values of 'columns', 'local-column', and 'foreign-column'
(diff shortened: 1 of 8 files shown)


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1195,6 +1202,18 @@
...
+(defn projectable-json-fields
+  "Returns a list of fields of queryable json fields whose
+   field type supports json subtree extraction.
+
+   Currently only queryable json fields whose :field value
+   is a keyword or raw SQL support json subtree extraction."
+  [{:keys [projections]}]
+  (->> projections
+       (filter (comp #(and (= :queryable-json (:type %))
+                           (:projectable-json? %)) val))
```
rbrw: Think val should be on a separate line so it is more obviously a comp argument than an additional and clause.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1775,13 +1800,37 @@
...
+   :queryable? false
+   :field (hcore/raw
+            (str
+              ; Extract the original field as a string
+              (->> json-path
+                   first
+                   projections
+                   stringify-field)
+              ; append any required ->'key' operators
+              (when (> (count json-path) 1)
```
rbrw: Could say `(pos? (count json-path))`, but think it's more idomatic to say `(seq json-path)`.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1775,13 +1800,37 @@
...
 
+(defn create-json-subtree-projection
+  [dotted-field projections]
+  (let [json-path (str/split dotted-field #"\.")
+        stringify-field (fn [{:keys [field field-type]}]
+                          (case field-type
+                            :keyword (name field)
+                            :raw (:s field)
+                            (throw (IllegalArgumentException.
+                                           (tru "Cannot determine field type for field ''{0}''" field)))))]
```
rbrw: Indent is a bit off.

src/puppetlabs/puppetdb/query_eng/engine.clj
austb: I think that would only work for `> 0`, here I need to check that  there are actually json keys to look up.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1102,6 +1108,7 @@
                  :queryable? true
                  :field :timestamp}
     "facts" {:type :queryable-json
+             :projectable-json? false
```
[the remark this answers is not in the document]
austb: I don't feel too strongly about this, it'll work either way. I was just marking that as "does NOT support sub-projection". Let me know if you think it should be dropped.

PR #3063, opened 2019-09-09
(PDB-4456)Add import and export for catalog-inputs

Zak-Kent:


src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -449,6 +449,13 @@
     (-> wire-node
         (update "certname" anonymize-leaf :node context config))))
 
+(defn anonymize-catalog-inputs
+  "Anonymize a set of catalog inputs"
+  [config wire-catalog-inputs]
+  (let [context {"node" (get wire-catalog-inputs "certname")}]
+    (-> wire-catalog-inputs
+        (update "certname" anonymize-leaf :node context config))))
```
rbrw: Just a note that Zak and I think this might need additional anonymization (particularly for input names) before we can finish up here.

src/puppetlabs/puppetdb/anonymizer.clj
```diff
@@ -139,6 +139,8 @@
        :type (random-type-name)
        :title (random-string (max 10 (count value)))
        :parameter-name (random-string-alpha (max 10 (count value)))
+       :catalog-input-type (random-string-alpha (max 10 (count value)))
```
austb: How do we feel about not anonymizing types? It will be one of three things `hiera`, `function`, `template`.

austb: Nevermind, this only happens for `full` 👍 makes sense. Ignore me.

PR #3064, opened 2019-09-09
(maint) Re-add bouncycastle to uberjar

austb:
Longer term excluding the bouncycastle jar from the uberjar and loading
either the standard or the FIPs version at runtime is likely simpler,
but for now, the infrastructure for that does not exist so we'll revert
to building either a standard uberjar or a FIPs uberjar (with the fips
profile).


robdaemon:
Follow up based on Slack convo: please add a comment on that dep that we only want it included in FOSS and PE is handled differently

and then we'll merge this

PR #3066, opened 2019-09-10
(PDB-4278) Add metric that tracks "bashed" commands

Zak-Kent:
This commit adds a metric that tracks the number of commands which are
"bashed" in the command queue.

### DIFF
--- a/documentation/api/metrics/v1/mbeans.markdown
+++ b/documentation/api/metrics/v1/mbeans.markdown
@@ -111,6 +111,7 @@ following `<item>`s:
   previously enqueued commands
 * `queue-time`: histogram of the time commands have spent waiting in the queue
 * `depth`: number of currently enqueued commands
+* `ignored`: number of obsolete commands that have been ignored
 * `size`: histogram of submitted command sizes (i.e. HTTP Content-Lengths)
 
 For example: `puppetlabs.puppetdb.mq:name=global.seen`.
@@ -128,6 +129,7 @@ integer command version, and `<item>` must be one of the following:
 * `retried`: meter measuring commands scheduled for retrial
 * `retry-counts`: histogram of retry counts (until success or discard)
 * `discarded`: meter measuring commands discarded as invalid
+* `ignored`: number of obsolete commands that have been ignored
 * `processing-time`: timing statistics for the processing of
   previously enqueued commands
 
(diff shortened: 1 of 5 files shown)


Zak-Kent:
We still may want to add more documentation to this around how a command gets "bashed"

documentation/api/metrics/v1/mbeans.markdown
```diff
@@ -111,6 +111,7 @@ following `<item>`s:
   previously enqueued commands
 * `queue-time`: histogram of the time commands have spent waiting in the queue
 * `depth`: number of currently enqueued commands
+* `bashed`: number of commands which have been "bashed" in the command queue
```
rbrw: I'm not sure, but while "bashed" is quite specific, i wondered whether something like "suppressed" might be preferable as a public name, ie.. 'number of "stale/obsolete/obviated/whatever' commands that have been suppressed".  On the other hand, maybe metrics are "internal" enough that the name more directly linked to the operation is better...

Zak-Kent: That's a good point. I went with "ignored" because that's how we refer to it in the logs [here](https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/command.clj#L282)

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -462,7 +462,8 @@
           command-chan (async/chan
                          (queue/sorted-command-buffer
                           max-enqueued
-                          #(cmd/update-counter! :invalidated %1 %2 inc!)))
+                          #(cmd/update-counter! :invalidated %1 %2 inc!)
+                          #(cmd/update-counter! :ignored %1 %2 inc!)))
```
austb: I'd like to register my dislike of positional params.

robdaemon: https://stuartsierra.com/2019/09/15/clojure-donts-numbered-parameters

please change this to `(fn [...] ...)` :grin:

Zak-Kent: cool, that's a fair point. I'll update both counters to not use them in a separate commit

robdaemon:
cool, :+1: merge away as soon as Travis goes green

PR #3071, opened 2019-09-16
(PDB-4505) Log progress during resource-events-pk migration

rbrw:
Log "Migrated X of Y events" roughly once per minute during the
migration so that you can fairly quickly estimate about how long it
might take, and can see that it's making progress.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1540,22 +1540,29 @@
   ;; we've encountered during the migration to avoid inserting
   ;; duplicates into the new table.
 
-  (jdbc/call-with-query-rows
-   ["select
+  (let [event-count (-> "select count(*) from resource_events"
+                        jdbc/query-to-vec first :count)
+        last-logged (atom (.getTime (java.util.Date.)))]
+    (jdbc/call-with-query-rows
+     ["select
        report_id, certname_id, status, \"timestamp\", resource_type, resource_title, property,
        new_value, old_value, message, file, line, containment_path, containing_class, corrective_change
      from resource_events"]
-   (fn [rows]
-     (reduce (fn [hashes-seen row]
-               (conj hashes-seen
-                     (let [hash-str (hash/resource-event-identity-pkey row)]
-                       (when-not (hashes-seen hash-str)
-                         (jdbc/insert-multi! "resource_events_transform"
-                                             (list (assoc row :event_hash
-                                                          (sutils/munge-hash-for-storage hash-str)))))
-                       hash-str)))
-             #{}
-             rows)))
+     (fn [rows]
+       (reduce (fn [hashes-seen [row i]]
+                 (let [now (.getTime (java.util.Date.))]
+                   (when (> (- now @last-logged) 60000)
+                     (log/info (trs "Migrated {0} of {1} events" i event-count))
+                     (reset! last-logged now)))
+                 (conj hashes-seen
+                       (let [hash-str (hash/resource-event-identity-pkey row)]
+                         (when-not (hashes-seen hash-str)
+                           (jdbc/insert-multi! "resource_events_transform"
+                                               (list (assoc row :event_hash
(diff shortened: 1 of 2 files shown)


rbrw:
Not sure whether "one minute" is what we want, or even whether we want to log with a fixed cadence, but we can figure that out before we merge.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1540,22 +1540,32 @@
...
-                                                          (sutils/munge-hash-for-storage hash-str)))))
-                       hash-str)))
-             #{}
-             rows)))
+     (fn [rows]
+       (reduce (fn [hashes-seen [row i]]
+                 (let [now (.getTime (java.util.Date.))]
+                   (when (> (- now @last-logged) 60000)
+                     ;; Manually stringify numbers to avoid
+                     ;; locale-specific formatting
```
robdaemon: that's not i18n friendly. Let the locale formatting happen.

rbrw: Don't feel strongly about it, but we've done that in other places (command.clj), and I believe we did it so that people could more easily extract the numbers automatically.  Of course in the longer run, the "right" answer would be something like https://github.com/puppetlabs/structured-logging (or just that), but that's still not friendly if people just wanted to use sed.  Though of course you could also argue that in that case, they have to make their script locale specific...

robdaemon: yeah, but you can't extract the numbers easily with it wrapped in a `(trs ...)` because it might get translated. structured logging is definitely a good thing. For this purpose though, this message is for a user to know the progress. We should follow proper i18n guidelines here.

Sharpie: The addition of commas, regardless of locale conventions, makes the output less useful for diagnosis as any script that parses the logs has to discard the commas in order to perform any sort of calculation. The `{0}` and `{1}` should be interpolated after `(trs ...)` finishes, so the numbers should remain.

robdaemon: Any sort of translation can change the ordering of the values depending on locale. In some locales, it'll use a `.` instead of a `,` for a thousands-separator. For example "1 of 100" is not how it translates everywhere.

If you expect to do analysis of logs via scripting, you must use structured logging.

Sharpie: Decimal points won't be an issue since row counts are integers. A reverse in the order of the fields is much easier to handle when knocking out a quick log parsing regex than eliminating separators.

We frequently analyze logs via scripting despite the presence of structured logging. With that in mind, I assert that the benefits of making debugging and analysis simpler by omitting separators in this case is much greater than the benefits of adding them.

robdaemon: This log message is intended for users to read so that they can assess the progress of a migration. Given that, the benefits of user readability are more important that having to write a regex to strip them for any future scripting. See "Locale-specific Number formatting" in our "Developer’s Checklist for Internationalizing Software" on Confluence.

rbrw: @Sharpie presumably it wouldn't help to suggest that the locale be changed to one that matches what the script expects because you're thinking of cases where the log's are being processed after the fact?  Also of course, if/when we do real localization, then it's not just the numbers that might change; it's all the text.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1540,22 +1541,31 @@
...
-                       hash-str)))
-             #{}
-             rows)))
+     (fn [rows]
+       (reduce (fn [hashes-seen [row i]]
+                 (let [now (.getTime (java.util.Date.))]
+                   (when (> (- now @last-logged) 60000)
+                     (maplog :info
+                             {:migration 69 :at i :of event-count}
+                             #(trs "Migrated {0} of {1} events" (:at %) (:of %)))
```
robdaemon: I like this **a lot.**

PR #3073, opened 2019-09-17
(maint) Add ezbake configuration warnings

austb:


austb:
We should let the Jenkins tests run to verify that ezbake accepts the comment style. It's the same style as puppetserver uses in `ezbake.conf`, so I anticipate it will work.

PR #3074, opened 2019-09-17
(PDB-4504) Use batched inserts for add-resource-events-pk

Sharpie:
This commit updates the add-resource-events-pk migration to perform
insertions in batches of 1000 when re-writing the resource_events
table to add a primary key.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1509,7 +1509,10 @@
 
   {::vacuum-analyze #{"factsets"}})
 
-(defn add-resource-events-pk []
+(defn add-resource-events-pk
+  ([]
+   (add-resource-events-pk 1000))
+  ([batch-size]
   (jdbc/do-commands
    "CREATE TABLE resource_events_transform (
       event_hash bytea NOT NULL PRIMARY KEY,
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1546,16 +1546,31 @@
...
+                          (jdbc/insert-multi! :resource_events_transform
+                                              new-cols
+                                              batch)
+                          ;; Return hashes for each row directly instead of
+                          ;; reading them back from the DB via the return
+                          ;; value of insert-multi!
+                          (map :event_hash batch))
+           dedupe-and-insert (fn [hashes-seen batch]
+                               (->> batch
+                                    (pmap update-row)
```
Sharpie: @rbrw You mentioned there may be some issues with using `pmap` vs `map`. It's a minor optimization compared to the boost we get from batching, but hashing is CPU-intensive enough that it seems worth doing in parallel if we can do so easily.

Sharpie: For the test dataset of 5.7 million rows, `pmap` appeared to cut about 3.5 minutes off of a 15 minute batched migration.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1546,16 +1546,31 @@
...
+                                    (filter #(-> % first hashes-seen nil?))
+                                    insert->ids
+                                    (reduce conj hashes-seen)))]
+       (reduce dedupe-and-insert
+               #{}
+               (partition 1000 1000 [] rows)))))
```
Sharpie: Using a fixed batch size of 1000 here. For the 5.7 million row dataset, this dropped migration time from about 45 minutes to about 15 minutes. I haven't played with the sensitivity of the batch size extensively, but going to 10,000 rows per batch only yielded another 1.5 minutes in speedup. Larger batches will require more RAM to process through PuppetDB.

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Yeah, let's check.  I think the concern may have been the fact that pmap is a bit blunt, and uses the common agent pool.  If that's it, then ideally you'd want to parallelize via a separate pool, though it might be less of a concern during startup.  I think part of that (serious) general concern is that in general you might end up with all the threads blocked (affecting agents, core.async, etc. too), and possibly even deadlocking the program?

src/puppetlabs/puppetdb/scf/migrate.clj
rbrw: Think I'd be inclined to stick with 1k if the further increase is that marginal.

Sharpie:
One outstanding question is why PG-JDBC requires `reWriteBatchedInserts` to be enabled in order to get a batched insert. We may want to enable this by default, but only for the `PDBMigrationsPool`.

Sharpie:
Updated to:

  - De-duplicate within batches using `group-by` before removing duplicates seen in prior batches via `filter`.
  - Guard `insert-multi!` against empty batches that occur when all rows in the iteration end up filtered as duplicates.
  - Made the batch size a parameter of the migration with a default value of 1000. The test is updated to run with a batch size of 2 to ensure two iterations occur.

Sharpie:
Patch updated to drop the use of `pmap`. Further benchmarking showed the change produces nearly a 3x speedup for the 5.7 million event test dataset even without `reWriteBatchedInserts`.

Using `pmap` to parallelize the hash computation and adding `reWriteBatchedInserts` to the connection options each add about a 10% speedup relative to the original migration.

src/puppetlabs/puppetdb/scf/migrate.clj
Sharpie: Removed `pmap`.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1546,16 +1549,38 @@
...
-             #{}
-             rows)))
+     (let [old-cols '(:report_id :certname_id :status :timestamp :resource_type
+                      :resource_title :property :new_value :old_value :message
+                      :file :line :containment_path :containing_class
+                      :corrective_change)
+           new-cols (conj old-cols :event_hash)
+           update-row (apply juxt (conj old-cols
+                                        (comp sutils/munge-hash-for-storage
+                                              hash/resource-event-identity-pkey)))
```
rbrw: If you'd rather leave it alone, no worries, but I'd be tempted to change old-cols to a vector, create new-cols as `(apply vector :event_hash old-cols)` (or could just use cons, if it doesn't need to be a vector), and then use `(apply juxt (comp ...) old-cols)`.  That'd be more conventional, but more notably, clojure-mode won't try to indent it wrong every time paredit, etc. adjusts the form, and it avoids people having to remember conj's "direction-sensitivity".

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1546,16 +1549,38 @@
...
+           ;; Retrieve event_hash string from PGobject produced by sutils
+           row->id (fn [row] (-> row first .getValue))
+           insert->hash (fn [batch]
+                          (when-not (empty? batch)
```
rbrw: If you decide to change the above, perhaps while you're here `(when (seq? batch) ...)`.

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -25,9 +25,11 @@
 (use-fixtures :each tdb/call-with-test-db)
 
 (defn apply-migration-for-testing!
-  [i]
+  [i & args]
   (let [migration (migrations i)
-        result (migration)]
+        result (if (empty? args)
+                 (migration)
+                 (apply migration args))]
```
rbrw: Think you may not need the `if` here -- `(apply migration args)` will do the right thing either way.

src/puppetlabs/puppetdb/scf/migrate.clj
Sharpie: Switched to a vector.

src/puppetlabs/puppetdb/scf/migrate.clj
Sharpie: Makes sense, flipped it.

test/puppetlabs/puppetdb/scf/migrate_test.clj
Sharpie: Thought this blew up when `nil` was passed, but looks like it's doing the right thing. Changed.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1541,31 +1544,55 @@
...
-                                               (list (assoc row :event_hash
-                                                            (sutils/munge-hash-for-storage hash-str)))))
-                         hash-str)))
+  (jdbc/call-with-query-rows
+   ["select
+     report_id, certname_id, status, \"timestamp\", resource_type, resource_title, property,
+     new_value, old_value, message, file, line, containment_path, containing_class, corrective_change
+   from resource_events"]
+   (fn [rows]
+     (let [event-count (-> "select count(*) from resource_events"
```
robdaemon: This is going to run the `select count(*)` for every batch. This should be outside of the `call-with-query-rows` like it was originally.

Sharpie: Hmm, doesn't `call-with-query-rows` invoke the `(fn [rows] ...)` once and pass a sequence of rows returned by the query?

The batching bit happens inside the `let` where `rows` is passed to `reduce`.

rbrw: Yeah, unless I'm misreading, I think this should be fine since `rows` is the entire `seq` of rows, and there's only one of those -- now if it were inside the inter reduce...

PR #3075, opened 2019-09-19
(docs) Add documentation for configuring masterless SSL

austb:


austb:
Those docs provide some fairly opinionated recommendations for how to configure masterless Puppet - PuppetDB terminus communication. I'd say it's fairly non-controversial because it recommends strongly that you use Puppet Server CA, but it could use an extra pair of eyes to make sure we're ok with it.

PR #3080, opened 2019-09-24
(maint) Specify Docker volumes explicitly

Iristyle:
- Dockerfile was copying 2 files that were unused. Restore the default
       logging configuration files to write to files in addition to STDOUT
  
  Since PDB was already configured to remove timestamps over STDOUT in the logback.xml, keep it configured as such (Docker also emits timestamps, so dupes are not needed)

- It turns out that LCOW has a bug with permissions on Docker VOLUMEs which is why Postgres Linux container won't start by default. This issue is filed as https://github.com/moby/moby#39922

       For now, our hosts have granted permissions to the group that runs
       all containers (NT VIRTUAL MACHINE\Virtual Machines) so that symlinks
       can be created inside NTFS backed VOLUMEs for Linux containers, as
       is required for Postgres container initialization

- Based on the container configuration, there are 3 separate locations
       where persistent data is located, so explicitly specify those:

  * generated certs (when puppetserver is connected)
  * puppetdb data
  * log files

- This data will be updated to consume only a single VOLUME in a future commit to make VOLUME management simpler and align around the same configuration as PE

 - Reconfigure logs for /opt/puppetlabs/server/data/puppetdb/logs rather than /var/logs/puppetlabs/puppetdb so that we can remove one of the defined Docker VOLUME declarations

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -20,7 +20,6 @@ services:
       - 5432
     volumes:
       - ./puppetdb/postgres-custom:/docker-entrypoint-initdb.d
-      - ${VOLUME_ROOT:-.}/pgdata:/var/lib/postgresql/data
     dns_search: 'test'
     networks:
       puppetdb_test:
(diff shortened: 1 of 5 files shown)


Iristyle:
Since reconfiguring SSL is a bit more complex, I'm going to open another PR to get that VOLUME tied up / consistent with other projects.

Iristyle:
Follow on PR is #3082

PR #3082, opened 2019-09-24
(maint) Update docker ssl volume

Iristyle:
* Reconfigure Docker logging to use standard config file paths
* Add Puppetserver to testing as this will be important for SSLDIR validation
* Remove the call to ssl-setup and replace with shell code
* Configure SSLDIR to use  `/opt/puppetlabs/server/data/puppetdb/certs` (just like pe-puppetdb-extensions)


Iristyle:
a332123 passed in Travis, failed due to transient DNS in Azure

PR #3085, opened 2019-09-26
(maint) Use all hostnames for Docker DNS_ALT_NAME

Iristyle:
- Docker supplies the HOSTNAME value, which may be different from how
   host recognizes its hostname setting (short and fully qualified).

   While these values may be duplicated, that's OK b/c the ssl.sh
   script doesn't care if names are duplicated.

 - NOTE: hostname -s / hostname -f doesn't properly return under LCOW
   like it does on Linux, and always returns the short name


Iristyle:
This now requests alt names like this in Travis:

```
(/ssl.sh) * DNS_ALT_NAMES: 'puppetdb,puppetdb,puppetdb.test,'
```

When previously it requested them like (i.e. in a prior PR like https://travis-ci.com/puppetlabs/puppetdb/jobs/239224509):

```
(/ssl.sh) * DNS_ALT_NAMES: 'puppetdb,'
```

PR #3087, opened 2019-09-26
(maint) Make Docker SSL configuration flexible

Iristyle:
- In some instances, users may be deriving from this container to
   specify well-known cert files (as is the case with Bolt specs).

   In this case, users are setting USE_PUPPETSERVER=false, but still
   desire SSL to be configured in Jetty, based on well known certs
   that have been supplied.

 - Adjust the script for these cases:

   * Only generate certs through puppetserver when no cert files are
     on disk *and* USE_PUPPETSERVER=true

   * When a private key exists at $SSLDIR/private_keys/$CERTNAME.pem:

     * always generate a corresponding pk8
     * always create / overwrite symlinks for Jetty
     * always enable SSL in Jetty
     * always correct ownership / perms

     This will handle scenarios where SSL files have been generated,
     or when they've been supplied by a user (as in the Bolt case)

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/30-configure-ssl.sh
+++ b/docker/puppetdb/docker-entrypoint.d/30-configure-ssl.sh
@@ -1,15 +1,21 @@
 #!/bin/sh
 
+# when certs need to be generated and haven't been user supplied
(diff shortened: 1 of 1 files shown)


Iristyle:
Also note that the Bolt scenario of extending the container should eventually be replaced when we provide more flexible container customization through a `docker-entrypoint.d` or similar folder as a mount.

But for now, this gets the job done!

PR #3090, opened 2019-09-26
(docs) explain docs redirect

jbondpdx:
Adding a README to explain to users why they might have been redirected here, and how they can find what they really need.


jbondpdx:
Hmm, @austb , I admit I was a bit stressed and didn't think of that. But I think it will be OK; it won't be in the ToC, so someone would have to type in that URL intentionally, which I wouldn't really expect.

PR #3091, opened 2019-09-26
(PDB-4502) Add a custom header for uncompressed command size

Zak-Kent:
The 'Content-Length' header isn't sent when commands are compressed in the
terminus. As a result the PDB size metrics aren't updated and the
'max-command-size' configuration option which relies on 'Content-Length'
doesn't function as intended.

This commit adds a custom 'X-Uncompressed-Length' header to requests
sent from the terminus. The value of this header is the size in bytes of
the command before it is compressed. This value is used to update
the size metric and to decide when to discard commands based on
'max-command-size'.

### DIFF
--- a/src/puppetlabs/puppetdb/middleware.clj
+++ b/src/puppetlabs/puppetdb/middleware.clj
@@ -265,15 +265,24 @@
   [app reject-large-commands? max-command-size]
   (fn [req]
     (if (= :post (:request-method req))
-      (let [length-in-bytes (request/content-length req)]
+      (let [length-in-bytes (or (when-let [length (get-in req [:headers "x-uncompressed-length"])]
+                                  (try
+                                    (Long/parseLong length)
+                                    (catch NumberFormatException e
+                                      (log/warn (trs "The X-Uncompressed-Length value {0} cannot be converted to a long."
+                                                     length)))))
+                                (request/content-length req))]
 
-        (when length-in-bytes
-          (log/debug (trs "Processing command with a content-length of {0} bytes" length-in-bytes))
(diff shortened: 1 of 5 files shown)


Zak-Kent:
This fix doesn't guard against clients which may submit data to pdb outside of the terminus but should get the most common use case working again. I'm in favor of documenting the need to include the custom header. People can then include this header if they have their own clients submitting data to pdb and care about the size metric and rejecting large commands. Solving this issue only on the pdb side for all requests is doable but is more complex and this first pass might be good enough.

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -265,7 +265,8 @@
   [app reject-large-commands? max-command-size]
   (fn [req]
     (if (= :post (:request-method req))
-      (let [length-in-bytes (request/content-length req)]
+      (let [length-in-bytes (or (request/content-length req)
+                                (Long. (get-in req [:headers "x-uncompressed-length"])))]
```
austb: This `Long` constructor can throw an error `java.lang.NumberFormatException`. We should probably guard against that, log a warning message about that problem and use a sane default for `length-in-bytes` (0 is fine by me if that's what it was ending up as before).

And so we are extra confident on this piece of potentially user-input, it might be a good idea to additionally verify that `length-in-bytes` is actually an integer and someone isn't trying to send us π bytes.

Zak-Kent: that's a good point, I'll update it. I think it's ok to let `length-in-bytes` be nil here that's what it has been. If we set it to 0 it will throw off the metrics

PR #3097, opened 2019-10-01
(PDB-4509) Throw error for invalid database configuration

austb:
If a puppetdb was applying migrations on a database with `standard_conforming_strings = off` it would previously error with

```
2019-09-18T13:31:10.415-04:00 INFO  [p.p.s.migrate] Applying database migration version 43
2019-09-18T13:31:10.456-04:00 ERROR [p.p.s.migrate] Caught SQLException during migration
java.sql.BatchUpdateException: Batch entry 1 CREATE AGGREGATE sha1_agg (BYTEA)
    (
      sfunc = dual_sha1,
      stype = bytea,
      initcond = '\x00'
    ) was aborted: ERROR: invalid byte sequence for encoding "UTF8": 0x00
```

Now it will recognize the invalid configuration and exit after printing
```
********************
*
* Invalid database configuration settings: 'standard_conforming_strings' (expected 'on', got 'off')
*
********************
```

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -312,6 +312,41 @@
     (future-cancel command-loader))
   context)
 
+;; A map of required postgres settings and their expected value.
+;; PuppetDB will refuse to start unless ALL of these values
+;; match the actual database configuration settings.
+(def required-pg-settings
+  {:standard_conforming_strings "on"})
+
+(defn verify-database-setting
+  "Given m, a map of database settings, verify 'expected', which has the form
+  [setting-name expected-value]."
+  [m [setting value]]
+  (let [map-path [setting :setting]]
+    (when-not (= (get-in m map-path) value)
+      {setting {:expected value :actual (get-in m map-path)}})))
+
+(defn request-database-settings
+  [db-conn-pool]
+  (jdbc/with-db-connection
+    db-conn-pool
+    (jdbc/query "show all;" )))
+
+(defn verify-database-settings
+  "Ensure the database configuration does not have any settings known
+  to break PuppetDB. If an invalid database configuration is found, throws
+  {:kind ::invalid-database-configuration :failed-validation failed-map}"
+  [database-settings]
+  (let [munged-settings (apply merge
+                               (map (fn [h]
+                                      {(keyword (:name h)) (dissoc h :name)})
+                                    database-settings))
+        invalid-settings (remove nil? (map #(verify-database-setting munged-settings %) required-pg-settings))]
+    (when (seq invalid-settings)
+      (throw (ex-info "One or more postgres settings failed validation"
+                      {:kind ::invalid-database-configuration
+                       :failed-validation (apply merge invalid-settings)})))))
+
 (defn initialize-schema
   "Ensures the database is migrated to the latest version, and returns
   true iff any migrations were run.  Throws
@@ -337,7 +372,9 @@
   `write-db-config`.  Blocks until the initialization is complete, and
   then returns an unspecified value.  Throws exceptions on errors.
   Throws {:type ::unsupported-database :current version :oldest
-  version} if the current database is not supported."
+  version} if the current database is not supported. Throws
+  {:kind ::invalid-database-configuration :failed-validation failed-map}
+  if the database contains a disallowed setting."
   [write-db-config config]
   ;; A C-c (SIGINT) will close the pool via the shutdown hook, which
   ;; will then cause the pool connection to throw a (generic)
@@ -356,6 +393,7 @@
                         (.setName "PuppetDB migration pool closer"))]
       (.addShutdownHook runtime on-shutdown)
       (try
+        (verify-database-settings (request-database-settings {:datasource db-pool}))
         (loop [i 0
                last-ex nil]
           (let [result (try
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -312,6 +312,37 @@
...
+  [db-conn-pool]
+  (let [database-settings (jdbc/with-db-connection db-conn-pool (jdbc/query "show all;" ))
+        munged-settings (apply merge
+                                (map (fn [h]
+                                       {(keyword (:name h)) (dissoc h :name)})
+                                     database-settings))
+        invalid-settings (doall (map #(verify-database-setting munged-settings %) required-pg-settings))]
+    (if (seq (filter some? invalid-settings))
+             (throw+ {:type ::invalid-database-configuration :failed-validation
+                      (apply merge invalid-settings)}))))
```
rbrw: Hmm -- might be reasonable to have infrastructure like this from the start, but I offhand, I might have been tempted to just begin with something simpler for this case, i.e.

```
(defn verify-database-settings
  ...
  (let [{:keys[standard_conforming_strings]} settings]
    (when-not (= "on" standard_conforming_strings)
      (throw
        (ex-info "PuppetDB requires the PostgreSQL standard_conforming_strings setting to be \"on\""
                 {:kind ::invalid-database-configuration
                  ...))))
```
Though I suppose that wouldn't make it easy to report multiple errors in one pass.  That could be fixed by just throwing one exception at the end (causing the shutdown), after logging/trs the messages, but it's still not as flexible as what you have.  Not sure -- need to think about it.

rbrw: In any case, I'd argue for avoiding any new slingshot code (since I'd really like to remove it altogether "soon"), i.e. perhaps just be explicit, i.e. throw ex-info like that ^ and then catch ExceptionInfo and use ex-data below.

rbrw: And I'd favor `:kind` rather than `:type` to match our existing precedents.  (Though maybe we use `:type` elsewhere too?)

rbrw: Oh, and I wondered if the setting values might be case sensitive and/or already normalized, i.e. whether "on" is the only value we need to check.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -312,6 +312,37 @@
...
+;; A map of required postgres settings and their expected value.
+;; PuppetDB will refuse to start unless ALL of these values
+;; match the actual database configuration settings.
+(def required-pg-settings
+  {:standard_conforming_strings "on"})
+
+(defn verify-database-setting
+  "Given m, a map of database settings, verify 'expected', which has the form
+  [setting-name expected-value]."
+  [m expected]
```
rbrw: Could just do `[m [setting value :as expected]]` here, or `[[setting value] expected ...` in the let.  Though in the former case, maybe we don't really need `expected` at all, unless we want it just for clarification.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -312,6 +312,37 @@
...
+  to break PuppetDB. If an invalid database configuration is found, throws
+  {:type ::invalid-database-configuration :failed-validation failed-map}"
+  [db-conn-pool]
+  (let [database-settings (jdbc/with-db-connection db-conn-pool (jdbc/query "show all;" ))
+        munged-settings (apply merge
+                                (map (fn [h]
+                                       {(keyword (:name h)) (dissoc h :name)})
+                                     database-settings))
+        invalid-settings (doall (map #(verify-database-setting munged-settings %) required-pg-settings))]
+    (if (seq (filter some? invalid-settings))
```
rbrw: I'd suggest a `when` here -- not critical, but it can help make it clear that this is intentionally "one-armed", and could also be `(remove nil? ...`.  I'd thought about suggesting mapv instead of `(doall (map ...` since it's also eager, but I suspect you don't need either, i.e. if it matters at all, the the merge will force the issue below.

src/puppetlabs/puppetdb/cli/services.clj
austb: `:type` is for slingshot, we use `:kind` for `ex-info` it looks like (but it's not a requirement). I went ahead and switched it over to ex-info. It was easier to get it working by switching the other error over than trying to figure out how to get slingshot and ex-info to play well together inside the same `try+` (also it's just `try` now 🎉 ). So that change is in a second commit.

PostgreSQL does the normalization `"OFF"` and `"ON"` are lowercased when changed.

I went ahead and left it in this flexible form as it'll make it easier to add settings in the future. I imagine we'll at least want to verify `server_encoding` is `UTF8`.

Iristyle:
I believe this PR has resulted in the problem with no longer retrying the `PDBMigrationsPool` connection, but instead failing fast.

Previously:

```
ERROR [p.p.c.services] Will retry database connection after temporary failure:
java.sql.SQLTransientConnectionException: PDBMigrationsPool - Connection is not available, request timed out after 3008ms.
```

New fail-fast behavior:

```
ERROR [p.t.internal] Error during service start!!!
java.sql.SQLTransientConnectionException: PDBMigrationsPool - Connection is not available, request timed out after 3011ms.
```

A pin to a package created just prior to this PR was merged demonstrates that reverting works:
https://github.com/puppetlabs/pe-puppetdb-extensions/pull/499

PR #3104, opened 2019-10-04
(PDB-4342) make node lifetimes configurable

Zak-Kent:


Zak-Kent:
Closing because this has the same functionality of the facts expiration command but under a different name. We were planning on doing this to clean up some of the commands the console folks were using to keep their inventory nodes alive. Due to the reorg not sure we'll get back around to it because the other approach works even if it is awkwardly named.

PR #3105, opened 2019-10-04
(PE-26658) fips profile for pdb testing

robdaemon:
Update FIPS profile to latest known working config. Adds security
policies for JDK8 and JDK11.

Adds a FIPS test also

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ aliases:
     ext/bin/test-config --set pgver "$pgver"
     ext/bin/test-config --set pgport 34335
     ext/bin/check-spec-env "$PDB_TEST"
-    ext/bin/boxed-core-tests -- lein test
+    ext/bin/boxed-core-tests -- lein with-profiles "${LEIN_PROFILES:-dev}" test
     ext/bin/run-external-tests
 
   - &run-integration-tests |
@@ -56,7 +56,7 @@ aliases:
     PDB_TEST_RICH_DATA="$(ext/travisci/spec-includes "$PDB_TEST" rich)"
     export PDB_TEST_RICH_DATA
     ext/bin/check-spec-env "$PDB_TEST"
-    ext/bin/boxed-integration-tests -- lein test :integration
+    ext/bin/boxed-integration-tests -- lein with-profiles "${LEIN_PROFILES:-dev}" test :integration
 
   - &run-spec-tests |
     set -e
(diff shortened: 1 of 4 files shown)


robdaemon:
Do not merge this yet - once tk-jetty9 and clj-http-client changes get merged, we'll spin a new clj-parent that uses them and I can remove the snapshot versions from this PR.

project.clj
```diff
@@ -247,11 +246,32 @@
...
+                     :jvm-opts ~(let [version (System/getProperty "java.version")
+                                      [major minor _] (clojure.string/split version #"\.")
+                                      unsupported-ex (ex-info "Unsupported major Java version. Expects 8 or 11."
+                                                       {:major major
+                                                        :minor minor})]
+                                  (condp = (java.lang.Integer/parseInt major)
+                                    1 (if (= 8 (java.lang.Integer/parseInt minor))
+                                        ["-Djava.security.properties==dev-resources/jdk8-fips-security"]
+                                        (throw unsupported-ex))
+                                    11 ["-Djava.security.properties==dev-resources/jdk11-fips-security"]
```
austb: Do we need to package this security file in ezbake? Or will it ship with another package/be created by the PE module?

.travis.yml
```diff
@@ -86,6 +86,10 @@ jobs:
       env: PDB_TEST=core+ext/openjdk8/pg-9.6
       script: *run-core-and-ext-tests
 
+    - stage: ❧ pdb tests with FIPS
+      env: PDB_TEST=core+ext/openjdk8/pg-9.6 LEIN_PROFILES=FIPS
+      script: *run-core-and-ext-tests
+
```
austb: I think we have very limited usage of SSL in our core tests, this might better test our setup in integration tests (but that relies on being able to run puppetserver and Puppet in FIPs mode too. And our certs over in `test-resources/` might not be adequate for FIPs testing, so running integration tests in FIPS mode should not hold up this PR and is probably an idea for post-Lovejoy

project.clj
robdaemon: no, that's handled by puppetlabs/bouncy-castle-vanagon#4 and puppetlabs/puppet-enterprise-modules#673

.travis.yml
robdaemon: I'd like to keep at least one set of unit tests that run in this mode, just as an early warning system. puppetserver had unit tests that failed quickly in FIPS mode. Adding integration tests too is a great idea.

Our certs should work - they're read as PEM files and a FIPS keystore is created in memory. I can run puppetdb in FIPS mode with a pgbox set of certs.

PR #3111, opened 2019-10-11
(PDB-4547) Add wireformat validation for deactivate node cmds

Zak-Kent:
I can handle merge ups when this goes in. It's going to conflict with the other pr against 6.3.x

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -58,6 +58,7 @@
             [puppetlabs.puppetdb.catalogs :as cat]
             [puppetlabs.puppetdb.reports :as report]
             [puppetlabs.puppetdb.facts :as fact]
+            [puppetlabs.puppetdb.nodes :as nodes]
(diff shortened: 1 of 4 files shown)


src/puppetlabs/puppetdb/command.clj
```diff
@@ -363,12 +364,15 @@
     (log-command-processed-messsage id received start-time :deactivate-node certname)))
 
 (defn deactivate-node [{:keys [payload version] :as command} start-time db]
-  (-> command
-      (assoc :payload (case version
-                        1 (deactivate-node-wire-v1->wire-3 payload)
-                        2 (deactivate-node-wire-v2->wire-3 payload)
-                        payload))
-      (deactivate-node* start-time db)))
+  (let [validated-payload (upon-error-throw-fatality
```
austb: This makes the `upon-error-throw-fatality` check in `deactivate-node-wire-v1->wire-3` redundant, so that can probably get removed

Zak-Kent: updated

PR #3117, opened 2019-10-15
(PDB-3300) Add admin delete node command

Zak-Kent:
This command allows for the immediate deletion of all data associated with the
certname in the command payload.

### DIFF
--- a/documentation/api/admin/v1/cmd.markdown
+++ b/documentation/api/admin/v1/cmd.markdown
@@ -9,13 +9,12 @@ canonical: "/puppetdb/latest/api/admin/v1/cmd.html"
 
 
 The `/cmd` endpoint can be used to trigger PuppetDB maintenance
-operations.  Only one maintenance operation can be running at a time.
-Any request received while an operation is already in progress will
-return an HTTP conflict status (409).
+operations or to directly delete a node.  Admin commands are processed
+synchronously seperate from other PuppetDB commands.
 
 ## `POST /pdb/admin/v1/cmd`
 
-The maintenance operations must be triggered by a POST.
+Admin commands must be triggered by a POST.
 
 ### Request format
 
@@ -23,7 +22,32 @@ The POST request should specify `Content-Type: application/json` and
 the request body should look like this:
 
 ``` json
-{"version" : 1, "payload" : [REQUESTED_OPERATION, ...]}
+{"command": "...",
+ "version": 123,
+ "payload": <json object>}
+```
+
+`command` is a string identifying the command.
+
+`version` is a JSON integer describing what version of the given
+command you're attempting to invoke. The version of the command
+also indicates the version of the wire format to use for the command.
+
+`payload` must be a valid JSON object of any sort. It's up to an
+individual handler function to determine how to interpret that object.
+
+### URL parameters
+
+* The POST endpoint accepts no URL parameters.
+
+## List of admin commands
+
+## "clean", version 1
+
+``` json
+{"command" : "clean",
+ "version" : 1,
+ "payload" : [REQUESTED_OPERATION, ...]}
 ```
 
 where valid `REQUESTED_OPERATION`s are `"expire_nodes"`,
@@ -41,10 +65,6 @@ will be the [`node-purge-gc-batch-limit`][config-purge-limit].
 
 An empty payload vector requests all maintenance operations.
 
-### URL parameters
-
-* The POST endpoint accepts no URL parameters.
-
 ### Response format
 
 The response type will be `application/json`, and upon success will
@@ -54,8 +74,9 @@ include this JSON map:
 {"ok": true}
 ```
 
-If any other maintenance operation is already in progress the HTTP
-response status will be 409 (conflict), will include a map like this
+Only one maintenance operation can be running at a time.  If any other
+maintenance operation is already in progress the HTTP response status will be
+409 (conflict), will include a map like this
 
 ``` json
 {"kind": "conflict",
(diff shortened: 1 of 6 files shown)


robdaemon:
In the `scf/storage.clj` file (not part of this PR), we may want to extend `delete-certname!` to delete the appropriate reports and resource-events also. Right now, that's going to depend on cascades, but that's not as efficient as something like (note this is written off the top of my head, may not be correct as far as column names):

```sql
DELETE FROM resource_events WHERE report_id IN (SELECT id FROM reports WHERE certname = 'foo');
DELETE FROM FROM reports WHERE certname = 'foo';
```

Zak-Kent:
👍 makes sense @robdaemon, I'll take a look

documentation/api/admin/v1/delete.markdown
```diff
@@ -0,0 +1,54 @@
...
+
+The delete operation must be triggered by a POST.
+
+### Request format
+
+The POST request should specify `Content-Type: application/json` and
+the request body should look like this:
+
+``` json
+{"version" : 1, "command" : "delete", "payload" : {"certname" : CERTNAME_TO_DELETE}}
```
austb:
```suggestion
{"version" : 1, "command" : "delete", "payload" : {"certname" : "<CERTNAME_TO_DELETE>"}}
```
We use that format elsewhere (ast docs, etc) to signify, it's a string and user supplied.

austb: The other style we use in command documentation is
```suggestion
{"version" : 1, "command" : "delete", "payload" : {"certname" : <string>}}
```

documentation/api/admin/v1/delete.markdown
```diff
@@ -0,0 +1,54 @@
...
+
+* The POST endpoint accepts no URL parameters.
+
+### Response format
+
+The response type will be `application/json`, and upon success will
+include this JSON map:
+
+``` json
+{"ok": true}
```
austb: One other concern I have about the wire format and this response, is both implied to me that its a command that gets enqueued on the command queue. I would then assume if I just stopped the puppet-agent daemon on that machine, and then sent this command, it would be processed after all the previous commands were processed from the last puppet run, but that might not be the case if I'm running any queue depth.

I'm just throwing things out wildly, but we could start with a simpler wire format
```
{"version": 1, "certname": <string>}
```
which would allow us to take up that more complicated wire format in v2 if we ever wanted to treat it as a command.

If we simplify the wire format and don't enqueue it on the command queue, it'd be nice to return a bit of information to them, ie. the JSON representation of "deleted 1 certname, dead-server.company.net in 747ms" since it won't be part of the command metrics.

austb: On the other side, instead of changing the wire format, we could do what the wire format implies and enqueue it on the command queue (this wouldn't be a command we need to sync we can just document that it is ephemeral). This style would make it easier for someone to write a Bolt plan to, first go out and disable the agent service on all the machines they want to delete and then enqueue the delete commands.

documentation/api/admin/v1/delete.markdown
Zak-Kent: 👍 I'm cool with the second one

documentation/api/admin/v1/delete.markdown
Zak-Kent: The other admin /cmd endpoint uses a similar wire format so I was trying to stay consistent with that. I was thinking that having the :payload key would allow us to include keys for more targeted deletions (say just factsets/reports or a vector of certnames) later on, but you're right we could just rev the endpoint and add it when we decide we want/need it.  

I agree with the part about returning more detailed information about what we deleted in the response. I'll update that. 

Also an interesting idea about submitting the command to the queue, that would account for some of the edge cases if a command for the certname was in the queue already, etc. The ticket seems like it was asking for a way to immediately delete data without having to hit the db with a sql statement directly. I'm kinda leaning towards having it delete the thing directly, though we could ask around and see what the folks using it would prefer.

austb: Ah I see, that's a general endpoint `/pdb/admin/v1/cmd` for they need to specify the specific command `"expire_nodes", "purge_nodes", "purge_reports"` etc. Maybe this command belongs on that cmd endpoint then? I don't like the response from that one either, but it does provide the protection of
> Only one maintenance operation can be running at a time. Any request received while an operation is already in progress will return an HTTP conflict status (409).

If we leave it on this endpoint I'd suggest at least dropping the key `"command" : "delete"`  because the url `/admin/v1/delete` specifies the command is "delete". If we add other entities to delete in the future we'll have to rev the cmd version no matter what and can add it back in then.

Zak-Kent: that's a good idea, I'll move it to the general `/cmd` endpoint

rbrw: Hmm, yeah, I hadn't really thought about it hard yet, but I think I'd probably been vaguely imagining it might just be a `/cmd` operation, but looking back at how `/cmd` now works, I suppose you could argue that might add complexity (including perhaps wrt careful error handling) without sufficient justification (i.e. if you can mix this in with the other commands).  

On the other hand if it's "sufficiently similar" it might be fine, and easily share the infrastructure.  It also crossed my mind that at some point "delete by query" might be a natural extension if/when we ever need it, and that would be more complicated.

I suppose a slight variation on what we have here would be a `admin/v1/delete-nodes` which would only need the certname(s) specification(s) in the body, and could also support (and I suppose could even just start with) one of our query-style shortcut urls, i.e. `admin/v1/delete-nodes/CERTNAME.  Then we wouldn't need any command body to start with.

...of course people might well suggest that this should be a an http `DELETE` on something like `.../node/CERTNAME`, and likely wouldn't be happy with a `GET` or `POST`.

Zak-Kent: How important do y'all feel locking the `"delete"` command down to one request at a time is? 

I just realized we probably don't want to use the existing `clean-lock` that gates admin `"clean"` commands because the lock also blocks during normally scheduled gc operations which could block `"delete"` commands when we didn't intend to. 

If we do think it's important to limit the amount of concurrent `"delete"` commands we process we can gate it behind a lock of its own or a a semaphore if we want to let a few go at once, etc.

PR #3118, opened 2019-10-15
(PDB-4506) Make migrations interruptable via C-c (SIGINT)

rbrw:


robdaemon:
I'm noticing we only have two migrations that return a `:vacuum-analyze` map, do any need that added to them? I don't think anyone's likely to notice the change in the field :(

rbrw:
Not sure -- we added all this because we definitely wanted/needed it for the old facts storage migrations, and we probably do want any newly created tables to be analyzed "soon", but any that receive much change will be auto-analyzed eventually.  It was just the small, mostly static tables that were a serious problem.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1656,26 +1643,46 @@
...
+        (analyze small-tables)
         false))))
 
+(defn migrate! [db-connection-pool]
+  "Migrates database to the latest schema version. Does nothing if
+  database is already at the latest schema version.  Requires a
+  connection pool because some operations may require an indepdendent
+  database connection.
+
+  This function will return true iff any migrations were run."
```
robdaemon: s/iff/if/

rbrw: Hah, oh, that was the "if and only if" (math) abbreviation, which I tend to use, but perhaps (and possibly to your point) I shouldn't here.

PR #3127, opened 2019-10-24
(PDB-4255) fix deprecation warning from i18n

robdaemon:
Update i18n plugin to remove deprecation warning about hook

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -20,7 +20,7 @@
    :password :env/nexus_jenkins_password
    :sign-releases false})
 
-(def i18n-version "0.8.0")
+(def i18n-version "0.9.0")
 
 (def need-permgen?
   (= "1.7" (System/getProperty "java.specification.version")))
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -218,6 +218,8 @@
             [lein-parent "0.3.7"]
             [puppetlabs/i18n ~i18n-version]]
 
+  :prep-tasks [["i18n" "make"] "compile" "javac"]
```
austb: I don't recall how long the i18n make takes, but we have a CI job that will automatically update the POT file, we might save ourselves some duplication of effort and compilation time if we omit this and just let CI keep doing what it does.

robdaemon: It's near instantaneous, because the files it creates are untouched.

I made the change like this to not remove any existing functionality.

austb: Ah ok, yep I took a look at clj-i18n, that's not the actual command that generates the POT file

PR #3128, opened 2019-10-24
(PDB-4255) remove taoensso warnings

robdaemon:
taoensso/nippy was only used in benchmark - it's not a crucial part of
the benchmarking process, so replaced it with built-in EDN reader and
writer.

This will clear this warning at startup:

    WARNING: pos-int? already refers to: #'clojure.core/pos-int? in namespace: taoensso.encore, being replaced by: #'taoensso.encore/pos-int?
    WARNING: bytes? already refers to: #'clojure.core/bytes? in namespace: taoensso.encore, being replaced by: #'taoensso.encore/bytes?

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -167,9 +167,8 @@
                  ;; Various
                  [cheshire]
                  [clj-stacktrace]
-                 [clj-time]
+                 [clj-time "0.15.2"]
                  [com.rpl/specter "0.5.7"]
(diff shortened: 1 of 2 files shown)


robdaemon:
The use of nippy is old, old, old. I sort of lose track of it in git history. Judging by the message, this warning has likely been there a long time.

robdaemon:
hold off on merging for a bit, I wanna do a bit more archaeology.

PR #3129, opened 2019-10-25
(PDB-4255) cleaning up warnings

robdaemon:
Cleared up illegal reflective access warnings - these can occur
currently on uses of (with-open) because the underlying channel
implementation in JDK11 hits an internal class. If you annotate these
uses with ^Closeable, the warnings stop.

    WARNING: Illegal reflective access by clojure.lang.InjectedInvoker/0x00000008401e8040 (file:/home/rroland/.m2/repository/org/clojure/clojure/1.10.1/clojure-1.10.1.jar) to method sun.nio.ch.ChannelInputStream.close()
    WARNING: Please consider reporting this to the maintainers of clojure.lang.InjectedInvoker/0x00000008401e8040
    WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
    WARNING: All illegal access operations will be denied in a future release

Cleared up performance reflection warnings also after enabling
*warn-on-reflection!* in lein:

    Performance warning, puppetlabs/puppetdb/queue.clj:269:3 - hash collision of some case test constants; if selected, those entries will be tested sequentially.


rbrw:
Hmm, glad you tracked this down -- I wonder if this is work on an upstream fix: https://clojure.atlassian.net/browse/CLJ-2066

In any case, at least for the `with-open` calls, I'm not sure what I think yet, but I wondered about just fixing `with-open`, so that if/when the problem is fixed upstream, we can just drop our version without having had to change the code.  For example, say we put something like this in puppetlabs.puppetdb.utils:

```
(defmacro with-open
  [bindings & body]
  (assert (vector? bindings))
  (if (empty? bindings)
    `(do ~@body)
    (do (assert (even? (count bindings)))
        (let [bindings (vec (mapcat (fn [[var val]] [var ^Closeable val])
                                    (partition 2 bindings)))]
          `(clojure.core/with-open ~bindings ~@body)))))
```

and then in the namespaces where we use it:

```
(ns ...
  (:refer-clojure :exclude (with-open))
  ...
  (:require
   [puppetlabs.puppetdb.utils :refer [with-open]]))
```

PR #3132, opened 2019-10-27
(maint) Setup Docker PuppetDB LOGDIR / add logging

Iristyle:
- Configure open source PDB similarly to pe-puppetdb, to:

   * be consistent between environments
   * use LOGDIR env var like other containers
   * make a transition from open source to PE container easier

 - Add the LCOW volume fix script as well


Iristyle:
Azure didn't run its check for some reason...

PR #3134, opened 2019-10-27
(maint) Re-enable Azure PRs on 6.3.x branch

Iristyle:
- Undoes #2861 / 39f6bbbd467d0970140a7cddcdf9d42ebdb64d06 which disabled
   builds on this branch.

   Not sure why we did that at the time, but since we're still shiping
   2019.1.2 off of 6.3 and other PRs have targeted the 6.3 branch,
   let's turn Azure back on for pull requests


Iristyle:
Additionally verified that this will merge up cleanly to master.

Merging!

PR #3133, opened 2019-10-27
(maint) Emit docker build command line

Iristyle:

PR #3135, opened 2019-10-28
(maint) Catch 6.3 Docker assets up to master

Iristyle:
This isn't pretty.

The @puppetlabs/pupperware team made the mistake of not targeting many Docker-related at PRs at the 6.3.x branch, and instead targeted master. This stemmed from a fundamental misunderstanding of what version of PDB was shipping in the 2019.1.* PE packages.

This creates at least a minor issue in making sure that 6.3 open source containers are configured the same way as PE 2019.1.* containers (necessary for being able to migrate from open source to PE).

Since the water is under the bridge, there's only a couple of ways of syncing the branches up:

* Cherry-pick over all the relevant commits from master to a new branch that can be merged into 6.3.x branch (which is what this does). The commits don't maintain their original SHA identities, but at least there's some semblance of history (even if it ends up being not quite the same as master).

* Copy all the relevant docker files over from master to a single commit targeted at 6.3.x. This loses all history, but makes the commit simpler.

I've opted for something that looks like history - unless there are any objections, it seems like the right way to go.

In both cases, merging the resulting branch up to master is intended be clean w/out any need to resolve conflicts.


Iristyle:
This looks good to go! I also tested the merge locally and there were 3 conflicts based on how the history was laid down (and because 6.3.x isn't currently merged up to master so doesn't have #3132 or #3133 yet), but there's no unexpected changes. I'll put up the PR for the merge-up momentarily so @puppetlabs/puppetdb team doesn't have to deal with it.

PR #3137, opened 2019-10-29
(PDB-4552) Add support for checking x-pdb-sync-ver header

Zak-Kent:
This commit adds checking for the x-pdb-sync-ver header which is sent from
all pe-puppetdb sync requests. If there is a mismatch detected the incoming
sync request will receive a 409 conflict response until it is upgraded and
starts emitting a matching sync version. This is needed during pe-puppetdb
ha-sync to detect and avoid syncing with a pdb that has been recently upgraded
in a way that's not backwards compatible.

### DIFF
--- a/src/puppetlabs/puppetdb/middleware.clj
+++ b/src/puppetlabs/puppetdb/middleware.clj
@@ -25,7 +25,8 @@
             [bidi.schema :as bidi-schema]
             [puppetlabs.puppetdb.schema :as pls]
             [schema.core :as s]
-            [puppetlabs.puppetdb.command :as cmd]))
+            [puppetlabs.puppetdb.command :as cmd]
+            [puppetlabs.puppetdb.constants :as constants]))
 
 (def handler-schema (s/=> s/Any {s/Any s/Any}))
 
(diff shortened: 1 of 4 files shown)


src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -298,6 +299,40 @@
...
+   present but improperly formatted returns a 400 error response."
+  [app]
+  (fn [{:keys [headers] :as req}]
+    (let [maybe-sync-ver (when-let [v (get headers "x-pdb-sync-ver")]
+                           (try
+                             (Integer/parseInt v)
+                             (catch NumberFormatException e
+                               {:error true :input v})))]
+      (cond
+        (nil? maybe-sync-ver)
```
rbrw: Hmm, wondered about rearranging this a bit, pulling up the common case, checking for the "completely unexpected" case, etc.
```
(cond
  (not maybe-sync-ver) (app req)

  (integer? maybe-sync-ver)
  (if (= maybe-sync-ver (get-sync-ver))
    (app req)
    (error-response...))

  (:error maybe-sync-ver)
  (error-response ...)

  :else something-is-terribly-wrong)
```

src/puppetlabs/puppetdb/middleware.clj
```diff
@@ -298,6 +299,40 @@
...
+
+(defn verify-sync-version
+  "Check that the x-pdb-sync-ver header of an incoming requests matches the
+   pdb-sync-ver locally. This header is only sent from pe-puppetdb sync requests
+   and a mismatch indicates that one pdb has upgraded in an incompatible way
+   before the other. In this case all sync requests will return 409s until both
+   sides have been upgraded and have matching sync versions. If the header is
+   present but improperly formatted returns a 400 error response."
+  [app]
+  (fn [{:keys [headers] :as req}]
```
rbrw: Not sure, but contemplated using :strs, i.e. `{{:strs [x-pdb-sync-ver]} :headers}` and and then could have something like `(some-> x-pdb-sync-ver Integer/parseInt)` in the `try` instead of the `when-let/get`.

test/puppetlabs/puppetdb/middleware_test.clj
```diff
@@ -247,3 +247,42 @@
                        :query-params {"foo" "bar"}}]
           (is (= post-req (middleware-fn post-req)))
           (is (= get-req (middleware-fn get-req))))))))
+
+(deftest verify-sync-version-test
```
rbrw: Have some questions about the tests, but think I'll chat with you about those independently.

src/puppetlabs/puppetdb/middleware.clj
Zak-Kent: yep, that looks good. I'll update this

src/puppetlabs/puppetdb/middleware.clj
Zak-Kent: makes sense, I'll adjust

PR #3141, opened 2019-10-30
(PDB-4255) update taoensso libs to remove warnings

robdaemon:
Update taoensso/nippy and taoensso/encore to remove warnings:

    WARNING: pos-int? already refers to: #'clojure.core/pos-int? in namespace: taoensso.encore, being replaced by: #'taoensso.encore/pos-int?
    WARNING: bytes? already refers to: #'clojure.core/bytes? in namespace: taoensso.encore, being replaced by: #'taoensso.encore/bytes?

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,5 @@
 (def pdb-version "6.7.4-SNAPSHOT")
-(def clj-parent-version "4.2.6")
+(def clj-parent-version "4.2.10")
 
 (defn true-in-env? [x]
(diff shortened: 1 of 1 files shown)


rbrw:
No strong opinion, but wondered about putting at least the shared bit into clj-parent instead -- good/bad idea?  (Only wondering if that might mean we can avoid having to remember to undo this later if/when it's updated there.)

robdaemon:
as soon as we merge puppetlabs/clj-parent#219, I'll redo this change to just update clj-parent

PR #3142, opened 2019-10-31
(PDB-4577) fix tests for pgsql 12

robdaemon:
Apply an ordering to the indexes-sql results, so that all versions of
postgresql return this the same.

pgsql 12 has a different default ordering than 9.x and 11.


rbrw:
Looks like this needs a rebase/resolution -- other than that, seems ready to go.

robdaemon:
rebased, and added a pg12 test to .travis.yml

PR #3147, opened 2019-11-07
(PDB-4579) fix connection tests

robdaemon:
PostgreSQL JDBC driver has a bug where the isValid() call on a
connection does not respect timeouts. This causes our connection pool to
not detect a dead connection, which will eventually exhaust our
connection pool entirely. See pgjdbc/pgjdbc#1557


robdaemon:
This is targeting 6.3.x because 5.2.x uses a much older JDBC driver that doesn't support the isValid call at all.

robdaemon:
rebased and updated to use a newer clj-parent that has this fix

PR #3148, opened 2019-11-07
(PDB-4579) fix connection tests

robdaemon:
PostgreSQL JDBC driver has a bug where the isValid() call on a
connection does not respect timeouts. This causes our connection pool to
not detect a dead connection, which will eventually exhaust our
connection pool entirely. See pgjdbc/pgjdbc#1557

This fix is specific to PuppetDB 5.2.x and should not be applied to
newer branches.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -197,7 +197,7 @@
                     ;; Database connectivity
                     [com.zaxxer/HikariCP]
                     [honeysql "0.6.3"]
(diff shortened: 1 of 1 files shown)


Zak-Kent:
Just rekicked the tests travis took a while so not sure if the failures were real

robdaemon:
Okay switched this up - instead of using a connection test query, this is using the same JDBC driver as 6.3.x and beyond, but the version that supports JRE 7.

PR #3149, opened 2019-11-11
Test merge master

austb:


project.clj
```diff
@@ -160,6 +159,9 @@
                  [puppetlabs/structured-logging]
                  [puppetlabs/tools.namespace "0.2.4.1"]
                  [puppetlabs/trapperkeeper]
+                 ;; trapperkeeper began pulling in clj-parent's 1.5.3
+                 ;; which is not compatible with java 11
+                 [org.flatland/ordered "1.5.7"]
```
austb: This is probably not quite right for master? I'll double check

robdaemon:
terminus commit looks reasonable to me :+1:

PR #3150, opened 2019-11-14
(PDB-3902) Add null? operator support for inventory query 

Zak-Kent:


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1969,8 +1972,21 @@
...
             [["null?" column-name value]]
-            (let [cinfo (get-in query-rec [:projections column-name])]
-              (map->NullExpression {:column cinfo :null? value}))
+            (let [maybe-dotted-path (str/split column-name #"\.")
+                  cinfo (get-in query-rec [:projections (first maybe-dotted-path)])]
+
+              ;; only matches inventory queries with a '(fs.stable||fs.volatile)' :field value
+              ;; that have a dotted path query like: facts.foo.bar
+              (if (and (= :queryable-json (:type cinfo))
+                       (h/raw? (:field cinfo))
```
austb: Does this introduce the requirement that queryable json fields must be raw or was that already in place?

I worry that the raw check will lead to us re-introducing this bug if we try to make a non-raw field queryable json.

src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -1274,8 +1274,11 @@
   NullExpression
   (-plan->sql [{:keys [column] :as expr}]
     (s/validate column-schema column)
-    (let [lhs (-plan->sql (:field column))
-          json? (= :jsonb-scalar (:type column))]
+    (let [queryable-json? (= :queryable-json (:type column))
+          lhs (if queryable-json?
+                (-> column :field h/sqlraw->str) ; jsonb-null? expects field to be a str
```
austb: This is throwing a stack track because `:field` is a keyword for resource parameters.

```
curl -X POST http://localhost:8080/pdb/query/v4   -H 'Content-Type:application/json'   -d '{"query": "resources[] { type = \"Class\" and parameters.lib_path is null}" }'
```

You might want a helper function like
```
(defn queryable-json-field->str [{:keys [field field-type]}]
  (case field-type
    :keyword (name field)
    :raw (:s field)
      (tru "Cannot determine field type for field ''{0}''" field)))))]
```

I pulled that out of my PR for fact projections, it relies on adding a key called `:field-type` to queryable-json to identify how to stringify it.
https://github.com/puppetlabs/puppetdb/pull/3055/files#diff-fefe75691a2bbea085773295ea2b7c49R1811-R1816

austb:
```suggestion
               (name (extract-sql (:field column))
```
this will take both keywords and hsql raw to string values.

PR #3151, opened 2019-11-18
(PDB-4590) store reports with no events properly

robdaemon:
A report with no events does not get the latest_report pointer in its
certname entry set. This occurred because of a nesting error.

The schema for reports does not allow an empty set of events, but that
was only used at testing time. In production, this happens frequently
(due to truncation of the resource events table, and other conditions.)

### DIFF
--- a/src/puppetlabs/puppetdb/reports.clj
+++ b/src/puppetlabs/puppetdb/reports.clj
@@ -187,7 +187,7 @@
                      :corrective_change))))
 
 (pls/defn-validated munge-resource-events-for-v8
-  [resource-events :- resource-events-expanded-query-schema]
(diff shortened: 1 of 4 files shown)


src/puppetlabs/puppetdb/reports.clj
```diff
@@ -187,7 +187,7 @@
                      :corrective_change))))
 
 (pls/defn-validated munge-resource-events-for-v8
-  [resource-events :- resource-events-expanded-query-schema]
+  [resource-events :- (s/maybe resource-events-expanded-query-schema)]
```
robdaemon: As we learned with this defect, this can be nil in production!

PR #3154, opened 2019-11-22
Test merge master

robdaemon:
[facebones ignore]

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage.clj
+++ b/src/puppetlabs/puppetdb/scf/storage.clj
@@ -220,6 +220,10 @@
   "Delete the given host from the db"
   [certname]
   {:pre [certname]}
+  ;; With partitioning, we must execute this delete on every active partition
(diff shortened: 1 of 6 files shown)


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -220,6 +220,10 @@
   "Delete the given host from the db"
   [certname]
   {:pre [certname]}
+  ;; With partitioning, we must execute this delete on every active partition
+  (doseq [table (jdbc/query-to-vec "select tablename from pg_tables where tablename like 'resource_events_%'")]
+    (jdbc/delete! (:tablename table) ["certname_id in (select id from certnames where certname=?)" certname]))
```
robdaemon: I added this so that we delete within the partitioned tables appropriately (this is a good candidate for making a generic function / macro like `with-partitions`)

PR #3156, opened 2019-11-26
Updates for puppetdb container versioning

underscorgan:
Moving forward we want to change from tagging containers built from source with the latest tagged version of puppetdb to only tagging containers built from released packages with a version number. We will still produce builds from source, but those containers will be tagged `edge`.

This also includes some updates to make the build process consistent between puppetdb and puppetserver source containers.

See also https://github.com/puppetlabs/puppetserver/pull/2188


nwolfe:
Tiny little thing, but I noticed that if we name it `Dockerfile-release` it's not recognized as a Dockerfile (by GitHub, IDEs, etc), which means we don't get syntax highlighting and certain IDE features.

If we instead name files like `foo.Dockerfile` they *are* recognized properly!

So this new one would be `release.Dockerfile` if we wanted to follow that.

Iristyle:
Yeah, I agree with the bit about the file extension - `release.Dockerfile` would be good here if possible

docker/puppetdb/Dockerfile
```diff
@@ -1,57 +1,32 @@
-FROM clojure:openjdk-8-lein AS builder
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends make
-# Install only dependencies
-WORKDIR /app
-COPY project.clj /app/
-COPY resources/puppetlabs/puppetdb/bootstrap.cfg \
```
Iristyle: Looks like `bootstrap.cfg` might have been missed?

```
20 Execution error (IllegalArgumentException) at sun.reflect.NativeConstructorAccessorImpl/newInstance0 (NativeConstructorAccessorImpl.java:-2).
20 Problem loading service '../../puppetlabs/puppetdb/bootstrap.cfg' from /etc/puppetlabs/puppetdb/bootstrap.cfg:1:
20 Invalid line in bootstrap config file:
20 
20 	../../puppetlabs/puppetdb/bootstrap.cfg
20 
20 All lines must be of the form: '<namespace>/<service-fn-name>'.
```

underscorgan: hmm, we shouldn't need to include that separately since this is building a puppetdb package from source now (and it's working on non-windows), will hop onto the azure builder to investigate

docker/puppetdb/Dockerfile
```diff
@@ -65,48 +40,25 @@ LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
...
-CMD ["services"]
-
-COPY docker/puppetdb/healthcheck.sh /
-RUN chmod +x /healthcheck.sh
-# The start-period describes how long it takes PuppetDB to come
-# up in the worst case (LCOW). The other timing parameters are set so that it
-# takes at most 2 minutes to realize that PuppetDB has failed.
-# Probe failure during --start-period will not be counted towards the maximum number of retries
-HEALTHCHECK --start-period=3m --interval=10s --timeout=10s --retries=12 CMD ["/healthcheck.sh"]
+COPY resources/puppetlabs/puppetdb/bootstrap.cfg /etc/puppetlabs/puppetdb/bootstrap.cfg
```
Iristyle: This was added for LCOW per the commit message... but the behavior seems to be different between Travis and Azure - we should figure out why.

docker/puppetdb/Dockerfile
underscorgan: turns out `resources/ext/config/bootstrap.cfg` is a symlink point to `resources/puppetlabs/puppetdb/bootstrap.cfg`, which is why it was failing on windows. Updated the `COPY`s below to copy over `bootstrap.cfg` explicitly

docker/puppetdb/Dockerfile
underscorgan: This was also related to the symlinks in the git repo causing issues on windows. Should be resolved by updated `COPY`ing

underscorgan:
@Iristyle this is working on windows now! Since this is going in after the dockerfile improvements you were working on, can you make sure everything looks reasonable? Tried to group / reorder things where possible.

PR #3161, opened 2019-12-06
(PDB-4606) partition the reports table

robdaemon:
Partition the reports table on producer_timestamp. This creates
a partition per day, and allows garbage collection to occur by
dropping expired tables. This follows the same pattern as PDB-4420.

The report hash has been changed to remove resource_events from the
hash calculation. The resource_events table can be garbage collected
separately from the reports table, and if the hash is still including
the resource_events, we end up with duplicate reports during a pdb
sync. The hash is rewritten during the migration.

There is no trigger to direct inserts to the appropriate partition.
This is because the usage of the reports table relies on the behavior
of a RETURNING clause on the INSERT INTO statement. You cannot rewrite
the query and combine it with a RETURNING clause. All INSERTs occur
directly into the appropriate partition.

During garbage collection, an UPDATE statement is ran across the
certnames table to ensure latest_report_id is set to null when the
report is deleted. Previously, this relied on a cascade during the
delete.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1590,21 +1590,20 @@
    "DROP AGGREGATE IF EXISTS md5_agg(BYTEA)"
    "DROP FUNCTION IF EXISTS dual_md5(BYTEA, BYTEA)"))
 
-(defn autovacuum-vacuum-scale-factor-factsets-catalogs-certnames-reports []
+(defn autovacuum-vacuum-scale-factor-factsets-catalogs-certnames []
   (jdbc/do-commands
    "ALTER TABLE factsets  SET ( autovacuum_vacuum_scale_factor=0.80 )"
    "ALTER TABLE catalogs  SET ( autovacuum_vacuum_scale_factor=0.75 )"
-   "ALTER TABLE certnames SET ( autovacuum_vacuum_scale_factor=0.75 )"
-   "ALTER TABLE reports   SET ( autovacuum_vacuum_scale_factor=0.01 )"))
+   "ALTER TABLE certnames SET ( autovacuum_vacuum_scale_factor=0.75 )"))
 
 ;; for testing via with-redefs
 ;; used to account for the possibility of the name column being added in the
 ;; original version of migration 69 before a user has applied migration 73
 (defn migration-69-stub [])
 
-(defn reporting-partitioned-tables
+(defn resource-events-partitioning
   ([]
-   (reporting-partitioned-tables 500))
+   (resource-events-partitioning 500))
   ([batch-size]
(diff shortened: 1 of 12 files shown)


robdaemon:
I've rebased this and applied the fix from resource-events partitioning to reports.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1590,21 +1590,20 @@
    "DROP AGGREGATE IF EXISTS md5_agg(BYTEA)"
    "DROP FUNCTION IF EXISTS dual_md5(BYTEA, BYTEA)"))
 
-(defn autovacuum-vacuum-scale-factor-factsets-catalogs-certnames-reports []
+(defn autovacuum-vacuum-scale-factor-factsets-catalogs-certnames []
   (jdbc/do-commands
    "ALTER TABLE factsets  SET ( autovacuum_vacuum_scale_factor=0.80 )"
    "ALTER TABLE catalogs  SET ( autovacuum_vacuum_scale_factor=0.75 )"
-   "ALTER TABLE certnames SET ( autovacuum_vacuum_scale_factor=0.75 )"
-   "ALTER TABLE reports   SET ( autovacuum_vacuum_scale_factor=0.01 )"))
```
npwalker: Usually we wouldn't remove anything from a past migration... what's the benefit since the partitioning migration will make a new table I don't think this customization will apply to the new reports table?

robdaemon: The test for this ensures that at the end of all the migrations, the autovacuum scale factors are applied. With the change to the reports table via partitioning, that value has no effect on the reports table. The statement was removed and the test was updated accordingly.

We have a few other places where we've removed previous migrations or made changes in them because applying them is costly if they're superseded by later migrations.

test/puppetlabs/puppetdb/integration/reports.clj
```diff
@@ -237,8 +237,8 @@
             logs  (get-href pdb (get-in report [:logs :href]))]
 
         (is (= #{{:name "total", :value 1, :category "events"}
-                 {:name "changed", :value 1, :category "resources"}
-                 {:name "total", :value 1, :category "changes"}
+                 {:name "changed", :value 0, :category "resources"}
+                 {:name "total", :value 0, :category "changes"}
```
austb: @robdaemon I'm about ready to merge this PR, but I can't quite figure out what caused the numbers in this file to change. Do you know/recall why this was necessary?

robdaemon: I changed some of the fixtures, so that changed these values. I'd have to dig through it to find the specific ones

austb:
Force pushed to rebase

PR #3197, opened 2020-02-14
(Merge on 18 Feb) PDB 6.9.0 release notes

clairecadman:
This commit adds a bug fix note for the PDB 6.9.0 release.


austb:
The team didn't have all the Jira fix versions set before stop ship so we ended up having to merge release notes for 6.9.0 in #3198

Sorry about that we will have our fix versions set next time.

PR #3209, opened 2020-03-12
(maint) Fix testutil that clears db

austb:
If clear-db-for-testing! is called on a migrated db, it will attempt to
delete the sha1_agg aggregate function via the drop-function! testutils
method. Because aggregates and function are deleted differently (ie.
"DROP AGGREGATE" instead of "DROP FUNCTION") this would cause the test
to fail with
```
java.sql.BatchUpdateException: Batch entry 0 DROP FUNCTION IF EXISTS sha1_agg(bytea) CASCADE was aborted: ERROR: "sha1_agg" is an aggregate function
  Hint: Use DROP AGGREGATE to drop aggregate functions.
```
This adds a DROP AGGREGATE section to the clear-db-for-testing!
function. It also changes the sql-current-connection-function-names to
exclude aggregates and adds a new version of that function to lookup
aggregate names.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/storage_utils.clj
+++ b/src/puppetlabs/puppetdb/scf/storage_utils.clj
@@ -86,6 +86,31 @@
                    "INNER JOIN pg_namespace pn ON (pp.pronamespace = pn.oid) "
                    "INNER JOIN pg_language pl ON (pp.prolang = pl.oid) "
                    "WHERE pl.lanname NOT IN ('c') "
+                   (if (> 11 (-> @db-metadata :version first))
+                     ;; proisagg boolean was replaced with prokind char in pg11
+                     "AND pp.proisagg = 'f'"
+                     "AND pp.prokind = 'f'")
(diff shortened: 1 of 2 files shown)


src/puppetlabs/puppetdb/scf/storage_utils.clj
```diff
@@ -86,6 +86,25 @@
                    "INNER JOIN pg_namespace pn ON (pp.pronamespace = pn.oid) "
                    "INNER JOIN pg_language pl ON (pp.prolang = pl.oid) "
                    "WHERE pl.lanname NOT IN ('c') "
+                   "AND pp.proisagg = 'f'"
```
Zak-Kent, line 89: Looks like _proisagg_ was replaced by _prokind_ in [PG 11+](https://www.postgresql.org/docs/11/release-11.htm) which looks like why some of the travis cells are failing. Maybe we can use the **db-metadata** func in storage-utils to sniff out the version of PG we're using and have these funcs switch the column name when needed.

PR #3218, opened 2020-03-27
(PDB-4637) Prevent concurrent migrations

rbrw:


test/puppetlabs/puppetdb/config_test.clj
```diff
@@ -186,7 +186,32 @@
...
+                                  :username "someone"
+                                  :password "something"}}
+          migrator (update no-migrator :database assoc
+                           :migrator-username "admin"
+                           :migrator-password "admin")]
+
+      (testing "migrator-username"
+        (let [config (configure-db no-migrator)]
+          (is (= "someone" (get-in config [:database :migrator-username])))
+          (is (= "someone" (get-in config [:read-database :migrator-username]))))
```
Zak-Kent, line 203: I wonder about including the migrator-username/password in the read-database section of the config. I'm not sure how we handle DB migrations if someone is running a hot standby using the [read-database](https://puppet.com/docs/puppetdb/latest/configure.html#read-database-settings) config section. Do we know if migrations are performed on the hot standby in this setup? If they're not we might not want to allow the migrator user to show up in this config section.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1724,40 +1724,50 @@
...
-            ;; Don't let this rethrow, so that it won't suppress any
-            ;; pending exception.
-            (log/error
-             ex (trs "Unable to restore db access after migrations"))))))))
+  ;; When a non-migrator-name is given, do our part to attempt to
+  ;; prevent concurrent migrations or any access to a database that's
+  ;; at an unexpected migration level.  See the connectionInitSql
+  ;; setting for related efforts.
+  (let [coordinate? non-migrator-name
+        orig-user (and coordinate? (jdbc/current-user))]
```
Zak-Kent, line 1732: Might be good to add a comment here about the current-user being the migrator user in this case because the migrations use a separate connection pool set up with that user and how any jdbc calls in the form will use the migrator user.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -668,3 +673,22 @@
...
         "    and pg_stat_activity.usename = ?")
    [db user]))
+
+(defn revoke-role-db-access
+  [role db]
+  ;; revoke commands can't be parameterized with the pgjdbc driver right now
+  (prn :revoke (format "revoke connect on database %s from %s restrict"
+                       (double-quote db) (double-quote role)))
+  (do-commands
+   (format "revoke connect on database %s from %s restrict"
```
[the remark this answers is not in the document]
Zak-Kent: I'm wrong about this... this func is called using the migrations pool which is set up using the migration user. Leaving the comments here for a reminder in case I forget :)

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -668,3 +673,22 @@
...
+           (double-quote db) (double-quote role))))
+
+(defn restore-role-db-access
+  [role db]
+  ;; grant commands can't be parameterized with the pgjdbc driver right now
+  (do-commands
+   (format "grant connect on database %s to %s"
```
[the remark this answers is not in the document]
Zak-Kent: See comment above. The migration user is used to set up the migration pool that calls the func so this isn't an issue.

test/puppetlabs/puppetdb/migration_coordination_test.clj
```diff
@@ -1,19 +1,25 @@
...
+  (:import
+   (org.postgresql.util PSQLException)))
 
 (deftest schema-mismatch-causes-new-connections-to-throw-expected-errors
   (doseq [db-upgraded? [true false]]
-    (tdb/with-test-db
-      (svc-utils/call-with-puppetdb-instance
-       (-> (svc-utils/create-temp-config)
+    (with-test-db
+      (call-with-single-quiet-pdb-instance
```
Zak-Kent, line 21: I still have to dig into why a bit more, but I think this macro is seeing some of the expected connection errors in these tests and dumping all the logs for the failed connection retries. We may want to go back to `call-with-puppetdb-instance` here and in the test below to suppress the logs or see if we can update `*notable-log-event?*` so that it suppresses the logs for expected errors.

test/puppetlabs/puppetdb/config_test.clj
rbrw: Hmm, no strong opinion.  I'd just kept treating the read and write sections mostly the same, asn I think we had been, and fwiw, just having the migrator info in the read db shouldn't actually cause anything to happen.  The actual migration should still be guarded by `[database] migrate`.

rbrw: But I'm happy to see about adjusting it to elide the migrator settings altogether if we think that's preferable.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -647,3 +654,41 @@
...
+   (str "select pg_terminate_backend (pg_stat_activity.pid)"
+        "  from pg_stat_activity"
+        "  where pg_stat_activity.datname = ?"
+        "    and pg_stat_activity.usename = ?")
+   [db user]))
+
+(defn revoke-role-db-access
+  [role db]
+  ;; revoke commands can't be parameterized with the pgjdbc driver right now
+  (prn :revoke (format "revoke connect on database %s from %s restrict"
```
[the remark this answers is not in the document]
rbrw: Yep, vestigial.

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -647,3 +654,41 @@
...
+
+(defn current-user []
+  (-> (query-to-vec "select user") first :current_user))
+
+(defn-validated has-database-privilege?
+  [user db privilege] :- s/Bool
+  (-> ["select has_database_privilege(?, ?, ?)" user db privilege]
+      query-to-vec first :has_database_privilege))
+
+(defn disconnect-db-role [db user]
```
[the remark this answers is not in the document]
rbrw: Yeah, this function should never be called unless you know you want to disconnect and have the values.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -1640,87 +1643,136 @@
...
-        (trs "We are unable to create the recommended pg_trgm indexes due to\nthe extension not being installed correctly.")
-        " "
-        (trs " Run the command:\n\n    CREATE EXTENSION pg_trgm;\n\nas the database super user on the PuppetDB database to correct\nthis, then restart PuppetDB.\n")))))
-  (ensure-report-id-index))
+  []
+  (set/union
+   (maybe-create-trgm-indexes)
+   (ensure-report-id-index)))
+
+(defn note-migrations-finished [] true)
```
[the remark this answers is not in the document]
rbrw: Done (or will be in the next update).

src/puppetlabs/puppetdb/jdbc.clj
```diff
@@ -647,3 +654,46 @@
     (str "in ("
          (str/join "," (repeat (count coll) inner))
          ")")))
+
+(defn current-user []
+  (-> (query-to-vec "select user") first :user))
```
Zak-Kent, line 659: I'm getting _[{:current_user "pdb_test_migrator"}]_ as the result for this "select user" query locally. May need to change _:user_ to _:current_user_ here.

Zak-Kent: Travis looks like it got _:user_ for the column name in the result so maybe it's due to a diff version of PG or something. I'm seeing _current_user_ be returned as the column name for the query _"select user"_ on n2 as well. Maybe we could use the query _"select current_user"_ to avoid the column name switching.

test/puppetlabs/puppetdb/testutils/db.clj
```diff
@@ -128,51 +135,63 @@
...
-       (format "create database %s template %s" db-name template-name)))
-    (db-user-config db-name)))
+       (format "drop database if exists %s" db-q)
+       (format "create database %s template %s" db-q template-name)
+       ;; Needed by migration coordination
+       (format "revoke connect on database %s from public" db-q)
+       (format "grant connect on database %s to %s with grant option" db-q migrator-q)
+       (format "set role %s" migrator-q)
+       (format "grant connect on database %s to %s" db-q user-q)
+       "commit"))
```
Zak-Kent, line 184: I think the extra "commit" statement here is causing the _WARNING:  there is no transaction in progress_ pg logs we've been seeing. I think we may need a "begin" at the top of the statements here if we want them grouped into a transaction.

Zak-Kent:
The recent testing changes look good overall.

This is the list of things I think we still need:
    * logging when we kick non-migrators out and block connections, etc
    * fix log spew issues seen in: updated tests in _cli.services-test_ and test I added in _migration-coordination-test_
    * alias column name in the _jdbc/current-user_ func to avoid PG version issues  
    * update docs to provide instructions on how to use new features 
    *  fix the "there is no transaction in progress" pg warnings (see related pr comment) 

Everything looks good otherwise. In addition to running the added tests I was able to successfully check the behavior by hand setting up different combinations of pg users with/without the expected perms and diff combinations of settings in the database section of our ini file. Also checked that all db objects were owned by the non-migrator user like we'd expect.

test/puppetlabs/puppetdb/testutils/db.clj
rbrw: Ahh, right - left over from some diagnostics.  I thought I'd removed that.

test/puppetlabs/puppetdb/testutils/db.clj
```diff
@@ -128,51 +135,62 @@
...
+  [{:keys [user migrator-username subname] :as config}]
+  (jdbc/with-db-connection config
+    (let [db (subname->validated-db-name subname)
+          priv? jdbc/has-database-privilege?
+          migrator migrator-username]
+      (when (priv? "public" db "connect")
+        (throw (Exception. "public connect")))
+      (when-not (priv? user db "connect")
+        (throw (Exception. "no user connnect")))
+      (when (priv? user db "connect with grant option")
```
Zak-Kent, line 153: Should the user only have a connect without the grant option here? I thought the migrator was the only one that should have "connect with grant option"

Zak-Kent: ah, never mind I read that wrong

PR #3219, opened 2020-03-31
(maint) Bump tk-metrics to 1.2.2

austb:
This version was released publicly to clojars.

1.2.1 was an internal-only release

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -92,7 +92,7 @@
     ;; (logstash depends on jackson-databind)
     [com.fasterxml.jackson.core/jackson-databind "2.9.10"]
     ;; CVE-2020-7943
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -167,7 +167,6 @@
                     [puppetlabs/stockpile "0.0.4"]
                     [puppetlabs/tools.namespace "0.2.4.1"]
                     [puppetlabs/trapperkeeper]
-                    [puppetlabs/trapperkeeper-metrics "1.2.1" :exclusions [ring/ring-defaults org.slf4j/slf4j-api]]
```
austb: This line was redundant, we `concat pdb-dep-pins` above.

PR #3221, opened 2020-03-31
(maint) acceptance test: fix pry error

austb:
This is only affecting 5.2.x


austb:
Acceptance tests passed, this should be good to go. Travis tests are still running, so we can wait for those to be safe.

PR #3222, opened 2020-04-01
(maint) Update release test logic to work with `set -e`

underscorgan:


.github/workflows/docker.yml
```diff
@@ -34,6 +34,7 @@ jobs:
         env:
           IS_RELEASE: true
         working-directory: docker
+        shell: bash {0}
```
rbrw, line 37: Assuming the `run:` code below is the only code affected, then given my admittedly still limited experience with actions, this seems fine.  Though I wondered about this alternative:
```
  if make prep; then
      make lint build test
  else
      echo "Skipping ..."
  fi
```
Which would leave `set -e` in place for any future additions which might be a little safer?  Not sure, but I at least wouldn't have realized what that `bash {0}` does offhand.

underscorgan:
@rbrw updated!

PR #3224, opened 2020-04-04
(maint) Update Docker ssl.sh script and remove consul

Iristyle:
(maint) Remove references to Consul from container

 - Decision has been made to use Kubernetes, in which case we can rely
   on etcd instead
 
(maint) Update container to latest ssl.sh script

 - The latest script doesn't support passing an arg as $1, so instead
   use the CERTNAME= convention.

   The latest script has master_running built in to the flow, so that
   test doesn't need to be duplicated here
 
(maint) Change how container sets DNS_ALT_NAMES

 - Don't perform magic name injection - instead use just what has been
   provided.

 - Also, don't check whether value is set or not - an empty value is
   already handled in the ssl.sh script

   Added in #3107 but not necessary
 
(maint) Make sure Dockerfile has all ENV vars

 - DNS_ALT_NAMES is added to the Dockerfile for documentation
 - Add notes for HEALTHCHECK and EXPOSE
 
(maint) Format Docker README config consistently …

 - Use same table layout as other projects
 
(maint) Add more JAVA_ARGS to Docker container …

 - Fully specify additional arguments that are part of pe-puppetdb

   * -XX:+UseParallelGC for better GC perf / memory reclamation
   * -Xlog:gc for GC logs
   * -Djdk.tls.ephemeralDHKeySize=2048 for RHEL8 / CentOS8 compliance


docker/puppetdb/docker-entrypoint.d/30-configure-ssl.sh
```diff
@@ -10,7 +10,7 @@ if [ ! -f "${SSLDIR}/certs/${CERTNAME}.pem" ] && [ "$USE_PUPPETSERVER" = true ];
   if [ -n "${DNS_ALT_NAMES}" ]; then
     DNS_ALT_NAMES="${HOSTNAME},$(hostname -s),$(hostname -f),${DNS_ALT_NAMES}"
   fi
-  /ssl.sh "$CERTNAME"
+  CERTNAME="$CERTNAME" /ssl.sh
```
Iristyle, line 13: Might want to tweak how we set `DNS_ALT_NAMES` and make it consistent with puppetdb, pe-puppetdb, pe-console-services, pe-orchestration-services and pe-boltserver

Iristyle: Did this

PR #3226, opened 2020-04-08
Test merge master

austb:


austb:
Just thought I should call out that https://github.com/puppetlabs/puppetdb/pull/3226/commits/abdc936be752d8520f0ca63e60a46e9009b1a379 rewrites the SQL that deletes reports since in master we will be dropping partitions not truncating a table.

PR #3229, opened 2020-04-14
(maint) Write pe.conf file to pdbbox

austb:
This is nearly identical to the example.conf in the
pe-puppetdb-extensions repo, with host, port, and passwords interpolated
to work with the pgbox that is generated.

Adds a pe.conf file to pdbbox-init, which will work with
pe-puppetdb-extensions. It still relies on the SSL in ./test-resources/
because the SSL files generated by the pdbbox-init scipt are not
generated by Puppet and don't work quite right when running pe-puppetdb

### DIFF
--- a/ext/bin/pdbbox-init
+++ b/ext/bin/pdbbox-init
@@ -191,6 +191,129 @@ ssl-key = $abs_pdbbox/ssl/pdb.key.pem
 disable-update-checking = true
 EOF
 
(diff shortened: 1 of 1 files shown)


austb:
Turned out to be fairly simple to create a wrapper script for the PE setup, replacing this with https://github.com/puppetlabs/pe-puppetdb-extensions/pull/536

PR #3232, opened 2020-04-16
(maint) Remove LCOW support / convert to WSL2

Iristyle:
Picks up Docker changes from #3231 properly


Zak-Kent:
I'm not sure if this is needed on 6.3.x for the docker builds but we've killed off 6.3.x for other PDB work so haven't been merging up changes though that branch. Maybe we could cherry pick the docker changes to master if they're needed there as well .

underscorgan:
Yeah, in the future @Iristyle we should only need to target master here

PR #3234, opened 2020-04-22
(PDB-4637) Prevent concurrent migrations (on master)

rbrw:
Merge all the concurrent migrations work to master.


rbrw:
Note that this is effectively three commits, which is fairly easy to see in say `gitk`.  The first is on master to make the "defer pool init" adjustments we did on 5.2.x.  It's effectively a cherry-pick, except that the changes were different enough for that not to work well mechanically.  Putting that first decreased the conflicts in the next commit, the actual merge commit that merges up from 5.2.x.  Finally, there's a commit to introduce the custom exit statuses via the newer `request-shutdown`.  Suggest something like `git show -w MERGE_COMMIT_HASH` to look at the merge commit.

PR #3240, opened 2020-04-28
(PDB-3832) Update permissions for ini files to match puppetlabs-puppetdb module

underscorgan:
blocked on https://github.com/puppetlabs/ezbake/pull/571

### DIFF
--- a/resources/ext/ezbake.conf
+++ b/resources/ext/ezbake.conf
@@ -38,11 +38,18 @@ ezbake: {
                   "     fi",
                   "     /sbin/service puppetdb start > /dev/null 2>&1",
                   " fi)"
-              ]
+              ],
+              postinst-install: [
+                "find /etc/puppetlabs/puppetdb/conf.d/ -type f -name '*.ini' -exec chown {{user}}: {} \\; -exec chmod 600 {} \\; ||:"
+              ],
             },
     debian: { dependencies: ["puppet-agent (>= 5.99.0)"],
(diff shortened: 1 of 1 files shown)


underscorgan:
wow, this fell off my radar for way too long, but just ran a test and it seems to work
```
[root@dilute-datum ~]# ll /etc/puppetlabs/puppetdb/conf.d/config.ini
-rw------- 1 puppetdb puppetdb 472 Jan 29 18:48 /etc/puppetlabs/puppetdb/conf.d/config.ini
```

austb:
We just added a file named `auth.conf`, so we might want to update the command to change that as well

jplindquist:
Is there an update on this?  I noticed the same thing still occurring during every update cycle, permissions are getting reset to match what the module is expecting
```
notice | /Stage[main]/Puppetdb::Server::Global/File[/etc/puppetlabs/puppetdb/conf.d/config.ini]/owner | current_value 'root', should be 'puppetdb' (noop) (corrective)
-- | -- | --
notice | /Stage[main]/Puppetdb::Server::Global/File[/etc/puppetlabs/puppetdb/conf.d/config.ini]/group | current_value 'root', should be 'puppetdb' (noop) (corrective)
notice | /Stage[main]/Puppetdb::Server::Global/File[/etc/puppetlabs/puppetdb/conf.d/config.ini]/mode | current_value '0644', should be '0600' (noop) (corrective)
```

PR #3244, opened 2020-05-05
(docs) Add release notes for 6.10.1

austb:


documentation/release_notes/release_notes_latest.markdown
```diff
@@ -21,6 +21,12 @@ canonical: "/puppetdb/latest/release_notes.html"
 
 ---
 
+## PuppetDB 6.10.1
+
+### Bug fixes
+
+  - Fixed an issue that would cause PuppetDB to fail to start on a database that had migrations from before PuppetDB 4.0.0. [PDB-4709](https://tickets.puppetlabs.com/browse/PDB-4709)
```
clairecadman, line 28: @austb does 'migrations from before PuppetDB 4.0.0' mean a database that has not been upgraded since 4.0.0?

rbrw: Ahh, I believe that would just be the case for any database that had been in service before 4.0.0.  i.e. if the database were in use with say PuppetDB 3.2, which would have created those earlier migrations.  Any database that had only come in to service with 4.0.0 or newer wouldn't have the relevant migrations because it would have started off with a single "roll up" migration that consolidated the effects of all of the older ones.

austb: @clairecadman There were migrations that were only applied by PuppetDB versions before 4.0.0 that caused a bug in this last release. If your very first PuppetDB install was 4.0.0 or later, this bug doesn't apply.

clairecadman: Maybe "Fixed an issue that caused PuppetDB to fail to start on a database that had been in service before PuppetDB 4.0.0." is clearer?

PR #3247, opened 2020-05-05
(maint) with-pdbbox: use ::1 for localhost to fix macos tests

rbrw:
Looks like macos doesn't define ip6-localhost


rbrw:
Looks like `::1` might not be quite what we need.  Might revisit this later.

rbrw:
Ended up moving macos tests to github actions.

PR #3251, opened 2020-05-07
(PDB-4701) Add ast rewriter to default to agent reports in queries

rbrw:
Add maybe-add-agent-report-filter-to-query to take an incoming query
and ast filter and rewrite the filter to include an additional ["="
"type" "agent"] clause at the top level and in every subquery,
whenever the query involves reports, and the existing filter at that
level doesn't already mention the report "type".


austb:
Merged in https://github.com/puppetlabs/puppetdb/pull/3255

PR #3252, opened 2020-05-08
(maint) Dockerfile updates

Iristyle:
- Upgrade dumb-init to 1.2.2 (fixes signal race condition)
- Set ENV var LOGDIR in a way that it can be used elsewhere in Dockerfile
- Change the Docker "name" label


Iristyle:
Container job results at https://travis-ci.com/github/puppetlabs/puppetdb/jobs/329795290

docker/puppetdb/Dockerfile
```diff
@@ -8,7 +8,7 @@ ARG UBUNTU_CODENAME=bionic
 
 FROM ubuntu:18.04 as base
 
-ARG DUMB_INIT_VERSION="1.2.1"
+ARG DUMB_INIT_VERSION="1.2.2"
```
nwolfe, line 11: Not related to your change, but.... why is this an `ARG` ?  Doesn't seem like something we'd ever set externally.

docker/puppetdb/Dockerfile
```diff
@@ -24,19 +24,18 @@ ENV PUPPERWARE_ANALYTICS_ENABLED=false \
...
-# NOTE: Docker ENV values cannot consume other ENV values, so full path to $LOGDIR is specified for -Xloggc
+    PUPPETDB_REPORT_TTL=14d
+# note: LOGDIR cannot be defined in the same ENV block it's used in
 # this value may be set by users, keeping in mind that some of these values are mandatory
 # -Djavax.net.debug=ssl may be particularly useful to set for debugging SSL
-    PUPPETDB_JAVA_ARGS="-Djava.net.preferIPv4Stack=true -Xms256m -Xmx256m -XX:+UseParallelGC -Xlog:gc*:file=/opt/puppetlabs/server/data/puppetdb/logs/puppetdb_gc.log::filecount=16,filesize=65536 -Djdk.tls.ephemeralDHKeySize=2048" \
+ENV PUPPETDB_JAVA_ARGS="-Djava.net.preferIPv4Stack=true -Xms256m -Xmx256m -XX:+UseParallelGC -Xlog:gc*:file=$LOGDIR/puppetdb_gc.log::filecount=16,filesize=65536 -Djdk.tls.ephemeralDHKeySize=2048" \
 # used by entrypoint to determine if puppetserver should be contacted for config
 # set to false when container tests are run
     USE_PUPPETSERVER=true
```
nwolfe, line 34: Not a problem but this `USE_PUPPETSERVER` could be with the other `ENV` block since it's static.

And I just saw another Dockerfile in your PRs that had it up there 😄

docker/puppetdb/Dockerfile
Iristyle: There are 2 ways of doing a variable in a Dockerfile and ARG is the appropriate one for this context.

ARGs are for build-time variables (even if not specified as a `--build-arg` it's still an input and gives you one source of truth for that value in the Dockerfile). They are actually implemented as environment variables that don't show up in final image.

ENVs are for runtime variables - signals to end users that something that can be changed at launch time.

docker/puppetdb/Dockerfile
Iristyle: Yeah I did it that way to minimize the diff, but if you think it reads better, I can reorg the list. No biggie either way.

PR #3254, opened 2020-05-11
(maint) Switch Travis to Ubuntu 20.04 builder

Iristyle:
- Focal fossa time!
 - No need to upgrade Docker for buildkit as image has 19.03 already,
   so this should remove ~30s of setup time
 - Switch to Ruby 2.6.6 for specs, which is known good and installed
 - docker-compose in the 19.03 release included here is 1.23.1, so
   continue to upgrade it on Travis to 1.25.5
 - Separate out Travis operations so that logs have timings for lint, build and test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ dist: xenial
 sudo: required
 
 services:
-  # bionic uses 18.06 but need 19.03+ for buildkit so upgrade later in relevant cell
   - docker
 
 # The test specifications are all extracted from the PDB_TEST value.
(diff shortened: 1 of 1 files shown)


Iristyle:
make lint: 3.00s
make build: 212.26s
make test: 130.64s

.travis.yml
```diff
@@ -67,20 +66,17 @@ aliases:
     ext/bin/check-spec-env "$PDB_TEST"
     ext/bin/run-rspec-tests "$puppet_ref"
 
-  - &run-docker-tests |
-    set -ex
```
underscorgan, line 71: double-check whether we need `set -e`

Iristyle: Double checked.. its needed... and fixed.

Thanks!

PR #3265, opened 2020-05-26
(PDB-4743) Fix migration check when there's more than one db

rbrw:
Comparison was backward, and also didn't consult "migrate=?" settings.

fixes: 7a52d73801a38be79b725a25aa74bcfe1d847bcb


Zak-Kent:
Merging with 4 expected integ test failures in CI. A fix should be added to facter tomorrow which should resolve these test failures.

PR #3266, opened 2020-05-27
(PDB-4751) Improve console output during upgrades

rbrw:
Print friendlier, more concise output to the console during upgrades.
May want to make this nicer later.  One easy way to see the new
output:

  - Create a 5.2.x database
  - Add some commands via benchmark
  - ./pdb upgrade -c ... > tmp.log


rbrw:
We decided not to pursue this for now since it wasn't quite what was needed for the case in question.  See the ticket for more information.

PR #3269, opened 2020-05-28
(maint) Log shutdown errors

rbrw:
So they'll still be visible somewhere if stdout/stderr is not flushed
or lost.


Zak-Kent:
closing in favor of #3263. That pr had the same commit.

PR #3273, opened 2020-06-04
(PDB-4750) Add additional config tests for subsection behavior

Zak-Kent:
Once docs are finished I'll double check to see if we want to add any additional tests.


Zak-Kent:
config syntax changed

PR #3276, opened 2020-06-04
(PDB-4656) Schema check the write database(s) too

rbrw:


rbrw:
The "don't merge" is just because this is on top of #3275, and so I'll want to rebase it once that's settled.

rbrw:
Merged as part of #3283.

PR #3278, opened 2020-06-12
(PDB-4763) Add initial command broadcast docs

rbrw:


rbrw:
Still need to double-check that the inter-doc `#links` are correct.

PR #3286, opened 2020-07-09
(PDB-4808) Validate read db and migrate write dbs

rbrw:


rbrw:
This was the inspiration for the exception in the "(PDB-4812) convert-to-schema: always throw an exception" commit: https://github.com/plumatic/schema/blob/ddb54c87dea6926c6d73542e517b53685cc82a37/src/cljx/schema/core.cljx#L149

src/puppetlabs/puppetdb/command.clj
```diff
@@ -778,110 +796,156 @@
...
+   ;; The reason some of these aren't conditional is because they're
+   ;; established by init, tk won't call stop if init didn't
+   ;; finish, and in those cases a nil would indicate a bug.
+   (finally (some-> command-chan async/close!))
+   (finally (when command-shovel (.interrupt command-shovel)))
+   (finally (some-> broadcast-threadpool pool/shutdown-unbounded))
+   (finally (some-> consumer-threadpool pool/shutdown-gated))
+   (finally (stop-and-reset-pool! delay-pool))
+   (finally
+     ;; Wait up to ~5s for https://bugs.openjdk.java.net/browse/JDK-8176254
```
austb, line 916: We should ticket this for removal in Platform 7 as this bug doesn't affect Java 7

documentation/configure.markdown
```diff
@@ -394,6 +406,52 @@ Then specify `puppetdb_migrator` as the
...
+    username = <USERNAME2>
+    password = <PASSWORD2>
+
+When subsections are defined, the `[database]` section provides
+defaults that will apply whenever a value is not specified in a
+subsection.  These defaults will also be used for the read database
+whenever an explict [`[read-database]`]](#[read-database]--settings)
+isn't provided.
+
+PuppetDB will not migrate multiple databases.  This restriction may be
```
Zak-Kent, line 447: Reminder to update docs here because we're allowing multi db migrations with cmd broadcast now.

src/puppetlabs/puppetdb/pdb_routing.clj
```diff
@@ -92,51 +93,71 @@
...
+                  (pdb-core-routes config
+                                   augmented-globals
+                                   enqueue-command
+                                   query
+                                   clean
+                                   delete-node))
+         (mid/wrap-cert-authn cert-whitelist)
+         mid/wrap-with-puppetdb-middleware))
+
+    (enable-maint-mode)
```
Zak-Kent, line 121: I think we also need to guard the MaintenanceMode service that provides this method above. When looking at this I made the PDB service pop during init and saw null pointer exceptions from the maint mode service when it doesn't have a context. Adding the call-unless-shutting-down wrapper to the maint stuff helped prevent the null pointer exception during shutdown.

PR #3291, opened 2020-07-15
(PDB-4659) Add  acceptance tests for cmd broadcast

Zak-Kent:
Will need to rebase this once Rob's config work gets merged. Also needs a fix for a race in blocking cmd submission when in broadcast mode (see [PDB-4796](https://tickets.puppetlabs.com/browse/PDB-4796)).

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -570,7 +570,7 @@
   (update-counter! :invalidated command version dec!))
 
 (defn broadcast-cmd
-  [{:keys [certname command version id] :as cmd}
+  [{:keys [certname command version id callback] :as cmd}
    write-dbs pool response-chan stats shutdown-for-ex]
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/command.clj
```diff
@@ -586,7 +586,8 @@
         not-exception? #(not (instance? Throwable %))
         attempt-exec (fn [db status]
                        (try
-                         (attempt-exec-command cmd db status response-chan stats
+                         ;; handle callback below to avoid races during cmd broadcast
+                         (attempt-exec-command (assoc cmd :callback identity) db status response-chan stats
```
austb, line 590: Maybe, `(constantly true)` instead of `identity`, I didn't see any places where we store the result of the callback, so it probably doesn't matter what we return from callback, but if we do it could end up storing a duplicate of the cmd, which has a user-defined size.

PR #3298, opened 2020-07-29
(maint) travis: run drop-joins test in main stage

austb:
Run all puppetdb tests in one stage to save time. After this commit,
there are only two stages, clojure/ruby tests and container tests.


rbrw:
Hmm, not sure why we didn't do that to begin with...

PR #3309, opened 2020-08-11
(maint) Move macos tests to github actions

rbrw:


ext/bin/run-external-tests
```diff
@@ -37,7 +41,10 @@ run ext/test/schema-mismatch-causes-pdb-shutdown
 if test "$failures" -eq 0; then
     echo "failures: $failures" 1>&2
     exit 0
-else
+elif test "$TERM"; then
```
austb: If the lack of colorized output here and above is due to missing `tput` on macos something like
```
printf "\e[38;5;1mfailures: $failures\n"
```
looks to work in both linux and macos

rbrw: Oh, it's not missing, it just runs and fails with the message I quoted, which is odd because "TERM=dumb tput ..." works fine here.  Wondering if it's a macos-specific issue there.  I suppose we could use brew's tput (via ncurses? that's where mine comes from), but I assumed that colors wouldn't work in the gh log anyway, so didn't matter.

PR #3312, opened 2020-08-13
(PDB-4745) disable gc in config-expiration test

austb:
Two things acquire the `clean-lock`, periodic garbage collection and the
admin clean endpoint. Garbage collection uses `(.lock clean-lock)`,
which means it will block until the lock is available. The admin
endpoint uses `(when (.tryLock clean-lock) <clean>)` which means if the
lock is not available it will return `false` and skip the attempt to
perform the admin clean action.  Sometimes, when this test runs, gc can
be running in the background and cause the initial call to
`(cli-svc/clean pdb ["expire_nodes"])` to be skipped, which puts the db
into an incorrect state for the following test checks.

### DIFF
--- a/test/puppetlabs/puppetdb/acceptance/node_ttl.clj
+++ b/test/puppetlabs/puppetdb/acceptance/node_ttl.clj
@@ -62,6 +62,7 @@
       (svc-utils/call-with-single-quiet-pdb-instance
        (-> (svc-utils/create-temp-config)
            (assoc :database *db*)
+           (assoc-in [:database :gc-interval] 0)
            (assoc-in [:database :node-ttl] lifetime-cfg)
            (assoc-in [:database :node-purge-ttl] lifetime-cfg))
        (fn []


austb:
If you want to run it a number of times, you can apply this git patch to the PR
```
diff --git a/test/puppetlabs/puppetdb/acceptance/node_ttl.clj b/test/puppetlabs/puppetdb/acceptance/node_ttl.clj
index 851b9c1bd..c6eb2e9f0 100644
--- a/test/puppetlabs/puppetdb/acceptance/node_ttl.clj
+++ b/test/puppetlabs/puppetdb/acceptance/node_ttl.clj
@@ -56,6 +56,10 @@
                         :body))))))))))

 (deftest configure-expiration-behavior
+  (doseq [i (range 0 20)]
+    (println "-------------------------------------")
+    (println (format "Interval %d of %d" (inc i) 20))
+    (println "-------------------------------------")
   (let [lifetime-ms 1
         lifetime-cfg (str lifetime-ms "ms")]
     (with-test-db
@@ -166,7 +170,7 @@
              (deactivate "foo" (now))
              (is (= [{"certname" "foo" "expired" nil}] (nodes)))
              (cli-svc/clean pdb [])
-             (is (= [] (nodes))))))))))
+             (is (= [] (nodes)))))))))))

 (deftest test-zero-gc-interval
   (with-redefs [puppetlabs.puppetdb.cli.services/purge-nodes! (tu/mock-fn)]
```

rbrw:
OK, right, so I suppose we could try to address the "returns false" question somehow, but don't really see the point, since that's irrelevant to the test, i.e. agree that it's better to just turn off gc here.

PR #3314, opened 2020-08-14
Honor env var PUPPETDB_JAVA_ARGS

rstruber:
Creates support for the environment variable PUPPETDB_JAVA_ARGS through to the JVM args at application startup.

### DIFF
--- a/docker/puppetdb/Dockerfile
+++ b/docker/puppetdb/Dockerfile
@@ -31,7 +31,7 @@ ENV PUPPERWARE_ANALYTICS_ENABLED=false \
 # note: LOGDIR cannot be defined in the same ENV block it's used in
 # this value may be set by users, keeping in mind that some of these values are mandatory
 # -Djavax.net.debug=ssl may be particularly useful to set for debugging SSL
-ENV PUPPETDB_JAVA_ARGS="-Djava.net.preferIPv4Stack=true -Xms256m -Xmx256m -XX:+UseParallelGC -Djdk.tls.ephemeralDHKeySize=2048"
+ENV PUPPETDB_JAVA_ARGS="-Djava.net.preferIPv4Stack=true -Xms256m -Xmx256m -XX:+UseParallelGC -Xlog:gc*:file=$LOGDIR/puppetdb_gc.log::filecount=16,filesize=65536 -Djdk.tls.ephemeralDHKeySize=2048"
 
 LABEL org.label-schema.maintainer="Puppet Release Team <release@puppet.com>" \
       org.label-schema.vendor="Puppet" \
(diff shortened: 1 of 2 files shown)


docker/puppetdb/Dockerfile
```diff
@@ -31,7 +31,7 @@ ENV PUPPERWARE_ANALYTICS_ENABLED=false \
 # note: LOGDIR cannot be defined in the same ENV block it's used in
 # this value may be set by users, keeping in mind that some of these values are mandatory
 # -Djavax.net.debug=ssl may be particularly useful to set for debugging SSL
-ENV PUPPETDB_JAVA_ARGS="-Djava.net.preferIPv4Stack=true -Xms256m -Xmx256m -XX:+UseParallelGC -Xlog:gc*:file=$LOGDIR/puppetdb_gc.log::filecount=16,filesize=65536 -Djdk.tls.ephemeralDHKeySize=2048"
```
rstruber, line 34: This arg caused an error on startup in the JVM for "not known argument".

Iristyle: Hmmm - we're successfully using that option in the PE version of PuppetDB. We should try and find the equivalent for this version of Java if it's not working here if it's a different Java runtime.

This is the code from pe-puppetdb-extensions

```
# NOTE: Docker ENV values cannot consume other ENV values, so full path to $LOGDIR is specified for -Xloggc
# this value may be set by users, keeping in mind that some of these values are mandatory
# -Djavax.net.debug=ssl may be particularly useful to set for debugging SSL
    PUPPETDB_JAVA_ARGS="-Djava.net.preferIPv4Stack=true -Xms256m -Xmx256m -XX:+UseParallelGC -Xlog:gc*:file=/opt/puppetlabs/server/data/puppetdb/logs/puppetdb_gc.log::filecount=16,filesize=65536 -Djdk.tls.ephemeralDHKeySize=2048" \
```

That first note is actually wrong about ENV vars not being able to use other ENV vars... they just have to be defined earlier like this:

```
ENV foo
ENV bar=$foo
```


/cc @puppetlabs/puppetdb

Iristyle:
This is looking very familiar @underscorgan.  Ah, because we did a similar thing in `pe-puppetdb` with copying in `/etc/puppetlabs` - https://github.com/puppetlabs/pe-puppetdb-extensions/pull/458/commits/6199e62aab2183e0613a3e95c3785f8f51002931

Did we really miss doing the same thing in this container?

docker/puppetdb/Dockerfile
rstruber: I'll see about getting you more detailed output. My dockerfile doesn't extend this one too much.

underscorgan:
@Iristyle d'oh looks like we did...

Iristyle:
Taking over this work in #3378 - figured out the correct logging arg there

PR #3317, opened 2020-08-17
(maint) pdbbox-init: use mktemp for all auth data

rbrw:
Using install may have been fine, but the handling of permissions is
not documented, and install might not be as portable, so just rely on
mktemp, which we already require, and which at least Linux and OpenBSD
does explicitly document the handling of permissions.

### DIFF
--- a/ext/bin/pdbbox-init
+++ b/ext/bin/pdbbox-init
@@ -79,18 +79,27 @@ migrator_passwd="$(dd if=/dev/urandom bs=1 count=32 | base64)"
 
 # This code relies on echo being a function in bash, not a subprocess
 # and "here documents" to prevent passwords from being visible in
-# "ps", etc.
+# "ps", etc.  We also assume that mktemp perms are always go-wrx,
+# across all the relevant platforms.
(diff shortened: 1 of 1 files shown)


ext/bin/pdbbox-init
```diff
@@ -79,26 +79,35 @@ migrator_passwd="$(dd if=/dev/urandom bs=1 count=32 | base64)"
...
- install -m 0600 /dev/null test-pass-admin
- echo -n "$admin_passwd" > test-pass-admin)
-
-(cd "$PGBOX"
- cat > pgpass <<-EOF
+# "ps", etc.  We also assume that mktemp perms are always go-wrx,
+# across all the relevant platforms.
+
+tmp_pass="$(mktemp "$tmp_dir/tmp-test-pass-XXXXXX")"
+chmod 0600 "$tmp_pass" # Not entirely safe, but expected to be redundant
```
austb, line 86: Why is the `chmod` necessary? It looks like mktemp already documents that it creates files with those permissions.

PR #3329, opened 2020-08-26
(PDB-4730) Use Puppet 7 HTTP Client

austb:


puppet/lib/puppet/util/puppetdb/http.rb
```diff
@@ -43,7 +43,7 @@ def self.with_http_error_logging(server_url, route, &cb)
           "after #{config.server_url_timeout} seconds. #{SERVER_URL_FAIL_MSG}")
         return e
 
-      rescue SocketError, OpenSSL::SSL::SSLError, SystemCallError, Net::ProtocolError, IOError, Net::HTTPNotFound => e
+      rescue Puppet::HTTP::HTTPError => e
```
austb, line 46: This should catch `Puppet::SSL::SSLError` as well.

puppet/lib/puppet/util/puppetdb/http.rb
```diff
@@ -131,17 +131,18 @@ def self.failover_action(path_suffix, server_urls, sticky, ssl_context = nil, ht
...
-        route = concat_url_snippets(server_url.request_uri, path_suffix)
+        route = concat_url_snippets(server_url.to_s, path_suffix)
 
         request_exception = with_http_error_logging(server_url, route) {
-          if ssl_context.nil?
-            http = Puppet::Network::HttpPool.http_instance(server_url.host, server_url.port)
-          else
-            http = Puppet::Network::HttpPool.connection(server_url.host, server_url.port, ssl_context: ssl_context)
-          end
+          # TODO: ssl_context?
```
joshcooper, line 137: For the default case where you just want to use the existing set of certs, you don't need to pass an ssl_context. The client will take care of that. But if you had a specific requirement to use a particular set of certs, then you could build a context and pass it to the `Puppet::HTTP::Client#get`, etc methods.

puppet/lib/puppet/indirector/resource/puppetdb.rb
```diff
@@ -27,21 +27,22 @@ def search(request)
...
 
-        unless response.is_a? Net::HTTPSuccess
+        unless response.success?
           # Newline characters cause an HTTP error, so strip them
-          raise "[#{response.code} #{response.message}] #{response.body.gsub(/[\r\n]/, '')}"
+          raise Puppet::Error, "[#{response.code} #{response.reason}] #{response.body.gsub(/[\r\n]/, '')}"
         end
       rescue => e
-        raise Puppet::Error, "Could not retrieve resources from the PuppetDB at #{self.class.server}:#{self.class.port}: #{e}"
+        raise Puppet::Error, "Could not retrieve resources from the PuppetDB at  #{e}"
```
joshcooper, line 45: If you want to preserve the `server:port` you could keep a reference to the `uri` variable passed to the block https://github.com/puppetlabs/puppetdb/pull/3329/files#diff-492bee772fbfabea0f556ca28ba42f74R30?

.travis.yml
```diff
@@ -74,56 +74,43 @@ jobs:
...
     - stage: ❧ pdb tests
-      name: int pup-master srv-master jdk-8 pg-11 rich
-      env: PDB_TEST=int/openjdk8/pup-master/srv-master/pg-11/rich
-      script: *run-integration-tests
-
-    - stage: ❧ pdb tests
-      name: int pup-master srv-master jdk-8 pg-9.6 rich
-      env: PDB_TEST=int/openjdk8/pup-master/srv-master/pg-9.6/rich
+      name: int pup-master srv-6.x jdk-11 pg-11 rich
+      env: PDB_TEST=int/openjdk11/pup-master/srv-main/pg-11/rich
```
austb, line 102: the `srv-` setting should be `main`

Magisus:
I'm working on the puppetserver portion of this, and I think the changes here will fix some of the errors I'm seeing in the acceptance suites (notably https://github.com/puppetlabs/puppetdb/pull/3329/files#diff-b43043c3acd01a66d0886dd6f3b01d71R138). @austb are you aware of anything more that needs to happen with this?

puppet/spec/unit/indirector/catalog/puppetdb_spec.rb
```diff
@@ -50,7 +51,7 @@ def save
     end
 
     it "should log a deprecation warning if one is returned from PuppetDB" do
-      response['x-deprecation'] = 'A horrible deprecation warning!'
+      response.nethttp['x-deprecation'] = 'A horrible deprecation warning!'
```
joshcooper, line 54: you should be able to access the response headers through the `response[]` directly

puppet/spec/unit/indirector/facts/puppetdb_spec.rb
```diff
@@ -168,10 +169,14 @@ def find_facts()
       body = [{"certname" => "some_node", "environment" => "production", "name" => "a", "value" => "1"},
               {"certname" => "some_node", "environment" => "production", "name" => "b", "value" => "2"}].to_json
 
-      response = Net::HTTPOK.new('1.1', 200, 'OK')
+      response = Puppet::HTTP::Response.new(Net::HTTPOK.new('1.1', 200, 'OK'), "mock url")
       response.stubs(:body).returns body
 
-      http.stubs(:get).with("/pdb/query/v4/nodes/some_node/facts", subject.headers, options).returns response
+      http.stubs(:get).with do |uri, opts|
+        "/pdb/query/v4/nodes/some_node/facts" == uri.path &&
```
joshcooper, line 176: Does this need to be `expect("/pdb/query/v4/nodes/some_node/facts").to eq(uri.path)` etc?

austb: Our tests are still using Mocha I believe, so the yield needs to return a boolean https://mocha.jamesmead.org/Mocha/Expectation.html#with-instance_method

Zak-Kent (approved):
👍 merging now to get further validation in PE while we fix up the one broken integ test.

puppet/spec/unit/indirector/facts/puppetdb_spec.rb
joshcooper: Ahhh yeah good ol' mocha

PR #3330, opened 2020-08-28
(PDB-4834) Add a debug AST/SQL logging option

Zak-Kent:
Probably want to hide whitespace changes during review.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng.clj
+++ b/src/puppetlabs/puppetdb/query_eng.clj
@@ -169,7 +169,8 @@
    :node-purge-ttl Period
    :add-agent-report-filter Boolean
    (s/optional-key :warn-experimental) Boolean
-   (s/optional-key :pretty-print) (s/maybe Boolean)})
+   (s/optional-key :pretty-print) (s/maybe Boolean)
+   (s/optional-key :log-queries) Boolean})
 
 (pls/defn-validated stream-query-result
   "Given a query, and database connection, return a Ring response with the query
--- a/src/puppetlabs/puppetdb/query_eng.clj
+++ b/src/puppetlabs/puppetdb/query_eng.clj
@@ -95,64 +95,76 @@
          :when (:queryable? projection-value)]
      (keyword projection-key))))
 
+(defn maybe-log-sql [logging-options gen-sql]
+  (if-let [log-id (:log-id logging-options)]
+    (let [sql-output (gen-sql)]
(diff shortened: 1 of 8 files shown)


src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -182,14 +194,21 @@
...
           :or {warn-experimental true
-               pretty-print false}} options
+               pretty-print false
+               log-queries false}} options
+         log-id (when log-queries (.toString (java.util.UUID/randomUUID)))
          {:keys [remaining-query entity]} (eng/parse-query-context version query warn-experimental)
          munge-fn (get-munge-fn entity version paging-options url-prefix)]
+
+     (when log-queries
+       ;; log the AST of the incoming query
```
Zak-Kent, line 206: Do we want a way to differentiate between queries submitted externally vs. those submitted via the TK PuppetDBServer service's query method when logging?

Zak-Kent:
I still need to write config docs for this but will hold off till next week in case we want to change the names of things.

src/puppetlabs/puppetdb/query_eng.clj
austb: I don't think it matters, we want to log the queries regardless of where they come from. But I think the only places the TK service method is used, for better or for worse, is in HA sync and the admin export comand.

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -95,64 +95,75 @@
...
+           (jdbc/valid-jdbc-query? (:results-query %))
+           (or (not (:include_total query-options))
+               (jdbc/valid-jdbc-query? (:count-query %)))]}
 
-  (cond
-    (= :aggregate-event-counts entity)
-    (aggregate-event-counts/query->sql version query query-options)
+   (maybe-log-sql
+    logging-options
+    #(cond
```
austb, line 119: Do you think that `(fn []` provides any useful clarity that the anonymous function has no arguments?

Also if wouldn't be clojure PR review without indentation comments, it looks like this is only 1 space indented.

Zak-Kent: 👍 I could see changing it to `(fn []` for clarity sake. I wasn't sure about the one space indent at first when my editor kept insisting on it, but then saw [this](https://github.com/bbatsov/clojure-style-guide#arguments-indentation) in the style guide.

PR #3334, opened 2020-09-15
(PDB-4897) Correct :source-tables for catalog inputs queries

rbrw:
Fix the :source-tables metadata for the catalog inputs queries They
were incorrectly using :source-table and providing a string, not a
set, but given the current uses of the data, that may not have
affected anything.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -901,7 +901,7 @@
      :entity :catalog-input-contents
      :alias "catalog_input_contents"
      :subquery? false
-     :source-table "catalog_inputs"}))
+     :source-tables #{"catalog_inputs"}}))
 
 (def catalog-inputs-query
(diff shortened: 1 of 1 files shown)


src/puppetlabs/puppetdb/query_eng/engine.clj
```diff
@@ -901,43 +901,43 @@
...
+                         :from [:catalog_inputs]
+                         :group-by [:certname_id]} :ci]
+                       [:= :certnames.id :ci.certname_id]]
+                :where [:<> :certnames.catalog_inputs_timestamp nil]}
+    :relationships certname-relations
+
+    :entity :catalog-inputs
+    :alias "catalog_inputs"
+    :subquery? false
+    :source-table #{"certnames"}}))
```
austb, line 940: Does this need to be `:source-tables` as well?

rbrw: Yep.

austb (approved):

PR #3335, opened 2020-09-15
(PDB-4881) Add primary key to catalog_inputs table

austb:
Before there was no primary key, which meant the `pg_repack` could not
be used on the table.

By adding a primary key we can now use `pg_repack` to maintain the
table. Also by making the compound index on (type, name, certname_id) we
allow queries like this to use an index.
```
catalog-inputs-contents[] {
  type = "hiera"
  name = "puppetdb::globals::version"
}
```


austb:
Superceded by #3337

PR #3336, opened 2020-09-16
(PDB-4895) Add a top level hash to catalog_inputs

austb:
Don't merge because https://github.com/puppetlabs/puppetdb/pull/3335 needs to go first


Zak-Kent (approved):
🦖 LGTM! We might still want to test the compound PK index creation time with some generated catalog_inputs data in PE but we can do that after this is merged in a PE build.

Zak-Kent:
Also the one error in the travis tests looked like a transient in the container build.

rbrw:
Re-raised as #3337.

PR #3341, opened 2020-10-08
Add basic partition drop and command timeout tests

rbrw:


rbrw:
May want to rework a bit based on #3346.

PR #3344, opened 2020-10-09
(PDB-4908) prune-daily-partitions: don't nest do-commands

rbrw:
Accidentally introduced by the refactoring related to lock-statuses
cf. f240694492b0bf783190cfdb5630ad1218400afa

Causes the normal periodic garbage collections to fail like this:

  ERROR puppetlabs.puppetdb.cli.services - Error while sweeping reports and resource events
  java.sql.BatchUpdateException: Batch entry 0 0 was aborted: ERROR: syntax error at or near "0"
    Position: 1  Call getNextException to see other errors in the batch.

The garbage collections at startup and those triggered by the admin
endpoint should be unaffected.


Zak-Kent (approved):
👍

PR #3346, opened 2020-10-14
(PDB-4931) Incrementally drop partitions oldest-first

rbrw:


src/puppetlabs/puppetdb/config.clj
```diff
@@ -101,7 +101,7 @@
   "Includes the common database config params, also the write-db specific ones"
   (merge per-database-config-in
          (all-optional
-           {:gc-interval (pls/defaulted-maybe s/Int 60)
+           {:gc-interval (pls/defaulted-maybe s/Num 60)
```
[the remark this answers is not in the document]
Zak-Kent: We also hit another null pointer exception in the shutdown when this happens so we may have another case where we need to guard against other services not getting the config they need. We can sort that later just wanted to note it so we don't forget. 

`2020-10-15 13:57:56,452 ERROR [async-dispatch-3] [p.t.internal] Encountered error during shutdown sequence
java.lang.NullPointerException: null
	at clojure.core$reset_BANG_.invokeStatic(core.clj:2381)
	at clojure.core$reset_BANG_.invoke(core.clj:2376)
	at clojure.core$update.invokeStatic(core.clj:6198)
	at clojure.core$update.invoke(core.clj:6188)
	at puppetlabs.puppetdb.pdb_routing$reify__53626$service_fnk__22529__auto___positional$reify__53633.enable_maint_mode(pdb_routing.clj:91)`

src/puppetlabs/puppetdb/config.clj
```diff
@@ -363,6 +363,18 @@
 (defn default-events-ttl [config]
   (update config :resource-events-ttl #(or % (:report-ttl config))))
 
+(defn gc-interval->ms [config]
+  (update config
+          :gc-interval
+          (fn [sec]
```
Zak-Kent, line 369: We might want to call the _sec_ arg here _min_?

Zak-Kent:
Incremental drop partition commit looks good.

Zak-Kent (approved):
👍 LGTM. We might want to run the travis tests on pdbext when this goes in to check if we need to make test changes related to the gc-interval.

test/puppetlabs/puppetdb/command_broadcast_test.clj
```diff
@@ -156,7 +156,7 @@
 (deftest cmd-broadcast-with-all-pgs-down
   (with-open [pg1 (int/setup-postgres)
               pg2 (int/setup-postgres)
-              pdb1 (int/run-puppetdb pg1 {:database {"gc-interval" 0}
```
Zak-Kent, line 159: Looks like we only need to pass gc-interval in as a string in tests if it's a float.

PR #3361, opened 2020-10-29
(maint) fix call to `concat_url_snippets`

ciprianbadescu:
previously `concat_url_snippets` was called only with `path`
part of `server_url`, without passing `host` and `port` settings


ciprianbadescu:
ping @puppetlabs/night-s-watch

austb (approved):

PR #3362, opened 2020-10-29
(FEAT) Add (certname,endtime) index to reports

npwalker:
Prior to this commit, we had a certname index on reports
and an endtime index on reports.  For queries that want
certname sorted by endtime they were scanning the
endtime index and removing rows that don't match the certname
causing the query to read way more data pages into memory than
is desirable.

After this commit, we remove the certname index and replace it
with a certname,endtime index.  This allows for finding a certname
in the index and also getting the reports for it in the endtime
order that the query desires.  This drastically reduces the number
of data pages that need to be read into memory.

Also, any queries that previously used the certname index can still
use the certname,endtime index to satisfy their needs.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -1964,6 +1964,23 @@
              (jdbc/double-quote idx-name)
              (jdbc/double-quote table)))))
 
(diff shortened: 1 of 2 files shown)


austb:
Cherry-picked this commit to https://github.com/puppetlabs/puppetdb/pull/3410, added a test, and fixed the issue with the test for migration 74.

PR #3366, opened 2020-11-06
(PDB-4961) State that postgres <11 is unsupported

austb:
Since this is now a non-fatal error, we've gone further than deprecating
postgresql less than 11. In PuppetDB 7+ it will be officially
unsupported and may break in any release.


Zak-Kent (approved):
👍

PR #3379, opened 2020-11-19
(maint) travis: remove integration tests

austb:
See [RE-13716](https://tickets.puppetlabs.com/browse/RE-13716) for more
information on the issue that forced us to upgrade bouncycastle to 1.65.

Puppetserver is not compatible with bouncycastle 1.65, and the upgrade
path is not as simple as ours. As a result our integration tests will
not pass. Since this branch will EOL on Jan 31, they don't plan to
upgrade at all, so we should stop attempting to run integration tests.


Zak-Kent (approved):
🔥

PR #3380, opened 2020-12-02
(PDB-4974) Wait for schedulers to shut down (and replace at-at)

rbrw:
Although we'd expected to be waiting for our schedulers (at-at pools)
to stop before returning from our start/stop methods, we weren't
because at-at runs the relevant ScheduledThreadPoolExecutor shutdown
function in a future and returns immediately.

Since our needs are fairly straightforward, and since the relevant
blocking function in at-at is private, just manage the
ScheduledThreadPoolExecutors ourselves.

cf. PDB-4932

### DIFF
--- a/pdb
+++ b/pdb
@@ -3,8 +3,8 @@
 set -e
 
 jar="${PDB_JAR:-target/puppetdb.jar}"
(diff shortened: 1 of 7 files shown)


rbrw:
If we do decide to pursue this, among other things, we'll need to figure out the right thing to do about the two tests that are commented out.

Zak-Kent (approved):
🦖 LGTM!

PR #3382, opened 2020-12-04
(PDB-4975) Deliver on promise even if gc disabled

austb:
When gc is disabled, PuppetDB would never deliver on the
initial-gc-finished promise. This would cause sync to never run, and
puppetdb to never exit maintenance mode on startup. Now we'll deliver on
the promise if gc is disabled or once all initial garbage collections
have finished. Before this, gc would deliver at the end of its first
initial gc, even if there were multiple databases.


austb:
Highly recommended to view diff with whitespace disabled.

Zak-Kent (approved):
👍

PR #3383, opened 2020-12-07
(PDB-4971) Remove Ruby 2.7 deprecation warnings

austb:
* Remove dependency on json_pure
    * This is an old an unused dependency that was causing deprecation warnings on Ruby 2.7
* Replace URI with CGI
    * The URI functions we were using were obsoleted by CGI replacements
* rspec: fix a deprecation warning
    * The name shadowing was producing a warning


Zak-Kent (approved):
🦖

PR #3384, opened 2020-12-08
(PDB-4971) Explicitly splat the http options

austb:
In Ruby 3.0 they are aiming to separate positional and keyword
arguements, so in 2.7 they deprecated the implicit use of keyword
arguments as the last parameter. Adding the "hash splat" operator to the
function call makes it explicit that the hash is to be parsed as keyword
arguments.


Zak-Kent (approved):
👍

PR #3389, opened 2020-12-18
(PDB-4948) Allow GC to cancel blocking queries

Zak-Kent:


test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -480,3 +480,70 @@
...
+    (let [config (-> (create-temp-config)
+                     (assoc :database *db*)
+                     (assoc-in [:database :gc-interval] "0.01"))
+          store-report #(sync-command-post (svc-utils/pdb-cmd-url)
+                                           example-certname
+                                           "store report"
+                                           cmd-consts/latest-report-version
+                                           (change-report-time example-report %))
+          before-gc (CyclicBarrier. 2)
+          after-gc (CyclicBarrier. 2)
```
austb, line 495: Are both cyclicbarriers necessary? It looks like in the test we could only wait for after-gc and drop before-gc

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1433,6 +1433,79 @@
...
+          ;; Binding *db* to the passed in db var will cause the killer thread
+          ;; to get a new Hikari connection
+          (binding [jdbc/*db* db]
+            ;; Q: is there any possible way that all queries could be in a blocked
+            ;; state where we're unable to get another connection from the pool
+            ;; to issue the kill command?
+            (while (not @gc-finished?)
+              (jdbc/do-commands kill-blocking-qs)
+              (Thread/sleep 1000)))
+          ;; Q: do we need to worry about other exceptions here as well?
```
austb, line 1475: We could wrap it all with `with-nonfatal-exceptions-suppressed` but I don't know if that will have any difference either way the Thread is going to die and with-noisy-failure will ensure it is logged.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1433,6 +1433,79 @@
...
+    (let [gc-pid (-> "select pg_backend_pid();"
+                     jdbc/query-to-vec
+                     first
+                     :pg_backend_pid)
+          gc-finished? (atom false)
+          query-killer (spawn-query-killer db gc-pid gc-finished?)]
+      (try
+        (.start query-killer)
+        (drop-one candidate)
+        (reset! gc-finished? true)
```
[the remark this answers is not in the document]
Zak-Kent: Moving the `gc-finished?` reset into the finally block makes sense to me. One advantage of keeping the interrupt would be that it can break us out of the sleep in the killer thread's loop right away which could mean a shorter wait on the `.join` call. That's a good point about the `jdbc` stuff possibly not supporting thread interruption. We might want to add a timeout directly to the query in this case and double check what we're doing in `sync` wrt thread interruption.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1433,6 +1433,79 @@
...
+        (finally
+          ;; make sure the GC killer thread has broken out of its loop
+          (.interrupt query-killer) ; Q: do we need an interrupt as well as the gc-finished? atom?
+          (.join query-killer gc-query-killer-timeout-ms)
+          (when (.isAlive query-killer)
+            ;; if the join above timed out and the killer thread is still alive
+            ;; log an ERROR because it could indicate a memory leak
+            (log/error "Unable to clean up the gc killer thread.
```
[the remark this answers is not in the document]
Zak-Kent: It will only kill queries when it detects that the gc pid it was started with is blocked on an AccessExclusiveLock. It's possible if the killer thread got wedged and the same pid was used and blocked on that lock at a later point that it could wake up and kill queries it shouldn't. We could request a shutdown in this case if we wanted to be extra careful to never allow this to happen.

test/puppetlabs/puppetdb/cli/services_test.clj
Zak-Kent: I think you're right we only need one in this case. That was copy pasta :)

src/puppetlabs/puppetdb/scf/storage.clj
Zak-Kent: 👍 I'll double check what `with-nonfatal-exceptions-suppressed` is doing and see if we want it here.

src/puppetlabs/puppetdb/scf/storage.clj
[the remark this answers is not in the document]
austb: I think since it has to be the same pid and an AccessExclusiveLock just logging an error should be ok. I don't know of anything else that would grab an exclusive lock like that except pgrepack and it'd do the same thing.

austb (approved):

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1434,6 +1433,72 @@
...
+    (Thread.
+     #(with-noisy-failure
+        (try
+          ;; you can't use *db* value from the GC thread because the connection
+          ;; it has may already be blocked attempting to drop the partition.
+          ;; Binding *db* to the passed in db var will cause the killer thread
+          ;; to get a new Hikari connection
+          (binding [jdbc/*db* db]
+            (while (not @gc-finished?)
+              (kill-blocking-qs current_database
```
Zak-Kent, line 1467: If we did decide we wanted a 5-10 second grace period before attempting to cancel blocking queries we could add a sleep here before the kill query is issued.

PR #3390, opened 2020-12-18
(maint) Log when deleting nodes

austb:
In the success case this will create two log messages
```
INFO Starting delete of all data for host foo.example.com
INFO Finished delete of all data for host foo.example.com
```

If it errors it will throw and not log the finished messages so the logs
will look like.
```
INFO  Starting delete of all data for host foo.example.com
ERROR Failed to delete all data for host foo.example.com
<with an exception here>
```


Zak-Kent (approved):
👍

PR #3393, opened 2021-01-14
 (PDB-4811) Add tk-auth to tk-metrics endpoint

austb:
Assuming puppetdb's jetty is configured to use the same certs, metrics
can be access via authenticated GET requests like the curl command
below.

```
curl -G https://localhost:8081/metrics/v2/list \
  --cacert test-resources/ca.pem \
  --cert test-resources/localhost.pem \
  --key test-resources/localhost.key
```

### DIFF
--- a/docker/puppetdb/conf.d/auth.conf
+++ b/docker/puppetdb/conf.d/auth.conf
@@ -1,0 +1,29 @@
+authorization: {
+    version: 1
+    rules: [
+        {
+            # Allow nodes to access the metrics service
+            # for puppetdb, the metrics service is the only
+            # service using the authentication service
+            match-request: {
+                path: "/metrics/v2/([^/]+)$"
+                type: regex
+                method: [get, head, post, put]
+            }
+            allow: "*"
+            sort-order: 500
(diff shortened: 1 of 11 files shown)


austb:
@Iristyle just wanted to give you a heads up that we are adding an `auth.conf` to puppetdb. It's all the same kinds of rules that puppetserver uses, but it also has to be added to the container. I think I have it working now where the healthcheck will work unauthenticated.

docker/puppetdb/conf.d/auth.conf
```diff
@@ -0,0 +1,50 @@
...
+            sort-order: 500
+            name: "puppetlabs status service - full"
+        },
+        {
+            match-request: {
+                path: "/status/v1/simple"
+                type: path
+                method: get
+            }
+            allow-unauthenticated: true
```
Iristyle, line 21: Seems fine to me -- similar PR landing in pe-puppetdb-extensions?

austb: Yep, working on that one right now

Zak-Kent (approved):
👍

PR #3396, opened 2021-01-25
(PDB-4984) Allow drop-joins to opimize count queries

austb:
When a count function is provided, LEFT JOIN's cannot affect the total
number of results returned (unless used in a WHERE clause), so we can
allow drop-joins to remove unused LEFT JOIN's even if a count is
requested by the query.


Zak-Kent (approved):
🦖 LGTM!

rbrw (approved):

PR #3397, opened 2021-01-25
(maint) Log queries at INFO level

austb:
Query Logging is enabled by a config variable, and disabled by default,
so we can log them at info level without causing unecessary log growth.


Zak-Kent (approved):
👍

PR #3398, opened 2021-01-26
(PDB-3159) Enable performance dashboard over https

florindragos:


rbrw:
Hmm, this might need a rebase?  Wondering if it might have been started against main, but raised against 6.x (which is where I suspect we want it).

resources/public/dashboard/index.html
```diff
@@ -91,6 +91,9 @@
 
 <script src="d3.v5.js"></script>
 <script src="charts.js"></script>
+<!-- token needed here ? -->
+<!-- <script src="d3.v5.js?token="></script>
+<script src="charts.js?token="></script> -->
```
austb, line 96: Just want to make sure we remember to remove these before merging.

austb (changes requested):
This looks good, once those comments are removed this should be ready to merge.

We have quite a few features that are PE-only, but the code exists in part or in whole in the open source product, so having the token parameter here seems fine. I also tested that the page works with the certificate-allowlist and everything looks to function as expected there as well.

resources/public/dashboard/index.html
florindragos: got catch. I forgot about them 👍

austb (approved):

PR #3413, opened 2021-02-04
(PDB-4934-2) PuppetDB does not support username@hostname auth for Azure PostgreSQL

IrimieBogdan:
### The problem
For some databases the user we use for connecting to the databse can be different from the one reported by the database. For example the Azure managed PostgreSQL uses `username@hostname` to connect to a database, but the user that is reported by the database is `username`.  

PuppetDB does not make a distinction between the two users and will fail when it compares the connection user to the user reported by the database.

### The solution
Added two new users `connection-migrator-username` and `connection-username`. The new users are used to establish connections to the database when the connection username is different form the database username (this is the case for managed PostgreSQL in Azure)

`database.ini` example

```
[database]

# The database address, i.e. //HOST:PORT/DATABASE_NAME
subname = //<databse_url>:5432/puppetdb

# Credentials used to establish migration connections
connection-migrator-username = puppetdb@ghost-db

# Credentials used to establish connections
connection-username = puppetdb@ghost-db

# Connect as a specific user
username = puppetdb

# Use a specific password
password = <password_value>

# How often (in minutes) to compact the database
# gc-interval = 60
```

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -620,7 +620,7 @@
   ;; SQLException.
   (log/info (trs "Ensuring {0} database is up to date" db-name))
   (let  [connection-migrator (:connection-migrator-username db-config)
-         connection-migrator-password (:connection-migrator-password db-config)
+         migrator-password (:migrator-password db-config)
          migrator (:migrator-username db-config)]
     (with-open [db-pool (-> (assoc db-config
                                    :pool-name (if db-name
@@ -629,7 +629,7 @@
                                    :connection-timeout 3000
                                    :rewrite-batched-inserts "true"
                                    :user connection-migrator
-                                   :password connection-migrator-password)
+                                   :password migrator-password)
                             (jdbc/make-connection-pool database-metrics-registry))]
       (let [runtime (Runtime/getRuntime)
             on-shutdown (doto (Thread.
@@ -773,15 +773,12 @@
                   pool-name (if (::conf/unnamed config)
                               "PDBWritePool"
                               (str "PDBWritePool: " name))
-                  connection-username (:connection-username config)
-                  connection-password (:connection-password config)
-                  ]]
+                  connection-username (:connection-username config)]]
       (swap! pools conj
              (-> config
                  (assoc :pool-name pool-name
                         :expected-schema desired-schema
-                        :user connection-username
-                        :password connection-password)
+                        :user connection-username)
                  (jdbc/pooled-datasource database-metrics-registry))))
     {:write-db-names db-names
      :write-db-cfgs (mapv #(get databases %) db-names)
@@ -901,6 +898,7 @@
          (trs "WARNING: multiple write database support is experimental")))
       (log/warn (trs "multiple write database support is experimental")))
 
+    ; Initialize migrator pool for each DB we want to write to
     (doseq [[name config] write-dbs-config]
       (init-with-db name config))
 
@@ -910,8 +908,7 @@
                                       :pool-name "PDBReadPool"
                                       :expected-schema (desired-schema-version)
                                       :read-only? true
-                                      :user (:connection-username read-database)
-                                      :password (:connection-password read-database))
+                                      :user (:connection-username read-database))
                                (jdbc/pooled-datasource database-metrics-registry))
                    :error .close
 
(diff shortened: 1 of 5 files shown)


rbrw (commented):
From a first pass, this looks just fine.  With respect to tests, I imagine you may be able to just augment some of the existing tests, or copy and adapt them.  Could also check the git history for previous config and/or db username changes, for prior art.

src/puppetlabs/puppetdb/config.clj
```diff
@@ -398,6 +402,16 @@
...
 
+(defn ensure-connection-users-info [config]
+  ;; This expects to run after ensure-migrator-info and prefer-db-user-on-username-mismatch,
+  ;; so the :user and :migrator-user should already be set
+  (assert (:user config))
+  (assert (:migrator-username config))
+  (-> config
+      (update :connection-username #(or % (:user config)))
+      (update :connection-migrator-username #(or % (:migrator-username config))))
+  )
```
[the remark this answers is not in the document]
IrimieBogdan: Pulled the paren up.

src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -619,15 +619,17 @@
...
+         migrator (:migrator-username db-config)]
     (with-open [db-pool (-> (assoc db-config
                                    :pool-name (if db-name
                                                 (str "PDBMigrationsPool: " db-name)
                                                 "PDBMigrationsPool")
                                    :connection-timeout 3000
                                    :rewrite-batched-inserts "true"
-                                   :user migrator
-                                   :password (:migrator-password db-config))
+                                   :user connection-migrator
```
IrimieBogdan, line 631: @rbrw 
Maybe we can achieve the same thing by changing https://github.com/puppetlabs/puppetdb/blob/2d547241c619b6531dd5e98597add84d27ef6f17/src/puppetlabs/puppetdb/jdbc.clj#L642

test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -546,3 +546,23 @@
...
+          validation-fn (fn [config metrics-registry]
+                          (is (= con-mig-user (:connection-migrator-username config)))
+                          (is (= con-mig-user (:user config)))
+                          (is (= mig-pass (:migrator-password config)))
+                          (is (= mig-pass (:password config)))
+                          (throw (Exception.
+                                   "everything ok exception")))]
+      (with-redefs [jdbc/make-connection-pool validation-fn]
+        (testing "should use connection user"
+          (is (thrown? Exception ((init-with-db "test-db" {:connection-migrator-username con-mig-user
```
Zak-Kent, line 565: I think there is an extra set of parens wrapping the **init-with-db** call.

Also, might be worth using **(is (thrown-with-msg? Exception #"everything ok exception" ...))** In the assertion here. It probably would never happen, but if **init-with-db** somehow threw an Exception before it got to the assertions in the mocked **make-connection-pool** func I think the test would still pass.

test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -546,3 +546,23 @@
...
+    (let [con-mig-user "conn-migration-user-value"
+          mig-pass     "migrator-password-value"
+          user         "user-value"
+          pass         "password-value"
+          validation-fn (fn [config metrics-registry]
+                          (is (= con-mig-user (:connection-migrator-username config)))
+                          (is (= con-mig-user (:user config)))
+                          (is (= mig-pass (:migrator-password config)))
+                          (is (= mig-pass (:password config)))
+                          (throw (Exception.
```
Zak-Kent, line 561: I didn't understand why we were throwing the exception at first but I think I get it now. Is this to pop out of the mocked function after we do our assertions and avoid any of the other calls downstream from this which are expecting a connection pool obj?

Zak-Kent (commented):
Overall LGTM!

test/puppetlabs/puppetdb/cli/services_test.clj
IrimieBogdan: * removed the extra parens
* added check for exception message

test/puppetlabs/puppetdb/cli/services_test.clj
IrimieBogdan: Yes, exactly, we want to stop the execution after we exit our mock `make-connection-pool` because the rest of the code from `init-with-db` expects a valid pooled DB spec map from `make-connection-pool`. If we don't stop the execution with an exception, we would have to mock more methods or the code will try to connect to an actual database.

IrimieBogdan: I have added a comment with a short description as to why we have the exception.

austb (approved):

PR #3416, opened 2021-02-08
(maint) Update container pk8 filename

Iristyle:
- This simply makes it consistent with the filename used in
   orchestration services and console services. Changes will be made to
   ssl.sh to produce this file, and it makes sense to have all
   containers have the same naming convention

### DIFF
--- a/docker/puppetdb/docker-entrypoint.d/20-configure-ssl.sh
+++ b/docker/puppetdb/docker-entrypoint.d/20-configure-ssl.sh
@@ -10,7 +10,10 @@ fi
 
 # cert files are present from Puppetserver OR have been user supplied
 if [ -f "${SSLDIR}/private_keys/${CERTNAME}.pem" ]; then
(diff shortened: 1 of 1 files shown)


docker/puppetdb/docker-entrypoint.d/20-configure-ssl.sh
```diff
@@ -10,7 +10,10 @@ fi
 
 # cert files are present from Puppetserver OR have been user supplied
 if [ -f "${SSLDIR}/private_keys/${CERTNAME}.pem" ]; then
-  openssl pkcs8 -inform PEM -outform DER -in ${SSLDIR}/private_keys/${CERTNAME}.pem -topk8 -nocrypt -out ${SSLDIR}/private_keys/${CERTNAME}.pk8
+
+  openssl pkcs8 -topk8 -nocrypt -inform PEM -outform DER \
+    -in "${SSLDIR}/private_keys/server.key" \
+    -out "${SSLDIR}/private_keys/server.private_key.pk8"
```
Iristyle, line 16: Note that this container doesn't even use this file.

PR #3417, opened 2021-02-09
(PDB-3418) Add new-fact-time metric

sebastian-miclea:
Added new-fact-time metric

### DIFF
--- a/test/puppetlabs/puppetdb/scf/storage_test.clj
+++ b/test/puppetlabs/puppetdb/scf/storage_test.clj
@@ -90,6 +90,9 @@
       first
       :c))
 
+(defn new-facts-time []
+  (metrics.timers/number-recorded  (:add-new-fact @storage-metrics)))
+
 (deftest-db large-fact-update
   (testing "updating lots of facts"
     (let [certname "scale.com"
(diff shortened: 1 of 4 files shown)


test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -292,6 +293,19 @@
             (is (= (assoc fact-map  "one more" "here")
                    (factset-map "some_certname")))))
 
+        (defn new-facts []
+          (time! (get-storage-metric :add-new-fact) [ "new-fact-time" ]))
+        (testing "replace-facts with only additions"
```
Zak-Kent, line 298: We need to change the way we're trying to access the storage-metrics in this test. Looking at the example in the [catalog-duplicates](https://github.com/puppetlabs/puppetdb/blob/6.x/test/puppetlabs/puppetdb/scf/storage_test.clj#L1012) test, I can see that it's accessing the [storage-metrics](https://github.com/puppetlabs/puppetdb/blob/6.x/src/puppetlabs/puppetdb/scf/storage.clj#L172) atom and pulling the metrics out of it once a catalog is submitted. I think we should do something similar in this test. 

We could grab the value out of the new **_:add-new-fact_** metric first, assert it's zero, submit a factset, and then check that the metric was properly updated.

Zak-Kent (changes requested):

Zak-Kent (approved):
LGTM! 

I was going to suggest adding something to our docs for the new metric but looks like we don't maintain docs for all the metrics we collect 😿. The onlyinformationn I was able to find about the storage metrics in our docs is [here](https://puppet.com/docs/puppetdb/latest/api/metrics/v1/mbeans.html#storage-metrics). We can address adding more/better metrics documentation in a different ticket, if we decide that's something we want.

PR #3418, opened 2021-02-09
(PDB-2488) Invalid AST query is ignored

Filipovici-Andrei:
**Problem description**:
When submitting multiple queries in one request, the first one is resolved and the rest are ignored.
Example: `["from","facts"] ["not" ["=", "certname", "security-sensitive-host"]]`. This returns all the facts. 
Also if we add random strings after the query, they get ignored as well.
Example: `["from","facts"] this-is-trailing-garbage-and-not-part-of-the-query-that-gets-evaluated`. This returns all the facts.

With this PR, when multiple queries are submitted, an error message is returned, saying that no more than one 
query may be submitted in one request. 
When the query ends with a string, a `Malformed JSON` error is returned.

### DIFF
--- a/src/puppetlabs/puppetdb/http/query.clj
+++ b/src/puppetlabs/puppetdb/http/query.clj
@@ -14,13 +14,16 @@
             [puppetlabs.puppetdb.jdbc :as jdbc]
             [puppetlabs.puppetdb.http :as http]
             [puppetlabs.puppetdb.schema :as pls]
+            [ring.util.request :as request]
             [puppetlabs.puppetdb.query.paging :refer [parse-limit
                                                       parse-offset
                                                       parse-order-by
                                                       parse-order-by-json]]
             [puppetlabs.puppetdb.pql :as pql]
             [puppetlabs.puppetdb.time :refer [to-timestamp]]
-            [puppetlabs.puppetdb.utils :refer [update-when]]))
+            [puppetlabs.puppetdb.utils :refer [update-when
+                                               pprint-json-parse-exception]])
+  (:import (com.fasterxml.jackson.core JsonParseException)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Schemas
@@ -291,8 +294,11 @@
 (defn post-req->query
   "Takes a POST body and parses the JSON to create a pdb query map"
   [req parse-fn]
-  (-> (with-open [reader (-> req :body clojure.java.io/reader)]
-        (json/parse-stream reader true))
+  (-> (let [req-body (request/body-string req)]
+        (try
+          (json/parse-string req-body true)
+        (catch JsonParseException e
+          (throw (IllegalArgumentException. (tru (pprint-json-parse-exception e req-body)))))))
       (update :query (fn [query]
                        (if (vector? query)
                          query
--- a/src/puppetlabs/puppetdb/http/query.clj
+++ b/src/puppetlabs/puppetdb/http/query.clj
@@ -21,8 +21,8 @@
                                                       parse-order-by-json]]
             [puppetlabs.puppetdb.pql :as pql]
             [puppetlabs.puppetdb.time :refer [to-timestamp]]
-            [puppetlabs.puppetdb.utils :refer [update-when
-                                               pprint-json-parse-exception]])
+            [puppetlabs.puppetdb.utils :refer [update-when]]
+            [puppetlabs.puppetdb.utils.string-formatter :refer [pprint-json-parse-exception]])
   (:import (com.fasterxml.jackson.core JsonParseException)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -298,7 +298,7 @@
         (try
           (json/parse-string req-body true)
         (catch JsonParseException e
-          (throw (IllegalArgumentException. (tru (pprint-json-parse-exception e req-body)))))))
+          (throw (IllegalArgumentException. (pprint-json-parse-exception e req-body))))))
       (update :query (fn [query]
                        (if (vector? query)
                          query
--- a/src/puppetlabs/puppetdb/pql.clj
+++ b/src/puppetlabs/puppetdb/pql.clj
@@ -1,15 +1,13 @@
 (ns puppetlabs.puppetdb.pql
   (:import com.fasterxml.jackson.core.JsonParseException)
   (:require [clojure.string :refer [join]]
-            [clojure.tools.logging :as log]
             [instaparse.core :as insta]
             [instaparse.failure :as failure]
             [instaparse.print :as print]
-            [puppetlabs.i18n.core :as i18n]
+            [puppetlabs.i18n.core :refer [tru]]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.pql.transform :as transform]
-            [puppetlabs.i18n.core :refer [trs tru]]
-            [puppetlabs.puppetdb.utils :refer [pprint-json-parse-exception]]))
+            [puppetlabs.puppetdb.utils.string-formatter :refer [pprint-json-parse-exception]]))
 
 (defn transform
   "Transform parsed PQL to AST."
@@ -66,7 +64,7 @@
     (with-open [string-reader (java.io.StringReader. query)]
       (doall (json/parsed-seq string-reader)))
   (catch JsonParseException e
-    (throw (IllegalArgumentException. (tru (pprint-json-parse-exception e query)))))))
+    (throw (IllegalArgumentException. (pprint-json-parse-exception e query))))))
 
 (defn parse-json-query
   "Parse a query string as JSON. Multiple queries or any other
@@ -78,7 +76,7 @@
       (if (= query-count 1)
         (first parsed-query)
         (throw (IllegalArgumentException.
-                 (i18n/tru "Only one query may be sent in a request. You sent {0}." query-count)))))))
+                 (tru "Only one query may be sent in a request. You sent {0}." query-count)))))))
 
 (defn parse-pql-query
   "Parse a query string as PQL. Parse errors will result in an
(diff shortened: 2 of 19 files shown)


Filipovici-Andrei:
Should we change the error message when we receive a malformed JSON? 
Instead of returning `Malformed JSON for query: ["from", "nodes"] string`, we could return something like
`Malformed JSON: Unrecognized token 'string': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: (StringReader); line: 1, column: 25]` (I used the message from the JsonParseException)

This can be improved to look similar to what the PQL parser returns.

Zak-Kent (commented):
@Filipovici-Andrei I think returning information about where the parsing went wrong makes sense to me if we can easily expose that information to the caller.  

I remember we talked a little before about the code [here](https://github.com/puppetlabs/puppetdb/blob/6.x/src/puppetlabs/puppetdb/http/query.clj#L294-L295) which will parse the json-stream for :post requests and then we put it back into a **java.io.StringReader** so we can call **parse-seq** on it in this change. I think the way you have it now might be what we want, I just wanted to call it out again in case there are any performance concerns. @rbrw @austb do y'all have any opinions on this? I think to change the approach in the pr to do the validation at a higher level would require a significant amount of change/churn from what I can tell looking at it briefly. 

Other than the one comment about the **if** block and the question above the PR looks good to me.

src/puppetlabs/puppetdb/pql.clj
```diff
@@ -57,16 +57,28 @@
...
+      (doall (json/parsed-seq string-reader)))
+  (catch JsonParseException e
+    (throw (IllegalArgumentException.
+             (tru "Malformed JSON for query: {0}" query))))))
+
+(defn parse-json-query
+  "Parse a query string as JSON. Multiple queries or any other
+  data, after the query, will result in an IllegalArgumentException"
+  [query]
+  (if query
```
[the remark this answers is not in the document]
Filipovici-Andrei: This is for the case when `extract-query` uses `parse-json-query` directly, not through `parse-json-or-pql-to-ast` which has the regex to differentiate between AST and PQL. 
If the request does not have a `body`, then `java.io.StringReader. query` throws NullPointerException.

Filipovici-Andrei:
> @Filipovici-Andrei I think returning information about where the parsing went wrong makes sense to me if we can easily expose that information to the caller.
> 
> I remember we talked a little before about the code [here](https://github.com/puppetlabs/puppetdb/blob/6.x/src/puppetlabs/puppetdb/http/query.clj#L294-L295) which will parse the json-stream for :post requests and then we put it back into a **java.io.StringReader** so we can call **parse-seq** on it in this change. I think the way you have it now might be what we want, I just wanted to call it out again in case there are any performance concerns. @rbrw @austb do y'all have any opinions on this? I think to change the approach in the pr to do the validation at a higher level would require a significant amount of change/churn from what I can tell looking at it briefly.
> 
> Other than the one comment about the **if** block and the question above the PR looks good to me.

We need to perform both parsings, because `json/parse-stream` parses the request's body into query, order_by, limit, etc. fields. If the query's body is unclosed with `"`, then it's content will not be validated. That's why we need to parse it independently, again.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -644,3 +643,18 @@
 
 (defn schedule-with-fixed-delay [s f initial-delay delay]
   (.scheduleWithFixedDelay s f initial-delay delay TimeUnit/MILLISECONDS))
+
+(defn pprint-json-parse-exception
+  "Returns a parsed JsonParseException message"
+  [error query]
+  (let [lines (str/split-lines (.getMessage error))
+        positions (re-find #"line: (\d+), column: (\d+)" (lines 1))
```
Zak-Kent, line 651: I noticed this regex was getting the index values you wanted and also picking up the `"line: 1 column: 5"` part of the string. I saw it getting the following array in the `reports_test.clj/query-report-with-malformed-json` test: `["line: 1, column: 5" "1" "5"]`. If it's easy to restrict the regex to get only the index values we might want that. 

We could also use destructuring here in the `let` block to pull out the `line-index`and `column-index` values from the `re-find`. Right now with it would look something like this: 
`[_ line-index column-index] (re-find #"line: (\d+), column: (\d+)" (second lines))`
but we could drop the throw away `_` binding there if we restrict the regex. We'd still need to convert the strings to ints later on though, so the destructuring might not be ideal here unless we can make sure we'll only ever get the index values back from the regex and convert them to ints before binding them.

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -644,3 +643,18 @@
...
+        line-index (Integer. (positions 1))
+        column-index (Integer. (positions 2))
+        error-line-index (- line-index 1)
```
Zak-Kent, line 654: `(dec line-index)` is another way to write this

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -644,3 +643,18 @@
...
+        error-column-index (- column-index 2)]
+
+    (str "Json parse error at line " line-index ", column " column-index ":\n\n"
+         ((str/split-lines query) error-line-index) "\n"
+         (apply str (concat (repeat error-column-index " ") "^")) "\n\n"
+         (lines 0))))
```
Zak-Kent, line 660: `(first lines)` is another way to write this

src/puppetlabs/puppetdb/utils.clj
```diff
@@ -644,3 +643,18 @@
...
+
+(defn pprint-json-parse-exception
+  "Returns a parsed JsonParseException message"
+  [error query]
+  (let [lines (str/split-lines (.getMessage error))
+        positions (re-find #"line: (\d+), column: (\d+)" (lines 1))
+        line-index (Integer. (positions 1))
+        column-index (Integer. (positions 2))
+        error-line-index (- line-index 1)
+        error-column-index (- column-index 2)]
```
Zak-Kent, line 655: How did you know we needed to subtract two off the column index to get the `^` in the line underneath to line up with the error in the output? Will this always be the case? It might be useful to add a little more to the doc string of this func to explain how it pulls out the error message and adds a `^` pointer to the location that the JsonParseException is referencing.

Zak-Kent (commented):
Overall looks really good and the error output is nice. I tested this running invalid queries against it locally and it was obvious where the error was in each.

src/puppetlabs/puppetdb/utils.clj
Filipovici-Andrei: I don't think there's a way to return the capturing groups without the full match

src/puppetlabs/puppetdb/utils.clj
Filipovici-Andrei: I subtract 2 from the column index because the indexes returned by the error don't start from 0 and I need to replace the last `space` that I print with the `^` mark.

src/puppetlabs/puppetdb/utils/string_formatter.clj
```diff
@@ -0,0 +1,101 @@
...
+
+(defn pprint-json-parse-exception
+  "Returns a parsed JsonParseException message.
+  From the second line of the error message, the line and
+  column index are extracted, so that we can place an arrow,
+  pointing at that position. "
+  [error query]
+  (let [error-lines (string/split-lines (.getMessage error))
+        [_ line column] (re-find #"line: (\d+), column: (\d+)" (second error-lines))
+        line-index (Integer. line)
```
Filipovici-Andrei, line 89: Probably the conversion to Integer and the subtraction can be performed nicer, but I don't know how.

jonathannewman: An alternative to the integer constructor:  https://github.com/puppetlabs/clj-kitchensink/blob/main/src/puppetlabs/kitchensink/core.clj#L208-L216 but will return nil.  
I believe if the parse fails now, it will throw an exception.  I wonder what is the user visible output in that case.

src/puppetlabs/puppetdb/utils/string_formatter.clj
```diff
@@ -0,0 +1,101 @@
...
+        column-index (Integer. column)
+        ;subtracting 1 from the line and column index,
+        ;because they start from 1, not 0
+        query-line-index (dec line-index)
+        query-column-index (dec column-index)]
+
+    (str "Json parse error at line " line-index ", column " column-index ":\n\n"
```
jonathannewman, line 96: @Zak-Kent are we using i18n consistently in PDB?  If so, these should likely be using tru (trs?) : https://github.com/puppetlabs/clj-i18n#developer-usage

Zak-Kent: Good catch, I wasn't even thinking about the i18n stuff. It looks like the the result of this func is wrapped in a `tru` in this PR but we could move that wrapping inside this func so we don't need to remember it in the future.

src/puppetlabs/puppetdb/pql.clj
```diff
@@ -8,7 +8,8 @@
             [puppetlabs.i18n.core :as i18n]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.pql.transform :as transform]
-            [puppetlabs.i18n.core :refer [trs tru]]))
+            [puppetlabs.i18n.core :refer [trs tru]]
```
Zak-Kent, line 11: duplicate require

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -291,8 +294,11 @@
 (defn post-req->query
   "Takes a POST body and parses the JSON to create a pdb query map"
   [req parse-fn]
-  (-> (with-open [reader (-> req :body clojure.java.io/reader)]
-        (json/parse-stream reader true))
+  (-> (let [req-body (request/body-string req)]
+        (try
+          (json/parse-string req-body true)
+        (catch JsonParseException e
+          (throw (IllegalArgumentException. (tru (pprint-json-parse-exception e req-body)))))))
```
jonathannewman, line 301: Wrapping here won't identify the string for translation properly and thus won't be translated.  The extraction process expects strings to be embedded directly in the `tru` `trs`, etc functions.

Zak-Kent: ah, yep

src/puppetlabs/puppetdb/utils/string_formatter.clj
Zak-Kent: I think if we wrap that top line we're creating here in a `tru` that might be what we want. It would look like this: 
```(str (tru "Json parse error at line {0}, column {1}:\n\n" line-index column-index) ...)``` I think we can get away with not translating the rest because it's coming from the query the user provided and the stack trace from the JsonParseException.

Zak-Kent (approved):
👍 LGTM

PR #3419, opened 2021-02-10
(PDB-5017) Add missing arg to admin purge_resource_events path

Zak-Kent:
Prior to this commit there was an arg missing from the
scf-store/delete-resource-events-older-than! function which is
used when the admin endpoint gets a purge_resource_events clean
command.


austb (approved):
🦖

PR #3420, opened 2021-02-11
(PDB-4172) Allow metrics during startup

oanatmaria:
An unit test, that checks that metrics can be retrieved in maintenance mode, was added.


Zak-Kent (approved):
👍

PR #3421, opened 2021-02-15
(PDB-4268) Add new metric for concurrent enqueue

oanatmaria:
A new metric **(:concurrent-depth),** which counts the number of `/cmd` API requests that are waiting to write to the disk, was added.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -121,6 +121,7 @@
 ;;         :retry-counts <histogram>
 ;;         :size <histogram>
 ;;         :ignored <counter>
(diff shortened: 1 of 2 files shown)


Zak-Kent (commented):
I pushed a branch using this work with an added throw away commit that has examples of some of what I was talking about with reworking the test. See [pdb-4268-test-changes](https://github.com/Zak-Kent/puppetdb/commit/8e7d26bdff01f8a7f1cf9da7659355b27f6e650d)

Zak-Kent (approved):
👍 LGTM!

Similar to the new-fact-time metric we'll want to add better docs for all the metrics we collect but we can do that in a separate ticket.

PR #3425, opened 2021-02-23
(PDB-3635) Add ability to disable storage of resource events

florindragos:


src/puppetlabs/puppetdb/command.clj
```diff
@@ -467,7 +470,9 @@
                "deactivate node" exec-deactivate-node
                "configure expiration" exec-configure-expiration
                "replace catalog inputs" exec-replace-catalog-inputs)]
-    (exec cmd start db conn-status)))
+    (if (= command "store report")
```
Zak-Kent, line 473: It looks like `exec-store-report` is only called in the `exec-command` func. If we make the config value the first argument to `exec-store-report` we could then use `(partial exec-store-report blocklist-config)` in the `case` statement above and wouldn't need to add the if condition for `"store report"` at the bottom of this func.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1739,19 +1740,22 @@
   ([{:keys [certname producer_timestamp] :as report} :- reports/report-wireformat-schema
    received-timestamp :- pls/Timestamp
    db conn-status]
-   (add-report! report received-timestamp db conn-status true))
+   (add-report! report received-timestamp db conn-status true {}))
```
Zak-Kent, line 1743: we'll need to adjust how this adds the extra arity to the `add-report!` func. We'll still need to support calls with only 5 args. I think if we remove the defaulting for the `blocklist-config` add another block like the one below in addition to the final 6 arg one you added it should work. 
```
([{:keys [certname producer_timestamp] :as report} :- reports/report-wireformat-schema
    received-timestamp :- pls/Timestamp
    db conn-status update-latest-report?]
   (add-report! report received-timestamp db conn-status update-latest-report? {}))
 ```

**Review by Zak-Kent — COMMENTED** [review]
The general approach looks good to me. I think we might need to account for cases where `resource-events-ttl` isn't set based on a few null pointer errors I noticed in the travis tests but I didn't dig around further.

File: src/puppetlabs/puppetdb/command.clj [hunk 3]
```diff
@@ -529,12 +534,15 @@
...
+  ([{:keys [callback] :as cmd} db conn-status response-pub-chan stats-atom
    shutdown-for-ex]
+   attempt-exec-command cmd db conn-status response-pub-chan stats-atom shutdown-for-ex {})
+  ([{:keys [callback] :as cmd} db conn-status response-pub-chan stats-atom
+   shutdown-for-ex blocklist-config]
   (try
     (let [result (call-with-quick-retry quick-retry-count
                                         shutdown-for-ex
-                                        #(exec-command cmd db conn-status (now)))]
+                                        #(exec-command cmd db conn-status (now) blocklist-config))]
```
Zak-Kent, line 545: We might want to rename `blocklist-config` to something more general now that it's holding blocklist settings and resource-events-ttl.

Zak-Kent (commented):
Approach looks good overall. We'll still want a test that checks setting `resource-event-ttl` to zero disables the storage. I think you might be able to follow the pattern in [integration/resource_events.clj](https://github.com/puppetlabs/puppetdb/blob/6.x/test/puppetlabs/puppetdb/integration/resource_events.clj) to setup a pdb with the storage disabled and then submit a report with resource-events and check that they're not stored. Or it could be tested in the http tests if there is an easy enough way to change the pdb config in the macros they use. There were a few stray commented lines in the pr that need to be dropped as well.

We'll also want to investigate what sync does if one pdb has the `resource-event-ttl` set to zero when the other side has a different ttl. We can add a sync tests in another pr/ticket. I can help make the follow up ticket for that work this week.

src/puppetlabs/puppetdb/command.clj
florindragos: changed to `options-config`

src/puppetlabs/puppetdb/command.clj
```diff
@@ -456,28 +458,30 @@
...
   "Takes a command object and processes it to completion. Dispatch is
    based on the command's name and version information"
-  [command db blocklist-config]
+  [command db options-config]
   (when-not (:delete? command)
     (let [start (now)]
       (-> command
-          (prep-command blocklist-config)
+          (prep-command options-config)
           (exec-command db (atom {}) start)))))
```
austb, line 485: For `store-report` we need to pass the options-config to `exec-command` as well.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -456,28 +458,30 @@
 (defn exec-command
   "Takes a command object and processes it to completion. Dispatch is
    based on the command's name and version information"
-  [{:keys [command version] :as cmd} db conn-status start]
+  ([{:keys [command version] :as cmd} db conn-status start]
+    (exec-command cmd db conn-status start {}))
```
austb, line 462: `exec-store-report` needs the `options-config` every time so we should require it as an argument

PR #3429, opened 2021-02-25
(PDB-4524) PuppetDBCommandDispatcher: support pause and resume 

IrimieBogdan:
Added `pause-execution` and `resume-execution` service functions to allow pausing of command execution.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -511,7 +511,12 @@
 
   (response-mult [this]
     "Returns a core.async mult to which {:id :exception} maps are written after
-     each command has been processed. " ))
+     each command has been processed.")
+
+  (pause-execution [this]
+    "Pause command execution")
+  (resume-execution [this]
+    "Resume command execution"))
 
 (defn make-cmd-processed-message [cmd ex]
   (conj
@@ -848,7 +853,8 @@
                response-pub (async/pub response-chan-for-pub :id) :error async/unsub-all
                delay-pool (scheduler delay-pool-size) :error #(shut-down-command-scheduler-or-die
                                                                % request-shutdown)
-               concurrent-writes (get-in config [:command-processing :concurrent-writes])]
+               concurrent-writes (get-in config [:command-processing :concurrent-writes])
+               blocked? (atom false)]
     (async/tap response-mult response-chan-for-pub)
     (assoc context
            :delay-pool delay-pool ;; For general scheduling, not just delays
@@ -858,7 +864,8 @@
            :response-chan response-chan
            :response-mult response-mult
            :response-chan-for-pub response-chan-for-pub
-           :response-pub response-pub)))
+           :response-pub response-pub
+           :blocked? blocked?)))
 
 (defn start-command-service
   [context config {:keys [dlo] :as globals} request-shutdown]
@@ -918,9 +925,10 @@
       ;; bubble up and be reported.
 
       ;; FIXME: check that nil command-chan is actually fatal if appropriate
-      (let [shovel #(with-monitored-execution shutdown-for-ex
+      (let [blocked? (context :blocked?)
+            shovel #(with-monitored-execution shutdown-for-ex
                       (try
-                        (pool/dochan command-pool handle-cmd command-chan)
+                        (pool/dochan command-pool handle-cmd command-chan blocked?)
                         (catch ExceptionInfo ex
                           (when-not (= :puppetlabs.puppetdb.threadpool/rejected
                                        (:kind (ex-data ex)))
@@ -941,7 +949,7 @@
 (defn stop-command-service
   [{:keys [consumer-threadpool broadcast-threadpool command-chan
            response-chan response-chan-for-pub response-mult response-pub
-           delay-pool command-shovel]
+           delay-pool command-shovel blocked?]
     :as context}
    request-shutdown]
   ;; Must also handle a nil context.
@@ -956,6 +964,9 @@
    (finally (when command-shovel (.interrupt command-shovel)))
    (finally (some-> broadcast-threadpool pool/shutdown-unbounded))
    (finally (some-> consumer-threadpool pool/shutdown-gated))
+
+   ; we don't unblock the worker pool from dochan because the command-shovel
+   ; will be interrupted and we don't want to process any request before that
    (finally (some-> delay-pool (shut-down-command-scheduler-or-die request-shutdown)))
    (finally
      (when command-shovel
@@ -969,6 +980,14 @@
    (finally (some-> response-mult async/untap-all))
    (finally (some-> response-chan-for-pub async/close!))
    (finally (some-> response-chan async/close!))))
+
+(defn pause-command-service
+  [blocked?]
+  (locking blocked? (reset! blocked? true) (.notifyAll blocked?)))
+
+(defn resume-command-service
+  [blocked?]
+  (locking blocked? (reset! blocked? false) (.notifyAll blocked?)))
 
 (defn throw-unless-ready [context]
   (when-not (seq context)
@@ -1032,4 +1051,12 @@
    [this]
    (throw-unless-ready (service-context this))
    (throw-if-shutdown-pending (get-shutdown-reason))
-   (-> this service-context :response-mult)))
+   (-> this service-context :response-mult))
+
+  (pause-execution
+    [this]
+    (pause-command-service (-> this service-context :blocked?)))
+
+  (resume-execution
+    [this]
+    (resume-command-service (-> this service-context :blocked?))))
--- a/test/puppetlabs/puppetdb/command_test.clj
+++ b/test/puppetlabs/puppetdb/command_test.clj
@@ -72,7 +72,8 @@
             [puppetlabs.puppetdb.testutils :as tu]
             [puppetlabs.puppetdb.time :as t]
             [puppetlabs.puppetdb.client :as client]
-            [puppetlabs.puppetdb.threadpool :as pool])
+            [puppetlabs.puppetdb.threadpool :as pool]
+            [puppetlabs.puppetdb.utils :refer [await-ref-state]])
   (:import
    (clojure.lang ExceptionInfo)
    (java.nio.file Files)
@@ -1882,6 +1883,43 @@
 
            (testing "concurrent-depth metric should be 0"
              (is (= 0 (concurrent-depth))))))))))
+
+(deftest block-command-processing
+  (with-test-db
+    (svc-utils/with-pdb-with-no-gc
+      (let [dispatcher (get-service svc-utils/*server* :PuppetDBCommandDispatcher)
+            stats (-> dispatcher service-context :stats)
+            new-producer-ts (now)
+            base-cmd (get-in wire-catalogs [9 :basic])
+            enqueue-catalog (fn []
+                              (enqueue-command
+                                dispatcher
+                                (command-names :replace-catalog)
+                                9
+                                (:certname base-cmd)
+                                nil
+                                (->  base-cmd
+                                     (assoc :producer_timestamp new-producer-ts)
+                                     tqueue/coerce-to-stream)
+                                ""))]
+
+        (testing "should pause execution"
+          (pause-execution dispatcher)
+          (enqueue-catalog)
+
+          (is (= 1 (-> stats
+                       (await-ref-state #(= 1 (% :received-commands))
+                                        default-timeout-ms :timeout)
+                       :received-commands))))
+
+          (is (= :timeout (await-ref-state stats #(= 1 (% :executed-commands)) 500 :timeout)))
+
+        (testing "should resume execution"
+          (resume-execution dispatcher)
+
+          (is (not= :timeout (await-ref-state stats #(= (% :executed-commands) 1)
+                                              default-timeout-ms :timeout))))))))
+
 
 (deftest missing-queue-files-accounted-for-in-stats
   ;; Use a lock to hold up command processing while we trash the
--- a/test/puppetlabs/puppetdb/threadpool_test.clj
+++ b/test/puppetlabs/puppetdb/threadpool_test.clj
@@ -3,7 +3,8 @@
             [clojure.test :refer :all]
             [puppetlabs.puppetdb.threadpool :refer :all]
             [puppetlabs.puppetdb.testutils :as tu]
-            [puppetlabs.puppetdb.utils :refer [await-ref-state]]
+            [puppetlabs.puppetdb.utils :refer [await-ref-state
+                                               noisy-future]]
             [puppetlabs.trapperkeeper.testutils.logging :refer [atom-logger
                                                                 with-logging-to-atom
                                                                 with-log-suppressed-unless-notable]]
@@ -31,8 +32,9 @@
                        [:done id]))]
 
       (with-open [gtp (gated-threadpool 1 "test-pool-%d" 1000)]
-        (let [fut (future
-                    (dochan gtp worker-fn in-chan))]
+        (let [blocked? (atom false)
+              fut (future
+                    (dochan gtp worker-fn in-chan blocked?))]
           (async/>!! in-chan {:id :a})
           (is (= [:done :a] (async/<!! out-chan)))
           (is (= 1 @counter))
@@ -53,8 +55,9 @@
                        [:done (:foo some-command)]))]
 
       (with-open [gtp (gated-threadpool 1 "test-pool-%d" 1000)]
-        (let [fut (future
-                    (dochan gtp worker-fn in-chan))]
+        (let [blocked? (atom false)
+              fut (future
+                    (dochan gtp worker-fn in-chan blocked?))]
 
           (async/>!! in-chan {:foo 1})
           (async/close! in-chan)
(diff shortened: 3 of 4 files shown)


src/puppetlabs/puppetdb/command.clj
```diff
@@ -990,6 +999,8 @@
 
   (start
    [this context]
+   (pause-execution this)
+   (resume-execution this)
```
IrimieBogdan, line 1003: Used this for testing

austb:
I talked with Rob today, and with a lock like this we'd need to ensure that the resume is called from the same thread as the pause is called from, which may not be the case if/when we hook this up to an API endpoint. So we think this needs to use a piece of global state, likely a boolean in an atom, and then use Java monitors (which are really just conditioned variables that Java picked a new name for) to handle the coordination.

The result I think would be a set of one-liners in clojure in the various locations.
```
;; To start blocking the command processor
(locking blocked? (reset! blocked? true) (.notifyAll blocked?))
```
```
;; To unblock the command processor
(locking blocked? (reset! blocked? false) (.notifyAll blocked?))
```
```
;; Before we get a command to process we would check the locked status
(locking blocked? (while @blocked? (.wait blocked?))
```

The other caveat we came up with is that we'll want to ensure that when shutting down the trapperkeeper service that we unblock the threads as well so that they aren't stuck in that stopped state. To ensure that we don't attempt to process another command while shutting down, we would also want to ensure that the threadpool is shutdown first.

[docs for the wait and notify functions can be found on the Java Object page](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#wait--)

IrimieBogdan:
@austb thanks for the suggestion and the detailed response! Made the changes, plase let me know what you think about the new version.

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -438,3 +438,6 @@
         table]
        jdbc/query-to-vec
        (map :indexdef)))
+
+; used to pause/resume command processing/execution
+(def blocked? (atom false))
```
IrimieBogdan, line 443: Save the atom in the context

src/puppetlabs/puppetdb/command.clj
```diff
@@ -956,6 +964,7 @@
    (finally (when command-shovel (.interrupt command-shovel)))
    (finally (some-> broadcast-threadpool pool/shutdown-unbounded))
    (finally (some-> consumer-threadpool pool/shutdown-gated))
+   (finally (when blocked? (locking blocked? (reset! blocked? false) (.notifyAll blocked?))))
```
rbrw, line 967: After a bit more thought (and if I'm thinking straight), I believe we can just remove this, given the `(.interrupt command-shovel)` above.  i.e. the interrupt should cause it to throw back to the catch in the shovel thread and then exit (even) if the thread actually is blocked in `.wait`, and other than the shovel thread itself, there's nothing else to notify.  It also might be better to stay blocked during shutdown anyway, i.e. shutting down perhaps shouldn't (even temporarily) unblock us, from a semantic perspective.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -992,7 +1009,8 @@
    [this context]
    (call-unless-shutting-down
     "command service start"  (get-shutdown-reason) context
-    #(start-command-service context (get-config) (shared-globals) request-shutdown)))
+    #(start-command-service context (get-config) (shared-globals) request-shutdown))
+   )
```
rbrw, line 1013: Trailing paren.

src/puppetlabs/puppetdb/command.clj
```diff
@@ -1032,4 +1050,16 @@
...
+    [this]
+    (let [context (service-context this)
+          blocked? (context :blocked?)]
+      (pause-command-service blocked?)))
+
+  (resume-execution
+    [this]
+    (let [context (service-context this)
+          blocked? (context :blocked?)]
+      (resume-command-service blocked?))))
```
rbrw, line 1065: Not critical, but could rewrite these a bit -- one option: `(pause-command-service (-> this service-context :blocked?))`.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1883,6 +1883,49 @@
            (testing "concurrent-depth metric should be 0"
              (is (= 0 (concurrent-depth))))))))))
 
+(deftest block-command-processing
+  (with-test-db
+    (svc-utils/call-with-puppetdb-instance
+      (assoc (svc-utils/create-temp-config)
+        :database *db*
+        :command-processing {:concurrent-writes 1})
```
rbrw, line 1891: Hmm, not sure we need the concurrent-write limit?  That's on the ingest side iirc.  And I suspect you might be able to use `with-pdb-with-no-gc` and avoid needing to explicitly specify the database/config at all, or have the thunk/wrapper i.e. the `(fn [] ...)`.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1883,6 +1883,49 @@
...
+                                  ""))
+              cmds-enqueued (CountDownLatch. 1)]
+
+          (with-redefs [cmd/process-message (fn [cmdref
+                                                q command-chan dlo
+                                                delay-pool broadcast-pool write-dbs
+                                                response-chan stats
+                                                blocklist-config maybe-send-cmd-event!
+                                                shutdown-for-ex]
+                                             (.countDown cmds-enqueued))]
```
rbrw, line 1915: Looks like the indentation may be a bit off inside this `fn`, and if we don't need any of the args, then I might be tempted to just say `(fn [& _] (.countDown cmds-enqueued))`.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1883,6 +1883,49 @@
...
+            (testing "should pause execution"
+              (pause-execution dispatcher)
+
+              (enqueue-catalog)
+              (is (= false (.await cmds-enqueued 1000 TimeUnit/MILLISECONDS)))
+              )
```
rbrw, line 1921: Trailing paren.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1883,6 +1883,49 @@
...
+                                                delay-pool broadcast-pool write-dbs
+                                                response-chan stats
+                                                blocklist-config maybe-send-cmd-event!
+                                                shutdown-for-ex]
+                                             (.countDown cmds-enqueued))]
+            (testing "should pause execution"
+              (pause-execution dispatcher)
+
+              (enqueue-catalog)
+              (is (= false (.await cmds-enqueued 1000 TimeUnit/MILLISECONDS)))
```
rbrw, line 1920: Not critical, but could even just say `(is (not (.await ...)))` unless we care about false specifically.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1883,6 +1883,49 @@
...
+              )
+
+            (testing "should resume execution"
+              (resume-execution dispatcher)
+              (is (= true (.await cmds-enqueued default-timeout-ms TimeUnit/MILLISECONDS)))
+              )))))))
```
rbrw, line 1926: I might just do something like this to fix the trailing parens:
```
(is (= true (.await cmds-enqueued default-timeout-ms
                    TimeUnit/MILLISECONDS))))))))))
```

test/puppetlabs/puppetdb/testutils.clj
rbrw, line 443: This might be vestigial now?

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1883,6 +1883,49 @@
            (testing "concurrent-depth metric should be 0"
              (is (= 0 (concurrent-depth))))))))))
 
+(deftest block-command-processing
```
rbrw, line 1886: I'm not positive yet, but I suspect we might be able to remove the need for the process-message `with-redefs` here by just waiting on changes to the `(-> cmd-service-context :stats)` atom, i.e. it has `:received-commands` and `:executed-commands` counts, and we could just pause commands, submit the command, wait for `:received-commands` to go to 1, then make sure `:executed-commands` stays zero for more than a second or two, unpause commands, and then wait for `:executed-commands` to go to 1.  If we end up thinking that seems reasonable, then `await-ref-state` might be useful.

rbrw (commented):
Didn't quite have time to finish this today (for example, haven't looked at the threadpool test at all yet), but thought I'd post what I had.  I should have time to finish tomorrow.

src/puppetlabs/puppetdb/command.clj
IrimieBogdan: changed

test/puppetlabs/puppetdb/command_test.clj
IrimieBogdan: changed

test/puppetlabs/puppetdb/command_test.clj
IrimieBogdan: changed

test/puppetlabs/puppetdb/command_test.clj
IrimieBogdan: Fixed the trailing parens and removed `(= true ...)` and rewrite it as
```
              (is (.await cmds-enqueued default-timeout-ms TimeUnit/MILLISECONDS)))))))))
```

test/puppetlabs/puppetdb/command_test.clj
IrimieBogdan: Changed to use `with-pdb-with-no-gc` as I don't really need to make any changes to database config.

src/puppetlabs/puppetdb/command.clj
rbrw: OK, so then, if I follow you, perhaps we can remove that `finally` code without any issue.

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1883,6 +1884,37 @@
...
+        (testing "should pause execution"
+          (pause-execution dispatcher)
+          (enqueue-catalog)
+
+          (is (= :timeout (await-ref-state stats #(= (% :executed-commands) 1), 500 :timeout))))
+
+        (testing "should resume execution"
+          (resume-execution dispatcher)
+
+          (is (not= :timeout (await-ref-state stats #(= (% :executed-commands) 1), default-timeout-ms :timeout))))))))
```
rbrw, line 1915: Suggest wrapping this somehow, at say 80-100 chars.  Also suggest removing commas (they're generally omitted).

test/puppetlabs/puppetdb/command_test.clj
```diff
@@ -1883,6 +1884,37 @@
...
+                                nil
+                                (->  base-cmd
+                                     (assoc :producer_timestamp new-producer-ts)
+                                     tqueue/coerce-to-stream)
+                                ""))]
+
+        (testing "should pause execution"
+          (pause-execution dispatcher)
+          (enqueue-catalog)
+
```
rbrw, line 1909: Particularly given that CI has sometimes been *surprisingly* slow (which is how we ended up with such a high `default-ttimeout-ms`), might be worth adding another check that we're "closer" to the state we want before we block the 500ms, i.e. maybe add something like this here:
```
(is (= 1 (await-ref-state stats #(= 1 (% :received-commands))
                          default-timeout-ms :timeout))))))))
```

test/puppetlabs/puppetdb/testutils.clj
rbrw: Yeah, I'd suggest just adding it to each test so we don't have to worry about any possible future test concurrency, etc.  Could put it in a let binding, or, since we don't really need it, perhaps also fine to just put it in line for now: `(dochan gtp worker-fn in-chan (atom false))`

test/puppetlabs/puppetdb/threadpool_test.clj
```diff
@@ -32,14 +32,39 @@
...
+                    (dochan gtp worker-fn in-chan tu/blocked?))]
           (async/>!! in-chan {:id :a})
           (is (= [:done :a] (async/<!! out-chan)))
           (is (= 1 @counter))
           (async/close! in-chan)
           (is (not= ::timed-out (deref fut tu/default-timeout-ms ::timed-out)))
           (is (true? (future-done? fut)))))))
 
+
+  (testing "block command"
```
rbrw, line 44: No strong feelings, but if there's no notable extra cost, I might even consider pulling this out into its own top-level test so it's more clearly isolated, easier to run independently, etc. e.g.
```
(deftest dochan-blocking-and-unblocking ...)
```

test/puppetlabs/puppetdb/threadpool_test.clj
```diff
@@ -32,14 +32,39 @@
...
+          out-chan (async/chan 10)
+          worker-fn (wrap-out-chan
+                      out-chan
+                      (fn [{:keys [id]}]
+                        (swap! counter inc)
+                        [:done id]))
+          blocked? (atom true)]
+
+        (with-open [gtp (gated-threadpool 1 "test-pool-%d" 1000)]
+          (let [fut (future
```
rbrw, line 56: Suggest in most cases using a `noisy-future` so we'll make sure to see any errors, unless we know we're going to always deref it, and we want to be able to see the exception then.

test/puppetlabs/puppetdb/threadpool_test.clj
```diff
@@ -32,14 +32,39 @@
...
+                      (dochan gtp worker-fn in-chan blocked?))]
+            (async/>!! in-chan {:id :a})
+            (let [channel (future (async/<!! out-chan))
+                  ]
```
rbrw, line 60: Trailing bracket.

test/puppetlabs/puppetdb/threadpool_test.clj
```diff
@@ -32,14 +32,39 @@
...
+              (is (= (deref channel 1000 "Did not wake up!") "Did not wake up!") )
+              )
```
rbrw, line 62: Trailing paren.

test/puppetlabs/puppetdb/threadpool_test.clj
```diff
@@ -32,14 +32,39 @@
...
+                        [:done id]))
+          blocked? (atom true)]
+
+        (with-open [gtp (gated-threadpool 1 "test-pool-%d" 1000)]
+          (let [fut (future
+                      (dochan gtp worker-fn in-chan blocked?))]
+            (async/>!! in-chan {:id :a})
+            (let [channel (future (async/<!! out-chan))
+                  ]
+              (is (= (deref channel 1000 "Did not wake up!") "Did not wake up!") )
```
rbrw, line 61: Extra space at the end of the line, and also suggest putting the expected value first in `is` tests, i.e. `(is (= EXPECTED ACTUAL))`.  I believe that's more idiomatic, and I *think* some tooling might expect it (when rendering results, etc.).

test/puppetlabs/puppetdb/threadpool_test.clj
```diff
@@ -32,14 +32,39 @@
...
+              )
+            (is (= 0 @counter))
+            (async/close! in-chan)
+            (is (= ::timed-out (deref fut 1000 ::timed-out)))
+            (is (true? (future-cancel fut)))))))
```
rbrw, line 66: I might lean toward just `(is (future-cancel fut))`.

test/puppetlabs/puppetdb/command_test.clj
IrimieBogdan: I can add the check, but don't know if it make a difference because the `stats` atom is incremented in [enqueue-command ](https://github.com/puppetlabs/puppetdb/blob/b2b783d4f46daa35433e9d90645076b26faaa2b1/src/puppetlabs/puppetdb/command.clj#L1033 ) which is the first thing we do in the test `enqueue-catalog -> enqueue-command`

IrimieBogdan: re-wrote it to
``` clojure
          (is (= 1 (-> stats
                       (await-ref-state #(= 1 (% :received-commands))
                                        default-timeout-ms :timeout)
                       :received-commands))))
```

test/puppetlabs/puppetdb/command_test.clj
IrimieBogdan: Removed commas and wrapped thee line.

test/puppetlabs/puppetdb/testutils.clj
IrimieBogdan: Removed `(def blocked? (atom false))` and made an `atom` in each test that required it.

test/puppetlabs/puppetdb/threadpool_test.clj
IrimieBogdan: Replaced `future` with `noisy-future`

test/puppetlabs/puppetdb/threadpool_test.clj
IrimieBogdan: Extracted the test in a top level test, added test for unblocking, decreased the timeout for blocking to `500 ms`.

test/puppetlabs/puppetdb/threadpool_test.clj
IrimieBogdan: removed

test/puppetlabs/puppetdb/threadpool_test.clj
IrimieBogdan: changed

test/puppetlabs/puppetdb/threadpool_test.clj
IrimieBogdan: changed

test/puppetlabs/puppetdb/threadpool_test.clj
IrimieBogdan: `(future-cancel fut)` was removed because now we unblock and the future can finish. Waiting for the future to finish was done with `(is (future-done? fut))`

rbrw (approved):

PR #3432, opened 2021-03-05
(PDB-5022) Introduce eastwood

rbrw:


eastwood.clj
```diff
@@ -0,0 +1,45 @@
...
+ {:linter :constant-test
+  :for-macro 'clojure.core/coll?
+  :if-inside-macroexpansion-of #{'puppetlabs.structured-logging.core/maplog}
+  :within-depth 2
+  :reason "maplog checks the logger type at compile time."})
+
+(disable-warning
+ {:linter :deprecations
+  :symbol-matches
+  #{#"^#'puppetlabs\.dujour\.version-check/check-for-updates!$"
```
austb, line 23: We should be using the non-deprecated version of this command now

PR #3435, opened 2021-03-10
(PDB-5059) Fix facts blocklist logic in command submission

austb:
The options config passed to `prep-replace-facts` has more keys in it
than just blocklist options now. Update the function to only attempt to
filter blocked facts when the config option `facts-blocklist` is
present. We don't need to also check `facts-blocklist-type` because it
is defaulted in the config so it will always be present (except in some
tests). Also if a bug exists where a user sets a blocklist but we fail
to default the blocklist type, it would be better to error than store
potentially sensitive or gargantuan information in the database.

### DIFF
--- a/src/puppetlabs/puppetdb/command.clj
+++ b/src/puppetlabs/puppetdb/command.clj
@@ -349,17 +349,18 @@
 
 (defn prep-replace-facts
   [{:keys [version received] :as command}
-   {:keys [facts-blocklist facts-blocklist-type] :as blocklist-config}]
-  (let [blocklisting? (seq blocklist-config)
-        rm-blocklisted (when blocklisting?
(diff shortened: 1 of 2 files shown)


austb:
The error in tests that this prevents can be seen by running
```
lein test :only puppetlabs.pe-puppetdb-extensions.sync.pull-changes-test/pull-factsets-test
```
```
165378 [blocking-sync-1] ERROR p.p.sync.services - Deferring sync of 'replace facts' for f.local, due to: java.lang.IllegalArgumentException: No matching clause: 
java.lang.IllegalArgumentException: No matching clause: 
```
After `lein install`ing the PR, the error will go away.

Zak-Kent (approved):
🦖

PR #3437, opened 2021-03-12
(PDB-5059) Add a helper for command processing options

austb:
Extract this to a helper, like `shared-globals` because initial sync
will need to select the same options map that command processing uses in
order for initial sync to provide the correct settings to
`process-command!`.


Zak-Kent (approved):
👍

Zak-Kent (approved):
👍

PR #3438, opened 2021-03-18
(maint) Update ssl.sh script / entrypoint

Iristyle:
- Includes a number of new features in ssl.sh, including:

   * Much better logging during failed HTTP requests
   * Produce pk8 private key files for apps like orch, console services,
     pdb (so that their entrypoints can remove code to generate them)
   * Set permissions on cert files - 700 for dirs, 600 for files
     (so that entrypoint scripts are not responsible)
   * Always generate canonical symlinks to certs up front (rather than
     as files are generated), which repairs bad links on disk
   * Add ability to retry HTTP requests and do this when:
       - Grabbing CA
       - Grabbing CRL
       - Retrieving signed certs
   * Warning when new a different certname is used when a cert already
     exists on disk

 - Additionally simplifes the entrypoint script to always run ssl.sh
   when USE_PUPPETSERVER is true, given ssl.sh is idempotent.

 - Always set perms on SSLDIR, noting that this won't work in k8s
   when using non-root containers

### DIFF
--- a/docker/puppetdb/Dockerfile
+++ b/docker/puppetdb/Dockerfile
@@ -54,7 +54,7 @@ CMD ["foreground"]
 HEALTHCHECK --start-period=5m --interval=10s --timeout=10s --retries=6 CMD ["/healthcheck.sh"]
 
 # hadolint ignore=DL3020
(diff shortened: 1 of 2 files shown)


austb:
@Iristyle Does this need a review from pupperware/someone else, or is this ready to merge?

Iristyle:
Ready to merge! Similar updates have landed elsewhere

PR #3439, opened 2021-03-23
(PDB-5055) Add :explain query parameter

oanatmaria:
This PR exposes a new query parameter (:explain) introduced in : https://github.com/puppetlabs/puppetdb/pull/3433

### DIFF
--- a/test/puppetlabs/puppetdb/http/nodes_test.clj
+++ b/test/puppetlabs/puppetdb/http/nodes_test.clj
@@ -442,6 +442,14 @@
         (is (> (count (clojure.string/split-lines results))
                (count (clojure.string/split-lines normal-results))))))))
 
+(deftest-http-app query-with-explain-printing
+  [[version endpoint] endpoints
+   method [:get :post]]
+    (testing "should support explain when quering in reports"
+      (let [results (slurp (:body (query-response method endpoint nil {:explain "analyze"})))]
+        (is (not (empty? (json/parse-string results))))
+        (is (= true (contains? (first (json/parse-string results)) "query plan"))))))
+
 (deftest-http-app aggregate-functions-on-nodes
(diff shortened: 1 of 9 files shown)


Zak-Kent:
I'll need to spend some more time tomorrow to look at what's going on, but I noticed a validation error in the `row->resource` func when testing the new `explain: analyze` query option in a POST request against the `resources` & `fact-contents` endpoints (may be happen with other endpoints as well but haven't checked). [gist with error](https://gist.github.com/Zak-Kent/da7123d6c29f3ff3db8280ff6adc2aae). Everything worked as expected when I hit `query/v4` directly with a `"from" "facts"` query.

test/puppetlabs/puppetdb/http/nodes_test.clj
```diff
@@ -436,6 +436,15 @@
         (is (> (count (clojure.string/split-lines results))
                (count (clojure.string/split-lines normal-results))))))))
 
+(deftest-http-app query-with-explain-printing
+  [[version endpoint] endpoints
```
Zak-Kent, line 440: We may want to move this test so that it hits a different endpoint to cover the issue we saw with the `munge-fn` erroring before. Because the nodes `nodes` endpoint uses `identity` for the `munge-fn` it wouldn't cause the same error seen when testing against the `fact-contents` endpoint. See [here](https://github.com/puppetlabs/puppetdb/blob/6.x/src/puppetlabs/puppetdb/query_eng.clj#L64) for the atom where the `munge-fn` is pulled out per entity type.

oanatmaria: I added a test in `resources_test.clj` because the munge function of `fact-contents` was valid for the query plan.

Zak-Kent (approved):
LGTM!

PR #3441, opened 2021-03-26
(PDB-5063) Return only error message on API

IrimieBogdan:
When an exception is thrown on the server,
only return the error message to the API.
The stack trace is no longer returned,
it is logged as an error.


austb:
The responses I'm getting with this change seem exactly like what we want

PR #3444, opened 2021-03-29
(PDB-5063) Return only error message on API

oanatmaria:
When an exception is thrown on the server,
only return the error message to the API.
The stack trace is no longer returned,
it is logged as an error.


austb (approved):
:+1: tested after installing in pe-puppetdb-extensions and the response was also shortened

PR #3445, opened 2021-03-30
(PDB-5053) Save only the events that have not expired

sebastian-miclea:


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1359,6 +1359,20 @@
                                    reports/resources->resource-events
                                    (map normalize-resource-event)))))
 
+(defn filter-filter
+  [timestamp resource-events-ttl]
+  (let [after (time/after? (org.joda.time.DateTime. timestamp)
```
Zak-Kent, line 1364: Don't need the let block here, you can call the `time/after?` func directly and its result will be returned.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1359,6 +1359,20 @@
                                    reports/resources->resource-events
                                    (map normalize-resource-event)))))
 
+(defn filter-filter
```
Zak-Kent, line 1362: Could rename this to `resource-event-expired?`

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1359,6 +1359,20 @@
...
+  [timestamp resource-events-ttl]
+  (let [after (time/after? (org.joda.time.DateTime. timestamp)
+               (.minus (now) resource-events-ttl))]
+    after)
+  )
+
+(defn filter-expired-resources
+   [resource-events-ttl resource-list]
+   (if (some? resource-events-ttl)
+     (let [result (filter #(filter-filter (:timestamp %) resource-events-ttl) resource-list)]
```
Zak-Kent, line 1372: Don't need the let here

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1359,6 +1359,20 @@
...
+(defn filter-filter
+  [timestamp resource-events-ttl]
+  (let [after (time/after? (org.joda.time.DateTime. timestamp)
+               (.minus (now) resource-events-ttl))]
+    after)
+  )
+
+(defn filter-expired-resources
+   [resource-events-ttl resource-list]
+   (if (some? resource-events-ttl)
```
Zak-Kent, line 1371: I'm not sure if we need the `some?` check here

test/puppetlabs/puppetdb/admin_test.clj
```diff
@@ -26,26 +27,84 @@
...
+    (-> resource
+        (assoc :timestamp timestamp)
+        (update :events (fn [events]
+                                   (mapv #(assoc % :timestamp timestamp)
+                                         events)))
+        )
+    )
+  )
+
+(defn change-report-time
```
Zak-Kent, line 53: We should move this helper func into `testutils.clj` and then change the existing call in `services_test.clj` to use the same helper.

test/puppetlabs/puppetdb/admin_test.clj
```diff
@@ -26,26 +27,84 @@
...
+                                  (change-report-time res timestamp)
+                                  )
+                                )
+                              report-list)]
+    expected-report
+    )
+  )
+
 (deftest test-basic-roundtrip
+  (db/with-test-db
```
Zak-Kent, line 89: I'm not sure if we need this `with-test-db` wrapper. Did you notice the tests doing something unexpected without it?

test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -493,27 +497,38 @@
...
+          six-days (.minus current-time (time/parse-period "6d"))
+          seven-days (.minus current-time (time/parse-period "7d"))
+          basic-date1 (time/unparse (time/formatters :basic-date) six-days)
+          basic-date2 (time/unparse (time/formatters :basic-date) seven-days)
+          iso_date1 (time/unparse (time/formatters :date-hour-minute-second) six-days)
+          iso_date2 (time/unparse (time/formatters :date-hour-minute-second) seven-days)
+          ago-mock (fn [_]
+                    (.minus current-time (time/parse-period "2d")))]
+      (with-redefs [svcs/invoke-periodic-gc invoke-periodic
+                    time/ago ago-mock]
```
Zak-Kent, line 511: Why did we need to redef the `ago` func here? I tried running the test without doing so and the test hung and I had to manually kill the pids with `pg_cancel_backend` for the queries the test issued. The test was also still failing for me locally when I ran it without redefing `ago`. If trying to fix up this test is too much of a pain feel free to ignore it for now, I can help look at it more tomorrow. I think there are some tricky bits there, iirc

Zak-Kent:
Looks good so far. I pushed a branch to my fork of PDB which has an example of how we could test the feature you're adding. See [this commit](https://github.com/Zak-Kent/puppetdb/commit/c52d83289295e74e3c180bc4ef08c2cff52afa00). The example there will get PDB going, allow us to change the config, and has an example of how to submit a report. We'll still need to update the relevant timestamps and then assert that expired events were filtered out as expected.

test/puppetlabs/puppetdb/cli/services_test.clj
sebastian-miclea: The ago redef is needed in order to have the gc-bulldozer working properly because it uses both `report-ttl` and `resource-event-ttl` when cleaning up the old reports and resource event partitions. A time shift is simulated so reports are created and valid, but when gc is called, the resources are already expired and allowing it to cut the execution of the long running pids.

src/puppetlabs/puppetdb/scf/storage.clj
sebastian-miclea: I've noticed that in some tests `resource-events-ttl` is nil. Not sure wether it can happen in real life so just wanted to be sure it won't try to do unexpected operations on nil.

src/puppetlabs/puppetdb/scf/storage.clj
sebastian-miclea: Updated, forgot to remove them after debugging.

test/puppetlabs/puppetdb/cli/services_test.clj
Zak-Kent: I'm still going to need time to poke at this test. I'm wondering if changing the timestamps for the events in the reports that are submitted get us around needing to redef the `ago` call?

Zak-Kent (commented):
Tested this locally and the filtering looks like it's doing what we want. In addition to the test which needs updates, there are some minor spacing differences and trailing parens we'll still want to clean up. We generally try to follow the style guidelines described [here](https://github.com/bbatsov/clojure-style-guide).

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -440,3 +441,37 @@
         table]
        jdbc/query-to-vec
        (map :indexdef)))
+
+(defn update-timestamp [resource resource-list time]
```
Zak-Kent, line 445: Correct me if I'm confused here, but this looks like it's updating the timestamps of events in the fixture data based on their index in the `resources` vector. Then we're using the fact that we're getting results ordered by this timestamp back from the db so that the test fixture order matches the query result. I think this is happening due to the `sort-by` found in [testutils/reports/munge-resource-events](https://github.com/puppetlabs/puppetdb/blob/6.x/test/puppetlabs/puppetdb/testutils/reports.clj#L149) but not totally sure yet. We might want to take some time at the meeting tomorrow to double check this and see if we might be able to carry over the difference between the existing timestamps in the fixture data when we do the update.

test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -493,27 +476,38 @@
...
```
Zak-Kent, line 490: I think we should be able to avoid updating the timestamps on the reports in this test if we redef the `scf/storage/resource-events-expired?` func that's added in this pr. Redefining that func to `(constantly false)` allowed me to run the tests using the older timestamps. I was still seeing some transient failures due to the expected postgres error not always getting delivered from the future, but that appears to be unrelated to your changes.

test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -482,7 +463,9 @@
   (with-unconnected-test-db
     (let [config (-> (create-temp-config)
                      (assoc :database *db*)
-                     (assoc-in [:database :gc-interval] "0.01"))
+                     (assoc-in [:database :gc-interval] "0.01")
+                     (assoc-in [:database :report-ttl] "10d")
```
Zak-Kent, line 467: Should be able to pull out the updates to the ttls here when we redef the func in the comment below.

test/puppetlabs/puppetdb/testutils.clj
```diff
@@ -440,3 +441,31 @@
...
+  (let [current-time  (time/to-date-time time)
+        index (.indexOf resource-list resource)
```
Zak-Kent, line 447: This can likely be written using the `map-indexed` func. Might also be useful to mention in the comment below that we can't have identical timestamps due to the sorting which is done in the [testutils/reports/munge-resource-events](https://github.com/puppetlabs/puppetdb/blob/6.x/test/puppetlabs/puppetdb/testutils/reports.clj#L149) function.

test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -44,14 +44,14 @@
...
-             :refer [block-until-results default-timeout-ms temp-file]]
+             :refer [block-until-results default-timeout-ms temp-file change-report-time]]
             [puppetlabs.puppetdb.cheshire :as json]
             [puppetlabs.puppetdb.testutils.queue :as tqueue]
             [clojure.string :as str])
   (:import
-   [clojure.lang ExceptionInfo]
-   (java.util.concurrent CyclicBarrier TimeUnit)
-   [java.util.concurrent.locks ReentrantLock]))
+    [clojure.lang ExceptionInfo]
```
Zak-Kent, line 52: spacing

Zak-Kent (approved):
👍

PR #3447, opened 2021-03-31
(PDB-4771) Add command to import and timeshift datasets

florindragos:


ext/bin/pdb-dataset
```diff
@@ -0,0 +1,139 @@
...
+              dumpfile="$1"
+              shift
+            else
+              echo "Unknown argument:" $1
+              misuse
+            fi
+    esac
+done
+
+case $command in
```
rbrw, line 43: Also, it looks like this case statement is indented 2 spaces, but the previous one is indented 4.  I'm not sure what we want to do more broadly, but might be worth being consistent within the same file, so I'd suggest adjusting one or the other, so they match.

rbrw (commented):
Mostly minor comments.  Thanks.

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
...
+
+    echo "Checking for existing puppetdb database" 1>&2
+    db_intialized=$(psql "puppetdb" -t -c "SELECT 1 FROM information_schema.tables WHERE table_name = 'schema_migrations';")
+
+    if [ "$db_intialized" ]; then
+      echo 'puppetdb database already exists' 1>&2
+      exit 2
+    fi
+
+    echo "Restoring database from backup"
```
rbrw, line 66: Perhaps send this to stderr too.

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
...
+DROP TABLE IF EXISTS max_resource_event;
+EOF
+
+    echo "Updating data timestamps"
+    pgbox env psql "puppetdb" -a -f "$PGBOX/updatetime.sql"
+
+    echo "Running vacuum full on puppetdb database"
+    pgbox env vacuumdb -f puppetdb
+
+    echo "Finished"
```
rbrw, line 133: Perhaps echo to stderr for these three as well.

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
...
+    --timeshift-to)
+      test $# -gt 0 || misuse
+      timeshift="$2"
+      shift 2
+      ;;
+    *)
+      if [ -z $command ]; then
+        command="$1"
+        shift
+      elif [ -z $dumpfile ]; then
```
rbrw, line 33: Recommend quoting `$dumpfile`.

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
...
+command=''
+while test $# -gt 0; do
+  case "$1" in
+    --timeshift-to)
+      test $# -gt 0 || misuse
+      timeshift="$2"
+      shift 2
+      ;;
+    *)
+      if [ -z $command ]; then
```
rbrw, line 30: Recommend quoting `$command`.

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
...
+misuse()
+{
+  usage 1>&2
+  exit 1
+}
+
+timeshift=''
+timeshiftsql=''
+dumpfile=''
+command=''
```
rbrw, line 21: I'd suggest checking for the command first, outside the while loop, i.e. requiring for now that the subcommand to be the first argument.  This prevents any ambiguity about whether an argument is a command or a positional argument (like a datafile).

For example (assuming the error checking of command later is sufficient), perhaps just:
```
test $# -gt 0 || misuse
command="$1"
shift 1
```
And accordingly, I'd imagined that the other arguments like `--timeshift-to` and `DUMPFILE` are `import` specific arguments, not a general args, so I might move the while loop into the `import)` handling lower down.
  
i.e. we don't have any "global" options right now (like git's `--git-dir`) that might go before the subcommand.

Though since we only have one subcommand right now, it's not critical; we can always move the loop into the `import)` section later if you prefer.

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
...
+  timestamp = timestamp + ((select minute_diff from time_diff) * INTERVAL '1 minute');
+
+DROP TABLE IF EXISTS time_diff;
+DROP TABLE IF EXISTS max_report;
+DROP TABLE IF EXISTS resource_events_time_diff;
+DROP TABLE IF EXISTS max_resource_event;
+EOF
+
+    echo "Updating data timestamps"
+    pgbox env psql "puppetdb" -a -f "$PGBOX/updatetime.sql"
```
rbrw, line 128: Wondered if we might want consider changing this to a `... psql ... < "$PGBOX/updatetime.sql"` given the discussion in the `psql` manpage (in the docs for the `-c` argument).  If so, we might also want to add a `BEGIN` and `COMMIT` to the beginning and end of the file, so all the changes still happen in a single transaction.

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
...
+    echo "Checking for existing puppetdb database" 1>&2
+    db_intialized=$(psql "puppetdb" -t -c "SELECT 1 FROM information_schema.tables WHERE table_name = 'schema_migrations';")
+
+    if [ "$db_intialized" ]; then
+      echo 'puppetdb database already exists' 1>&2
+      exit 2
+    fi
+
+    echo "Restoring database from backup"
+    pgbox env pg_restore --role=postgres --no-owner --no-acl -d puppetdb "$dumpfile" || echo $?
```
rbrw, line 67: I suspect we don't need the `pgbox env` here (or anywhere else in the file), given the current requirement that we already be in a PDBBOX environment, since `pdbbox-env` already runs everything via `pgbox env`.  (See the last line in ext/bin/pdbbox-env`.)

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env bash
+
+set -ueo pipefail
+
+usage() { echo 'Usage: PDBBOX=BOX_DIR pdb-dataset import --timeshift-to now DUMPFILE'; }
+
+if test "$#" -lt 1 -o -z "$PDBBOX"; then
+  usage 1>&2
```
Filipovici-Andrei, line 8: you should call misuse here

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
...
+        misuse
+      fi
+  esac
+done
+
+tmpdir="$(mktemp -d "pdb-dataset-XXXXXX")"
+tmpdir="$(cd "$tmpdir" && pwd -P)"
+trap "$(printf 'rm -rf %q' "$tmpdir")" EXIT
+
+case "$command" in
```
Filipovici-Andrei, line 47: Could we change this case to an if and break early? Also for readability we should enclose some ob these blocks into functions

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,140 @@
...
+      else
+        echo "Unknown argument: $1" 1>&2
+        misuse
+      fi
+  esac
+done
+
+tmpdir="$(mktemp -d "pdb-dataset-XXXXXX")"
+tmpdir="$(cd "$tmpdir" && pwd -P)"
+trap "$(printf 'rm -rf %q' "$tmpdir")" EXIT
```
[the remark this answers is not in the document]
rbrw: It's the `%q`, which (if I got the actual use right here) makes sure that even if the path contains spaces it'll be properly escaped for the eventual trap call.  cf. `help printf` in bash, or there should also be a bit in `man bash` with more details.

ext/bin/pdb-dataset
florindragos: maybe something like `cat "$PGBOX/updatetime.sql" | psql "puppetdb" -a` ?

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,139 @@
...
+      timshiftsql="${timeshift}()"
+    elif [[ $timeshift =~ ^[0-9]{4}-[0-9]{2}-[0-9]{2}$ ]]; then
+      timshiftsql="TIMESTAMP '${timeshift}'"
+    else
+      echo "Invalid value for timeshift parameter:" $timeshift
+      misuse
+    fi
+
+    export PGBOX="$PDBBOX/pg"
+    export PGUSER=postgres
```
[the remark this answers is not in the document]
florindragos: removed `pdbbox-env` from all the commands and the PG env var exports

ext/bin/pdb-dataset
florindragos: this makes sense. moved everything under the `import)` handling

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,141 @@
...
+
+DROP TABLE IF EXISTS time_diff;
+DROP TABLE IF EXISTS max_report;
+DROP TABLE IF EXISTS resource_events_time_diff;
+DROP TABLE IF EXISTS max_resource_event;
+COMMIT;
+EOF
+
+    echo "Updating data timestamps" 1>&2
+    psql "puppetdb" -a -f "$PGBOX/updatetime.sql"
```
Zak-Kent, line 129: `PGBOX` is unbound. I think we'll need to change this to the `$tmpdir` created above.

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,141 @@
...
+    echo "Checking for existing puppetdb database" 1>&2
+    db_intialized=$(psql "puppetdb" -t -c "SELECT 1 FROM information_schema.tables WHERE table_name = 'schema_migrations';")
+
+    if [ "$db_intialized" ]; then
+      echo 'puppetdb database already exists' 1>&2
+      exit 2
+    fi
+
+    echo "Restoring database from backup" 1>&2
+    pg_restore --role=postgres --no-owner --no-acl -d puppetdb "$dumpfile" || echo $?
```
Zak-Kent, line 63: When testing this without a `dumpfile` argument the `pg_restore` command failed and then the `db_exists` check below output `'[' '' ']'` which made it pass the following check when I think it was intended to fail. This was after I created an empty puppetdb database with `createdb -E UTF8 -O puppetdb puppetdb`. I created the empty `puppetdb` to get around the error mentioned above in the `db_initialized` check.

Zak-Kent (commented):
I should have a bit more time tomorrow to keep testing this. I was running the script with a pg_dump taken from a local PDB filled up with some benchmark data when I noticed the things mentioned in the comments.

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,141 @@
...
+
+DROP TABLE IF EXISTS resource_events_time_diff;
+
+SELECT (DATE_PART('day', ${timshiftsql} - (select max from max_resource_event)) * 24 +
+        DATE_PART('hour', ${timshiftsql} - (select max from max_resource_event))) * 60 +
+        DATE_PART('minute', ${timshiftsql} - (select max from max_resource_event)) as minute_diff
+INTO TEMPORARY TABLE resource_events_time_diff;
+
+UPDATE reports
+  SET producer_timestamp = producer_timestamp + ((select minute_diff from time_diff) * INTERVAL '1 minute'),
```
Zak-Kent, line 105: I was able to get this running to the point where it restored a puppetdb database from a pg_dump and then attempted to run the `updatetime.sql` script on the restored data. Doing this hits an issue with the constraints on existing report partitions. 

This `updatetime.sql` script was written before we had partitioned `reports` and `resource_events` tables so we'll need to change how the updating of timestamps is handled because the existing constraints on each partition won't allow you to update a row in place and increase its timestamp outside of the range allowed in the partition. After the script errors the transaction updating all the timestamps rollsback and you end up with a restored db with old timestamps. See example constraint error below: 

```
ERROR:  new row for relation "reports_20210416z" violates check constraint "reports_20210416z_producer_timestamp_check"
DETAIL:  Failing row contains (22, \x62c7899ab6e6219bd2766bd053e4ca9fa8b8b905, 2bce4bec-ce8b-4d36-91d5-e204af508e98, host-4, 3.7.2, 4, a3adc55e-514f-4483-80d3-a258646ced2c, 2021-05-13 15:47:47.293-07, 2021-05-13 15:47:52.293-07, 2021-05-18 12:33:09.63-07, f, 1, 1, null, null, 2021-05-13 15:47:57.293-07, [{"name": "anchor", "value": 0.00046438, "category": "time"}, {"..., [{"file": null, "line": null, "tags": ["notice"], "time": "2015-..., [{"file": "/home/wyatt/.puppet/modules/concat/manifests/fragment..., 76e7a26a-f7a2-44bc-b0fd-8db6edd8ec9e, on_failure, 87bf8011-918c-40b3-aded-cee4ec9d5536, 1, f, t, null, agent).
```

ext/bin/pdb-dataset
florindragos: good catch. I missed it when moving to tmpdir

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,141 @@
...
+test $# -gt 0 || misuse
+command="$1"
+shift 1
+
+case "$command" in
+  import)
+    timeshift=''
+    timeshiftsql=''
+    dumpfile=''
+
```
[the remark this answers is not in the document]
florindragos: adding it back. i was doing `export PGUSER=puppetdb` in an earlier version but removed it when I removed the `pdbbox-env` calls as I thought it's not needed anymore

ext/bin/pdb-dataset
```diff
@@ -0,0 +1,141 @@
...
+      echo 'puppetdb database already exists' 1>&2
+      exit 2
+    fi
+
+    echo "Restoring database from backup" 1>&2
+    pg_restore --role=postgres --no-owner --no-acl -d puppetdb "$dumpfile" || echo $?
+
+    db_exists=$(psql "puppetdb" -t -c "SELECT 1 FROM information_schema.tables WHERE table_name = 'schema_migrations';")
+
+    if [ "$db_exists" ]; then
```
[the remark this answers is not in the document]
florindragos: fixed

Zak-Kent (approved):
I was able to test that this tool restored a pg_dump and attempted to shift the timestamps for the data contained in the dump. It hits an issue described in [PDB-5130](https://tickets.puppetlabs.com/browse/PDB-5130) if the pg_dump was taken from a newer PDB that has partitioned report and resource-events tables. 

I'm going to merge this pr as-is so we can close out PDB-4771 and then we can follow up with the work described in PDB-5130. Currently this tool will work as intended with older datasets like the SLV one. 

In the follow up work we'll also want to address the comment from @Filipovici-Andrei in this PR.

PR #3455, opened 2021-04-06
(PDB-5025) Added skipped unit tests for some issues

Filipovici-Andrei:
Changed Pupperware branch from master to main
Skipped unit tests for issues:
- order_by fails in some subqueries
- to_string fails for integer numbers 
- to_string adds multiple spaces when extracting month from date

### DIFF
--- a/test/puppetlabs/puppetdb/http/facts_test.clj
+++ b/test/puppetlabs/puppetdb/http/facts_test.clj
@@ -1,12 +1,10 @@
 (ns puppetlabs.puppetdb.http.facts-test
   (:require [cheshire.core :as json]
-            [clojure.java.jdbc :as sql]
             [clojure.java.io :as io]
             [clojure.test :refer :all]
             [flatland.ordered.map :as omap]
             [puppetlabs.puppetdb.scf.storage :as scf-store]
             [puppetlabs.puppetdb.cli.services :as cli-svc]
-            [puppetlabs.puppetdb.scf.storage-utils :as sutils]
             [puppetlabs.puppetdb.examples :refer :all]
             [puppetlabs.puppetdb.http :as http]
             [puppetlabs.puppetdb.http.server :as server]
@@ -2255,6 +2253,21 @@
                       [{"certname" "foo1" "value" "testing.com"}
                        {"certname" "foo2" "value" "testing.com"}
                        {"certname" "foo3" "value" "testing.com"}]))))))
+
+(deftest-http-app  ^:skipped test-for-known-issues
+  [[version endpoint] fact-contents-endpoints
+   method [:get :post]]
+
+   (populate-for-structured-tests reference-time)
+
+   ; enable this test after ticket https://tickets.puppetlabs.com/browse/PDB-5104 is solved
+   (testing "to_string function on integers"
+      (is (= (query-result method endpoint
+                           ["extract" [["function" "to_string" "value" "FM9999"]]
+                            ["and"
+                            ["=" "name" "uptime_seconds"]
+                            ["=" "certname" "foo1"]]])
+           #{{:to_string "1000"}}))))
 
 (def no-parent-endpoints [[:v4 "/v4/factsets/foo/facts"]])
 
--- a/test/puppetlabs/puppetdb/http/nodes_test.clj
+++ b/test/puppetlabs/puppetdb/http/nodes_test.clj
@@ -433,6 +433,20 @@
           (are-error-response-headers headers)
           (is (re-find msg body)))))))
 
+;TODO fix all skipped tests
+(deftest-http-app ^:skipped tests-for-PDB-5026
+  [[version endpoint] endpoints
+  method [:get :post]]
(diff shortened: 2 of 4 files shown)


Zak-Kent:
We'll need to update the `test-selectors` in `project.clj` [here](https://github.com/puppetlabs/puppetdb/blob/6.x/project.clj#L331-L333) so they know to skip the `^:skipped` metadata tag when running the default tests. The snippet below should work.  

```
  :test-selectors {:default (fn [m] (not (or (:integration m) (:skipped m))))
                   :unit (fn [m] (not (or (:integration m) (:skipped m))))
                   :integration :integration
                   :skipped :skipped}
```
See [leiningen/sample-project.clj](https://github.com/technomancy/leiningen/blob/master/sample.project.clj#L354-L359) for more info.

I'm not sure if we have tickets for all of the issues you've identified but if we don't we should make some and add the `query-engine-improvements` label to them. I can help take a closer look at the added tests and check for existing tickets tomorrow. It would be good to have a comment with a ticket number above each skipped tests so we know which tickets go with the work.

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -2256,6 +2254,26 @@
                        {"certname" "foo2" "value" "testing.com"}
                        {"certname" "foo3" "value" "testing.com"}]))))))
 
+(deftest-http-app  ^:skipped tests-for-known-issues
+  [[version endpoint] fact-contents-endpoints
+   method [:get :post]]
+
+  (testing "to_string function on integers"
+    (is (= (query-result method endpoint
+                         ["extract" [["function" "to_string" "value"]]
```
Zak-Kent, line 2263: I was able to get this query to hit a validation error from [munge-result-row](https://github.com/puppetlabs/puppetdb/blob/6.x/src/puppetlabs/puppetdb/query/fact_contents.clj#L26-L32) if I added a format string to the function call. 
```
[["function" "to_string" "value" "FM9999"]]
```
See the [functions-formatting](https://www.postgresql.org/docs/11/functions-formatting.html) Postgres docs for more info. 

By adding a `:to_string` key to the `fact-contents` [row-schemas](https://github.com/puppetlabs/puppetdb/blob/6.x/src/puppetlabs/puppetdb/query/fact_contents.clj#L10-L24) the query was able to return the result we would expect. 

For this test we'll want to add the format string in the snippet above, it will still be failing, but we can work on it from there. I also had to add a call to submit factsets before running queries in the test and changed the certname to `foo1` to match the test data. The call to add data looks like: `(populate-for-structured-tests reference-time)`

Next we'll want to make two tickets: 
1. a ticket to update the AST docs to add examples of passing in format strings when using the `to_string` function on numeric and timestamp fields. 

2. a ticket to audit all of the schemas used to munge rows for different entity types for any endpoint that allows a user to use the `to_string` function. We need to make sure validation doesn't fail because we can return a `:to_string` row that the schemas aren't expecting. 

If you're up for it and wouldn't mind trying to write the tickets I can help fill in detail tomorrow if needed. We can then add a comment above this skipped test with the ticket number for 2 above and update it when we do the work.

test/puppetlabs/puppetdb/http/facts_test.clj
```diff
@@ -2256,6 +2254,26 @@
...
+                          ["and"
+                           ["=" "name" "uptime_seconds"]
+                           ["=" "certname" "foo"]]])
+           #{{:to_string "1000"}})))
+
+  (testing "order_by in a subquery"
```
Zak-Kent, line 2269: We might only need the similar test you added to the nodes-test ns.

test/puppetlabs/puppetdb/http/nodes_test.clj
```diff
@@ -425,6 +425,19 @@
           (are-error-response-headers headers)
           (is (re-find msg body)))))))
 
+;TODO fix all skipped tests
+(deftest-http-app ^:skipped tests-for-known-issues
+  [[version endpoint] endpoints
+  method [:get :post]]
+
+  (testing "in a reports subquery"
```
Zak-Kent, line 433: This issue is covered by [PDB-5026](https://tickets.puppetlabs.com/browse/PDB-5026). Adding a comment with the ticket number here should help us remember to enable this test once the issue if fixed.

docker/Gemfile
```diff
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 gem 'pupperware',
     :git => 'https://github.com/puppetlabs/pupperware.git',
-    :branch => 'master',
+    :branch => 'main',
```
Iristyle, line 5: Took care of this one in https://github.com/puppetlabs/puppetdb/pull/3456

Filipovici-Andrei: Ok, thanks. I'll remove it.

Zak-Kent (approved):
LGTM! We will merge once the release is tagged

PR #3461, opened 2021-04-13
(maint) Update container with latest ssl.sh

Iristyle:
- Symlinks in the same directory for canonical filenames server.pub,
   server.key and server.crt - works better when initContainer mounts
   to a different path structure than a destination container in k8s.


Iristyle:
Ssl.sh change already approved in Pupperware + fixes for Travis reliability ... merging!

PR #3462, opened 2021-04-14
(PDB-2898) Disallow paging opts outside PQL query

oanatmaria:
**What was done:** Disallow paging query params (limit, order_by and offset) for PQL queries. These options can now only be provided by including them in the query body.


rbrw (approved):
Looks good.

PR #3463, opened 2021-04-14
(PDB-5097) PuppetDB export time shift utility

Filipovici-Andrei:
This is just the beginning of this utility.
Now it only validates the input parameters and returns a "Not implemented" message.


Zak-Kent (approved):
LGTM, tested this locally and it looked what we would want wrt the params and error handling

PR #3471, opened 2021-04-22
(maint) Use Docker 19.03 in Travis again

Iristyle:
- Instead of incurring the upgrade cost to Travis 20.10, use the
   existing 19.03 version.

   In practice, upgrading Docker in Travis turned out to be unreliable.

   To enable buildx requires installing the plugin from the github
   releases page. Using the experimental cli flag was supposed to enable
   the plugin for 19.03, but didn't seem to work in the Travis env.


Iristyle:
The Travis / buildx stuff is working properly, so this is safe to merge.

PR #3474, opened 2021-04-23
(maint) Demarcate fact-path gc with log messages

austb:
fact path gc is an operation we _know_ can be slow, so this will help us
in identifying if gc is stuck on this operation.


austb:
Ignore whitespace changes when you review this is recommended.

Zak-Kent (approved):
👍

PR #3476, opened 2021-05-05
(SEC-446) Run lein-nvd in CI lint step

austb:


austb:
Don't merge until this contains a bump to clj-parent as well, which will be necessary for the nvd check to pass.

rbrw (approved):

PR #3477, opened 2021-05-07
(PDB-5096) Add command performance test

oanatmaria:


ext/bin/check-command-perf
```diff
@@ -0,0 +1,95 @@
...
+  printf "." 1>&2
+done
+
+start=`date +%s`
+
+printf "\n" 1>&2
+
+printf "Comand ingestion in progress..." 1>&2
+
+processed=$(curl -X GET http://localhost:8080/metrics/v1/mbeans/puppetlabs.puppetdb.mq:name=global.processed 2>/dev/null | jq | grep Count | sed 's/[^0-9]//g' | bc)
```
Filipovici-Andrei, line 76: This can be moved into a function.

Filipovici-Andrei (commented):
I think that this script can be divided into helper functions for validating parameters, starting PuppetDB, timing command processing and writing results.

ext/bin/check-command-perf
```diff
@@ -0,0 +1,95 @@
...
+  printf "An archive that contains the stockpile must be provided!\n" 1>&2
+  misuse
+else
+  stockpile="$1"
+fi
+
+printf "Unarchiving the stockpile in: ${target_dir} \n" 1>&2
+
+tar zxvf "$stockpile" -C "$target_dir" 2> /dev/null
+files_queue="$(ls "$target_dir" | wc -l | bc)"
```
Filipovici-Andrei, line 53: I think this would be better named commands_count or file_count

ext/bin/check-command-perf
```diff
@@ -0,0 +1,95 @@
...
+./pdb services -c "$PDBBOX/conf.d" > tmp/log.txt 2>&1 &
+id="$(jobs -p)"
+
+finished=$(cat tmp/log.txt | (grep "Finished database garbage collection" || true) | wc -l)
+
+printf "Waiting for PDB to start..." 1>&2
+while [ $finished == 0 ]
+do
+  finished=$(cat tmp/log.txt | (grep "Finished database garbage collection" || true) | wc -l)
+  printf "." 1>&2
```
Filipovici-Andrei, line 67: Will this print a lot of dots quickly?

ext/bin/check-command-perf
```diff
@@ -0,0 +1,95 @@
...
+
+printf "Unarchiving the stockpile in: ${target_dir} \n" 1>&2
+
+tar zxvf "$stockpile" -C "$target_dir" 2> /dev/null
+files_queue="$(ls "$target_dir" | wc -l | bc)"
+
+printf "Running PDB\n" 1>&2
+
+mkdir tmp
+./pdb services -c "$PDBBOX/conf.d" > tmp/log.txt 2>&1 &
```
Filipovici-Andrei, line 58: You should make sure that the log.txt is empty or it doesn't exist.

ext/bin/check-command-perf
```diff
@@ -0,0 +1,95 @@
...
+do
+	processed=$(curl -X GET http://localhost:8080/metrics/v1/mbeans/puppetlabs.puppetdb.mq:name=global.processed 2>/dev/null | jq | grep Count | sed 's/[^0-9]//g' | bc)
+	printf "." 1>&2
+done
+
+printf "\n" 1>&2
+
+processing_time=$(curl -X GET http://localhost:8080/metrics/v1/mbeans/puppetlabs.puppetdb.mq:name=global.processing-time 2>/dev/null | jq)
+
+kill $id
```
Filipovici-Andrei, line 88: I don't think this will work. I tried it with a ping command and got `kill: illegal pid: [1]  + 10514 running    ping -i 5 google.com`. The only think that worked was `kill %1`.

ext/bin/check-command-perf
oanatmaria: If log.txt already exists, it will be overwritten.

ext/bin/check-command-perf
```diff
@@ -0,0 +1,95 @@
...
+
+if test "$#" -lt 0 -o -z "$PDBBOX"; then
+  printf '$PDBBOX variable is not set. Do you have a pdbbox created?\n' 1>&2
+  misuse
+fi
+
+if ! test -e "$jar"; then
+    printf 'Unable to find the puppetdb jar %q; have you run "lein uberjar"?\n' \
+           "$jar" 1>&2
+    misuse
```
Filipovici-Andrei, line 32: I don't think it's necessary to print the usage for this specific messages. Or at least add an additional empty line between the error and the usage. Another empty line before the error message would be appreciated.

ext/bin/check-command-perf
```diff
@@ -0,0 +1,95 @@
...
+fi
+
+target_dir="${PDBBOX}/var/stockpile/cmd/q"
+stockpile=""
+
+if [ "$(ls -A $target_dir)" ]; then
```
Filipovici-Andrei, line 38: you should also check that the `target_dir` exists. Otherwise it will fail silently at the unarchiving step

ext/bin/check-command-perf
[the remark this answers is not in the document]
rbrw: Perhaps something like this:
```
cmds-processed() {
    local mbeans='http://localhost:8080/metrics/v1/mbeans'
    curl -sSX GET "$mbeans/puppetlabs.puppetdb.mq:name=global.processed" \
        | jq .Count
}
```

ext/bin/check-command-perf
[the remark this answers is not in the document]
oanatmaria: On my MacOs the result without bc is `\s\s\s\s\s\s33` and the problem was to match it with the processed metric.

ext/bin/check-command-perf
```diff
@@ -0,0 +1,103 @@
...
+misuse() { usage 1>&2; exit 2; }
+
+jar="${PDB_JAR:-target/puppetdb.jar}"
+
+if [ -z "$PDBBOX" ]; then
+    echo "PDBBOX is not set in the environment. Do you have a pdbbox created?" 1>&2
+    misuse
+fi
+
+if [ "$#" -le 0 ]; then
```
Filipovici-Andrei, line 28: Why not -eq?

ext/bin/check-command-perf
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+usage() {
+    cat <<EOF
+Usage: [PDB_JAR=JAR] PDBBOX=BOX_DIR $(basename "$0") STOCKPILE_ARCHIVE"
```
Filipovici-Andrei, line 7: You should add a new line at the top. This way the usage will be separated from the errors.

ext/bin/check-command-perf
```diff
@@ -0,0 +1,103 @@
...
+tmpdir="$(cd "$tmpdir" && pwd)"
+
+./pdb services -c "$PDBBOX/conf.d" >"$tmpdir/log.txt" 2>&1 &
+pdb_id=$!
+
+echo "Waiting for PDB to start..." 1>&2
+
+while ! grep -Fq "Finished database garbage collection" "$tmpdir/log.txt"; do
+    sleep 1
+    if [ "$(grep -F "Error during service start!!!" "$tmpdir/log.txt")" ]; then
```
Filipovici-Andrei, line 85: Unfortunately if the database server is not started, this will be stuck in a loop.  You also need to check for `Will retry database connection after temporary failure: java.sql.SQLTransientConnectionException: PDBMigrationsPool: default - Connection is not available, request timed out after 3006ms.`

Filipovici-Andrei:
You should also check that the metrics endpoint is working. Someone might be missing the configuration for it.

ext/bin/check-command-perf
```diff
@@ -0,0 +1,112 @@
...
+get-metric() {
+    local metric="$1"
+    local mbeans='http://localhost:8080/metrics/v1/mbeans'
+    curl -sSX GET "$mbeans/puppetlabs.puppetdb.mq:name=$metric"
+}
+
+target_dir="$PDBBOX/var/stockpile/cmd/q"
+stockpile=""
+
+if [ "$(ls "$target_dir" | wc -l)" -gt 0 ]; then
```
Filipovici-Andrei, line 62: if the path doesn't exist, wc -l is equal to 0

PR #3483, opened 2021-05-13
(PDB-5094) Added load testing tool

sebastian-miclea:
Added a tool that allows us to load test the endpoints. 
Requirements: Python3
Usage: locust/run-load-test
This will generate four csv files and one html with the result.

### DIFF
--- a/locust/load-test/load-test.py
+++ b/locust/load-test/load-test.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python3
+
+from locust import HttpUser, task, between
+import yaml, json, os
+
+class PuppetDbLoadTest(HttpUser):
+    def response_printer(self, opts, response):
+        if response.status_code == 0:
+            print(response.error)
+            exit(1)
+        elif response.status_code != 200:
+            print(
+            "Method: " + opts['method'],
+            "Query: " + str(opts['query']),
+            "Response status: " + str(response.status_code),
+            "Response body: " + response.text,
+            end="\n-------------------------------------------\n", sep="\n" )
+
+    def get_request(self, opts):
+        limit = opts.get('limit')
+        offset = opts.get('offset')
+        url = opts['path'] + "?query=" + json.dumps(opts['query'])
+        if limit:
+            url = url + "&limit=" + str(limit)
+        if offset:
+            url = url + "&offset=" + str(offset)
+        with self.client.request(opts['method'], url, str(opts['query'])) as response:
+            self.response_printer(opts, response)
+
+    def post_request(self, opts):
+        query = {'query': opts['query']}
+        limit = opts.get('limit')
+        offset = opts.get('offset')
+        if limit:
+            query['limit'] = limit
+        if offset:
+            query['offset'] = offset
+        headers = opts.get('headers')
+        with self.client.request(opts['method'], opts['path'], str(opts['query']), data=json.dumps(query), json=True, headers=headers) as response:
+            self.response_printer(opts, response)
+
+    @task
+    def swarm(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        with open(dir_path + '/config.yaml') as stream:
+            config = yaml.safe_load(stream)
+            for opts in config:
+                if opts['method'] == 'GET':
+                    self.get_request(opts)
+                elif opts['method'] == 'POST':
+                    self.post_request(opts)
+
+
+
--- a/locust/run-load-test
+++ b/locust/run-load-test
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+import sys, subprocess, os
+
+def usage():
+    usage = """
+This script only adds defaults if there are no options
+Usage: run-load-test [options]
+
+Defaults:
+  -H [--host]       http://localhost:8080
+  -u [--users]      1
+  -s [--spawn_rate] 1
+  -t [--run_time]   1m
+  -f [--locustfile] locust/load-test/load-test.py
+  --html            report.html
+"""
+    print(usage)
+
+def add_default_args(program_args):
+    defaults = [['--headless'], ['-H', 'http://localhost:8080'],
+           ["-u", '1'], ["-r", '1'], ["-t", '1m'], ['-f', locust_filepath], ['--html', 'report.html']]
+    additional_args = []
+    for default_arg in defaults:
+        found = False
+        for arg in program_args:
+            if arg == default_arg[0]:
+                found = True
+        if not found:
+            additional_args += default_arg
+
+    return additional_args + program_args
+
+subprocess.run([sys.executable, "-m", "pip", "install", "locust", "pyyaml", "-U"])
+
+missing_defaults = []
+dir_path = os.path.dirname(os.path.realpath(__file__))
+workdir = os.getcwd()
+locust_filepath = os.path.join(dir_path, 'load-test', 'load-test.py')
+
+sys_args = sys.argv[1:]
+if '--help' in sys_args:
+    subprocess.run(["locust", "--help"])
+    usage()
+    exit()
+
+locust_args = add_default_args(sys_args)
+subprocess.run(["locust"] + locust_args)
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,8 @@ checkouts
 
 /.eastwood
 /.nrepl-port
+
+# locust output files and python cache
+*.csv
+*report.html
+locust/load-test/__pycache__/
--- a/documentation/testing_storage_changes.markdown
+++ b/documentation/testing_storage_changes.markdown
@@ -39,3 +39,24 @@ yourself first.
 Fact storage benchmark profiles are configured in `test-configs` var in
 `src/puppetlabs/puppetdb/cli/fact_storage_benchmark.clj`. The profiles are run
 in order if none is specified at the command line.
+
+
+## Query performance testing with Locust
+
+If PuppetDB is running and listening on port 8080, you can run a
+simple query performance test using [Locust](https://locust.io/) like
+this:
+
+```
+locust/run_load_test
+```
+
+See `locust/run_load_test --help` for some additional
+information.
+
+Dependencies are installed during the run, but if you want 
+to install dependencies only for the current user you can run
+`pip install --user locust pyyaml`, which will install to
+~/.local, in which case you'll need to make sure the relevant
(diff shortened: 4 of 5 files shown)


ext/bin/load_test/load_test.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+
+from locust import HttpUser, task, between
+import yaml
+import json
+import os
+
+class QuickstartUser(HttpUser):
+    wait_time = between(1, 2.5)
```
IrimieBogdan, line 9: In order to make the runs predictable I would go with a fixed value for this (e.g. `1`) or maybe we can eliminate it altogether to submit commands as fast as possible.

ext/bin/load-test/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+  headers: ''
+- name: post call
+  path: "/pdb/query/v4/facts"
+  query: {"query": ["~", "certname", "host"]}
+  method: POST
+  headers: {content-type: application/json}
+- name: v4/facts GET2
+  path: "/pdb/query/v4"
+  query: query=["from", "resources", ["extract", ["certname"], ["and", ["and", ["=", "type", "Class"], ["or", ["=", "title", "Puppet_enterprise::Profile::Certificate_authority"], ["or", ["=", "title", "Puppet_enterprise::Profile::Master"], ["=", "title", "Puppet_enterprise::Profile::Console"]], ["=", "title", "Puppet_enterprise::Profile::Puppetdb"]], ["subquery", "nodes", ["and", ["null?", "deactivated", true], ["null?", "expired", true]]]], ["=", "node_state", "active"]]]]
+  method: put
```
Filipovici-Andrei, line 15: This should be a POST method

sebastian-miclea: This was meant to be an example. Updated.

ext/bin/run-load-test
```diff
@@ -0,0 +1,71 @@
...
+dir_path = os.path.dirname(os.path.realpath(__file__))
+workdir = os.getcwd()
+subprocess.check_call([sys.executable, "-m", "pip", "install", "locust", "pyyaml", "-U"])
+subprocess.run(["locust", "-f", dir_path + "/load-test/load-test.py", "--headless", "-H", host, "-u", str(users), "-r", str(spawn_rate), "-t", timelimit, "--csv", prefix, "--html", prefix + "_report.html"])
+
+report_files = ['_exceptions.csv', '_failures.csv', '_stats_history.csv', '_stats.csv', '_report.html']
+for file in report_files:
+    filename = prefix + file
+    source = workdir + '/' + filename
+    destination = dir_path + '/load-test/output/'
```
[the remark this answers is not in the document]
sebastian-miclea: I've updated .gitignore to also store the output dir so this won't happen.

Filipovici-Andrei (approved):

ext/bin/run-load-test
```diff
@@ -0,0 +1,71 @@
...
+from sys import exit, stderr, stdout
+from datetime import datetime
+import os, sys, getopt, subprocess, shutil
+
+def usage(stream):
+    usage = """
+Usage: run-load-test --host http://myhost:port [options]
+
+Options:
+  -h [--host]       Host to which the requests are made
```
IrimieBogdan, line 12: We should add that you need to specify the port as well

ext/bin/run-load-test
```diff
@@ -0,0 +1,71 @@
...
+workdir = os.getcwd()
+subprocess.check_call([sys.executable, "-m", "pip", "install", "locust", "pyyaml", "-U"])
+subprocess.run(["locust", "-f", dir_path + "/load-test/load-test.py", "--headless", "-H", host, "-u", str(users), "-r", str(spawn_rate), "-t", timelimit, "--csv", prefix, "--html", prefix + "_report.html"])
+
+report_files = ['_exceptions.csv', '_failures.csv', '_stats_history.csv', '_stats.csv', '_report.html']
+for file in report_files:
+    filename = prefix + file
+    source = workdir + '/' + filename
+    destination = dir_path + '/load-test/output/'
+    shutil.move(source, destination)
```
Filipovici-Andrei, line 71: new line

ext/bin/load-test/load-test.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
+
+from locust import HttpUser, task, between
+import yaml, json, os
+
+class PuppetDbLoadTest(HttpUser):
+    def response_printer(self, p, response):
+        if response.status_code != 200:
+            print(opts['name'] + "[" + str(response.status_code) + "] failed with:")
+            print(response.text)
```
Filipovici-Andrei, line 10: `response.text `doesn't have any relevant data, but `response.error` does. Putting the `status_code` between brackets doesn't really say what it is.
Something in this direction might be a little better:
`print(opts['name'] + " finished with status code: " + str(response.status_code) + ".\nFailed with:")
            print(str(response.error) + "\n")`

ext/bin/run-load-test
Filipovici-Andrei, line 12: I think that we should add an option for the port, or at least inform the user that if the port is not specified, port 80 is used by default.

ext/bin/run-load-test
Filipovici-Andrei, line 71: if the files already exist, shutil.move will throw an exception. I think that something like 
```
from pathlib import Path    
if Path(destination).is_file:
        os.remove(destination)
```
might work. Not sure about permissions.

Filipovici-Andrei (changes requested):

ext/bin/run-load-test
```diff
@@ -0,0 +1,73 @@
...
+
+def usage(stream):
+    usage = """
+Usage: run-load-test --host http://myhost:port [options]
+
+Options:
+  -h [--host]       Host to which the requests are made
+  -u [--users]      Number of users to simulate. Default 1
+  -s [--spawn_rate] The rate per second in which users are spawned. Default 1
+  -t [--timelimit]  How long the swarm should take. Default 1m (Examples: 30s, 1h30m)
```
IrimieBogdan, line 15: This argument should be named `--run-time` to be consistent with the argument from locust

ext/bin/run-load-test
sebastian-miclea: Updated.

ext/bin/load-test/load-test.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
+
+from locust import HttpUser, task, between
+import yaml, json, os
+
+class PuppetDbLoadTest(HttpUser):
+    def response_printer(self, p, response):
```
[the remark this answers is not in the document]
sebastian-miclea: updated

ext/bin/load-test/load-test.py
sebastian-miclea: Updated to a more detailed message.

ext/bin/run-load-test
sebastian-miclea: Updated to overwrite the files if they exist.

locust/load-test/load-test.py
```diff
@@ -0,0 +1,49 @@
...
+                if (opts['method'] == 'GET'):
+                    limit = opts.get('limit')
+                    offset = opts.get('offset')
+                    url = opts['path'] + "?query=" + json.dumps(opts['query'])
+                    if limit:
+                        url = url + "&limit=" + str(limit)
+                    if offset:
+                        url = url + "&offset=" + str(offset)
+
+                        # pry()
```
Filipovici-Andrei, line 34: forgot a pry

locust/load-test/load-test.py
```diff
@@ -0,0 +1,49 @@
...
+            "-------------------------------------------" + " \n")
+
+    @task
+    def swarm(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        with open(dir_path + '/config.yaml') as stream:
+            config = yaml.safe_load(stream)
+            for opts in config:
+                if (opts['method'] == 'GET'):
+                    limit = opts.get('limit')
```
Filipovici-Andrei, line 26: Could we extract the blocks under `if (opts['method'] == 'GET'):` and POST type into separate functions?

locust/run-load-test
```diff
@@ -0,0 +1,48 @@
...
+        found = False
+        for arg in program_args:
+            if arg == default_arg[0]:
+                found = True
+        if not found:
+            additional_args += default_arg
+
+    return additional_args + program_args
+
+subprocess.run([sys.executable, "-m", "pip", "install", "locust", "pyyaml", "-U"])
```
Filipovici-Andrei, line 34: is this still necessary, if in the docs, you suggest installing locust before running the load test?

locust/load-test/load-test.py
```diff
@@ -0,0 +1,49 @@
...
+        if response.status_code == 0:
+            print(response.error)
+            exit(1)
+        elif response.status_code != 200:
+            print(
+            "Request: " + opts['name'] + " \n" +
+            "Method: " + opts['method'] + " \n" +
+            "Response status: " + str(response.status_code) + " \n" +
+            "Response body: " + response.text + " \n" +
+            "-------------------------------------------" + " \n")
```
Filipovici-Andrei, line 17: you could use something like this
`            print("Request: " + opts['name'],
                  "Method: " + opts['method'],
                  "Response status: " + str(response.status_code),
                  "Response body: " + response.text,
                  end='\n-------------------------------------------\n', sep='\n')`

ext/bin/run-load-test
sebastian-miclea: Updated, now all arguments are passed directly to locust.

locust/load-test/load-test.py
sebastian-miclea: Removed, thanks.

locust/load-test/load-test.py
sebastian-miclea: Yes, I cleaned up the code a bit.

locust/run-load-test
sebastian-miclea: I've updated the docs. We need to update the locust gem because we use the `--html` option that was added in 1.4.2 which is quite recent.

locust/load-test/load-test.py
sebastian-miclea: Updated, thanks!

Filipovici-Andrei (approved):

PR #3487, opened 2021-05-14
(maint) enabled integration test 

Filipovici-Andrei:
Ticket PDB-3461 is done.
As a result we can uncomment integration test sensitive-parameter-redaction.


Zak-Kent (approved):
👍 thanks for fixing this!

PR #3490, opened 2021-05-18
(PDB-5026) Switch context in query when using from

oanatmaria:
**Description of the problem:** If the options (limit, offset, order_by) are provided to a subquery that uses "from", the context isn't switched to the entity from the "from" syntax. This behaviour results in an error when the subquery is made on the reports table. Eg. 
```
["from" "nodes"
    ["and"
        ["in"
         "certname"
         ["from"
         "reports"
              ["extract" "certname" ["=" "type" "agent"]]
              ["limit" 1]
              ["order_by" [["certname" "desc"]]]]]
        ["=" "node_state" "active"]]]
```
Results in the following error: 
```'type' is not a queryable object for nodes. Known queryable objects are 'cached_catalog_status', 'catalog_environment', 'catalog_timestamp', 'certname', 'deactivated', 'expired', 'facts_environment', 'facts_timestamp', 'latest_report_corrective_change', 'latest_report_hash', 'latest_report_job_id', 'latest_report_noop', 'latest_report_noop_pending', 'latest_report_status', 'report_environment', and 'report_timestamp'```
**Description of the change:** Ensure that context is correctly set and switched when a subquery with "from" form is used.

### DIFF
--- a/test/puppetlabs/puppetdb/http/nodes_test.clj
+++ b/test/puppetlabs/puppetdb/http/nodes_test.clj
@@ -433,19 +433,18 @@
           (are-error-response-headers headers)
           (is (re-find msg body)))))))
 
-;TODO fix all skipped tests
-(deftest-http-app ^:skipped tests-for-PDB-5026
+(deftest-http-app query-with-from
   [[version endpoint] endpoints
-  method [:get :post]]
+   method [:get :post]]
 
-  ; enable this test after ticket https://tickets.puppetlabs.com/browse/PDB-5026 is solved
-  (testing "in a reports subquery"
-    (let [expected ["web1.example.com" "web2.example.com" "puppet.example.com" "db.example.com"]
-          result (query-result method endpoint  ["in", "certname",
-                                                 ["from", "reports",
-                                                  ["extract", "certname"],
-                                                  ["order_by", ["certname"]]]])]
-      (is (= (mapv :certname result) expected)))))
+  (testing "should change context to reports"
+    (store-example-nodes)
+    (let [expected ["db.example.com" "puppet.example.com" "web1.example.com"]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/puppetlabs/puppetdb/http/nodes_test.clj [hunk 1]
```diff
@@ -433,19 +433,18 @@
...
-          result (query-result method endpoint  ["in", "certname",
-                                                 ["from", "reports",
-                                                  ["extract", "certname"],
-                                                  ["order_by", ["certname"]]]])]
-      (is (= (mapv :certname result) expected)))))
+   method [:get :post]]
+
+  (testing "should change context to reports"
+    (let [expected ["db.example.com" "puppet.example.com" "web1.example.com"]
+          _ (store-example-nodes)
```
Filipovici-Andrei, line 442: I think that if you don't use the return value from `store-example-nodes` you can omit  the `_`

rbrw: Sounds reasonable, and alternately, could just replace the `let` entirely, i.e.
```clj
  (testing "should change context to reports"
    (store-example-nodes)
    (is (= ["db.example.com" "puppet.example.com" "web1.example.com"]
           (->> ["in" "certname"
                 ["from" "reports"
                  ["extract" "certname"] ["order_by" ["certname"]]]]
                (query-result method endpoint)
                (mapv :certname result)
                sort)))))
```
But not critical either way.

In any case, I would suggest removing the commas, and flipping the test order from `(is (= actual expected))` to `(is (= expected actual))`, since that's what `clojure.test` expects.

rbrw (commented):
Aside from the minor test suggestions, looks good.

PR #3500, opened 2021-06-03
(PDB-5102) Add example of to_string function usage

oanatmaria:


documentation/api/query/v4/ast.markdown
```diff
@@ -209,7 +209,18 @@ specified column is not null.
...
+
+```
+["extract", [["function", "to_string", "producer_timestamp", "month"]]]
+```
+
+To get the last 2 digits of the year a report was submitted  from the Puppet Server:
+
+```
+["extract", [["function", "to_string", "producer_timestamp", "YY"]]]]
+```
```
Filipovici-Andrei, line 223: I think that having two examples for timestamps formatting, isn't necessary. I think that the second example should be about converting integers into strings. Something like this:
To convert the `uptime_seconds` fact from integer to string use the following query on the facts endpoint:
["extract", [["function", "to_string", "value"]], ["=","name", "uptime_seconds"]]

Filipovici-Andrei: Probably for integer conversion we need to provide a format mask? Always

PR #3501, opened 2021-06-17
(PE-32070) Update Java security files for new BCFIPS

Magisus:
There is an issue with bc-fips 1.0.2.1 involving circular provider
loading, that can be fixed by using the full class name for the
BCJSSE FIPS provider, rather than relying on an alias. This commit
updates our java-security files to use the full name. Note this only
caused issues on Java 11, but it is valid for both versions, so we are
updating it in both files for consistency.

### DIFF
--- a/dev-resources/jdk11-fips-security
+++ b/dev-resources/jdk11-fips-security
@@ -61,7 +61,7 @@
 # List of providers and their preference orders (see above):
 #
 security.provider.1=org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider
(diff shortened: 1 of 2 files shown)


Magisus:
Those failures look unrelated to this.

mwaggett (approved):

austb:
This looks like just dev-resources changes, so I don't think this needs to land for this release. Is that right?

Magisus:
Yeah that's true. This is prep for https://github.com/puppetlabs/clj-parent/pull/290

mwaggett (approved):

Magisus:
Rebased this.

PR #3506, opened 2021-06-25
(PDB-5104) Fix to_string function on facts and fact-contents endpoint

oanatmaria:
**Description of the problem (short):** Query with to_string function doesn't return anything on the facts endpoint and fails with the following error in fact-contents endpoint: 
```
Input to munge-result-row does not match schema:
           [(named {:to_string disallowed-key} row)]
```

**Description of the problem (long):** An optimisation was introduced, for queries that constrain the fact name to a
single value if they are made on the facts endpoint, in [this commit](https://github.com/puppetlabs/puppetdb/commit/8d5b509c9b6e5aa529bd0667694a8bd036100d42 ). Because the json -> operator for [the selection field](https://github.com/puppetlabs/puppetdb/blob/2126eb44069fb281330593b655fe95ffe97b72a2/src/puppetlabs/puppetdb/query_eng/engine.clj#L2740) was introduced, the parameters for the fact-name field (that will replace the "?" later) are added as the value of :selection-params. These :selection-params are extracted later in [extract-params method](https://github.com/puppetlabs/puppetdb/blob/2126eb44069fb281330593b655fe95ffe97b72a2/src/puppetlabs/puppetdb/query_eng/engine.clj#L1720), but are extracted later than they appear when the tree is traversed in [post-order](https://github.com/puppetlabs/puppetdb/blob/2126eb44069fb281330593b655fe95ffe97b72a2/src/puppetlabs/puppetdb/query_eng/engine.clj#L1726) and the :selected-params end up in an erroneous place in the final parameterized-plan's params. 

On the fact-contents endpoint the schema doesn't accept `to_string` as a column name. 

In addition to the unit tests, this change was tested manually with the following queries:
`["extract", [["function", "to_string", "value", "999999999"]], ["=","name", "uptime_seconds"]]`
```
["extract", [["function", "to_string", "value", "9999999"]], ["and",
  ["=", "name", "uptime_seconds"],
  ["in", "certname",
    ["extract", "certname",
      ["select_resources",
        ["and",
          ["=", "type", "Class"],
          ["=", "title", "Apache"]]]]]]]
```
**Description of the solution:** When traversing the tree in post-order, add :selection-params when they are firstly seen (in other words, modify the condition to add the :selection-params when the :selection map is firstly encountered, not in the end when the object is Query (object is Query only in the last iteration when root is visited)). For other endpoints, including fact-contents, allow `to_string` as a column name in schema.

### DIFF
--- a/src/puppetlabs/puppetdb/query.clj
+++ b/src/puppetlabs/puppetdb/query.clj
@@ -85,7 +85,8 @@
           (s/optional-key :min) s/Any
           (s/optional-key :max) s/Any
           (s/optional-key :avg) s/Any
-          (s/optional-key :sum) s/Any}))
+          (s/optional-key :sum) s/Any
+          (s/optional-key :to_string) s/Str}))
 
 (defn execute-paged-query*
   "Helper function to executed paged queries.  Builds up the paged sql string,
--- a/src/puppetlabs/puppetdb/query_eng/engine.clj
+++ b/src/puppetlabs/puppetdb/query_eng/engine.clj
@@ -1710,14 +1710,14 @@
     {:node (assoc node :value "?")
      :state (apply conj (:params node) state)}
 
-    ;; Handle a parameterized :selection -- the query's :selection
-    ;; must already have question marks in all the right places, and
-    ;; have the corresponding parameter values in :selection-params.
+    ;; Handle a parameterized :selection -- when the :selection
+    ;; appears in the tree traversal put the corresponding parameter values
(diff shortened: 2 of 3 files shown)


austb:
I didn't get to this until late this afternoon, but I am inclined to like that this solution worked without changing the pre/post order of the entire traversal. I'll need to take a closer look at the plan structure on Monday, but I think this is a preferable route to go.

austb (approved):

PR #3507, opened 2021-06-28
(PDB-5105) Add console and cd4pe queries

Filipovici-Andrei:
Add different .yaml files for the console, cp4pe and example queries.
Introduced an optional alias field in the .yaml files for easy identification of the request.
Requests query field is now optional.

### DIFF
--- a/locust/load-test/example.yaml
+++ b/locust/load-test/example.yaml
@@ -1,11 +1,11 @@
 # Example
+#  name: my custom call name, will be shown in the report
 #  path: path without the host. The host is set in cli as an option
 #  query: query should be hash for POST and string for GET
 #  method: GET/POST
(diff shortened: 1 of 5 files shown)


locust/load-test/example.yaml
```diff
@@ -10,7 +11,9 @@
   offset: 1
   method: GET
   headers: ''
+  alias: example 1
```
sebastian-miclea, line 14: I don't think this should be mandatory, maybe leave it optional, and default to query if it's not set.

sebastian-miclea (approved):

locust/load-test/example.yaml
Filipovici-Andrei: Fixed.

PR #3527, opened 2021-07-12
(PDB-5114) Make a first pass at profile directed query/command optimizations

rbrw:


austb (commented):
Looks good, but the ex-info errors are causing eastwood to fail. They'll either need a second arg or to be converted into IllegalArgumentExceptions

austb (approved):

PR #3533, opened 2021-07-15
(PDB-5168) Log PQL long parsing time

oanatmaria:
If the PQL query parsing it's taking too long, a warning message it's logged that includes the query and the time the parsing took for that query.

### DIFF
--- a/src/puppetlabs/puppetdb/http/query.clj
+++ b/src/puppetlabs/puppetdb/http/query.clj
@@ -6,6 +6,7 @@
   (:require [puppetlabs.puppetdb.cheshire :as json]
             [clojure.java.io]
             [clojure.core.match :as cm]
+            [clojure.tools.logging :as log]
             [clojure.walk :refer [keywordize-keys stringify-keys]]
             [puppetlabs.puppetdb.query-eng :as qeng]
             [clojure.set :as set]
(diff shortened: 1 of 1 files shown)


Filipovici-Andrei (approved):

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -304,13 +305,26 @@
...
+(defn time-and-parse-pql
+  "Time how long it takes to transform a PQL query to AST, if the time
+   is greater than 1s, a warning is logged. Returns the transformed
+   query."
+  [query]
+  (let [start (System/nanoTime)
+        result (pql/parse-pql-query query)
+        elapsed (/ (- (System/nanoTime) start) 1000000.0)
+        max-pql-limit 1000]
+    (if (> elapsed max-pql-limit)
```
rbrw, line 317: Perhaps `when`?

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -304,13 +305,26 @@
...
+      (log/warn (trs "Parsing PQL query `{0}` took: `{1}` ms" query elapsed)))
```
rbrw, line 318: If you don't feel strongly about it, then suggest removing the backticks for both values, using `pr-str` to formally quote the query, so it'll be easy to unambiguously copy paste from the logs, and putting the query at the end, since it's potentially long, variable length content, and doing so might also make it easier to copy/paste

i.e. I was pondering something like this:
```clj
(log/warn (trs "Parsing PQL took {0} ms: {1}" elapsed (pr-str query)))
```

src/puppetlabs/puppetdb/http/query.clj
rbrw: (Hmm...not sure how I posted that twice.)

PR #3539, opened 2021-07-26
(PDB-5135) Notify when resource_event size limit is reached

sebastian-miclea:

PR #3541, opened 2021-07-28
(PDB-5135) notify when resources are oversized

sebastian-miclea:


austb (changes requested):

sebastian-miclea:
This PR is closed because the termini implementation was preferred. PR https://github.com/puppetlabs/puppetdb/pull/3539

PR #3543, opened 2021-07-28
(PDB-5214) (Re)allow dotted projections with group_by and functions

rbrw:
Dotted projection arguments to group_by and functions were being
rejected.  Fix that, and add some initial tests.

  ["from" "inventory"
   ["extract" [["function" "count"] "facts.kernel"]
    ["=" "node_state" "active"]
    ["group_by" "facts.kernel"]]]


rbrw:
Note that this also completely reverts the AST query field parser work to reduce risk for pending releases, with the intention of restoring it immediately after they're finished.

oanatmaria (approved):

oanatmaria:
There are some failing integration tests due to a new behaviour introduced in puppet. The tests are fixed in [this PR](https://github.com/puppetlabs/puppetdb/pull/3544).

PR #3551, opened 2021-08-06
(PDB-5213) update clj-parent to 4.6.29

Filipovici-Andrei:


oanatmaria (commented):
`(def i18n-version "0.8.0")` shouldn't this be update to the newer version as well?

oanatmaria (approved):

PR #3566, opened 2021-09-02
Test clj-parent 4.7.3

rbrw:


austb:
These changes will be taken up in https://github.com/puppetlabs/puppetdb/pull/3564

PR #3569, opened 2021-09-10
(PDB-5256) Add details to message when AST validation fails

sebastian-miclea:
Example: 
`AST validation failed, but was successfully converted to SQL. Please file a PuppetDB ticket at https://tickets.puppetlabs.com
Unrecognized ast clause ["order_by" "certname"] in ast query ["extract" "certname" ["subquery" "reports" ["order_by" "certname"]]]`

Also updated docs that had invalid query as example.

### DIFF
--- a/src/puppetlabs/puppetdb/query_eng.clj
+++ b/src/puppetlabs/puppetdb/query_eng.clj
@@ -129,12 +129,13 @@
                         (log/error e (trs "Unknown exception when processing ast to add report type filter(s)."))
                         (throw e))
                       (log/error e msg)
-                      ::failed))
+                      {:type ::failed, :message msg}))
                   (catch Exception e
                     (log/error e (trs "Unknown exception when processing ast to add report type filter(s)."))
                     (throw e)))]
-        (if (= ast ::failed)
-          (throw (ex-info "AST validation failed, but was successfully converted to SQL. Please file a PuppetDB ticket at https://tickets.puppetlabs.com"
+        (if (and (map? ast) (= (:type ast) ::failed))
+          (throw (ex-info (trs "AST validation failed, but was successfully converted to SQL. Please file a PuppetDB ticket at https://tickets.puppetlabs.com \n{0}"
+                               (:message ast))
                           {:kind ::dr/unrecognized-ast-syntax
                            :ast query
                            :sql (eng/compile-user-query->sql query-rec query options)}))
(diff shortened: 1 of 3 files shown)


src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -129,15 +129,16 @@
...
-        (if (= ast ::failed)
-          (throw (ex-info "AST validation failed, but was successfully converted to SQL. Please file a PuppetDB ticket at https://tickets.puppetlabs.com"
+        (if (and (map? ast) (= (:type ast) ::failed))
+          (throw (ex-info (trs "AST validation failed, but was successfully converted to SQL. Please file a PuppetDB ticket at https://tickets.puppetlabs.com \n{0}"
+                               (:message ast))
                           {:kind ::dr/unrecognized-ast-syntax
-                           :ast query
-                           :sql (eng/compile-user-query->sql query-rec query options)}))
+                           :ast  query
+                           :sql  (eng/compile-user-query->sql query-rec query options)}))
```
austb, line 141: Looks like these two lines added an additional space, but nothing actually changed?

austb (approved):
This looks right to me, we don't seem to support the subquery style where `[certname]` is omitted in any other query.

test/puppetlabs/puppetdb/http/reports_test.clj
```diff
@@ -828,7 +828,7 @@
              ;; This is abusing the existence of PDB-4734 to throw an error from a malformed AST query
              (let [{:keys [status body]} (query-response method endpoint query)]
                 (is (= status http/status-internal-error))
-                (is (re-matches #"AST validation failed, but was successfully converted to SQL.*" body)))))))
+                (is (re-matches #"(?s)AST validation failed, but was successfully converted to SQL.*" body)))))))
```
oanatmaria, line 831: We should check that the message for "Unrecognized ast clause ..." is present here as well

src/puppetlabs/puppetdb/query_eng.clj
sebastian-miclea: Missed these. Thanks!

test/puppetlabs/puppetdb/http/reports_test.clj
sebastian-miclea: Updated. Thanks!

sebastian-miclea:
> This looks right to me, we don't seem to support the subquery style where `[certname]` is omitted in any other query.

Yes, I had a look with Oana and came to the conclusion that it won't work without `[certname]`

oanatmaria (approved):

PR #3573, opened 2021-09-17
(PDB-5278) Fix group by dotted fact path with forward slash

austb:
When grouping by a keyword, honeysql will convert the keyword to SQL by
calling `name` on it. But that will not return the entire fact name when
there's a forward slash in it because Clojure interprets everything
before the forward slash as the "namespace" and everything after as the
"name".

```
=> (name :facts.foo)
"facts.foo"
=> (name :facts.f/oo)
"oo"
=> (namespace :facts.f/oo)
"facts.f"
```

This commit changes the query engine to use sql raw instead of the
keyword to avoid splitting on forward slashes.


sebastian-miclea:
I had a look over this PR and tested on PDB-5262 and that case is also resolved.

sebastian-miclea (approved):

PR #3594, opened 2021-10-26
(PDB-5326) Add db storage for workspaces

stelcodes:
From first commit message:

This commit creates a new migration which prepares for the upcoming
workspaces feature by adding two new tables: workspaces and
workspace_membership. These tables track workspaces and which certnames
belong in which workspaces. Design specific notes:

- In the workspace_membership table, the certname column purposefully
  does not have a foreign key constraint because we want to potentially
  include deleted certnames in workspace_membership query results

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -2038,6 +2038,19 @@
   (jdbc/do-commands
    "ALTER TABLE certnames ADD COLUMN catalog_inputs_hash bytea"))
 
+(defn add-workspaces-tables
+  []
+  (jdbc/do-commands
+   ["CREATE TABLE workspaces ("
+    "  uuid UUID PRIMARY KEY,"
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -2038,6 +2038,19 @@
...
    "ALTER TABLE certnames ADD COLUMN catalog_inputs_hash bytea"))
 
+(defn add-workspaces-tables
+  []
+  (jdbc/do-commands
+   ["CREATE TABLE workspaces ("
+    "  workspace_id UUID PRIMARY KEY,"
+    "  updated TIMESTAMP WITH TIME ZONE NOT NULL)"]
+
+   ["CREATE TABLE workspace_membership ("
```
austb, line 2048:
```suggestion
   ["CREATE TABLE workspace_memberships ("
```

I think this table name should be plural as well. Doesn't really matter one way or the other but most of our table names are plural

PR #3600, opened 2021-11-17
(PDB-5378) Get latest report timestamp from certnames table

austb:
Scanning the report partitions for the latest report timestamp is
terribly inefficient. We have the latest report timestamp in the
certnames table so just use that.


stelcodes (approved):

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1750,13 +1754,13 @@
   ([certname :- String]
    (deactivate-node! certname (now)))
   ([certname :- String timestamp :- pls/Timestamp]
-   (if (have-newer-record-for-certname? certname timestamp)
-     (log/warn (trs "Not deactivating node {0} because local data is newer than {1}."
-                    certname timestamp))
-     (let [sql-timestamp (to-timestamp timestamp)]
+   (let [sql-timestamp (to-timestamp timestamp)]
```
austb, line 1757: I went with using `java.sql.Timestamp` because I could just move this let and re-use the conversion we were already doing. But it does feel like Timestamp class-creep. If we want to try to standardize on using a Joda DateTime for our comparisons in the hope that we eventually move to java time directly, I'm happy to make that change here.

rbrw (approved):

PR #3603, opened 2021-11-19
(PDB-5336) Create migration 75 schema diff test

stelcodes:
Add schema diff test to the existing migration 75 test.

### DIFF
--- a/test/puppetlabs/puppetdb/scf/migrate_test.clj
+++ b/test/puppetlabs/puppetdb/scf/migrate_test.clj
@@ -1410,12 +1410,53 @@
         (is (= expected
                (first hashes)))))))
 
+(def migration-75-expected-diff
+  "Table partitions for reports are created for a range of dates automatically
+  in migration 74 so we must programmatically create the expected diff based on
+  the current date and the aforementioned range."
+  (let [generic-table-diff {:left-only nil,
+                            :right-only
+                            {:numeric_scale nil,
+                             :column_default "'agent'::text",
+                             :character_octet_length 1073741824,
+                             :datetime_precision nil,
+                             :nullable? "NO",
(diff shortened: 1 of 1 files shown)


test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -1437,7 +1438,209 @@
...
+                   :column_default "'agent'::text",
+                   :character_octet_length 1073741824,
+                   :datetime_precision nil,
+                   :nullable? "NO",
+                   :character_maximum_length nil,
+                   :numeric_precision nil,
+                   :numeric_precision_radix nil,
+                   :data_type "text",
+                   :column_name "report_type",
+                   :table_name "reports_20211115z"},
```
austb, line 1470: These table names are going to be dependent on what day the test are run on due to the logic over in [`migrate.clj`](https://github.com/puppetlabs/puppetdb/blob/10435954d814f737a94752d264595d4a7e88383a/src/puppetlabs/puppetdb/scf/migrate.clj#L1891). We could either generate the list using the same date range logic, or regex match the `:table_name`. I would guess that generating the table names with the range logic is likely simpler.

stelcodes: Oh wow yeah that's a problem! Where is that date range logic? It looks like in that test we are only inserting a single report so why are there multiple days or report table partitions?

austb: it's here https://github.com/puppetlabs/puppetdb/blob/10435954d814f737a94752d264595d4a7e88383a/src/puppetlabs/puppetdb/scf/migrate.clj#L1891

austb (approved):

PR #3604, opened 2021-11-26
Add support for IPv6

jovandeginste:
I'm not sure if there are other missing parts, but so far it works for us with the changes in this PR.


jovandeginste:
Apologies for the force-pushes - I realized I also changed "old" (changelog) documentation, which is probably not to be touched.

stelcodes:
Hey @jovandeginste, thanks for this PR! The PuppetDB team is a bit overwhelmed with backlog at the moment, but we will discuss these changes as soon as we can.

jcpunk:
Any news on this?  I'd love to have IPv6 working with the puppetdb container.

PR #3606, opened 2021-12-06
(maint) travis: don't notify on repeated green builds

austb:


austb:
don't merge, added a failing test to test notifications

PR #3608, opened 2021-12-16
(PDB-5129) Update clj-parent to 4.9.0 for honeysql 1.0.461

rbrw:


test/puppetlabs/puppetdb/query_eng_test.clj
```diff
@@ -46,16 +47,17 @@
...
+       (-> {:projections {"foo" {:type :string
+                                 :queryable? true
+                                 :field :table.foo}}
+            :alias "thefoo"
+            :subquery? false
+            :where (->BinaryExpression := 1 1)
+            :selection {:from [:table]}
+            :source-tables #{:table}}
+           map->Query
+           (plan->sql {:node-purge-ttl (parse-period "14d")})))))
```
rbrw, line 60: I just threaded this while fixing the indentation.  The only substantive change here should be the change to the regex, i.e. `1 = 1` -> `? = ?`.

austb (approved):
I wasn't able to identify any other places we might need this inlining, but I don't really know if that is just because they could be difficult to find, or if we should feel confident about this change. I really wish we would stop doing our bespoke parameter parsing and actually _use_ honeysql instead of only using half of it an reinventing the other half to our detriment.

PR #3609, opened 2022-01-05
SCA With Snyk

LivingInSyn:
This PR does a few things:

* sets up scanning of the clojure code with snyk
* sets up scanning of the base ruby code with snyk
    * part of this is committing a gemfile.lock and removing the corresponding gitignore
* removes lein nvd in favor of snyk


Gemfile.lock
```diff
@@ -0,0 +1,195 @@
+GIT
+  remote: https://github.com/puppetlabs/puppet.git
+  revision: 5d7a81d87fc72c7c9a48d6779ae6ec950b5fdfc9
+  ref: 6.x
+  specs:
+    puppet (6.26.0)
```
stelcodes, line 6: Puppet 6.26.0 hasn't been released yet and that seems to be the cause of at least one of the failing tests

LivingInSyn: the gemfile.lock was created from main with `bundle lock`

LivingInSyn:
@austb removed the ruby scanning

austb (approved):
Looks good! We'll need to adjust our jenkins pipelines to remove the lein-nvd checks that run there and then this will be good to go

PR #3626, opened 2022-01-31
(PDB-5423) Only run "other" gc again after 24 hours

austb:
Run the fact path gc less frequently in order to lessen its impact on
postgres.

The fact path gc that is run during the "other" gc is quite expensive.
This can result in it taking a long time and/or putting Postgres under
load. The recent optimization to fact_contents did not provide any
improvements to the fact path gc operation, so this is ideally a
temporary fix while we further investigate why the operation is so
expensive and long running.


stelcodes:
Just a heads up, this change is going to conflict with #3623

rbrw (approved):

PR #3627, opened 2022-01-31
(maint) Catch everything in query middleware

austb:
In order to safely `assert` things in the query code, we need to ensure
that we do not leak AssertionErrors, which are not caught by `catch
Exception`. Add catch Throwable as a catch-all backstop to
any other thing that could be thrown.


stelcodes (approved):
We could obviously just use the `catch Throwable` catch clause, but it's pretty inconsequential.

PR #3643, opened 2022-02-23
(PDB-5452) Disable postgres jit for all connections

rbrw:
Problems with the jit have been reported with the jit, which was enabled by
deafult in pg 12.  In particular, it can cause some queries to
run *much* more slowly, even in situations where there don't appear to
be many rows involved.

So until the problem is resolved, disable the jit for all pdb
connections, by adding a "set session jit = off;" to the hikaricp
connection init string.

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -648,9 +648,14 @@
        ;; level to fail (because the pending transaction already
        ;; includes that select).
        (.setIsolateInternalQueries true))
-     (some->> (when expected-schema
-                (block-on-schema-mismatch expected-schema))
-              (.setConnectionInitSql config))
+     (->> [(when read-only? "set transaction read write;")
+           "update pg_settings set setting = false"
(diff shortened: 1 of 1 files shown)


austb (approved):

stelcodes:
Looks like the `core+ext` test is failing because the bouncycastle jar is not in the `.m2` directory like it's supposed to be. This is affecting all of our PR's from what I can tell:
```
Unable to find the bouncy castle provider jar /Users/runner/.m2/repository/org/bouncycastle/bcprov-jdk15on/1.68/bcprov-jdk15on-1.68.jar; have you run "lein deps"?
```
And it looks like the Travis tests with Postgres 9.6 are failing because the `jit` setting is unrecognized:
```
ERROR:  unrecognized configuration parameter "jit"
STATEMENT:  set session jit = off
```
We may have to check the postgres version and conditionally send the command.

rbrw:
Thanks, and maybe fixed.

stelcodes (approved):

PR #3645, opened 2022-02-23
(maint) Add clj-kondo linter config

stelcodes:
clj-kondo is a popular Clojure linter used by clojure-lsp. There are certain common macros in the PDB codebase that clj-kondo does not recognize, causing it to emit tons of error messages. The `:lint-as` clj-kondo config option can help avoid these superfluous error messages be telling clj-kondo how to lint these macros.


stelcodes:
Closing because #3673 fixes all clj-kondo related issues

PR #3646, opened 2022-03-02
(PT-1027) Add desired policy storage

rbrw:
This may be roughly right, and I'm raising it so that the other policy storage can go ahead and build upon it, relying on `config/experimental-features` and the support for migrations with `:strict-order?`.  I've marked it as a work in progress because I still need to add some tests, and we need to decide if we want to move the migration-table related changes to 6.x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,6 +96,11 @@ jobs:
       env: PDB_TEST=core+ext/openjdk11/pg-14
       script: *run-core-and-ext-tests
 
+    - stage: ❧ pdb tests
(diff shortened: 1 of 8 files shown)


rbrw:
Oh, and also need to decide if we really want/need the `changed` timestamp index right now.  We might not.

rbrw:
This also needs to wait until we resolve (and presumably merge up) #3647.

rbrw:
Closing because the effort has been suspended.

PR #3649, opened 2022-03-10
(PDB-5466) Allow database config passwords to be INI integers

stelcodes:
https://tickets.puppetlabs.com/browse/PDB-5466


stelcodes:
We are not going to merge this patch because the translation from integer to string may be lossy (ex +100 -> "100")

PR #3650, opened 2022-03-14
(maint) acceptance test: use https for package repos

austb:


austb:
this fixed all the failing tests besides centos8, which we will stop testing because it is EOL.

PR #3667, opened 2022-04-21
(PDB-5482) Log queries before parsing

stelcodes:
Adds query logging before parsing so parsing failures are easier to debug. Please see commit messages for more info.

### DIFF
--- a/src/puppetlabs/puppetdb/http/query.clj
+++ b/src/puppetlabs/puppetdb/http/query.clj
@@ -279,8 +279,9 @@
       (update-when [:distinct_resources] coerce-to-boolean)))
 
 (defn parse-json-sequence
-  "Parse a query string as JSON. Parse errors
-   will result in an IllegalArgumentException"
+  "Parse a query string as JSON. Parse errors will result in an
+  IllegalArgumentException. Should not be used as a parse-fn directly. Use
+  parse-json-query instead"
   [query]
   (try
     (with-open [string-reader (java.io.StringReader. query)]
@@ -291,7 +292,9 @@
 (defn parse-json-query
   "Parse a query string as JSON. Multiple queries or any other
   data, after the query, will result in an IllegalArgumentException"
-  [query]
+  [query query-uuid log-queries?]
+  (when log-queries?
+    (log/info (trs "Parsing PDBQuery:{0}:{1}" query-uuid (pr-str query))))
   (when query
     (let [[parsed & others] (parse-json-sequence query)]
       (when others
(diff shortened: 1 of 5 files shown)


src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -304,82 +305,83 @@
...
+     :get (get-req->query req)
+     :post (post-req->query req)
      (throw (IllegalArgumentException. (tru "PuppetDB queries must be made via GET/POST"))))
    param-spec))
 
 (defn extract-query
   "Query handler that converts the incoming request (GET or POST)
   parameters/body to a pdb query map"
   ([handler param-spec]
-   (extract-query handler param-spec parse-json-query))
```
austb, line 366: I think we may want to keep this `parse-json-query` default because of the REST-style routes like `http://localhost:8080/pdb/query/v4/inventory`. Their only valid query content is JSON, such as
```
curl -X GET 'http://localhost:8080/pdb/query/v4/inventory' \
  --data-urlencode 'limit=1000' \
  --data-urlencode 'query=["extract", ["facts.foo"]]'
```

If we attempt to put PQL there it fails to parse as JSON
```
curl -X GET 'http://localhost:8080/pdb/query/v4/inventory' \
  --data-urlencode 'limit=1000' \
  --data-urlencode 'query=inventory[]{}'
```


```
Json parse error at line 1, column 10:

inventory[]{}
```

By removing the two parse functions, it fails instead with the error `Input to paged-sql does not match schema: ...` because it has parsed the PQL and created an invalid query
```
["from","inventory",["from","inventory",["extract",[],["=","node_state","active"]]]]
```

stelcodes: You're totally right. I didn't realize that PQL queries were not allowed on entity endpoints but looking back at the handlers that makes sense. I totally got rid of the `parse-fn` removal commit and simplified the existing changes. I also added a test for that exact example ^^

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -304,48 +307,53 @@
...
+  (when log-queries?
+    (log/info (trs "Parsing PDBQuery:{0}:{1}" query-uuid (pr-str query))))
   (let [start (System/nanoTime)
         result (pql/parse-pql-query query)
         elapsed (/ (- (System/nanoTime) start) 1000000.0)
         max-pql-limit 1000]
-    (when (> elapsed max-pql-limit)
-      (log/warn (trs "Parsing PQL took {0} ms: {1}" elapsed (pr-str query))))
+    (when (and log-queries? (> elapsed max-pql-limit))
+      (log/warn (trs "Parsing PQL took {0} ms for PDBQuery:{1}:{2}" elapsed query-uuid (pr-str query))))
```
stelcodes, line 318: @austb One thing I'm not sure about: does it make sense to only print the PQL time logging when `log-queries` is true? Previously it always printed out but I feel like it should be behind that flag with the rest of the query logging

stelcodes, line 318: @austb  One thing I'm not sure about: does it make sense to only print the PQL time logging when log-queries is true? Previously it always printed out but I feel like it should be behind that flag with the rest of the query logging

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -370,10 +378,11 @@
       (if puppetdb-query
         req
         (let [query-map (create-query-map req param-spec parse-fn)
+              query-uuid (str (java.util.UUID/randomUUID))
               pretty-print (:pretty query-map
                                     (get-in req [:globals :pretty-print]))]
           (-> req
-              (assoc :puppetdb-query query-map)
+              (assoc :puppetdb-query query-map :puppetdb-query-uuid query-uuid)
               (assoc-in [:globals :pretty-print] pretty-print))))))))
```
stelcodes, line 386: I just pushed a change to this part, just simplifying it by adding the query-uuid in the existing assoc call

stelcodes:
Rebased onto origin/6.x recheck CI

austb (approved):

stelcodes:
A single travis test is failing because it's "exceeding maximum log length" so I'm ignoring that one -_-

austb:
that sounds reasonable to me. We should probably look to rein in the log length of puppetdb tests eventually.

PR #3669, opened 2022-04-26
(HDP-161) Remove `submit_only` server URL when it times out repeatedly

Magisus:
This commit adds logic to PDB's Ruby http utils to track timeouts for
the `submit_only` server URLs, which are used by HDP. If a given URL
times out twice in a row, it is removed from the loaded config, to avoid
tying up JRubies indefinitely in the case where HDP has gone down. By
removing the URL, PDB will no longer try to submit things to it,
allowing requests to the other configured URLs to proceed quickly.

The URL will be re-added to the config whenever the JRuby instance is
refreshed: with a restart of Puppet Server, when the JRuby instance is
recreated when the "max requests per instance" threshold is hit, or if
the pool is flushed manually. Otherwise, the misbehaving URL will be
kept out of the list permanently.

### DIFF
--- a/puppet/lib/puppet/util/puppetdb/http.rb
+++ b/puppet/lib/puppet/util/puppetdb/http.rb
@@ -12,6 +12,8 @@ module Puppet::Util::Puppetdb
 
     @@last_good_query_server_url_index = Atom.new(0)
 
+    @@submit_only_url_timeouts = {}
+
     # Concat two server_url snippets, taking into account a trailing/leading slash to
     # ensure a correct server_url is constructed
     #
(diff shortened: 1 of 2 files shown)


Magisus:
Finding out if I need to retarget this at 6.x.

Magisus:
Confirmed I need to port this to 2019.8. Gonna leave this open though, since I think it's different enough that we will probably need to merge both independently (and deal with the resulting conflicts).

Magisus:
Also looks like the Docker failure is an "always" thing right now that's unrelated to this change.

Magisus:
Backport to 6.x: https://github.com/puppetlabs/puppetdb/pull/3671

Magisus:
This was superseded by a mergeup of the 6.x PR to main.

PR #3670, opened 2022-04-28
(PDB-5483) Fix set transaction read write error on hot standby

jplindquist:
Fixes upgrade error introduced in #3643 (which seems to be the original intention):
 `adding a "set session jit = off;" to the hikaricp connection init string.`

Ref: https://www.postgresql.org/docs/current/hot-standby.html#HOT-STANDBY-USERS

```
Transaction management commands that explicitly set non-read-only state: 
(SET TRANSACTION READ WRITE) are explicitly denied while configured as hot standby, 
but you can change a run time parameter settings using SET
```

Tested this against both PostgreSQL 9.6.24 and latest 14.x

### DIFF
--- a/src/puppetlabs/puppetdb/jdbc.clj
+++ b/src/puppetlabs/puppetdb/jdbc.clj
@@ -648,10 +648,7 @@
        ;; level to fail (because the pending transaction already
        ;; includes that select).
        (.setIsolateInternalQueries true))
-     (->> [(when read-only? "set transaction read write;")
-           "update pg_settings set setting = false"
-           "  where name = 'jit';"
-           (when read-only? "set transaction read only;")
+     (->> ["set session jit = off;"
            (when expected-schema
              (block-on-schema-mismatch expected-schema))]
           (str/join " ")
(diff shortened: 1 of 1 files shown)


rbrw:
Thanks for the help.  I'm not positive (I'll have to double-check), but I think I may have tried something like that originally, but had to rework it to the more complicated approach after discovering that 9.6 wouldn't let you mention `jit` at all in a `set`, even if the call is conditionalized so that it'll never run.

At least, I have a note to that effect, but if that's actually right, and I didn't just misunderstand the situation, then why didn't the travis 9.6 tests fail...

In any case, I'll plan to investigate a bit more.

jplindquist:
Thank you!  I did try it with 9.6 via docker at least and it seemed to work with the conditional, effectively doing nothing and ignoring it, and applying the setting properly in latest 14.x.  I didn't check every version in between, but found the setting first introduced in 11.x so started there with the version comparison at least. 🤞

rbrw:
OK, thanks.  I'll investigate a bit more here, but if I can't determine what was going on before, we'll just assume I was off in the weeds somehow.  And for now, if 9.6, 11, and 14 are working, I think that will be sufficient.

rbrw:
Merged with some minor changes (listed in the commit message) as 44ba7a58ac73bc3963f70492f145031c86f52850

Thanks for the help.

PR #3671, opened 2022-04-28
(HDP-161) Remove `submit_only` server URL when it times out repeatedly (6.x)

Magisus:
This commit adds logic to PDB's Ruby http utils to track timeouts for
the `submit_only` server URLs, which are used by HDP. If a given URL
times out twice in a row, it is removed from the loaded config, to avoid
tying up JRubies indefinitely in the case where HDP has gone down. By
removing the URL, PDB will no longer try to submit things to it,
allowing requests to the other configured URLs to proceed quickly.

The URL will be re-added to the config whenever the JRuby instance is
refreshed: with a restart of Puppet Server, when the JRuby instance is
recreated when the "max requests per instance" threshold is hit, or if
the pool is flushed manually. Otherwise, the misbehaving URL will be
kept out of the list permanently.

### DIFF
--- a/puppet/lib/puppet/util/puppetdb/http.rb
+++ b/puppet/lib/puppet/util/puppetdb/http.rb
@@ -12,6 +12,11 @@ module Puppet::Util::Puppetdb
 
     @@last_good_query_server_url_index = Atom.new(0)
 
+    # Tracks when timeouts occur when making requests against the `submit_only_server_urls`.
+    # If one of these times out repeatedly, we remove it from the list so we stop trying to
+    # submit data to it.
+    @@submit_only_url_timeouts = {}
+
     # Concat two server_url snippets, taking into account a trailing/leading slash to
     # ensure a correct server_url is constructed
     #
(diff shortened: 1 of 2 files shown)


Magisus:
Note: each Jruby only processes on request at a time, so there are no race conditions in this code. However each JRuby will will track timeouts and remove the URL from its loaded config independently.

Magisus:
The Docker tests appear to be failing all the time right now, unrelated to this change.

puppet/lib/puppet/util/puppetdb/http.rb
```diff
@@ -141,7 +168,14 @@ def self.failover_action(path_suffix, server_urls, sticky, ssl_context = nil, ht
           end
 
           response = Timeout.timeout(config.server_url_timeout) do
-            http_callback.call(http, route)
+            response = http_callback.call(http, route)
+
+            if config.submit_only_server_urls.include?(server_url)
+              # Clear any previously registered timeout for this URL
+              @@submit_only_url_timeouts[server_url] = false
```
austb, line 175: I believe we will need to copy this logic in `broadcast_action` as well. It is configurable in the `action` method below which one a command will use.

Magisus: Okay, yeah found a testing hole now that you point this out, also. Working to fix that up.

Magisus:
Okay I made sure to add tests for both `broadcast_command=true/false`. The ones for `false` are a little funky, since it apparently doesn't try the `submit_only` url unless the ones before it both fail, which means that all URLs need to fail to get to testing the `submit_only` timeout.

PR #3672, opened 2022-04-29
Merge 6.x -> main; Modifications for HDP-161

Magisus:


Magisus:
I dropped all the changes to the terminus in the initial mergeup, then reapplied them as their own commit after doing the rework. Lemme know if something looks off about that.

austb (approved):

PR #3673, opened 2022-05-05
(PDB-5484) Add clj-kondo

rbrw:


rbrw:
Recommend looking at this in something like gitk that keeps the commits in branch order, and likely with ignore blanks turned on.  There are also references to related commits, providing some context, in some of the commit messages.

I've marked it work-in-progress because if nothing else, I'm not sure about keeping the final commit, and we also need to be sure the new logic is accurate in the important respects (also wondered if we could simplify it further).  If we're not sure, or don't want to think about it, I'll just drop it.

.clj-kondo/hooks.clj
```diff
@@ -0,0 +1,135 @@
...
+(defn rewrite-defspec
+  [name num-tests-or-options? property]
+  {:node (list-node (list (token-node 'def) name property))})
+
+(defn defspec
+  [{{[_ name & others] :children} :node}]
+  (case (count others)
+    1 (apply rewrite-defspec name nil others)
+    2 (apply rewrite-defspec name others)
+    (throw (ex-info "Unrecognized defspec form"))))
```
stelcodes, line 76: `ex-info` needs second argument

.clj-kondo/hooks.clj
```diff
@@ -0,0 +1,135 @@
...
+  {:node (list-node
+          (list (token-node 'defn)
+                name
+                (vector-node [])
+                (list-node (list* (token-node 'doseq)
+                                  bindings
+                                  body))))})
+
+(defn rewrite-service
+  [name doc proto dependencies & methods]
```
stelcodes, line 30: `_doc` and `_proto`?

.clj-kondo/hooks.clj
```diff
@@ -0,0 +1,135 @@
...
+(defn defservice
+  [{{[_ name & others] :children} :node}]
+  (apply rewrite-service name (arrange-service-args others)))
+
+(defn service
+  [{{[_ & body] :children} :node}]
+  (apply rewrite-service (gensym) (arrange-service-args body)))
+
+(defn rewrite-defspec
+  [name num-tests-or-options? property]
```
stelcodes, line 68: `_num-tests-or-options?`?

.clj-kondo/hooks.clj
```diff
@@ -0,0 +1,135 @@
...
+
+(defn one-binding-then-body
+  [{{[_ binding & body] :children} :node}]
+  {:node (list-node
+          (list* (token-node 'let)
+                 (vector-node [binding nil])
+                 body))})
+
+(defn with-coordinated-fn
+  [{{[_ proceed waiter & body] :children} :node}]
```
stelcodes, line 94: `_waiter`?

stelcodes (commented):
Some small sugguestions for `.clj-kondo/hooks.clj`

src/puppetlabs/puppetdb/query/paging.clj
```diff
@@ -94,14 +91,12 @@
   that satisfy `order-by-expr?`"
   [order_by]
   {:post [(every? order-by-expr? %)]}
-  (when-let [bad-order-by (some
-                           (fn [x] (when-not (contains? x :field) x))
-                           order_by)]
+  (when (some (fn [x] (when-not (contains? x :field) x))
+              order_by)
     (throw (IllegalArgumentException.
             (tru "Illegal value ''{0}'' in :order_by; missing required key 'field'."))))
```
stelcodes, line 97: This should be changed. We do need the value from the original `when-let`, we just mistakenly weren't using it. The `tru` call here is missing its `0` input parameter. Our error messages must have been missing this value for a while and no one noticed. It should be something like this:
```
  (when-let [bad-order-by (some #(when-not (contains? % :field) %) order_by)]
            (tru "Illegal value ''{0}'' in :order_by; missing required key 'field'."
                 bad-order-by))))
```

src/puppetlabs/puppetdb/query_eng.clj
```diff
@@ -252,11 +250,13 @@
     :else obj))
 
 (defn user-query->engine-query
-  ([version query-map options] (user-query->engine-query version query-map options false))
-  ([version query-map options warn-experimental]
+  ([_version query-map options] (user-query->engine-query _version query-map options false))
```
stelcodes, line 253: Since we are using `version` in this arity, I think it should be without the underscore

stelcodes (commented):
Got through a big chunk of the PR, going to finish the rest tomorrow

PR #3678, opened 2022-05-31
(PE-33853) Update clj parent to 5.0.0

rbrw:
...and switch to upstream tools.namespace.  If upstream still can't
handle spaces in the classpath, then let's pursue it there.

The only change in the puppetlabs fork was this in loader-classpath:

  -     #(java.io.File. (.getPath ^java.net.URL %))
  +     #(java.io.File. (.toURI ^java.net.URL %))

https://github.com/puppetlabs/tools.namespace/commit/6ab30bc2be773bb64850ce9460cec22f9d3fdf9e

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,6 +1,6 @@
 (def pdb-version "7.11.0-SNAPSHOT")
 
-(def clj-parent-version "4.9.8")
+(def clj-parent-version "5.0.0")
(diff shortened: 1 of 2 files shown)


jonathannewman:
Tests pass.  LGTM! Nice!

PR #3679, opened 2022-06-03
(PE-33901) Add dumpling to codeowners

KNBAnderson:


KNBAnderson:
I wasn't sure if this should be on 6x or on main

jonathannewman:
main should be fine.

CODEOWNERS
```diff
@@ -1,2 +1,2 @@
-* @puppetlabs/puppetdb
-/docker/ @puppetlabs/pupperware
+* @puppetlabs/dumpling
+/docker/ @puppetlabs/dumpling
```
jonathannewman, line 2: We can eliminate this specification as line 1 covers the whole repo.

jonathannewman (changes requested):
The scope of code owners haas duplication.  It can be more simple.

PR #3685, opened 2022-07-11
(maint) update clj-parent to 5.1.0

jonathannewman:
This updates clj-parent to 5.1.0, which contains a newer version
of jetty.(maint) update clj-parent to 5.1.0

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,6 +1,6 @@
 (def pdb-version "7.11.0-SNAPSHOT")
 
-(def clj-parent-version "5.0.0")
+(def clj-parent-version "5.1.0")
 
 (defn true-in-env? [x]
   (#{"true" "yes" "1"} (System/getenv x)))


jonathannewman:
Travis failure is on jdk8, which is not supported on this branch anyway.

mwaggett:
According to https://puppet.com/docs/puppet/7/server/install_from_packages.html#java-support, we do support Java 8 in the FOSS Puppet 7 stream.

jonathannewman:
@stelcodes the jdk8 failure seems to be a timeout in some channel read.  I'm unable to run these locally.  Can you look into what is going on?  It is very odd that an update in Jetty would break things.

stelcodes:
Looks like it was a transient issue. Yay?

jonathannewman:
Thanks @stelcodes !

PR #3687, opened 2022-07-11
(PE-34097) update clj-parent to 4.11.1

jonathannewman:
This brings in jetty, version 9.4.48 which contains a few security fixes as well as dependency updates.


steveax (approved):
👍

mwaggett (approved):

PR #3689, opened 2022-07-12
(PE-34097) update clj-parent to 5.1.1

jonathannewman:
This change in clj-parent brings in a new version of tk-jetty9 that uses a newer
version of Jetty (9.4.48).  This change includes several security fixes,
bug fixes as well as dependency updates.


steveax (approved):
👍

PR #3690, opened 2022-07-13
(maint) Disable travis container tests

stelcodes:
The container tests started mysteriously failing around April 4th 2022.
This is causing all CI runs to be failures. Since efforts to fix the
failing test have been unsuccessful and the PDB container is not a
priority, this commit disables the container tests.


steveax (approved):
👍

PR #3700, opened 2022-08-08
(PE-34257) update clj-parent to 5.2.1

jonathannewman:
This updates clj-parent to 5.2.1 to include a new version of the postgres driver
to address CVE-2022-31197


steveax (approved):
👍

PR #3701, opened 2022-08-08
(PE-34257) update clj-parent to 4.11.2

jonathannewman:
This updates the clj-parent reference to 4.11.2 to bring in a newer
version of the postgres driver to address CVE-2022-31197.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,6 +1,6 @@
 (def pdb-version "6.22.1-SNAPSHOT")
 
-(def clj-parent-version "4.11.1")
+(def clj-parent-version "4.11.2")
 
 (defn true-in-env? [x]
   (#{"true" "yes" "1"} (System/getenv x)))


steveax (approved):
👍

steveax (approved):
👍

mwaggett:
Are we okay with referencing private tickets in public repos /cc @rbrw ? I thought that was something we tried to avoid, but that could have changed without my knowledge.

jonathannewman:
We do this all the time.  clj-parent is a public repo @mwaggett

mwaggett:
I know the puppet repo is pretty stringent about it, and puppetserver has tried to be in the past. I don't know about puppetdb. People seem less likely to be poking around something like clj-parent for more info than any of our core components.

rbrw:
Not sure exactly yet what's being asked, but broadly speaking, I'd prefer to create PDB and SERVER tickets whenever it's feasible, so that, e.g. the release notes will be more meaningful to the public, etc. (git log too).

Though on the PDB side, the release notes do mention of PE specific changes with a `(PE-only)` and iirc PE ticket numbers.

PR #3703, opened 2022-08-08
Upgrade to honeysql v2

austb:
Blocked on https://github.com/puppetlabs/clj-parent/pull/351 which is blocked until after the upcoming release.


austb:
Jenkins tests are failing because the method of pinning honeysql in this PR doesn't work when building an ezbake package. Once we get the update into clj-parent and no longer need the pin, those tests will pass.

jonathannewman:
@austb are we ready to get this in?

jonathannewman:
I think continuing to target 8.x (main) for this change makes sense.  Seems risky for 7.x  / LTS.

PR #3710, opened 2022-09-09
(maint) dep-diff: allow ignoring bouncycastle provided scope

austb:
In Puppet Enterprise, the bouncycastle jars are not packaged in the
uberjar and are instead provided on the classpath depending on whether
or not it is a FIPS environment.

This allows the dep-diff script to ignore the provided scope on the
bouncycastle jars so that it just ensures we compare the versions of
bouncy castle.

### DIFF
--- a/ext/bin/dep-diff
+++ b/ext/bin/dep-diff
@@ -72,6 +72,18 @@ Usage:
         (exit 2))))
   ds)
 
+(defn strip-bc-scope [dep]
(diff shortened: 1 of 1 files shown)


steveax (approved):
👍 
Looks like it solved the issues in https://github.com/puppetlabs/pe-puppetdb-extensions/pull/749

PR #3713, opened 2022-09-16
(PE-34383) Test main branch with Java 17

austb:


justinstoller:
This looks good for the Java 17 bits. Jon brought it to my attention that the ticket for this work is https://tickets.puppetlabs.com/browse/PE-34383 if you want to link that up.
FWIW, I ended up upgrading puppet server to also use the jammy distro and lein 2.9.10, the latter change at least is something we want to happen soon (but can be another PR if you like).

PR #3728, opened 2022-10-25
(PDB-5547) Schedule gc jobs separately

austb:
Split the gc into multiple jobs to allow them to be scheduled at different intervals.

Run the gc routines on Postgres on a threadpool with a single worker thread to ensure gc jobs don't compete for PG's resources.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/services.clj
+++ b/src/puppetlabs/puppetdb/cli/services.clj
@@ -508,6 +508,8 @@
    context
    (finally (some-> (:job-pool context)
                     (shut-down-service-scheduler-or-die request-shutdown)))
+   (finally (some-> (:gc-pool context)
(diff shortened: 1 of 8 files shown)


src/puppetlabs/puppetdb/cli/services.clj
```diff
@@ -398,6 +420,15 @@
         (time! (get-metric :other-clean-time)
                (garbage-collect! db))
         (counters/inc! (get-metric :other-cleans)))
+      (when (request "gc_catalogs")
```
donoghuc, line 423: Trying to understand this comment, maybe that should be updated, or the `other` condition should be moved after these two?:
```
;; It's important that this go last to ensure anything referencing
;; an env or resource param is purged first.
```

rbrw (approved):

PR #3745, opened 2022-11-16
(PDB-5555) Lock tables consistently when inserting and purging reports

nicklewis:
Previously, we acquired table locks in reverse order when inserting vs.
purging reports. Insertion was certnames, reports, resource_events and
purging was resource_events, reports, certnames. This created the
possibility of a deadlock when two concurrent transactions each acquired
a portion of their locks at the same time. We now acquire all the locks
for report purging up front.

This introduces a helper function `acquire-locks!` which will lock a
given set of tables with specified modes, in a consistent order across
all callers. We now have a partial ordering (so far only consisting of
certnames, reports and resource_events) for tables, so that explicit
locks will always be acquired the same way.


src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -231,6 +231,18 @@
...
+  "This provides a partial ordering of tables, to ensure that we always lock
+  them in a consistent order."
+  ["certnames" "reports" "resource_events"])
+
+(defn acquire-locks!
+  "Acquires a lock on the specified table."
+  [table-modes]
+  {:pre [(every? string? (keys table-modes))
+         (every? string? (vals table-modes))]}
+  (let [ordered-tables (sort-by #(.indexOf table-lock-order %) (keys table-modes))]
```
nicklewis, line 244: This will put tables that aren't covered by the ordering at the start (`.indexOf == -1`) which seems vacuously correct.

nicklewis:
Is there a reason we only log those messages in the case where we use the bulldozer rather than all the time? That is, should they just be incorporated into `f`?

nicklewis:
Okay pushed a commit. I moved the _invocation_ of the bulldozer to cli.services which may or may not be desired. It makes sense to me because conceptually the operation of "drop these tables" is separate from the system-level feature of granting those steps priority access to the db. I'm happy to rework that if it's not right though.

nicklewis:
Retargeted!

PR #3748, opened 2022-11-18
(PDB-5554) Drop partition concurrently

jpartlow:
If we are working with a Postgresql 14+ database, we can use `drop partition concurrently` which has narrower locking constraints.


jpartlow:
I think I got this passing on 7.x with postgres11. Need to figure out how to test postgres14.

jpartlow:
Well, I suppose it would if I got the sql syntax correct. sigh.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1669,12 +1672,17 @@
...
-                    (finally
-                      (update-lock-status status-key dec))))
+                   (let [{current :version} (sutils/db-metadata)
+                          reduced-locking pg14-db]
+                     (update-lock-status status-key inc)
+                     (try!
+                       (jdbc/do-commands
+                         (if (neg? (compare current reduced-locking))
+                           (format "drop table if exists %s cascade" table)
+                           (let [base-table-name (str/replace table #"_[^_]+$" "")]
```
jpartlow, line 1682: There might be a better way to get this? I didn't see a utility method in partitioning.clj.

jpartlow: nm; I hadn't noticed that table-prefix was available.

jpartlow:
Turns out we partition via table inheritance and triggers rather than postgres declarative partitioning and so can't detach partitions this way.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -2108,7 +2360,9 @@
    77 add-catalog-inputs-pkey
    78 add-catalog-inputs-hash
    79 add-report-partition-indexes-on-certname-end-time
-   80 add-workspaces-tables})
+   80 add-workspaces-tables
+   81 migrate-resource-events-to-declarative-partitioning
```
jpartlow, line 2364: Is anyone who would be using these migrations guaranteed to be on at least Postgresql 11?

src/puppetlabs/puppetdb/scf/partitioning.clj
```diff
@@ -65,106 +60,31 @@
...
-                            ;; example: constraint will have 2019-09-21T00:00:00Z but upon querying, you'll see 2019-09-21 17:00:00-07
-                            ;; this is just the database performing i18n
-                            date-column (.format start-of-day date-formatter) date-column (.format start-of-next-day date-formatter)
-                            base-table)]
-                   (index-fn full-table-name table-name-suffix)))))
+        full-table-name (format "%s_%s" base-table table-name-suffix)
+        {current-pg-version :version} (sutils/db-metadata)
+        allow-expressions sutils/declarative-partitions-allow-expressions
+        ;; XXX although, if this is acceptable in pg11, then presumably date-formatter is doing all the work and the inclusion of TIMESTAMP WITH TIME ZONE is unnecessary here?
+        range-template (if (neg? (compare current-pg-version allow-expressions))
```
jpartlow, line 67: I suspect this either unnecessary or broken.

src/puppetlabs/puppetdb/scf/migrate.clj
jpartlow: Separate question; are the migrations wrapped in transactions? I feel like these two should be.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1676,12 +1679,17 @@
...
-                    (jdbc/do-commands
-                     (format "drop table if exists %s cascade" table))
-                    (finally
-                      (update-lock-status status-key dec))))
+                   (let [{current :version} (sutils/db-metadata)
+                          reduced-locking pg14-db]
+                     (update-lock-status status-key inc)
+                     (try!
+                       (if-not (neg? (compare current reduced-locking))
+                         (jdbc/do-commands-outside-txn
```
jpartlow, line 1687: What happens here if there's an outer transaction in play...particularly in regards to the following drop table?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1890,8 +1898,11 @@
     (try
       (store!)
       (catch org.postgresql.util.PSQLException e
-        ;; 42P01 undefined table
-        (if (= "42P01" (.getSQLState e))
+        ;; * 42P01 undefined table -- inheritance based partitions, or
+        ;; declarative partitions when no children yet attached
+        ;; * 23514 check constraint violation -- declarative partitions when row
+        ;; does not match check constraint of existing children partitions
+        (if (or (= "42P01" (.getSQLState e)) (= "23514" (.getSQLState e)))
```
jpartlow, line 1905: A new error test is introduced here because you can get back a 23514 attempting to add a row when declarative partitions are being used. Do we need to version guard this? My feeling was no, but I could be overlooking something.

test/puppetlabs/puppetdb/scf/storage_test.clj
```diff
@@ -1829,12 +1829,13 @@
...
-        ;; check that idx_reports_id index is present in on demand paritions
-        (is (= 1 (count partition)))
-        (dorun (->> partition
-                    (map tu/table-indexes)
-                    (map (partial assert-index-exists "idx_reports_id_")))))))
+            partitions (get-partition-names "reports")]
+        ;; check that primary key index is present in on demand paritions
+        ;; XXX Do we still care about this?
+        (is (= 1 (count partitions)))
+        (doseq [partition-name partitions
```
jpartlow, line 1836: Not sure if this test is still important since we aren't managing the children indices now.

jpartlow:
Alright, this was passing for me locally for Postgres 11 and 14, but it needs to be looked over carefully.

I still need to do a schema-diff and make sure the schema changes make sense, and probably add a test around that, at least.

Probably needs whacking with a clojure idiomatic clue bat as well.

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -1705,3 +1705,739 @@
...
+
+(deftest migration-81-resource-events-declarative-partitioning
+  (testing "resource_events table declarative partitioning migration"
+    (jdbc/with-db-connection *db*
+      (clear-db-for-testing!)
+      (fast-forward-to-migration! 80)
+      (let [before-migration (schema-info-map *db*)]
+        (apply-migration-for-testing! 81)
+        (clojure.pprint/pprint (diff-schema-maps before-migration (schema-info-map *db*)))
+        (is (= {:index-diff
```
jpartlow, line 1717: Still evaluating whether these are the correct changes to expect.

jpartlow: Ah, and I need to generate the expected diff a bit more dynamically to deal with the fact that the schema generates with current dates.

jpartlow:
I've gotten all the tests sorted out except for the macos-10.15 core-ext test run. That is failing in the services and migrate tests, and since it's specifically for gc/partition issues, I assume these are real, though I've no idea why just this one platform, unless perhaps it's the postgres version. One thing about the partition-gc-clears-queries-blocking-it-from-getting-accessexclusivelocks is puzzling me though; https://github.com/puppetlabs/puppetdb/blob/main/test/puppetlabs/puppetdb/cli/services_test.clj#L520-L521 -> it seems like we're storing two reports we expect to be gc'd but only drop one for comparison, but since this generally passes I figure I'm just not understanding something about the test.

jpartlow:
Also I don't think the Jenkins pipeline is actually passing; I think we're just getting back the fact that the ezbake job completed successfully, not that the downstream test jobs run with that build completed:

* https://jenkins-enterprise.delivery.puppetlabs.net/job/enterprise_puppetdb_init-ghpr_7.x/70/
* https://jenkins-enterprise.delivery.puppetlabs.net/job/enterprise_puppetdb_packaging-os-clj_lein-ezbake-stage_7.x/234

Which I think ran this failing set?
* https://jenkins-enterprise.delivery.puppetlabs.net/job/enterprise_puppetdb_integration-system-puppetdb_full-7.x/219/

test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -506,21 +506,38 @@
...
-                                count)))
-
-                 (is (= (->> report-parts (sort-by :table) (drop 1) set)
-                        (set (get-temporal-partitions "reports"))))
-                 (is (= (->> event-parts (sort-by :table) (drop 1) set)
-                        (set (get-temporal-partitions "resource_events")))))))))))))
+                 (let [expected-report-parts (->> report-parts (sort-by :table) (drop 2) set)
+                       expected-event-parts (->> event-parts (sort-by :table) (drop 2) set)]
+
+                   ;; There is some delay between garbage collection completing
```
jpartlow, line 512: Could do something like this to pass. But should garbage collection have already waited on this?

src/puppetlabs/puppetdb/scf/migrate.clj
jpartlow: I'm pretty sure these migration would fail on anything less than 11 due to lack of declarative partitioning, so, yeah, the question is how do we want to handle that for FOSS upgrades?

jpartlow: After chatting with Austin, will update oldest-allowed-db to 11.0 to block upgrades from 9.6 which doesn't support declarative partitions (and from 10 which is no longer supported by postgres).

src/puppetlabs/puppetdb/scf/partitioning.clj
jpartlow: Using 'foo'::timestamptz is not also not allowed in pg11. It has to be a literal constant. But since we are formatting them as ISO_OFFSET_DATE_TIME which has tz info ('2007-12-03T10:15:30+01:00') we shouldn't need the additional cast. I'll remove it and version test.

test/puppetlabs/puppetdb/cli/services_test.clj
jpartlow: Going over this with Austin, he pointed out that gc drops one table per run, so the test expectation is correct. So in ad30d43, to avoid the race, I added a before-gc cyclicbarrier to block next gc execution after after-gc point is reached.

src/puppetlabs/puppetdb/scf/storage.clj
jpartlow: Agreed that no version guard was needed here.

test/puppetlabs/puppetdb/scf/storage_test.clj
jpartlow: Postgres is managing the partition indices automatically, but since this is tied to a former escalation, https://tickets.puppetlabs.com/browse/PDB-4832, I'm going to be conservative and leave the test in place.

jpartlow:
In PG14, incrementally dropping partitions runs into the basic problem that once the partition is detached, it no longer shows up in the get-partition-names query, and prune-daily-partitions proceeds to drop the next partition instead.

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -2045,6 +2117,186 @@
...
+       "  (cached_catalog_status) WHERE (cached_catalog_status = 'on_failure'::text)"]
+      "CREATE INDEX reports_catalog_uuid_idx ON reports USING btree (catalog_uuid)"
+      "CREATE INDEX idx_reports_certname_end_time ON reports USING btree (certname,end_time)"
+      "CREATE INDEX reports_end_time_idx ON reports USING btree (end_time)"
+      "CREATE INDEX reports_environment_id_idx ON reports USING btree (environment_id)"
+
+      ; NOTE: The reports_hash_expr_idx changes to include
+      ; the producer_timestamp because declarative partitions must include the
+      ; range column in all unique indices.
+      "CREATE UNIQUE INDEX reports_hash_expr_idx ON reports USING btree (producer_timestamp, encode(hash, 'hex'::text))"
```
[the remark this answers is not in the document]
jpartlow: Thanks, I should be able to get to this tomorrow.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1716,38 +1792,46 @@
...
-
-
+  (let [effective-resource-events-ttl
+         (if (before? report-ttl resource-events-ttl) resource-events-ttl report-ttl)]
+    (if-not (detach-partitions-concurrently?)
+      ;; PG11
+      (delete-reports-older-than-in-pg-11! report-ttl effective-resource-events-ttl
+                                           incremental? update-lock-status)
+      ;; PG14+
+      (do
```
jpartlow, line 1802: So this is fixed to work on PG14, but may still be subtly broken as I alluded to in the commit. Since the detach happens outside of a transaction, I think a periodic-gc and a manual sweep-reports could collide. Also the dropping of the tables no longer as the update-lock-status call.

test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -572,15 +595,15 @@
        (fn []
          (testing "A report from 24 hours ago won't be gc'ed with a report-ttl of 1d"
            (store-report (-> 1 time/days time/ago time/to-string))
-           (svcs/sweep-reports! *db* {:incremental true
+           (svcs/sweep-reports! *db* {:incremental? false
```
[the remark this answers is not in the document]
jpartlow: Also, also, this test has gc running at 0.01, which is probably good because it alerted me to the overlap between periodic and manual sweep, but I don't know if that was intentional?

src/puppetlabs/puppetdb/scf/storage.clj
jpartlow: The drop-partition-tables now uses update-lock-status and call-with-lock-timeout like partition-daily-partitions.

However I think we still have a problem with detach in overlapping gc. An 'alter table [if exists] foo detach foo_x' can't be run idempotently. The if exists is only for whether the foo exists, not whether foo_x is still a partition of foo. Consequently detach-daily-partitions is vulnerable to get-partition-names being out of sync with detaching the partitions.

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1687,42 +1687,48 @@
...
                     (try!
                       (if just-detach?
-                        (jdbc/do-commands-outside-txn
-                          (format "alter table %s detach partition %s concurrently" table-prefix table))
-                        (jdbc/do-commands
-                          (format "drop table if exists %s cascade" table)))
+                        (let [detached-name (format "%s_detached_%s" table ts-string)]
+                          (jdbc/do-commands-outside-txn
+                            (format "alter table if exists %s detach partition %s concurrently" table-prefix table)
+                            (format "alter table if exists %s rename to %s" table detached-name))
```
jpartlow, line 1698: This is an attempt to reduce the window in which a new store thread fails because a new partition can't be created due to the existing table with same name.

It's messy, and I don't know how safe it is, in terms of time, what happens to indices, serial...

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1653,45 +1653,97 @@
...
+                       (log/warn (trs "More than 2 partitions remaining to prune: {0}"
+                                      (pr-str (butlast candidates)))))
+                     [dropped-or-detached]))
+                 (doall (map drop-one candidates)))]
+     (if-not daily-partition-drop-lock-timeout-ms
+       (drop)
+       (call-with-lock-timeout drop daily-partition-drop-lock-timeout-ms)))))
+
+(defn detach-daily-partitions
+  "Detaches declarative partitions concurrently in PG14+, which runs with a
```
jpartlow, line 1725: Is this correct as far as what we're trying to achieve here?

src/puppetlabs/puppetdb/scf/storage.clj
```diff
@@ -1716,38 +1807,47 @@
...
+        (jdbc/with-db-transaction []
+          ;; We update these tables in a different order than report insertion, creates
+          ;; the potential for deadlocks when two concurrent transactions have each
+          ;; acquired a portion of their locks. So first, we acquire locks on all the
+          ;; tables in the same order they are accessed during insertion. These match
+          ;; the locks that would implicitly be acquired by the operations we're going
+          ;; to perform: Access exclusive for reports and resource_events so we can
+          ;; drop partition tables, and row exclusive to update certnames.
+          (acquire-locks! {"certnames" "ROW EXCLUSIVE"
+                           "reports" "ACCESS EXCLUSIVE"
```
jpartlow, line 1836: I believe these locks have to remain access exclusive in order to drop the tables safely. So whatever benefit we are seeing is presumably because the detach means they should no longer be being queried?

src/puppetlabs/puppetdb/scf/migrate.clj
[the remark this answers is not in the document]
jpartlow: This should be fixed now.

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -1705,3 +1707,356 @@
...
+        :is_partial false,
+        :primary? true,
+        :user "pdb_test"},
+       :same nil}
+       {:left-only nil,
+        :right-only
+        {:schema "public",
+         :table (format "reports_%s" date-suffix),
+         :index (format "reports_%s_producer_timestamp_encode_idx" date-suffix),
+         :index_keys ["producer_timestamp" "encode(hash, 'hex'::text)"],
```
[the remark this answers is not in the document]
jpartlow: This should be fixed now.

test/puppetlabs/puppetdb/scf/migrate_test.clj
```diff
@@ -1705,3 +1707,356 @@
...
+                 :primary? false,
+                 :user "pdb_test"},
+                :right-only nil,
+                :same nil}
+               {:left-only nil,
+                :right-only
+                {:schema "public",
+                 :table "reports",
+                 :index "reports_hash_expr_idx",
+                 :index_keys ["producer_timestamp" "encode(hash, 'hex'::text)"],
```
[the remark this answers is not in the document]
jpartlow: This should be fixed now.

jpartlow:
The migration is primarily just renames and reattaching partitions. However a number of indices get recreated on reports, resource_events. Do we have any sense how long that may take on large tables? (secs, mins or hours?)

src/puppetlabs/puppetdb/scf/storage.clj
jpartlow: Adjusted this in 133c377

PR #3765, opened 2023-01-20
(PE-35151) Replace snyk action with mend

artlawson:


.github/workflows/mend.yaml
```diff
@@ -0,0 +1,53 @@
+name: mend_scan
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - 2021.7.x
```
jonathannewman, line 7: This should be 7.x -- 2021.7.x is Puppet Enterprise specific.

jonathannewman (approved):

PR #3769, opened 2023-01-30
(PDB-5560) PuppetDB Terminus updates for Ruby 3 and Puppet 8

austb:


raybellis:
Can you please backport this change to PuppetDB 7.x?    FreeBSD's package library now only supports Ruby 3.2 and we're not ready to move to Puppet 8 yet.

PR #3771, opened 2023-01-31
(maint) Fix race in accessexclusivelocks test

jpartlow:
Adds a CyclicBarrier before garbage collection starts to prevent the start of the *next* gc until we are ready. Without this, when the test awaits after-gc, it releases the invoke-periodic-gc to perform the next gc cycle (which is every hundredth of a second by the test config) and you get the race between the test validating which tables have dropped and the gc dropping the next table.

Also adds a barrier right before query submission, and a small sleep to hopefully let those threads submit their three queries. Otherwise, timing can be such that the bulldozer can go by first, find nothing, then the queries get submitted, then we hang for 20 min.

Unfortunately, we can't wait on a barrier absolutely after the query submissions because the whole point is for those queries to hang and be cleared by the bulldozer. We can't wait for them to come back.

### DIFF
--- a/test/puppetlabs/puppetdb/cli/services_test.clj
+++ b/test/puppetlabs/puppetdb/cli/services_test.clj
@@ -446,12 +446,16 @@
                                            cmd-consts/latest-report-version
                                            (change-report-time example-report %))
           after-gc (CyclicBarrier. 2)
+          before-gc (CyclicBarrier. 2)
+          queries-begun (CyclicBarrier. 4)
           original-periodic-gc svcs/invoke-periodic-gc
           invoke-periodic (fn [& args]
                             (if (some #{"purge_reports"} (nth args 2))
-                              (let [result (apply original-periodic-gc args)]
-                                (.await after-gc)
-                                result)
+                              (do
+                                (.await before-gc)
+                                (let [result (apply original-periodic-gc args)]
+                                  (.await after-gc)
+                                  result))
                               ;; if not purging reports, run gc normally
                               (apply original-periodic-gc args)))
           event-expired? (fn [_ _] true)]
@@ -463,6 +467,7 @@
            (with-log-level "puppetlabs.puppetdb.scf.storage" :info
              (with-logged-event-maps log
                ;; Wait for the first, full gc to finish.
+               (.await before-gc)
                (.await after-gc)
                (store-report "2011-01-01T12:00:01-03:00")
                (store-report "2011-01-02T12:00:01-03:00")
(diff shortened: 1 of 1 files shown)


test/puppetlabs/puppetdb/cli/services_test.clj
```diff
@@ -482,15 +487,24 @@
...
                                         (jdbc/with-transacted-connection *db*
-                                          (jdbc/do-commands "select event_hash, pg_sleep(1200) from resource_events")))]
+                                          (.await queries-begun)
+                                          (jdbc/do-commands "select event_hash, pg_sleep(12) from resource_events")))]
 
-                   ;; gc should clear the two queries from above and drop only the oldest partition
+                   ;; Intention is to ensure that the queries have begun before gc starts and the query bulldozer whizzes by,
+                   ;; because if the bulldozer zips through before the queries are actually made, nothing gets cleared.
+                   (.await queries-begun)
+                   (Thread/sleep 1000)
```
jpartlow, line 504: Is there a better way to do this?

jpartlow:
Without this, locally I see race issues between gc and testing that the reports have been submitted.

From a `time ext/bin/boxed-core-tests -- lein test :only puppetlabs.puppetdb.cli.services-test
time ext/bin/boxed-core-tests -- lein test :only puppetlabs.puppetdb.cli.services-test` run

[cli-services-7.x.log](https://github.com/puppetlabs/puppetdb/files/10551198/cli-services-7.x.log)

---

Without the sleep allowing the queries to begin, after adding the before-gc barrier, I instead start to see failures verifying queries were cleared:

```
lein test :only puppetlabs.puppetdb.cli.services-test/partition-gc-clears-queries-blocking-it-from-getting-accessexclusivelocks

FAIL in (partition-gc-clears-queries-blocking-it-from-getting-accessexclusivelocks) (services_test.clj:501)
expected: (thrown-with-msg? java.util.concurrent.ExecutionException #"FATAL: terminating connection due to administrator command" (deref report-query default-timeout-ms :timeout-getting-expected-query-ex))
  actual: nil

lein test :only puppetlabs.puppetdb.cli.services-test/partition-gc-clears-queries-blocking-it-from-getting-accessexclusivelocks

FAIL in (partition-gc-clears-queries-blocking-it-from-getting-accessexclusivelocks) (services_test.clj:505)
expected: (thrown-with-msg? java.util.concurrent.ExecutionException #"FATAL: terminating connection due to administrator command" (deref report-query2 default-timeout-ms :timeout-getting-expected-query-ex))
  actual: nil

lein test :only puppetlabs.puppetdb.cli.services-test/partition-gc-clears-queries-blocking-it-from-getting-accessexclusivelocks

FAIL in (partition-gc-clears-queries-blocking-it-from-getting-accessexclusivelocks) (services_test.clj:509)
expected: (thrown-with-msg? java.util.concurrent.ExecutionException #"FATAL: terminating connection due to administrator command" (deref resource-query default-timeout-ms :timeout-getting-expected-query-ex))
  actual: nil

lein test :only puppetlabs.puppetdb.cli.services-test/partition-gc-clears-queries-blocking-it-from-getting-accessexclusivelocks

FAIL in (partition-gc-clears-queries-blocking-it-from-getting-accessexclusivelocks) (services_test.clj:519)
expected: (<= 2 (->> (clojure.core/deref log) (map :message) (map (fn* [p1__58447#] (str/includes? p1__58447# "Partition GC terminated queries"))) (filter true?) count))
  actual: (not (<= 2 0))
```

PR #3774, opened 2023-02-07
(maint) Install puppetdb module 7.10.0 on platform5

jpartlow:
For 6.x, the acceptance suite running with INSTALL_TYPE upgrade_oldest will install puppet5 on el7 hosts in order to test upgrading from puppet5. However it currently installs the puppetlabs-puppetdb module latest to install puppetdb, and head of that module depends on puppetlabs-postgresql module <9 which pulls in 8.2.1 which makes use of Deferred, which isn't available in puppet5...

So if we're installing the puppetdb module for platform5, install version 7.5.0 which depends on postgresql module <8 that doesn't use Deferred.

### DIFF
--- a/acceptance/helper.rb
+++ b/acceptance/helper.rb
@@ -429,6 +429,15 @@ module PuppetDBExtensions
     end
   end
 
+  def install_puppetdb_module(hosts, puppet_platform)
+    command = "puppet module install puppetlabs/puppetdb"
+    # Use the older 7.10.0 version of puppetdb module on puppet5 because
+    # it depends on puppetlabs-postgresql < 8, which does not use Deferred
+    # (which does not exist in puppet5).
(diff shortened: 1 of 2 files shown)


jpartlow:
This passed INSTALL_TYPE upgrade_oldest for me locally for both centos7 and redhat8 (platform5 and platform6, respectively).

jpartlow:
The integration tests were all failing because puppetserver couldn't parse the puppetdb uris. I grabbed 625fecc6b to see if that fixes it. But there might be a reason we don't want that in 6.x?

jpartlow:
I assume the one core travis lein test :only puppetlabs.puppetdb.command-test/concurrent-catalog-resource-updates is a transient. It should be unrelated to the pr at least.

PR #3777, opened 2023-02-09
Merge 7.x to main 20230209

jpartlow:


project.clj
```diff
@@ -1,6 +1,6 @@
 (def pdb-version "8.0.0-SNAPSHOT")
 
-(def clj-parent-version "5.2.16")
+(def clj-parent-version "5.2.17")
```
jpartlow, line 3: I assume this is a good thing, since it's going forward?

PR #3778, opened 2023-02-14
(maint) Fix timezone issue in migration test

austb:
This test was failing on mismatching constraint names when run in the -08:00 offset

expected
```
{:constraint_name
 "(((producer_timestamp >= '2023-02-06 08:00:00+00'::timestamp with time zone) AND (producer_timestamp < '2023-02-07 08:00:00+00'::timestamp with time zone)))",
 :table_name "reports_20230207z",
```

actual
```
{:constraint_name
 "(((producer_timestamp >= '2023-02-06 16:00:00-08'::timestamp with time zone) AND (producer_timestamp < '2023-02-07 16:00:00-08'::timestamp with time zone)))",
 :table_name "reports_20230207z",
```

To fix this we take the date range, and ensure it is converted to UTC. This ensures that we get the proper UTC midnight datetime for our partition constraints.  Then we format the date using the systemDefault timezone (assuming that is what Postgres is also running in)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ aliases:
 
   - &run-integration-tests |
     set -e
+    rvm use "$(./ext/bin/prefixed-ref-from-spec "$PDB_TEST" ruby-)"
     ci/bin/prep-and-run-in travis "$PDB_TEST"
 
   - &run-spec-tests |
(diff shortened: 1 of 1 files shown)


jpartlow (approved):
I'm not sure why I didn't see this locally, but it make sense. Postgres will return schema based on system default tz and I was forcing it to format to utc.

jpartlow:
The main .travis.yml is specifying a 2.7 ruby version for 7.x; can we do something similar here?

PR #3779, opened 2023-02-14
(PDB-5553) Remove foreign keys from report partitions

austb:
We have not released this migration, otherwise a separate migration would be needed.

Postgres foreign keys are implemented with bidirectional triggers this means that even after we have detached the partition from the parent table, dropping it will still require access exclusive locks on all the tables that we reference with foreign keys.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -2159,8 +2159,8 @@
   (let [table "resource_events"
         date-column "timestamp"
         child-table-maps (partitioning/get-temporal-partitions table)]
-    ; Create new partitioned table with placeholder name and original schema
-    ; as documented in migration 73 (resource-events-partitioning)
+    ;; Create new partitioned table with placeholder name and original schema
+    ;; as documented in migration 73 (resource-events-partitioning)
     (jdbc/do-commands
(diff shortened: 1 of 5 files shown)


jpartlow (approved):
:+1:

src/puppetlabs/puppetdb/scf/migrate.clj
```diff
@@ -2242,26 +2253,29 @@
...
-    ; Moving existing children to partitioned table
+    ;; Moving existing children to partitioned table
     (reattach-partitions table child-table-maps)
 
-    ; Drop redundant child constraints
+    ;; Drop redundant child constraints
     (drop-child-partition-constraints date-column child-table-maps)
 
+    ;; Drop existing foreign keys from partitions
+    (drop-foreign-key-constraints child-table-maps)
```
jpartlow, line 2263: Just occurred to me that while it's true this migration isn't released externally yet, this will, I think, break z to z upgrades internally (pe-next and the like).

jpartlow:
The tests look reasonable. Re-running the lint job.
The only issue I see is what I raised above about PE latest z upgrades. I don't know if we've done this before and if so how much of an issue it's every been.

jpartlow:
Okie dokie; I think that just leaves the linter fix.

PR #3782, opened 2023-02-28
(PDB-3171) Improve performance for queries with large arrays

nicklewis:
Previously, the tree visitor code would *always* replace every node it 
visited, regardless of whether the value was actually changed. The code is
intended to allow the visitor to return nil to indicate the node shouldn't
be replaced. In practice, that doesn't actually work for a couple of
reasons. The first is that we weren't diligent about ensuring that every
visitor function returns `nil` when there is no change to be made. The
other is that the `node-visit` function had a bug which would cause it to
always return the original node value if the visitor function returned
`nil`.

The problem is compounded by the fact that whenever a child node is 
replaced, the *entire* parent is rewritten. So for large arrays, that means
that we recreate the whole array node once for every child. This caused
significant performance problems for queries with large arrays
(like a large "in" clause), because the parent array is entirely replaced
once for every single one of its children.

We now separately track the "current" node value and the "new" node value,
so that a sequence of visitors that each return `nil` will all be applied
to the initial node value, but we will still return `nil` for the final
replacement value.

Using the following query as a test, this change reduces query time from
~40sec to ~160ms (a 250x improvement).

["from", "nodes",
 ["in", "certname", ["array", [...10 thousand elements...]]]]

This PR also moves to vendoring fast-zip.visit as it is a single file and the
upstream library appears to be pretty inactive.


nicklewis:
See also https://github.com/puppetlabs/puppetdb/pull/3783 which fixes one (very important) instance of this problem that exists only on main.

project.clj
```diff
@@ -186,7 +186,7 @@
                  [com.rpl/specter "0.5.7"]
                  [com.taoensso/nippy :exclusions [org.tukaani/xz]]
                  [digest "1.4.3"]
-                 [fast-zip-visit "1.0.2"]
+                 [fast-zip "0.4.0"]
```
nicklewis, line 189: This was previously pulled in transitively by `fast-zip-visit`, so I just changed it to the version that was being pulled in. We might be able to update.

jonathannewman (approved):

nicklewis:
@austb I saw [this upstream issue])(https://github.com/akhudek/zip-visit/issues/8) which specifically referenced PuppetDB, so I figured that meant it was probably going to be some amount of work to upgrade `fast-zip`, so I just stuck with exactly the version we were using before.

PR #3786, opened 2023-03-01
(maint) remove daily GHA workflow from 7.x branch

austb:
scheduled github actions only run on the main branch, so this workflow is likely to cause merge-up conflicts without ever running tests.


jpartlow (approved):
:+1:

PR #3792, opened 2023-03-03
(PDB-5592) Generate catalogs

jpartlow:
Adds a generate command to generate sample data files to be used to seed a benchmark.
First pass at catalog generation to accompany it.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -159,6 +159,7 @@
                  [org.clojure/core.async]
                  [org.clojure/core.match "0.3.0-alpha4"]
                  [org.clojure/core.memoize]
+                 [org.clojure/data.generators "1.0.0"]
                  [org.clojure/java.jdbc]
                  [org.clojure/tools.macro]
                  [org.clojure/tools.namespace]
--- a/src/puppetlabs/puppetdb/cli/generate.clj
+++ b/src/puppetlabs/puppetdb/cli/generate.clj
@@ -14,6 +14,29 @@
    of 5 large catalogs and 10 small ones, you will need to run the tool twice
    with the desired parameters to create the two different sets.
 
+   ### Flag Notes
+
+   The num-resources flag is total and includes num-classes. So if you set
+   --num-resources to 100 and --num-classes to 30, you will get a catalog with a
+   hundred resources, thirty of which are classes.
+
+   A containment edge is always generated between the main stage and each
+   class. And non-class resources get a containment edge to a random class. So
+   there will always be a base set of containment edges equal to the resource
+   count. The --additional-edge-percent governs how many non-containment edges
+   are added on top of that to simulate some further catalog structure. There is
+   no guarantee of relationship depth (as far as, for example Stage(main) ->
+   Class(foo) -> Class(bar) -> Resource(biff)), but it does ensure some edges
+   between classes, as well as between class and non-class resources.
+
+   The --avg-blob-count and --blob-size parameters control inclusion of large
+   text blobs in catalog resources. By default an average of 1 ~100kb blob is
+   added per catalog. Blobs are distributed randomly through the set, so if you
+   set --avg-blob-count to 20 over --hosts 10, on averge there will be two per
+   catalog, but some may have none, others four, etc...
+
+   Set --avg-blob-count to 0 to exclude blobs altogether.
+
    TODO:
    * facts
    * reports"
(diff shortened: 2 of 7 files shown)


project.clj
```diff
@@ -159,6 +159,7 @@
                  [org.clojure/core.async]
                  [org.clojure/core.match "0.3.0-alpha4"]
                  [org.clojure/core.memoize]
+                 [org.clojure/data.generators "1.0.0"]
```
jpartlow, line 162: The version number seems to be required here. I have no idea why.

nicklewis: For dependencies that are specified in [clj-parent](https://github.com/puppetlabs/clj-parent), we can omit the version and it will be inherited. This one isn't in clj-parent so it has to be specified explicitly.

The parent project is defined just above this section:
https://github.com/puppetlabs/puppetdb/pull/3792/files#diff-274071745a4e2a04b647d79d500537e6dc13eee54f44d0426140026293701d1bL149-L150

src/puppetlabs/puppetdb/cli/generate.clj
```diff
@@ -0,0 +1,309 @@
...
+  ([prefix ordinal]
+   (pseudonym prefix ordinal 6))
+  ([prefix ordinal length]
+   (let [mid-length (max (- length (count prefix) (count (str ordinal))) 1)]
+     (format "%s-%s-%s" prefix (rnd/random-pronouncable-word mid-length) ordinal))))
+
+(defn build-to-size
+  "Build up list of strings to approximately size bytes."
+  [size word-fn]
+  (let [builder (fn [v]
```
nicklewis, line 45: This can be written using `loop` if you want to avoid the helper function.

```clj
(loop [v []]
  (if (< (reduce + (map count v)) size)
    (recur (conj v (word-fn)))
    v))
```

src/puppetlabs/puppetdb/cli/generate.clj
```diff
@@ -0,0 +1,309 @@
...
+
+   Randomly generates additional required-by, notifies, before and
+   subscription-of between a given percentage of both class and non-class resources.
+
+   Makes use of Catalog to ensure we're still building a directed acyclic graph, and to
+   track the relation type between edges."
+  [main-stage classes resources additional-edge-percent]
+  (let [additional-class-edges (quot (* additional-edge-percent (count classes)) 100)
+        additional-resource-edges (quot (* additional-edge-percent (count resources)) 100)
+        shuffled-classes (set (shuffle classes))
```
nicklewis, line 161: `set` appears to undo the shuffling process, because it all gets put into a hash.

project.clj
jpartlow: Ah, so these are shared dependencies that can be optionally included across subprojects without each subproject fiddling version numbers requiring multiple prs to update.

src/puppetlabs/puppetdb/cli/generate.clj
jpartlow: That's nicer; fixed this and build-parameters below.

src/puppetlabs/puppetdb/cli/generate.clj
jpartlow: Yep, that doesn't work quite right. Thanks for catching that!. I've shifted it over so I only force it into a set before performing the cset/difference.

jpartlow:
Current limitations:

- [ ] getting catalogs of around a particular size is a matter of dialing number of resources, resource size and extra edges and observing output size, and is therefore a bit fiddly
- [ ] if you want catalogs of significantly different size and shape, outside of standard distribution, you will need to run the tool multiple times (say to get 2 big catalogs, 10 medium, 3 small to serve as your base for benchmark which would then randomly grab from that set and permute).
- [ ] type length isn't controllable; that should be an easy thing to add
- [ ] doesn't model big blob parameters. For example the puppet-enterprise-uninstaller is an 82k file content blob in all the pe catalogs (including agent catalogs). Would need to add another knob or two to sprinkle those in (blob-size, blob-frequency?)
- [ ] would it be helpful to print out an end summary of what was generated since the inputs to outputs is a little vague?

jpartlow:
Blobs can be sprinkled in now.

One thing to consider is that without blobs, catalog sizes are pretty consistent because --num-resources is the same across catalogs and resource weights average out:

```
Generate from {:resource-size 200, :silent false, :additional-edge-percent 50, :verbose true, :num-hosts 5, :avg-blob-count 0, :help false, :blob-size 100, :num-resources 100, :num-classes 1
0, :title-size 20} and store at /tmp/pdb-generate-11928295403194073989
:catalogs: 5                                                                                   
                                               
|     :certname | :catalog-weight | :resource-count | :resource-weight | :min-resource | :mean-resource | :max-resource | :edge-count | :edge-weight |
|---------------+-----------------+-----------------+------------------+---------------+----------------+---------------+-------------+--------------|
| host-vetysa-0 |           43779 |             101 |            26681 |           100 |            264 |           453 |         150 |        16798 |
| host-bykife-1 |           44395 |             101 |            27594 |           123 |            273 |           408 |         150 |        16501 |
| host-mybavo-2 |           44026 |             101 |            26874 |           102 |            266 |           421 |         150 |        16852 |
| host-turisa-3 |           43802 |             101 |            26925 |            87 |            266 |           444 |         150 |        16577 |
| host-pekuko-4 |           44833 |             101 |            27431 |           134 |            271 |           436 |         150 |        17102 |
```

Currently I'm distributing blobs randomly afterwards, so there is variation in catalog resource weight, since one catalog may get none, another two for example:

```
Generate from {:resource-size 200, :silent false, :additional-edge-percent 50, :verbose true, :num-hosts 5, :avg-blob-count 1, :help false, :blob-size 100, :num-resources 100, :num-classes 1
0, :title-size 20} and store at /tmp/pdb-generate-3169369799079416170                                                                                                                         
:catalogs: 5                                                                                                                                                                                  
                                                                                                                                                                                              
|     :certname | :catalog-weight | :resource-count | :resource-weight | :min-resource | :mean-resource | :max-resource | :edge-count | :edge-weight |                                        
|---------------+-----------------+-----------------+------------------+---------------+----------------+---------------+-------------+--------------|                                        
| host-vexuko-0 |           44960 |             101 |            27811 |           105 |            275 |           459 |         150 |        16849 |                                        
| host-xadulo-1 |          215512 |             101 |           198414 |           118 |           1964 |         91684 |         150 |        16798 |                                        
| host-mijyju-2 |          148926 |             101 |           131844 |           107 |           1305 |        105752 |         150 |        16782 |                                        
| host-cudipu-3 |          240336 |             101 |           223409 |           116 |           2211 |        101376 |         150 |        16627 |                                        
| host-fuwehu-4 |           44408 |             101 |            26895 |            98 |            266 |           515 |         150 |        17213 |                                        
```

Would it be better to keep the blob distribution exact  so catalog sizes remain relatively uniform, and all major variation is dependent on multiple runs to generate catalogs of different shape?

Alternately, would it be better to similarly distribute resources randomly across the catalogs and thereby ensure  much more variable catalog sizes, even without blobs?

I'm kinda leaning towards #1 as being better for reproducibility.

jpartlow:
Jenkins testing is blocked by ruby version issues in the job setup that I'm looking into separately. Those tests aren't particularly relevant to this cli subcommand pr, so I'm going to go ahead and merge.

PR #3794, opened 2023-03-29
(PDB-5589) Drop obsolete partition indexes

jpartlow:
In the migration to declarative partitions for reports and resource_events, there are a few indexes which are left behind on the migrated partitions, and which do not appear on newly created partitions afterwards.

Because the existing report and resource_events partitions get migrated with their indexes as is, migrated reports partitions have left over idx_reports_id_[date] and reports_hash_expr_idx_[date] indexes, and migrated resource_events partitions have a left over resource_events_hash_[date] index.

For reports partitions:
* idx_reports_id_[date] (id) is superseded by primary key on (id, producer_timestamp)
* reports_hash_expr_idx_[date] (encode(hash)) is superseded by the unique reports_hash_expr_idx (encode(hash), producer_timestamp)

For resource_events partitions:
* resource_events_hash_[date] is superseded by the primary key on (event_hash, timestamp)

In both cases, newly created partitions (as partitions get added over time) do not have these obsolete indexes. Postgresql automatically creates indexes in new partitions based on the indexes created on the partitioned table.

The change in this commit just cleans up the migrated partitions to drop the obselete indexes so that they will match newly created partitions.

There is a separate issue in that the names of the indexes are different between migrated and new partitions because postgresql has its own naming scheme for the indexes it creates on new partitions. But since the actual index columns are identical, effectively, the indexes are the same, and this shouldn't matter.

The migrated partitions will cycle out within the ttl time period after migration, regardless.

### DIFF
--- a/src/puppetlabs/puppetdb/scf/migrate.clj
+++ b/src/puppetlabs/puppetdb/scf/migrate.clj
@@ -2120,7 +2120,7 @@
 (defn reattach-partitions
   "Detaches partition children from PostgreSQL inheritance and attaches them to
   the new partition table (under its placeholder name)."
-  [table child-table-maps]
+  [table child-table-maps {:keys [drop-indexes]}]
   (let [placeholder-name (format "%s_partitioned" table)]
     (apply jdbc/do-commands
       (flatten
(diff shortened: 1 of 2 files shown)


jpartlow:
Closing for now. If we want to include this change in a later release we could add an additional migration for it.

PR #3811, opened 2023-05-04
(PDB-5643) Use recursive PQL rules for and/or

austb:
Change PQL expr-or and expr-and grammar rules to use a recursive definition.

By using instaparse's `{ }` operator (meaning zero or more) in self-referential rules the grammar is ambiguous.

`a = 1 or b = 2 or c = 3` can be parsed as something that looks like
```
[:or
  a = 1
  b = 2
  c = 3]
```
Or it could be parsed as
```
[:or
  a = 1
  [:or
    b = 2
    c = 3]]
```
Both are valid given the self-referential grammar rule, but as clauses compound the number of valid parsings grows exponentially significantly affecting performance. As few as 16 or clauses can cause an OOM with 1GB heap.

The recursive grammar definition means only one (the second example above) version is valid. Every and/or operator will have two children.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -79,6 +79,7 @@
       :exclusions [com.andrewmcveigh/cljs-time]]
      [riddley "0.1.12"]
      [io.forward/yaml "1.0.5"]
+     [clj-commons/clj-yaml "1.0.26"]
 
      ;; Only needed for :integration tests
      [puppetlabs/trapperkeeper-filesystem-watcher nil]]


nicklewis:
This is what I came up with to collapse nested `or`/`and` nodes to return the same structure as before.

```diff
diff --git a/src/puppetlabs/puppetdb/pql/transform.clj b/src/puppetlabs/puppetdb/pql/transform.clj
index ef321471e..797f14657 100644
--- a/src/puppetlabs/puppetdb/pql/transform.clj
+++ b/src/puppetlabs/puppetdb/pql/transform.clj
@@ -33,13 +33,25 @@
   ;; Single arg? collapse
   ([data] data)
   ;; Multiple args? turn it into an or statement
-  ([data & args] (vec (concat ["or" data] args))))
+  ([data & args]
+   (->> (cons data args)
+        (mapcat #(if (and (vector? %) (= (first %) "or"))
+                   (rest %)
+                   [%]))
+        (cons "or")
+        (vec))))
 
 (defn transform-expr-and
   ;; Single arg? collapse
   ([data] data)
   ;; Multiple args? turn it into an and statement
-  ([data & args] (vec (concat ["and" data] args))))
+  ([data & args]
+   (->> (cons data args)
+                      (mapcat #(if (and (vector? %) (= (first %) "and"))
+                              (rest %)
+                              [%]))
+                      (cons "and")
+                      (vec))))
 
 (defn transform-expr-not
   ;; Single arg? Just collapse the :expr-not and pass back the data,
```

jpartlow (approved):
:+1: @austb was Nick presenting an alternative approach or something else that needed to be addressed?

jpartlow:
@austb Ok, I've pushed two additional commits, one adding Nick's patch to transform back to the original AST structure after PQL has parsed efficiently. And another to add a regression test for PQL parse time of a deep OR expressions.

If those look reasonable, than I think we can go ahead and merge this.

jpartlow:
Need to rebase.

jpartlow:
Ok, clj-yaml was only readded as a test dependency in main, but not tk in clj-parent 5.x has also been updated to remove that dependency, so I'm cherry-picking 1893f8f into 7.x.

PR #3814, opened 2023-05-22
(PE-35599) Add script to analyze index usage

austb:
The script is designed to easily get the index usage from a PE support script.

It is difficult to know which file a particular piece of data is in in the support script, so the file path is support script specific.

That being said, the data it needs is all standard postgres statistics, so it is trivial to reproduce the format well-enough that it will work on an open-source installation of PuppetDB

```
mkdir enterprise
echo puppetdb > enterprise/postgres_db_stats.txt
echo pg_stat_user_tables >> enterprise/postgres_db_stats.txt
psql puppetdb -c "select * from pg_stat_user_tables" >> enterprise/postgres_db_stats.txt
echo pg_stat_user_indexes >> enterprise/postgres_db_stats.txt
psql puppetdb -c "select * from pg_stat_user_indexes" >> enterprise/postgres_db_stats.txt
vim enterprise/postgres_db_stats.txt
./ext/bin/analyze-index-usage puppetdb ./
```

### DIFF
--- a/ext/bin/analyze-index-usage
+++ b/ext/bin/analyze-index-usage
@@ -1,0 +1,145 @@
+#!/usr/bin/env ruby
+#
+# intended to do some basic analysis of the index usage in a given
+# Puppet Enterprise Support Script
+
+def usage
+  <<~USAGETXT
+  Usage:
+      analyze-index-usage DATABASE_NAME SUPPORT_SCRIPT_DIR
+  USAGETXT
+end
+
+def helptext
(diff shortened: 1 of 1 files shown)


jpartlow (approved):
:+1: This will be handy for periodically checking index usage.

jpartlow:
I have a commit over in #3815 that increases the timeout for the unreleated query-timeout-tests that are failing. Going to go ahead and merge this.

PR #3815, opened 2023-05-26
(PDB-5593) Generate reports

jpartlow:
Adds report generation to the generate cli, which should complete the generation needed to run benchmark.


jpartlow:
I have some linting to fix, but the main test failures are odd since this is just cli/generate and it's based off current main.

jpartlow:
Oops, forgot to implement add-logs-to-reports.

jpartlow:
I have two generate tests to fix. But there are also some puppetlabs.puppetdb.query-timeout-test failures.

jpartlow:
Fixing some issues in report resource change generation.

jpartlow:
Ok, I think I'm down to a couple of puppetlabs.puppetdb.query-timeout-test failures that I'll look at tomorrow.

PR #3818, opened 2023-06-08
(PE-36194) benchmark: when submitting multiple messages backdate timestamps

austb:


src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -235,6 +236,9 @@
                 :parse-fn #(Integer/parseInt %)]
                ["-N" "--nummsgs NUMMSGS" "Number of commands and/or reports to send for each host"
                 :parse-fn #(Long/valueOf %)]
+               ["-e" "--end-commands-in PERIOD" "A period (like '3d') to use to set the ending date of a set of command"
```
jpartlow, line 239: "set of commands"

PR #3823, opened 2023-06-14
(PDB-5663) Sync version +1 to improve sync request timeout handling

rbrw:
Increment the sync version so that sync requests from newer servers that specify a query timeout will receive a more meaningful message that indicates the version mismatch, rather than a more generic crash on an unexpected parameter.


rbrw:
One way we might handle it.

PR #3827, opened 2023-06-26
(PDB-5636) add ubuntu 2204 as supported platform

jonathannewman:
This adds ubuntu 2204/jammy to the set of supported puppetdb versions starting now.


steveax (approved):
👍

PR #3828, opened 2023-06-26
(PDB-5599) Removes the catalog_resources_file_trgm index

jpartlow:
This was an expensive index which is no longer used. Also resets the autovacuum_analyze_scale_factor table storage parameter that had been set alongside the original index.


jpartlow:
Does this migration need ::vacuum-analyze set?

For that matter, did the declarative partition migrations?

PR #3833, opened 2023-06-30
(maint) testing: move jdk8 testing to 7.x only

austb:
Move testing of jdk8 to only on the 7.x branch. After upgrading hikari in the main branch, it is no longer compatible with jdk8.


jpartlow (approved):
:+1: This always warps my brain a bit. So it's the main branch's configuration github uses for actions, regardless of which branch is under test?

PR #3835, opened 2023-07-11
(PE-36193) PDB benchmark can use ssl

jpartlow:
(maint) Remove unused base-url functions

This look to have disappeared from use years ago when the cli tooling
changed.

---

    (PE-36193) Use puppetlabs.http.client in puppetdb.client

Replaces use of clj-http.client with our puppetlabs.hhtp.client that
provides for https communication when provided certs.  Specifically,
this is to allow benchmark to use https for it's pdb commands. Threads
pdb jetty ssl opts to through to the new http client from benchmark
invocation.

---

    (PE-36193) Update benchmark docs for https

* Adds documentation for using benchmark over https.
* Marks https as the preferred configuration if you are going to run
benchmark off the primary.
* The http configuration docs incorrectly specified that you could lock
down http access to a single agent by setting puppetdb's jetty.host to
the agent ip. The host setting tells puppetdb which interfaces to listen
on on the primary, not which ip's to allow requests from.
* Emphasized that opening http is not recommended as this is insecure and
allows http access from any source.

### DIFF
--- a/documentation/load_testing_tool.markdown
+++ b/documentation/load_testing_tool.markdown
@@ -65,14 +65,35 @@ which may significantly skew performance on the primary server. If you would lik
 the benchmark tool on an agent this can be achieved following the instructions
 below.
 
-* On the primary server, modify `/etc/puppetlabs/puppetdb/conf.d/jetty.ini`.
-In the `[jetty]` section, set either:
-    * `host=0.0.0.0 # http access from all agents`
-    * `host=<agent ip address> # access from specific agent`
+#### Running over https
+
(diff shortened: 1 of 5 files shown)


jpartlow:
I didn't find a great spot for adding tests for this. Setting up a full pdb stack with certs looked pretty fiddly as the helpers are just testing plain http I think. Open to suggestions.

jpartlow:
Jenkins tests are currently failing because of a transient issue getting gems from artifactory. Eric mentioned he's working on the rubygems caching with jfrog.

PR #3836, opened 2023-07-14
(maint) Remove beaker-hostgenerator --global-config ssh for smallstep

jpartlow:
7.x acceptance job is passing while main is not, and this is one visible
 difference, though it's not clear why it would matter.


jpartlow:
Thread: https://perforce.slack.com/archives/C049M3S5MB5/p1689270370227209

jpartlow:
osx lein failure in in the query-timeout-test is irrelevant here. Going ahead and merging this to see if it has any affect on the ci acceptance test.

PR #3841, opened 2023-07-24
(PE-36496) Generate fake - hopefully realistic - benchmark data

austb:
The command run to generate this data was

```
lein run generate \
        --num-classes 100 \
        --num-resources 800 \
        --title-size 40 \
        --resource-size 500 \
        --additional-edge-percent 15 \
        --blob-count 2 \
        --blob-size 40 \
        --num-facts 400 \
        --total-fact-size 15 \
        --max-fact-depth 7 \
        --num-packages 1000 \
        --num-reports 20 \
        --high-change-reports-percent 5 \
        --high-change-resources-percent 60 \
        --low-change-reports-percent 15 \
        --low-change-resources-percent 5 \
        --num-additional-logs 10 \
        --percent-add-report-logs 20
```


jpartlow:
The numbers you provided look good to me. However because you've only run the tool once with one set of catalog numbers, you've got 5 pretty even catalogs there (about 800kb each). The reports are spikier, since the bulk are low change, with a few high change.  But is it important to simulate an environment that has different catalog weights for different types of nodes? (And correspondingly different report spreads?) Maybe that's something we worry about later in testing.

At any rate, I think keeping a copy of data as a reference is a good idea. It's not terribly big, about 9.6M; but it could also be stored directly in orchestrator-scale-testing, if we didn't want it here in puppetdb. I'm fine with it being here though :shrug:.

jpartlow (approved):
:+1: Given that you are just adding sample data, I think we can absolve you of the test transients...

PR #3842, opened 2023-07-24
(maint) update ezbake to 2.5.2

jonathannewman:
This updates ezbake to 2.5.2, which does two things:
* it enables builds of the el9 platform
* it adds in tzdata-java as a dependency on the el platforms

tzdata-java was strangely removed as a dependency from openjdk, so this adds it in as a dependendency to ensure we don't release a broken package.


rileynewton (approved):
👍

PR #3857, opened 2023-08-16
(maint) Fix progressing-timestamp to progress

jpartlow:
The progressing-timestamp was being called with each update-host, regenerating the function each time with a base value. This ensures that we generate the function once so it can progress.

Also changed the default end-command-in days to 0 so that for cases where adding historical reports should take only minutes or hours don't end up shifted forward 7 days by default.


rbrw (approved):
Nice catch.

jpartlow:
Failures are a query monitoring timing issue and rhel9 platform which is work in progress.

PR #3862, opened 2023-08-23
(maint) benchmark: account for update time in sleep calculation

austb:
The calculation for ms-per-thread calculates the number of milliseconds that should elapse between each command on a single simulation thread. By sleeping for that number of milliseconds, benchmark would never be able to reach the number of commands per second that actually simulates the number of hosts. When rand-perc is 0 (the default), update-host is nearly instantaneous and the rate of commands is nearly the requested amount (often within 1%). When run with rand-perc of 100, update-hosts consumed a non-negligible amount of the time per simulation thread. When attempting to simulate 20,000 nodes, benchmark was only submitting about 16,500 commands per second.

This changes the simulation thread to calculate when it should output a host, performing the work necessary (if any) to update that host, and then sleeping until the previously calculated deadline if it hasn't already passed. If it has already passed, benchmark is not keeping up and it needs more resources (CPU/simulation threads/etc).

### DIFF
--- a/src/puppetlabs/puppetdb/cli/benchmark.clj
+++ b/src/puppetlabs/puppetdb/cli/benchmark.clj
@@ -460,9 +460,13 @@
      simulation-threads
      write-ch
      (map (fn [host-state]
-            (when-not num-msgs
-              (Thread/sleep (int (- ms-per-thread (rand)))))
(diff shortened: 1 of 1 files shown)


jpartlow (approved):
Ah, yup; good catch. :+1:

PR #3887, opened 2023-10-04
(PE-36947) Allow for Puppet::Util::Puppetdb to flush config

jpartlow:
This provides a hook for flushing the Puppetdb classes @config variable. This in turn means the next time Puppetdb.config is callled it will reload the configuration file.
This becomes important if in some command using this utility, we have accessed the Puppetdb.config, then go on to update puppet's puppetdb.conf with new information, and need to then take additional operations with Puppetdb.config that should be based on the new configuration state.

### DIFF
--- a/puppet/lib/puppet/util/puppetdb.rb
+++ b/puppet/lib/puppet/util/puppetdb.rb
@@ -23,6 +23,11 @@ module Puppet::Util::Puppetdb
   class SoftWriteFailError < Puppet::Error
   end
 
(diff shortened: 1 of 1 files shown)


jpartlow:
Oops, need to rebase this against 7.x.

jpartlow:
I kicked the jenkins test. Query timeout tests continue to fail on macos11.

tlehman (approved):

PR #3888, opened 2023-10-05
(PDB-5712) Move benchmark to JDK client

rbrw:


src/puppetlabs/puppetdb/client.clj
```diff
@@ -1,69 +1,122 @@
...
+
+(defn- post-body
+  [^HttpClient client
+   ^HttpRequest$Builder req-generator
+   body-publisher
+   response-body-handler]
+  (let [res (.send client (-> req-generator (.POST body-publisher) .build)
+                   response-body-handler)]
+    ;; REVIEW: is :status this enough for current purposes?
+    {::jdk-response res
```
jpartlow, line 85: This is probably my ignorance of Clojure syntax, but why two colons?

src/puppetlabs/puppetdb/client.clj
```diff
@@ -126,3 +179,5 @@
                    ssl-opts)]
      (when-not (= HttpURLConnection/HTTP_OK (:status result))
        (log/error result)))))
+
+(set! *warn-on-reflection* warn-on-reflection-orig)
```
jpartlow, line 183: Was this something you added while debugging or is it helpful to keep?

rbrw: So (particularly given the focus on performance), Austin and I have been trying out the idea of just keeping reflection eliminated in namespaces where the required changes aren't "too noisy".

There's no intention to do this eagerly, just whenever we like (and are presumably already making changes), and then we'll just leave the warnings turned on for those namespaces so we'll notice regressions.

PR #3896, opened 2023-10-24
(PE-35601) Add a cert number --offset to Benchmark

jpartlow:
Allows you to run two or more Benchmark instances in parallel, offsetting the generated cert numbers so that the commands don't collide in the database. Let's you test massively when number of simulated nodes exceeds what a single Benchmark can push.


jpartlow:
Working with a two benchmark test cluster.

PR #3913, opened 2023-11-27
(GH-3912) add include_catalog_edges config to terminus

### LINKED ISSUE #3912: Add configuration option to disable sending of catalog edges from terminus
Profiling catalog submission has shown that catalog edges can take up to one third of the total time to submit a catalog. They are not the most used, or important piece of catalog data, so many users (especially those with many nodes) would benefit from not storing edges and having faster catalog submission times.

austb:
Add configuration option to the terminus, `include_catalog_edges` that, when set to `false`, will disable sending any catalog edges to PuppetDB. Profiling of PuppetDB command processing has shown catalog edges can take as much as one third of the total time to process a catalog. Users with many nodes that don't use edges in their queries could benefit quite substantially from using this option. It is not enabled by default to maintain backwards compatibility during PuppetDB 8.

Resolves #3912.

### DIFF
--- a/documentation/puppetdb_connection.markdown
+++ b/documentation/puppetdb_connection.markdown
@@ -72,6 +72,16 @@ This setting can let the Puppet Server stay partially available during a PuppetD
 
 The default value is false.
 
+### `include_catalog_edges`
+
+This setting tells the PuppetDB terminus whether or not it should include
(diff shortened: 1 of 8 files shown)


jpartlow (approved):
:+1: This looks pretty straight forward.

PR #3917, opened 2023-11-30
(maint) Preserve random-hosts generated for initial-hosts-channel

jpartlow:
When benchmark starts, it generates a list of initial host-maps based on random selections from the catalog/factset/report sample data. This means that every time benchmark is run a new set of catalogs and facts are going to be pushed to PuppetDB. This can cause a great deal of initial load replacing the catalogs and facts wholesale until PuppetDB has caught up to processing one entire simulated node interval.

This makes it impractical to stop and restart benchmark when working with large simulated runs.

This patch attempts to improve this be nippy/freezing the initial host-maps to a --simulation-dir (if given). This then allows a subsequent run (if given --simulation-dir) to thaw those host-maps rather than rebuilding them.

XXX: There are some issues with this approach:

1. --numhosts changes. If the user changes the number of hosts for the subsequent run, we either have too many or too few frozen maps. The current match may deal with this adequately.

2. --offset changes. This is a subtler bug. Since we don't recent the thawed host names, they will be out of sync. And since we need offset ourselves to deal with --numhosts changes, there is further complication there. This has not been addressed yet.


jpartlow:
#3918 may be better than investing time in sorting out the --offset issue and ensuring this will work as expected when running multiple benchmarks.

jpartlow:
Doing some manual testing now.

jpartlow:
My initial implementation was loading all the preserved host-map files during initialization for simplicity sorting out changes to things like --nummsgs and which --catalog, --facts, --reports flags were set on re-run. But this blows up the heap for large runs (10,000 hosts of generate/realistic is about 4.6G of frozen host-map files, for example...).

Rewrote this to defer loading of the preserved-host-map to the simulation loop. This did get through 10,000 simulated hosts, but I think there's still a memory leak in it. Will look at it some more tomorrow.

Also need to fix up the tests again.

jpartlow:
Latest version defers loading of the preserved host-map file to the simulation-loop, outside of the sequence generated by random-hosts, and looks to resolve the oom issue.

I was able to run 20,000 nodes on my standard 8/16G test host with this. There might still be a timing or load issue with benchmark though, since, though I've got 20,000 preserved files after that run, I'm only seeing 19981 catalogs in puppetdb. I don't know whether this is a pre-existing problem though.

src/puppetlabs/puppetdb/cli/benchmark.clj
```diff
@@ -50,7 +50,21 @@
...
    ...
-   ```"
+   ```
+  
+   ### Preserving host-map data
+  
+   By default, each time Benchmark is run, it initializes the host-map catalog,
+   factset and report data randomly from the given set of base --catalogs
+   --factsets and --reports files. When re-running benchmark, this causes
+   excessive load on puppetdb due to the completely changed catalogs/factsets
```
jpartlow, line 60: Is it just the catalogs that are an issue here, or is it the facts as well?

jpartlow:
This patch worked to restart twin benchmarks pushing 50,000 each.

PR #3918, opened 2023-11-30
(maint) Sync initial-hosts on benchmark re-execution

jpartlow:
When benchmark starts, it generates a list of initial host-maps based on
random selections from the catalog/factset/report sample data. This
means that every time benchmark is run a new set of catalogs and facts
are going to be pushed to PuppetDB. This can cause a great deal of
initial load replacing the catalogs and facts wholesale until PuppetDB
has caught up to processing one entire simulated node interval.

This makes it impractical to stop and restart benchmark when working
with large simulated runs.

This patch attempts to improve this be encoding the certname of the
original sample file in the catalog version string. We then query
catalogs certname, version fields, and form an index of certname ->
original-certname. This index is then used to ensure that the set of
initial hostmaps is regenerated from the same base catalog and facteset
so there should not be excess churn when benchmark begins pushing
commands.

### DIFF
--- a/src/puppetlabs/puppetdb/cli/benchmark.clj
+++ b/src/puppetlabs/puppetdb/cli/benchmark.clj
@@ -59,6 +59,7 @@
             [me.raynes.fs :as fs]
             [clojure.java.io :as io]
             [clojure.stacktrace :as trace]
+            [clojure.string :as s]
             [clojure.walk :as walk]
             [puppetlabs.puppetdb.utils :as utils :refer [println-err]]
             [puppetlabs.kitchensink.core :as kitchensink]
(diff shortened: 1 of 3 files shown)


jpartlow:
This is working in limited local testing and is probably a more generically reliable approach.
Main downsides compared to #3917 are that it's more convoluted, and I don't know yet if it has scale issues (querying large catalog sets).
I haven't touched the tests yet.
I'm going to do some more testing on an ost instance to try and tease it if there's a performance problem.

jpartlow:
Still seeing queue spikes restarting benchmarks with an xl with 2 compilers and 100,000 simulated nodes.

jpartlow:
Since we can't deal with accumulated changes to the catalogs/facts during the tenure of the previous run without extensive querying of the database, closing in favor of a reworked #3917 that can preserve all files in the TempFileBuffer.

PR #3929, opened 2024-01-09
(PE-37477) Install puppetlabs-puppetdb from branch

donoghuc:
Install the puppetdb module from a branch to allow taking up a suitable puppetlabs-postgresql module.


donoghuc:
Seems to have worked from puppetserver https://github.com/puppetlabs/puppetserver/pull/2810

jonathannewman (approved):

PR #3932, opened 2024-01-10
Mergeup module fix

donoghuc:


acceptance/helper.rb
```diff
@@ -432,12 +432,9 @@ def get_package_version(host, version = nil)
   end
 
   def install_puppetdb_module(hosts, puppet_platform)
-    command = "puppet module install puppetlabs/puppetdb"
-    # Use the older 7.10.0 version of puppetdb module on puppet5 because
-    # it depends on puppetlabs-postgresql < 8, which does not use Deferred
-    # (which does not exist in puppet5).
-    command << " -v 7.10.0" if puppet_platform == :puppet5
-    on hosts, command
+    # While we sort out a new puppetlabs-puppetdb module release, point to a branch that allows us to take the latest puppetlabs-postgresql module
```
donoghuc, line 435: Note that we only upgrade from puppe 7 so we dont need the special puppetlabs-postgresql branch needed for upgrading from puppet 6 in the 7.x stream.

PR #3933, opened 2024-01-12
(PE-37466) Ensure query monitor forgets for all relevant routes

rbrw:


rbrw:
Suggest ignoring blankspace when looking at some of the commits.

jonathannewman:
Given the relative complexity of the code, and the multi-threaded nature of it, have you considered adding debug and trace level logging to both this code and the query monitor to help isolate issues when they occur?

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -396,68 +396,84 @@
...
+                   (http/error-response
+                    (if origin
+                      (tru "Query {0} from {1} exceeded timeout" id (pr-str origin))
+                      (tru "Query {0} exceeded timeout" id))
+                    HttpURLConnection/HTTP_INTERNAL_ERROR))))]
+         (if-not mon-id
+           (handler new-req)
+           (try!
+             (let [res (handler new-req)]
+               (when (and mon-id (not (:puppetlabs.puppetdb.query/streaming res)))
```
jonathannewman, line 470: `mon-id` is always set at this point, so I think you can make this a `when-not` on the second condition.

src/puppetlabs/puppetdb/http/query.clj
```diff
@@ -396,68 +396,84 @@
...
+                 (qmon/forget query-monitor mon-id))
+               res)
+             ;; REVIEW: throwable?
+             (catch Exception ex
+               (when mon-id (qmon/forget query-monitor mon-id))
```
jonathannewman, line 475: `mon-id` is always set at this point.

jonathannewman: Given it should always be forgotten, might be a little easier to follow if the `forget` is moved into a finally block.

jonathannewman: Is `forget` exception safe?

src/puppetlabs/puppetdb/http/query.clj
rbrw: Oh, right - thanks.  Think that was left over from the rearrangements.

src/puppetlabs/puppetdb/http/query.clj
rbrw: Hmm, I'll have to double-check, but I think this might be the case where we should only `forget` on the error path.  And I believe the `murphy/try!` should handle chaining any pending exception for us.

PR #3978, opened 2024-05-14
(maint) benchmark_test: track add/del operations to set max expected resources

austb:
We should be getting a balanced number of additions and deletions by using rand-nth, but if there's only 1 resource remaining deletes become a modification. We start with 4 resources, so if we get a string of 7 deletes followed by a string of 7 additions, we have a balanced number of operations, but we will end with 8 resources. This shifts our expected end number of resources to the right of a simple binomial distribution.


jpartlow (approved):
:+1:

Issue #3975, opened 2024-05-07
facts-blocklist doesn't work as expected with modern facts (hash structure).


sharewax:
## Describe the Bug
I'm trying to reduce database size and not to store facts which are not needed in puppetdb. For example - info in `networking.interfaces` for dynamic/virtual links (calico, veth, docker etc). Currently it's impossible to block it with current filters.
I have in config for example:
```
facts-blocklist-type = regex
facts-blocklist =  networking.interfaces.cali.*, networking.interfaces.veth.*, networking.interfaces.docker.*, networking.interfaces.macvtap.*, networking.interfaces.vnet.*, networking.interfaces.tunl.*
```
But this facts are still present in puppetdb. I've explained a little bit more in https://github.com/puppetlabs/facter/issues/2713
## Expected Behavior
I would like to have possibility to exclude storage some facts in puppetdb with modern structure (Hash) with regex support.

## Environment
 - Version 7.15.0
 - Platform CentOS7

## Additional Context
https://github.com/puppetlabs/facter/issues/2713

djuarezg:
I was getting frustrated with the blocklist as I was actually expecting it to work with structured facts... until I found this issue and then lost hope for the feature.

I needed the feature to exclude exactly the same type of structured facts as you do...

PR #3998, opened 2024-09-11
Add configuration options to filter facts out in puppetdb termini

traylenator:
This changeset adds two configuration options used by the facts PuppetDB indirector:

  * fact_names_blocklist
  * fact_names_blocklist_regex

They can be used to configure a list of fact names that will never be sent to PuppetDB, based on exact fact names or regular expressions.

This MR obsoletes - https://github.com/puppetlabs/puppetdb/pull/2634


traylenator:
CLA is already signed due to agreement between puppetlabs and CERN IT.

traylenator:
I realise since this patch was originally written PuppetDB itself can be configured to block facts however we have been running this way for years and dropping the facts earlier makes sense.

djuarezg:
For anyone that ends up finding this PR as well. PuppetDB itself cannot be configured to block structured facts due to https://github.com/puppetlabs/puppetdb/issues/3975.

PR #4000, opened 2024-09-13
(PDB-5793): Upgraded beaker-hostgenerator gem to add support for debian-12-x86_64

span786:
Upgraded beaker-hostgenerator gem to add support for debian-12-x86_64


steveax (approved):

Gemfile
```diff
@@ -65,7 +65,7 @@ if ENV['NO_ACCEPTANCE'] != 'true'
       # use the pinned version
       gem 'beaker', '~> 4.1'
     end
-    gem 'beaker-hostgenerator', '~> 2.2.3'
+    gem 'beaker-hostgenerator', '~> 2.4.0'
```
donoghuc, line 68: This specificity is just going to lead to issues later. I suggest:
```suggestion
    gem 'beaker-hostgenerator', '~> 2.4'
```
Which will allow any `2.x` gem after 2.4

donoghuc (changes requested):

Gemfile
steveax: My bad, didn't realize y'all would rather float on the `y`.

donoghuc: no worries! not sure how it got in that state to begin with, easy enough fix up :)

span786:
> I'm assuming this is platform 8 only?

Yes, it is for platform 8 only

Gemfile
span786: Updated the code to float the minor version

donoghuc (changes requested):
Looks like its targeted at 7 also https://github.com/puppetlabs/puppetdb/pull/4002/files . I think you should close this, merge the other one and do a mergeup. Typically for repos like this we like to target changes at the oldest stream and do mergeups.

